PR #1, opened 2023-03-01
Copy Polkadot+Kusama runtimes

ggwpez:


joepetrowski:
We should also move the system parachain runtimes

ggwpez:
> We should also move the system parachain runtimes

Sorry you were not invited to the channel, we now moved the discussion to https://forum.parity.io/t/polkadot-1-0-release/1675    
IIUC Gav correctly; the Cumulus runtimes are waiting for 2D weight completion. Then they can be moved into here.

joepetrowski:
> the Cumulus runtimes are waiting for 2D weight completion

I don't see an issue for this. Is someone working on it, or we need to find someone?

ggwpez:
> > the Cumulus runtimes are waiting for 2D weight completion
> 
> I don't see an issue for this. Is someone working on it, or we need to find someone?

*Someone*, yes… I assume it is mainly this thing https://github.com/paritytech/cumulus/pull/2157

bkchr:
> IIUC Gav correctly; the Cumulus runtimes are waiting for 2D weight completion. Then they can be moved into here.

This was about cumulus 1.0. However System parachains are not cumulus. So, please also move them here as @joepetrowski  said.

ggwpez:
In which structure do you want them @joepetrowski ?  
Like in the Current repo or all in the same root directory?  

- root
  - assets/
    - statemint
    - statemine
  - bridge-hubs/
    - bridge-hub-kusama
    - bridge-hub-polkadot
  - collectives
    - collectives-polkadot/

ggwpez:
Doing this results in a WASM build error. I assume something with the dependencies: `Could not find package with manifest path` from [here](https://github.com/paritytech/substrate/blob/e374a33fe1d99d59eb24a08981090bdb4503e81b/utils/wasm-builder/src/wasm_project.rs#L808) fails.  

Investigating this with my [feature](https://github.com/ggwpez/feature) tool reveals that `bridge-hub-kusama-runtime` depends on `polkadot-runtime` which is probably causing issues for the WASM builder:
```bash
$ feature trace bridge-hub-kusama-runtime polkadot-runtime --path-delimiter="\n -> "

bridge-hub-kusama-runtime
 -> cumulus-pallet-parachain-system
 -> cumulus-primitives-parachain-inherent
 -> cumulus-relay-chain-interface
 -> polkadot-service
 -> polkadot-runtime
```
How do you think we should resolve this?

joepetrowski:
> In which structure do you want them @joepetrowski ? Like in the Cumulus repo or all in the same root directory?
> 
> * assets/
>   
>   * statemint
>   * statemine
> * bridge-hubs/
>   
>   * bridge-hub-kusama
>   * bridge-hub-polkadot
> * collectives
>   
>   * collectives-polkadot/

I think they should be under some parent `runtimes`, because we do have some things that are used across all parachains but not the Relay Chain runtime. For example,

- [Collator Selection](https://github.com/paritytech/cumulus/tree/master/pallets/collator-selection)
- [constants, types, impls](https://github.com/paritytech/cumulus/blob/master/parachains/common/src)

So I would do them in the same structure as the Cumulus repo (maybe change `parachains` to `system-parachains`).

bkontur:
> Doing this results in a WASM build error. I assume something with the dependencies: `The crate package is contained in its own metadata` from [here](https://github.com/paritytech/substrate/blob/e374a33fe1d99d59eb24a08981090bdb4503e81b/utils/wasm-builder/src/wasm_project.rs#L808) fails.
> 
> Investigating this with my [feature](https://github.com/ggwpez/feature) tool reveals that `bridge-hub-kusama-runtime` depends on `polkadot-runtime` which is probably causing issues for the WASM builder:
> 
> ```shell
> $ feature trace bridge-hub-kusama-runtime polkadot-runtime --path-delimiter="\n -> "
> 
> bridge-hub-kusama-runtime
>  -> cumulus-pallet-parachain-system
>  -> cumulus-primitives-parachain-inherent
>  -> cumulus-relay-chain-interface
>  -> polkadot-service
>  -> polkadot-runtime
> ```
> 
> How do you think we should resolve this?

@ggwpez 

what I see from `cargo tree`:
```
bparity@bkontur-ThinkPad-P14s-Gen-2i:~/parity/cumulus/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama$ cargo tree --invert polkadot-runtime
polkadot-runtime v0.9.37 (https://github.com/paritytech/polkadot?branch=master#c75f92ae)
├── polkadot-client v0.9.37 (https://github.com/paritytech/polkadot?branch=master#c75f92ae)
│   └── polkadot-service v0.9.37 (https://github.com/paritytech/polkadot?branch=master#c75f92ae)
│       └── cumulus-relay-chain-interface v0.1.0 (/home/bparity/parity/cumulus/cumulus/client/relay-chain-interface)
│           └── cumulus-primitives-parachain-inherent v0.1.0 (/home/bparity/parity/cumulus/cumulus/primitives/parachain-inherent)
│               └── cumulus-pallet-parachain-system v0.1.0 (/home/bparity/parity/cumulus/cumulus/pallets/parachain-system)
│                   └── bridge-hub-kusama-runtime v0.1.0 (/home/bparity/parity/cumulus/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama)
└── polkadot-service v0.9.37 (https://github.com/paritytech/polkadot?branch=master#c75f92ae) (*)
```

`cumulus-relay-chain-interface` -> `polkadot-service`, and `polkadot-service ` has default feautres:
```
[features]
default = ["db", "full-node", "polkadot-native"]
...
polkadot-native = [ "polkadot-runtime", "polkadot-runtime-constants", "polkadot-client/polkadot" ]
```

so I tried this fix: https://github.com/paritytech/cumulus/pull/2274 (waiting for CI, local `cargo build` passed),
and with this fix `cargo tree` does not fetch `polkadot-runtime`

```
bparity@bkontur-ThinkPad-P14s-Gen-2i:~/parity/cumulus/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-kusama$ cargo tree --invert polkadot-runtime
error: package ID specification `polkadot-runtime` did not match any packages
```

but lets wait for CI if all features work

bkontur:
@ggwpez 
well, it does not work,
neither: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2483807
```
polkadot-service = { git = "https://github.com/paritytech/polkadot", branch = "master", default-features = false }
```
nor: https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2486986
```
polkadot-service = { git = "https://github.com/paritytech/polkadot", branch = "master", default-features = false, features = ["db", "full-node"] }
```
---

problem is that modules: `polkadot-cli` and `polkadot-service` both by-default adds `polkadot-native` which adds `polkadot-runtime`

so I guess, everywhere (binary, runtime, ...) where they are used, there is a `polkadot-runtime` also

---

so for this particular problem with `bridge-hub-kusama-runtime`, which is caused by `polkadot-service` from `cumulus-relay-chain-interface` where we only need `polkadot_service::Error`, maybe it could be possible to extract `polkadot_service::Error` from `polkadot-service`

ggwpez:
> maybe it could be possible to extract `polkadot_service::Error` from `polkadot-service`

Yea… maybe we can move it to it's own crate? I tried to remove the offending `ServiceError(#[from] polkadot_service::Error)` variant but there were too many errors. Seems to be used quite a lot through it's `#[from]` annotation.  

&nbsp;
> I think they should be under some parent `runtimes`, because we do have some things that are used across all parachains but not the Relay Chain runtime. For example,
> 
> * [Collator Selection](https://github.com/paritytech/cumulus/tree/master/pallets/collator-selection)
> * [constants, types, impls](https://github.com/paritytech/cumulus/blob/master/parachains/common/src)
> 
> So I would do them in the same structure as the Cumulus repo (maybe change `parachains` to `system-parachains`).

Okay. I initially thought that we only copy the minimally necessary files to build the runtime, but it looks like these files are de-facto part of the runtimes.

bkchr:
https://github.com/paritytech/cumulus/pull/2287

gilescope:
Ideally keep the history if at all possible as people always are trying to track back to see where a change was introduced and reading the context around that change.

relay/common/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+#![cfg_attr(not(feature = "std"), no_std)]
+
+pub mod assigned_slots;
+pub mod auctions;
+pub mod claims;
+pub mod crowdloan;
+pub mod elections;
+pub mod impls;
+pub mod paras_registrar;
+pub mod paras_sudo_wrapper;
```
bkchr, line 28: We should move these two to `polkadot-runtime-parachains`.

In general there are also other pallets part of this crate. I would personally not move them here, but I'm also not 100% convinced on my thinking :D

ggwpez: Can we do this as follow up? I would like to restrict this MR to moving files, so that reviewing stays easy.  
But in general yes, IMHO these should go somewhere else.

bkchr: This should be done in Polkadot before, as they should stay in Polkadot. Then you also don't need to do here anything.

ggwpez: Okay I can take a stab at it, if no one else does it before me.

bkchr: After some further internal discussion, we don't need this :see_no_evil:

ggwpez: Okay. Then that folder should be good to go. Only excluded stuff is now: metrics, rococo, test-runtime and westend.  
And on the Cumulus side i will move the complete `parachains` folder in following that logic.

ggwpez:
Patch of the last force-push (renaming only) [force-push.patch.txt](https://github.com/polkadot-fellows/runtimes/files/10979799/force-push.patch.txt).

Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[workspace.package]
+authors = ["Polkadot Fellowship"]
```
ggwpez, line 2: I guess this is fine for a start?  
The files still have the parity file header, so maybe we should include parity here as well?

Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+	"relay/common",
+	"relay/common/slot_range_helper",
+	"relay/runtimes/kusama",
+	"relay/runtimes/kusama/constants",
+	"relay/runtimes/metrics",
+	"relay/runtimes/parachains",
+	"relay/runtimes/polkadot",
+	"relay/runtimes/polkadot/constants",
+	"relay/runtimes/test-runtime",
+	"relay/runtimes/test-runtime/constants",
```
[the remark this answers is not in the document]
ggwpez: Removed.

I cannot merge this
image
or do you think we should first merge it into a `dev` branch? The only downside of merging to `main` now is that we will have some sync-up commits additionally in the file commit histories.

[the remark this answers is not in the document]
bkchr: No merging into master sounds right. We will also need to rebase it? So that in the future we can still use git to sync changes.

Issue #8, opened 2023-04-12
Publishing new releases


bkchr:
A new release should be published after a "special" commit got merged to the main branch. This "special" commit should alter the `CHANGELOG` file. When this change is detected, there should be an action that automatically builds all the runtimes and puts out a release. Would probably be super cool to take the diff in the `CHANGELOG` file to put it as changes into the release.

coderobe:
working on this atm

PR #9, opened 2023-07-04
Add release draft ci workflow

coderobe:
Initial workflow for https://github.com/polkadot-fellows/runtimes/issues/8

This builds all the runtimes and drafts a release once the `CHANGELOG` file is edited on the main branch. The workflow can also be triggered manually.
Included in the draft body are the contents of `CHANGELOG` as well as subwasm information for the runtimes built.

image


coderobe:
@bkchr anything specific you would like to see in the release draft body / other thoughts?

also, as i'm not in polkadot-fellows, me & the rest of the releng folk won't be able to see/manage/maintain release drafts & workflows - is this something we want to change in any way?

coderobe:
if we want any of the more intricate parts of the https://github.com/paritytech/polkadot release workflow we'd need to bring over a bunch of auxiliary scripts which i've avoided for now.

.github/workflows/release.yml
```diff
@@ -0,0 +1,159 @@
...
+      matrix:
+        runtime: [ # this is the same matrix as above
+          # relay/para, runtime path, package name, chain
+          ["relay", "polkadot", "polkadot-runtime", "polkadot"],
+          ["relay", "kusama", "kusama-runtime", "kusama"],
+          #["relay", "parachains", "polkadot-runtime-parachains", "polkadot"], # ?
+          #["relay", "metrics", "polkadot-runtime-metrics", "polkadot"], # ?         
+          ["system-parachains", "collectives/collectives-polkadot", "collectives-polkadot-runtime", "collectives-polkadot"],
+          ["system-parachains", "asset-hubs/asset-hub-polkadot", "statemint-runtime", "statemint"],
+          ["system-parachains", "asset-hubs/asset-hub-kusama", "statemine-runtime", "statemine"],
```
bkchr, line 132: Should be called asset-hub-* and not statemint/statemine anymore.

.github/workflows/release.yml
```diff
@@ -0,0 +1,159 @@
...
+    continue-on-error: true
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        runtime: [
+          # relay/para, runtime path, package name, chain
+          ["relay", "polkadot", "polkadot-runtime", "polkadot"],
+          ["relay", "kusama", "kusama-runtime", "kusama"],
+          #["relay", "parachains", "polkadot-runtime-parachains", "polkadot"], # ?
+          #["relay", "metrics", "polkadot-runtime-metrics", "polkadot"], # ?
```
bkchr, line 21: Can be removed.

bkchr (commented):
I think for an initial release it looks good!

@coderobe when including the CHANGELOG in the release, will this always only include the diff between the last release and the current?

.github/workflows/release.yml
coderobe: these come from the Cargo.toml of the runtimes, once this repo is synced to a more recent state this will be updated - until then this is required to build the current fellows runtimes revision.

Could you sync the main branch with the current parity revision? Then i can adjust this part.

coderobe:
> @coderobe when including the CHANGELOG in the release, will this always only include the diff between the last release and the current?

Currently the entire `CHANGELOG` is included in the release draft body. I can change that to only include the diff.

.github/workflows/release.yml
coderobe: Actually, once that is done i'm going to replace this whole thing with a dynamic matrix generated from something like `find */runtimes -name "lib.rs" -not -path "*/common/*" -not -path "*/constants/*" | awk -F/ '{sub("/src/lib.rs",""); print "[\"" gensub("/" $NF "$", "", "g") "\",\"" $NF "\"]"}'` which should reduce maintenance burden by not hardcoding any of these things

bkchr: Ahh okay nice!

bkchr:
@coderobe you can now update to latest main branch changes.

ggwpez:
Rebased. You can re-sign the commits if you want and force-push again @coderobe (or leave them unsigned).

bkchr:
@coderobe any update on this?

coderobe:
> @coderobe any update on this?

i was on vacation this week, i think this will be ready by monday afternoon or so

coderobe:
The initial CI workflow is now compatible with the current revision of the runtimes.

Thoughts, wishes, other feedback? Currently, the release draft generated looks like [this](https://github.com/coderobe/polkadot-fellows-runtimes/releases/tag/test1)

ggwpez:
Looks like a good start to me :tada:  
Can we maybe sort it such that Polkadot and Kusama are at the top? And mention the srtool version that was used to build them?

ordian:
would be nice to group parachain runtimes with corresponding relay chain, like this:
image

gilescope:
Currently we omit the `authorizeUpgrade` and `system.setCode` hashes for parachain runtimes. I think we should continue to do this.

Also bridge-hub-rococo is in the draft but that's a test network? (other test network runtimes are omitted from the draft)

gilescope (approved):

bkchr:
@ggwpez we need to remove the rococo bridge hub. This shouldn't be in this repo.

.github/workflows/release.yml
```diff
@@ -0,0 +1,152 @@
...
+            C() { >>CHANGELOG echo $@; }
+            WASM() { <${JSON} jq -r .runtimes.compact.subwasm.$1; }
+
+            C "### $(WASM core_version.specName)"
+            C '```'
+            C "🏋️ Runtime Size:           $(WASM size) bytes"
+            C "🔥 Core Version:           $(WASM core_version.specName)-$(WASM core_version.specVersion) $(WASM core_version.implName)-$(WASM core_version.implVersion).tx$(WASM core_version.transactionVersion).au$(WASM core_version.authoringVersion)"
+            C "🎁 Metadata version:       V$(WASM metadata_version)"
+            C "🗳️ system.setCode hash:    $(WASM proposal_hash)"
+            C "🗳️ authorizeUpgrade hash:  $(WASM parachain_authorize_upgrade_hash)"
```
joepetrowski, line 102: IMO these two fields are not really the purpose of srtool. They often get wrapped in other calls/instructions like `with_weight` or `Transact {..}`, so nobody sees this in a proposal anyway. They are also sensitive to changes in transaction version.

The Blake2 hash of the code itself should be enough. Beyond that it is the proposer's job to apply the code correctly (and OpenGov's to reject it if not applied correctly).

ggwpez:
> @ggwpez we need to remove the rococo bridge hub. This shouldn't be in this repo.

https://github.com/polkadot-fellows/runtimes/pull/16

.github/workflows/release.yml
```diff
@@ -0,0 +1,152 @@
...
+          key: srtool-target-${{ matrix.runtime[0] }}-${{ matrix.runtime[1] }}-${{ github.sha }}
+          restore-keys: |
+            srtool-target-${{ matrix.runtime[0] }}-${{ matrix.runtime[1] }}-
+            srtool-target-${{ matrix.runtime[0] }}-
+
+      - name: Build ${{ matrix.runtime[1] }}
+        id: srtool_build
+        uses: chevdor/srtool-actions@v0.7.0
+        with:
+          chain: ${{ matrix.runtime[1] }}
```
ggwpez, line 48: I think this is currently building with `--release` instead or `--production`.  
Is it possible to overwrite the profile here? Otherwise we could change the default in srtool.

gilescope: FYI I can't find anywhere where we're setting `WASM_BUILD_TYPE`. @chevdor and I think that the runtimes are currently built with the `--release` flag.

chevdor: `srtool` (the image) uses whatever is passed with the [PROFILE](https://github.com/paritytech/srtool/blob/master/scripts/build#L79) ENV or defaults to `release` indeed.

`srtool-actions` (the GHA) does not support passing the profile [yet](https://github.com/chevdor/srtool-actions/issues/18). As a result, `srtool`'s default will be used.

@gilescope started working on a [PR](https://github.com/chevdor/srtool-actions/pull/21) but that will have no effect if the wasm-builder does not support it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,152 @@
...
+
+            C "### $(WASM core_version.specName)"
+            C '```'
+            C "🏋️ Runtime Size:           $(numfmt --to iec-i --format "%.2f" $(WASM size)) ($(WASM size) bytes)"
+            C "🔥 Core Version:           $(WASM core_version.specName)-$(WASM core_version.specVersion) $(WASM core_version.implName)-$(WASM core_version.implVersion).tx$(WASM core_version.transactionVersion).au$(WASM core_version.authoringVersion)"
+            C "🎁 Metadata version:       V$(WASM metadata_version)"
+            C "🗳️ Blake2-256 hash:        $(WASM blake2_256)"
+            C "📦 IPFS:                   $(WASM ipfs_hash)"
+            C '```'
+          done
```
chevdor: Why not using a template such as `tera` as we usually do ?

That will make it more readable/maintainable with less bash-ery.
That makes it also be more robust and easier to debug. 

The templates already exist:
- https://github.com/paritytech/polkadot/blob/master/scripts/ci/changelog/templates/runtimes.md.tera
- https://github.com/paritytech/polkadot/blob/master/scripts/ci/changelog/templates/runtime.md.tera

You may want to tweak a few things but that should get you very close.

coderobe: I want to keep this quite self-contained. Personally I think the gazillion scripts and templates and generator dependencies and such in the polkadot repo are a pain to overview & maintain.

I'm trying to ensure that the bash-ery in this is kept to a minimum, if this ends up being more spaghetti than the polkadot templating I'm reworking it.

coderobe: I've just cleaned up the bash templating a little, it's just as readable as the tera templates now IMO

.github/workflows/release.yml
ggwpez: We should really fix this, since all the benchmarks are dont with `production` profile.  
This means that the weights for Polkadot/Kusama are currently all wrong.

ggwpez (approved):

.github/workflows/release.yml
bkchr: We also need to enable the following feature: https://github.com/paritytech/polkadot/blob/1a50cb1aede201e08df18d9deb501e32f833f40b/runtime/polkadot/Cargo.toml#L300

@coderobe could you please use `--profile production` and `--features on-chain-release-build`. We should pass this here explicitly to be sure.

bkchr:
@coderobe any update?

.github/workflows/release.yml
coderobe: @bkchr It now builds with the production profile. I don't know whether it's possible to configure that feature flag with srtool-actions, however.

ggwpez: Srtool should already use it by default https://github.com/paritytech/srtool/blob/a72f77ff200b3754034abb5aff9e2a25ea1c0f31/scripts/getBuildOpts.sh#L23

bkchr (approved):
Requires master merge to remove the rococo bridge hub. Otherwise looks good.

PR #17, opened 2023-08-08
Encointer Runtime

pifragile:
As Encointer is a system chain on Kusama, we suggest to include its runtime in this repo along the other system chains.


brenzi:
We suggest to use this PR to initiate the discussion about including Encointer in the fellowship repo. @joepetrowski suggested that this may be a good time, arguing that Polkadot is one "system" and that system chains, including Encointer, should be under the same roof.
So, even if this PR doesn't yet pass all the checks, please let us know what you think about the general idea

Encointer has a runtime upgrade ready to be submitted with the next system chain upgrade on Kusama (currently in testing on rococo). If possible, it would be released in this repo already. Otherwise we'd still like to join the next runtime upgrade whitelisting batch.

bkchr:
I think generally this is a good idea! 

How many of the encointer devs are part of the fellowship? It will also work without them being part of the fellowship, but having some idea what you approve (when it comes to future upgrades) sounds like a good idea.

brenzi:
currently no developers of encointer are in the fellowship. But if this will be merged, it sounds like a good idea to have a representation. Based on contributions to encointer, I guess 3 devs would qualify: @clangenb, @pifragile and myself. not sure if we need all 3, but

clangenb:
I would be happy to join the fellowship!

bkchr:
Fellowship is open to everyone, especially for getting into rank 1. I think the best would be that the ones of you that want to join to create some kind of list of work they have done and then reach out to me.

gilescope:
Should it be in a separate top level `common-good` dir rather than the `system-parachains` dir? (then again we define system chains now as anything with a para id of less than 2000)

brenzi:
No strong preferences here. 

*system chains*: sounds more like unopinionated necessities which would otherwise be part of the relaychain but are parachains for scalability reasons only (however, the discussion around a DEX on the asset hub shows that there won't be a clear line here)

Encointer is certainly opinionated, but like system chains, its utility is non-excludable, so the term common-good fits nicely.

On a more visionary note: Unlike current other system chains, Encointer would most certainly not be part of the relaychain otherwise. If Encointer weren't a parachain, it would more likely be an alternative relay-chain with personhood-based consensus than become part of Kusama or Polkadot relay chain

joepetrowski:
> Should it be in a separate top level `common-good` dir rather than the `system-parachains` dir? (then again we define system chains now as anything with a para id of less than 2000)

No, a parachain is either part of the Polkadot system (protocol) or it's not. No "common good" chains.

gilescope:
Ok `systems-parachains` it is.

brenzi:
After fellowship has accepted [RFC22](https://github.com/polkadot-fellows/RFCs/blob/main/text/0022-adopt-encointer-runtime.md) we will follow up on this PR soon:
* publish all dependency crates to crates.io
* update tomls here

pifragile:
Closing in favor of https://github.com/polkadot-fellows/runtimes/pull/80

pifragile:
Closing in favor of https://github.com/polkadot-fellows/runtimes/pull/80

PR #22, opened 2023-09-07
dummy chopsticks

samelamin:
We need a dummy github actions workflow file on main before we can run and/or make changes to it on a branch

See [comment](https://github.com/polkadot-fellows/runtimes/pull/19#issuecomment-1700480462)


samelamin:
Looks like new workflows are automatically run as part of the PR. In that case, ill close this and create the real workflow as a separate pr

Issue #23, opened 2023-09-08
Chopsticks + CI/CD


samelamin:
In addition to being the authors of [Chopsticks](https://github.com/AcalaNetwork/chopsticks), Acala also has a good e2e tests [repo](https://github.com/AcalaNetwork/e2e-tests) that is generic enough that all chains could use assuming they added the correct configs.

It would be good if we had a chopsticks test running on CI that validates that the new runtimes work as expected with the existing chain state 

I am planning on building a GitHub workflow that

1. Builds all the runtimes
2. Tests an XCM transfer between Kusama and assethub

Let me know if the above approach is unreasonable or if I should change it in any way

We can add more tests as we go along but just having that as a start would go a long way and hopefully help builders to ensure they test their changes against their live state and avoid future bricking of the chains

I _think_ using the e2e test repo is good if we want to ensure consistency across the board but if we want to only use chopsticks we can but the tests will be slightly harder to build

Is there a potential issue of using an external dependency like the e2e tests repo in CI? because we would be just as dependent on chopsticks unless the fellows fork it?

bkchr:
Somewhat a duplicate of: https://github.com/polkadot-fellows/runtimes/issues/171

PR #24, opened 2023-09-08
Add chopsticks workflow

samelamin:
This PR attempts to integrate chopsticks into our CI/CD process. See https://github.com/polkadot-fellows/runtimes/issues/23

 the goal of this is so we have a framework in place to ensure any and all commits are tested against a live chain state to catch any issues that other unit or integration tests might not catch.

Initially, I added a simple test to assert we can do xcm transfers between the relay chain and assethub and while this is a good first step, we can expand on this to cover all sorts of state transitions that might cause chains to brick,e.g. runtime upgrades, large migrations,etc

My only question/concern is that we might be tightly coupling our CI runs with the e2e test repo managed by acala but even if we break that dependency we are still dependent on chopsticks. Also, I do feel that having a defined best practice on testing chains might go a long way to stop and prevent chains from breaking in prod 

TLDR; chopsticks + CI/CD. 

Happy to hear feedback and change this approach to how people see fit


.github/workflows/chopsticks.yml
```diff
@@ -0,0 +1,51 @@
...
+          target: wasm32-unknown-unknown
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install e2e tests repo
+        run: git clone https://github.com/AcalaNetwork/e2e-tests && yarn --cwd e2e-tests/ && yarn --cwd e2e-tests/ update-env
+
+      - name: Build runtimes
+        run: cargo build --release --locked
```
ggwpez, line 43: Can probably restrict this to the runtimes with `-p`

.github/workflows/chopsticks.yml
```diff
@@ -0,0 +1,51 @@
...
+        with:
+          profile: minimal
+          toolchain: stable
+          target: wasm32-unknown-unknown
+
+      - name: Install nightly toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: nightly
```
ggwpez, line 33: Do we need nightly?

ggwpez (commented):
Very good start, thanks!

.github/workflows/chopsticks.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Build runtimes
+        run: cargo build --release --locked
+
+      - name: Run e2e tests
+        env:
+          KUSAMA_WASM: "../target/release/wbuild/kusama-runtime/kusama_runtime.compact.compressed.wasm"
+          POLKADOT_WASM: "../target/release/wbuild/polkadot-runtime/polkadot_runtime.compact.compressed.wasm"
+          STATEMINE_WASM: "../target/release/wbuild/asset-hub-kusama-runtime/asset_hub_kusama_runtime.compact.compressed.wasm"
+          STATEMINT_WASM: "../target/release/wbuild/asset-hub-polkadot-runtime/asset_hub_polkadot_runtime.compact.compressed.wasm"
+        run: find e2e-tests/tests/ -name '*.test.*' | xargs rm -rf && cp -r chopsticks/*  e2e-tests/tests/ && yarn --cwd e2e-tests test
```
ggwpez, line 51: What is it doing here with deletion? Can you please put a comment?

.github/workflows/chopsticks.yml
```diff
@@ -0,0 +1,45 @@
...
+
+      - name: Build runtimes
+        run: cargo build --release -p *-runtime --locked
+
+      - name: Run e2e tests
+        env:
+          KUSAMA_WASM: "../target/release/wbuild/kusama-runtime/kusama_runtime.compact.compressed.wasm"
+          POLKADOT_WASM: "../target/release/wbuild/polkadot-runtime/polkadot_runtime.compact.compressed.wasm"
+          STATEMINE_WASM: "../target/release/wbuild/asset-hub-kusama-runtime/asset_hub_kusama_runtime.compact.compressed.wasm"
+          STATEMINT_WASM: "../target/release/wbuild/asset-hub-polkadot-runtime/asset_hub_polkadot_runtime.compact.compressed.wasm"
```
joepetrowski, line 43: Should be Asset Hub throughout

chopsticks/xcm-transfer/shared.ts
```diff
@@ -0,0 +1,184 @@
...
+import type { TestType as KusamaAssetHubTestType } from './kusama-assethub.test'
+import type { TestType as PolkadotAssetHubTestType } from './polkadot-assethub.test'
+
+type TestType =
+  | KusamaAssetHubTestType
+  | PolkadotAssetHubTestType
+
+export default function buildTest(tests: ReadonlyArray<TestType>) {
+  for (const { from, to, test, name, ...opt } of tests) {
+    describe(`'${from}' -> '${to}' xcm transfer '${name}'`, async () => {
```
joepetrowski, line 16: "xcm transfer" is really vague; it doesn't say what it's actually testing.

chopsticks/xcm-transfer/kusama-assethub.test.ts
```diff
@@ -0,0 +1,25 @@
...
+import buildTest from './shared'
+
+const tests = [
+  {
+    from: 'kusama',
+    to: 'statemine',
+    name: 'KSM',
+    test: {
+      xcmPalletDown: {
+        tx: tx.xcmPallet.limitedTeleportAssets(kusama.ksm, 1e12, tx.xcmPallet.parachainV3(0, statemine.paraId)),
```
joepetrowski, line 15: This is not the function signature of `limitedTeleportAssets`. Where is this stuff defined?

.github/workflows/chopsticks.yml
samelamin: The e2e repo needs updating because its classed as Statemine/Statemint still there

chopsticks/xcm-transfer/shared.ts
samelamin: This is the shared file that all tests use, I can rename but the idea is to keep all xcm related chopsticks tests there

chopsticks/xcm-transfer/kusama-assethub.test.ts
samelamin: Here https://github.com/AcalaNetwork/e2e-tests/pull/108/files#diff-fe51f1149e20f2ecd06cde6383b42166885e45523430399b4f1cc022445292daR123

joepetrowski: This hardcodes in [stuff like `parents: 0`](https://github.com/AcalaNetwork/e2e-tests/pull/108/files#diff-fe51f1149e20f2ecd06cde6383b42166885e45523430399b4f1cc022445292daR130), so cannot be used to test teleporting between system parachains. I know this PR is just the CI/CD but it will need to cover a large set of test cases.

samelamin: I agree, very good point. I think we as a community can evolve the e2e test repo so it can handle far more test cases.

PR #31, opened 2023-09-18
add review-bot to require fellows as reviewers

### LINKED ISSUE #7: Merge/Approval rights for Fellows
Fellowship members should be getting approval/merge rights based on their rank like it is done on chain. So, to merge a pull request it will require X members to pass "the voting" to get the approval to merge. The list of fellowship members is maintained on chain, so it would introduce extra overhead to maintain the list on github as well. Like having teams for each rank and then having some CI check that succeeds when "the voting" has passed. As already said, the list and the associated rank is already on chain and we should use this directly in the CI. There exists the [pr-custom-review](https://github.com/paritytech/pr-custom-review) github action from Parity that already overs fine grained control on the approval process. This should be augmented with a script that is using smoldot to fetch the latest list of fellowship members, their ranks and their identities from the latest block. This will require that every fellowship member also provides their github account as part of their on chain identity. Then the CI job will be able to map from a github name to the fellowship rank to do "the voting".

Bullrich:
Created a Github Action that uses the [Review-Bot app](https://github.com/paritytech/review-bot) to require fellows to review pull requests before allowing the PR to be merged.

The user's information is fetched always from the chain after every event.

It looks in the fellows data for a field named GitHub and it extracts the handle from there. This resolves #7 (you can find more information about the request there)

This uses [`pull_request_target`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target) for the event, not `pull_request`. This is a security measure so that an attacker can't steal secrets by changing the workflow file

---

I added an example configuration for this repository so that we can tweak each individual requirements for the files. Feel free to suggest edits to fulfill your requirements.

- CI Files
  - All `.github/` files.
  - Requires 2 approvals of rank 6 or above
- Relay files
  - All files inside the `relay/kusama` or `relay/polkadot` folder with the exception of the files that end in `.adoc`.
  - Requires 2 approvals of rank 4 or above
- System Parachain Files
  - All the files inside `system/parachains` folder.
  - Requires 2 approvals of rank 2 or above
- General Files
  - All the files, with the exclusion of the files inside the other rules.
  - Requires 1 approval of rank 1 or above.

All these rules are of type `basic`. There [are more types of rules available](https://github.com/paritytech/review-bot/tree/v1.1.0#types) but it is intended for teams with different users, and because higher ranks also fulfill requirements for lower ranks, I thought it would be redundant to use those particular type of rules.

Side notes:
- We can move `.github/review-bot.yml` to a different file, there is [an option to move the location](https://github.com/paritytech/review-bot/tree/v1.1.0#inputs).

### DIFF
--- a/.github/review-bot.yml
+++ b/.github/review-bot.yml
@@ -4,7 +4,7 @@ rules:
       include:
         - ^\.github/.*
     type: basic
-    minFellowsRank: 6
+    minFellowsRank: 4
     min_approvals: 2
   - name: Relay and system files
     condition:
--- a/.github/review-bot.yml
+++ b/.github/review-bot.yml
@@ -23,13 +23,6 @@ rules:
     type: basic
     minFellowsRank: 2
     min_approvals: 2
-  - name: Target Files
-    condition:
-      include:
-        - ^target\/.*
-    type: basic
-    minFellowsRank: 4
-    min_approvals: 2
   - name: General Files
     condition:
       include:
--- a/.github/review-bot.yml
+++ b/.github/review-bot.yml
@@ -1,0 +1,44 @@
+rules:
+  - name: CI Files
+    condition:
+      include:
+        - ^\.github/.*
+    type: basic
+    minFellowsRank: 6
+    min_approvals: 2
+  - name: Relay files
+    condition:
+      include:
+        - ^relay\/kusama\/.*
+        - ^relay\/polkadot\/.*
+      exclude: 
+        - ^relay\/.+\.adoc$
+    type: basic
+    minFellowsRank: 4
+    min_approvals: 2
+  - name: System Parachain Files
+    condition:
+      include:
+        - ^system-parachains\/.*
+    type: basic
+    minFellowsRank: 2
+    min_approvals: 2
+  - name: Target Files
+    condition:
+      include:
+        - ^target\/.*
+    type: basic
+    minFellowsRank: 4
+    min_approvals: 2
+  - name: General Files
+    condition:
+      include:
+        - '.*'
+      exclude: 
+        - ^relay\/kusama\/.*
+        - ^relay\/polkadot\/.*
+        - ^\.github/.*
+        - ^system-parachains\/.*
+        - ^target\/.*
+    type: basic
+    minFellowsRank: 1
--- a/.github/review-bot.yml
+++ b/.github/review-bot.yml
@@ -12,6 +12,7 @@ rules:
         - ^relay\/kusama\/.*
         - ^relay\/polkadot\/.*
         - ^system-parachains\/.*
+        - ^CHANGELOG$
       exclude: 
         - ^relay\/.+\.adoc$
     type: fellows
--- a/.github/review-bot.yml
+++ b/.github/review-bot.yml
@@ -3,9 +3,9 @@ rules:
     condition:
       include:
         - ^\.github/.*
-    type: basic
-    minFellowsRank: 4
-    min_approvals: 2
+    type: fellows
+    minRank: 4
+    minApprovals: 2
   - name: Relay and system files
     condition:
       include:
@@ -14,9 +14,9 @@ rules:
         - ^system-parachains\/.*
       exclude: 
         - ^relay\/.+\.adoc$
-    type: basic
-    minFellowsRank: 3
-    min_approvals: 4
+    type: fellows
+    minRank: 3
+    minApprovals: 4
   - name: General Files
     condition:
       include:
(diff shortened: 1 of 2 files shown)


.github/review-bot.yml
```diff
@@ -0,0 +1,44 @@
...
+    minFellowsRank: 4
+    min_approvals: 2
+  - name: System Parachain Files
+    condition:
+      include:
+        - ^system-parachains\/.*
+    type: basic
+    minFellowsRank: 2
+    min_approvals: 2
+  - name: Target Files
```
xlc, line 26: there is no target files?

Bullrich: My bad, it was an ignored directory. I fixed it now.

.github/review-bot.yml
```diff
@@ -0,0 +1,37 @@
+rules:
+  - name: CI Files
+    condition:
+      include:
+        - ^\.github/.*
+    type: basic
+    minFellowsRank: 6
```
bkchr, line 7:
```suggestion
    minFellowsRank: 4
```

.github/review-bot.yml
```diff
@@ -0,0 +1,37 @@
...
+    condition:
+      include:
+        - ^relay\/kusama\/.*
+        - ^relay\/polkadot\/.*
+      exclude: 
+        - ^relay\/.+\.adoc$
+    type: basic
+    minFellowsRank: 4
+    min_approvals: 2
+  - name: System Parachain Files
```
bkchr, line 19: Should be merged with the relay chain

.github/review-bot.yml
```diff
@@ -0,0 +1,37 @@
...
+  - name: Relay files
+    condition:
+      include:
+        - ^relay\/kusama\/.*
+        - ^relay\/polkadot\/.*
+      exclude: 
+        - ^relay\/.+\.adoc$
+    type: basic
+    minFellowsRank: 4
+    min_approvals: 2
```
bkchr, line 18:
```suggestion
    minFellowsRank: 3
    min_approvals: 4
```

bkchr (commented):
In the future it would be nice to have some kind of "calculations" based on the rank. So that for example 4 rank two people get the same voting power as one rank 3 or whatever (this is just for illustration, not really put any thought into this).

.github/review-bot.yml
```diff
@@ -0,0 +1,37 @@
...
+  - name: System Parachain Files
+    condition:
+      include:
+        - ^system-parachains\/.*
+    type: basic
+    minFellowsRank: 2
+    min_approvals: 2
```
joepetrowski, line 25: In addition to what @bkchr, the standard to modify system parachain files should not be any lower than that of the Relay Chains.

.github/review-bot.yml
```diff
@@ -0,0 +1,37 @@
...
+      include:
+        - '.*'
+      exclude: 
+        - ^relay\/kusama\/.*
+        - ^relay\/polkadot\/.*
+        - ^\.github/.*
+        - ^system-parachains\/.*
+        - ^target\/.*
+    type: basic
+    minFellowsRank: 1
```
joepetrowski, line 37: `min_approvals`? I would also consider making the min rank be 3, but at least 2.
```suggestion
    minFellowsRank: 2
```

joepetrowski (commented):
This looks good. I thought for the release process we want to push a commit to `./CHANGELOG`? That should perhaps be a case and require approval from Fellows and above?

Bullrich:
> I thought for the release process we want to push a commit to `./CHANGELOG`? That should perhaps be a case and require approval from Fellows and above?

The last rule `General files` currently encapsulates the `CHANGELOG` file (it has `.*`) and it requires 1 review from rank `2` or above. Would this satisfy this requirement or would you recommend that I make a new rule to specifically capture this file?

mutantcornholio:
> In the future it would be nice to have some kind of "calculations" based on the rank. So that for example 4 rank two people get the same voting power as one rank 3 or whatever (this is just for illustration, not really put any thought into this).

TBH, I feel like this idea is a wrong turn that makes the bot more complex than it needs to be. Can we actually put some thought into it and provide usecases?

joepetrowski:
> > I thought for the release process we want to push a commit to `./CHANGELOG`? That should perhaps be a case and require approval from Fellows and above?
> 
> The last rule `General files` currently encapsulates the `CHANGELOG` file (it has `.*`) and it requires 1 review from rank `2` or above. Would this satisfy this requirement or would you recommend that I make a new rule to specifically capture this file?

IMO it should be a stronger hurdle than a single rank 2. Would recommend just putting it under the "Relay and system chains" requirements.

bkchr (approved):
For a first version I would say it is good enough. We can improve in the future.

joepetrowski (approved):

PR #34, opened 2023-09-19
Make System Parachains trusted Teleporters

NachoPal:
Make System Parachain trusted Teleporters of each other. System Parachains should be able to teleport their native asset between them.

Depends on: https://github.com/paritytech/polkadot-sdk/pull/1368


NachoPal:
bot fmt

joepetrowski (approved):
Should probably be on `oty-import-v1.1.0-rc2` as the base branch.

PR #37, opened 2023-09-22
Init StorageVersions (for `oty import v1.1.0 rc2`)

liamaharon:
Based on findings in https://github.com/polkadot-fellows/runtimes/pull/28#issuecomment-1730197874

- Initializes missing on-chain storage versions
- Applies Multisig migration to Asset Hub Kusama (note this migration for some reason was configured to return that it uses 100% of the block weight.. I will open a PR to fix that upstream)

### DIFF
--- a/system-parachains/collectives/collectives-polkadot/src/lib.rs
+++ b/system-parachains/collectives/collectives-polkadot/src/lib.rs
@@ -635,8 +635,6 @@ pub type UncheckedExtrinsic =
 /// All migrations executed on runtime upgrade as a nested tuple of types implementing
 /// `OnRuntimeUpgrade`. Included migrations must be idempotent.
 type Migrations = (
-	// v9420
-	import_kusama_fellowship::Migration<Runtime, FellowshipCollectiveInstance>,
 	// unreleased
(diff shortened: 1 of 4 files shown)


joepetrowski (approved):

system-parachains/collectives/collectives-polkadot/src/lib.rs
```diff
@@ -635,12 +632,55 @@ pub type UncheckedExtrinsic =
...
+
+impl frame_support::traits::OnRuntimeUpgrade for InitStorageVersions {
+	fn on_runtime_upgrade() -> Weight {
+		use frame_support::traits::{GetStorageVersion, StorageVersion};
+		use sp_runtime::traits::Saturating;
+
+		let mut writes = 0;
+
+		if Multisig::on_chain_storage_version() == StorageVersion::new(0) {
+			StorageVersion::new(1).put::<Multisig>();
```
bkchr, line 656:
```suggestion
			Multisig::current_storage_version().put::<Multisig>();
```

Plrease do this everywhere.

bkchr (changes requested):

bkchr (approved):
TY!

PR #46, opened 2023-09-28
Switch Collectives Slash Handler to Use Limited Teleport Assets

joepetrowski:
`teleport_assets` is marked [as deprecated](https://github.com/paritytech/polkadot-sdk/blob/fa4c672/polkadot/xcm/pallet-xcm/src/lib.rs#L789) and the message fails to execute on the Relay Chain due to incorrect weight. This migrates to `limited_teleport_assets` with an `Unlimited` weight limit.

I added the note to the CHANGELOG for the 1.0.0 release, as I don't think this release should be published without this change. I can create a 1.0.1 section instead if people want that.


xlc:
any regression tests?

joepetrowski:
> any regression tests?

There is already a test. It failing is what flagged the problem.

kianenigma (approved):
> This migrates to limited_teleport_assets with an Unlimited weight limit.



I presume there is no ramification to this as only trusted locations can teleport?

joepetrowski:
> > This migrates to limited_teleport_assets with an Unlimited weight limit.
> 
> I presume there is no ramification to this as only trusted locations can teleport?

Correct, it's also within system logic (called in a slash handler). This is not user-accessible.

KiChjang (approved):

ggwpez (approved):

bkchr (approved):

PR #52, opened 2023-09-29
Introduce Tinkernet multisig XCM configs to Kusama/Asset Hub Kusama

arrudagates:
This PR is a follow up of https://github.com/paritytech/polkadot/pull/7165, which was approved and merged, but then reverted due the usage of the `xcm-builder` crate to define the XCM configs, which are not generic and universal enough to fit within the scope of that crate. As a solution, this PR defines the configs directly in the two runtimes, which should solve the `xcm-builder` issue.

For a proper overview of the PR and it's goals, here is the description of the original PR:

> This PR introduces the XCM configs required for Saturn multisigs from Tinkernet to transact on Kusama/Rococo, this is done simply by providing implementations to convert the MultiLocation to both an AccountId and a Signed RuntimeOrigin.
> 
> For more information regarding why this approach of using a very specific derivation function and defining this function in the receiver chain's runtime is necessary for Saturn, and also general information about Saturn, you can check the discussion that was opened in the forum prior: https://forum.polkadot.network/t/saturn-xcm-multisig-integration-on-kusama-a-technical-discussion/2694
> 
> For a more high level overview of Saturn, check this article: https://invarch.medium.com/saturn-the-future-of-multi-party-ownership-ac7190f86a7b

Please feel free to discuss this in here or in the fellowship chat and thanks in advance for your time!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -441,6 +441,7 @@ dependencies = [
  "sp-consensus-aura",
  "sp-core",
  "sp-inherents",
+ "sp-io",
  "sp-offchain",
  "sp-runtime",
  "sp-session",
--- a/relay/kusama/src/xcm_config.rs
+++ b/relay/kusama/src/xcm_config.rs
@@ -21,19 +21,23 @@ use super::{
 	RuntimeCall, RuntimeEvent, RuntimeOrigin, StakingAdmin, TransactionByteFee, WeightToFee,
 	XcmPallet,
 };
+use core::marker::PhantomData;
 use frame_support::{
 	match_types, parameter_types,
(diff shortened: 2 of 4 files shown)


gavofyork:
This change should definitely be first mooted as a Fellowship RFC, then voted on by the Fellowship, and only when successful then can a PR be pushed.

arrudagates:
> This change should definitely be first mooted as a Fellowship RFC, then voted on by the Fellowship, and only when successful then can a PR be pushed.

Good to know, I'll write up an RFC proposing this!

arrudagates:
Closed since https://github.com/polkadot-fellows/RFCs/pull/34 was rewritten as a generic solution.

PR #57, opened 2023-10-06
Remove unused weight files

ggwpez:
Deleting the weight files of all pallets that were undeployed by the OpenGov migration.   
Kusama still had one duplicate file for the balances.


relay/kusama/src/lib.rs
```diff
@@ -309,7 +309,7 @@ impl pallet_balances::Config for Runtime {
 	type MaxLocks = MaxLocks;
 	type MaxReserves = MaxReserves;
 	type ReserveIdentifier = [u8; 8];
-	type WeightInfo = weights::pallet_balances::WeightInfo<Runtime>;
+	type WeightInfo = weights::pallet_balances_balances::WeightInfo<Runtime>;
```
bkchr, line 312: Why is this called balances balances?

bkchr (approved):

relay/kusama/src/lib.rs
ggwpez: Because there are two instances of the balances pallet; one for normal balances and one for NIS counter balances.  
But apparently it was forgotten to be changed after deploying the second instance.

bkontur: there is also possibility to setup nicer custom names for target weight files e.g.:
```
type WeightInfo = weights::pallet_balances_for_nis::WeightInfo<Runtime>;
...
type WeightInfo = weights::pallet_balances_native::WeightInfo<Runtime>;
```

if you tell me "what looks best", I could fix it here,
I've made the same for multiple pallet_assets instances here:
https://github.com/paritytech/cumulus/pull/2743/commits/116838170bc9e2977f3960aa841f02e1ccd1c8e8
https://github.com/paritytech/cumulus/pull/2743/commits/b63a8da1e63023dd239344f481717c3ca8fe0010

ggwpez: Thanks! `native` and `nis` sounds better indeed 👍

bkontur: @ggwpez I created this [PR](https://github.com/polkadot-fellows/runtimes/pull/58) for that.

Unrelated question, why only the kusama runtime has prefix `staging` [`name = "staging-kusama-runtime"`](https://github.com/polkadot-fellows/runtimes/blob/main/relay/kusama/Cargo.toml#L7) ?

bkchr: > Unrelated question, why only the kusama runtime has prefix `staging` [`name = "staging-kusama-runtime"`](https://github.com/polkadot-fellows/runtimes/blob/main/relay/kusama/Cargo.toml#L7) ?

Because we tried to put the stuff onto crates.io and someone else was already owning the crate name `kusama`. We should probably at some point come up with a better name.

ordian (approved):

joepetrowski (approved):

bkontur:
@NachoPal I think there is one more unused weight file: https://github.com/polkadot-fellows/runtimes/blob/oty-weight-cleanup/system-parachains/gluttons/glutton-kusama/src/weights/frame_system.rs should we remove it or add to `weights/mod.rs` and use it?

NachoPal:
> @NachoPal I think there is one more unused weight file: https://github.com/polkadot-fellows/runtimes/blob/oty-weight-cleanup/system-parachains/gluttons/glutton-kusama/src/weights/frame_system.rs should we remove it or add to `weights/mod.rs` and use it?

I think we can remove it

bkontur:
> > @NachoPal I think there is one more unused weight file: https://github.com/polkadot-fellows/runtimes/blob/oty-weight-cleanup/system-parachains/gluttons/glutton-kusama/src/weights/frame_system.rs should we remove it or add to `weights/mod.rs` and use it?
> 
> I think we can remove it

@NachoPal or what about setting [SystemWeightInfo = ()](https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/gluttons/glutton-kusama/src/lib.rs#L160)?

PR #61, opened 2023-10-16
Allow `freezes` and `holds`

liamaharon:
We need to enable freezes and holds now that pallets are migrating to use them.


bkchr (approved):

muharem (approved):

bkchr:
We already now have some values for freezes and holds on main.

bkontur:
`MaxHolds` was removed from `pallet_balances` and it is already merged in fellows's main and `MaxFreezes` will be refactored also: https://github.com/paritytech/polkadot-sdk/issues/2997

PR #64, opened 2023-10-17
upgraded review bot to v2.1.0

Bullrich:
Upgraded to version 2.1.0 which has paritytech/review-bot#94, a change in the logic of the action to overcome some problems with permissions coming from PRs from forks

For this, we needed to divide the actions into two files:
- A first action that triggers on PRs and reviews and uploads the PR number.
- A second action which is triggered under the completion of the first one and runs as the action normally runs (but won't have any problems regarding permissions because it is triggered from the master branch)

### DIFF
--- a/.github/workflows/review-bot.yml
+++ b/.github/workflows/review-bot.yml
@@ -1,23 +1,24 @@
-name: Review PR
+name: Review Bot
+
 on:
-  pull_request_target:
+  workflow_run:
+    workflows:
+      - Review-Trigger
     types:
-      - opened
-      - reopened
-      - synchronize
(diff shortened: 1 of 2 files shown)


bkchr (approved):

Bullrich:
> ### There was a problem running the action.
> ❌😵❌
> 
> Please find more information in the [logs](https://github.com/polkadot-fellows/runtimes/actions/runs/6548805677).

This is unexpected, but welcome.

It seems that a GitHub action that would modify a workflow file can not do it by default (which, for this particular case, makes sense to keep it).

For normal code it should work without any issues.

PR #71, opened 2023-10-22
Fix Asset Hub naming in code and release notes

liamaharon:
Cleans up some usage of old naming conventions.

I noticed also in the [last release](https://github.com/polkadot-fellows/runtimes/releases/tag/v1.0.0) the old naming conventions are used. I believe we should refer to these chains as Asset Hub now?
- Update Oct 24: Updated the release notes generation to replace "Statemint" and "Statemine" with "Polkadot Asset Hub (previous Statemint) and "Kusama Asset Hub (previously Statemine)". Release built with updated script: https://github.com/liamaharon/runtimes/releases/tag/v1.0.0


bkchr (approved):
The names are based on the `spec_name` found in the runtimes. We should rename the spec names.

system-parachains/asset-hubs/asset-hub-kusama/tests/tests.rs
```diff
@@ -15,7 +15,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Tests for the Statemine (Kusama Assets Hub) chain.
+//! Tests for the Kusama Assets Hub (previously known as Statemine) chain.
```
joepetrowski, line 18:
```suggestion
//! Tests for the Kusama Asset Hub (previously known as Statemine) chain.
```

system-parachains/asset-hubs/asset-hub-polkadot/tests/tests.rs
```diff
@@ -15,7 +15,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Tests for the Statemint (Polkadot Assets Hub) chain.
+//! Tests for the Polkadot Assets Hub (previously known as Statemint) chain.
```
joepetrowski, line 18:
```suggestion
//! Tests for the Polkadot Asset Hub (previously known as Statemint) chain.
```

joepetrowski (approved):

joepetrowski:
> The names are based on the `spec_name` found in the runtimes. We should rename the spec names.

Unfortunately a lot of wallets depend on this. I left [a note about it](https://github.com/polkadot-fellows/runtimes/blob/7bd91e826e4b2d95f6d2ee7f6bb0f0f47d095859/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L134) in the runtime. @tarikgul how much would break if we changed this now?

bkchr:
Then we need to adapt our release scripts.

TarikGul:
> > The names are based on the `spec_name` found in the runtimes. We should rename the spec names.
> 
> Unfortunately a lot of wallets depend on this. I left [a note about it](https://github.com/polkadot-fellows/runtimes/blob/7bd91e826e4b2d95f6d2ee7f6bb0f0f47d095859/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L134) in the runtime. @TarikGul how much would break if we changed this now?

I know many outside tooling is aware of the change and have been preparing. Even in our tooling we made sure to give support for whats to become legacy, and the new `spec_names`. 

> how much would break if we changed this now?

Within our tools (under integrations), nothing. Externally, it's hard to tell but I always expect a decent amount of users to be reactive as opposed to proactive, so I always expect some amount of breaking change for people that will require support.

My thoughts:

1. it "should" be a trivial change for users, 
2. This change has been known for a while so I would hope people have been preparing. 
3. If its in some type of release notes as a breaking change than its a fair change.

liamaharon:
@TarikGul @bkchr so consensus is to make that breaking `spec_name` change in this PR?

joepetrowski:
> @TarikGul @bkchr so consensus is to make that breaking `spec_name` change in this PR?

I'm OK with it.

bkchr:
If that breaks downstream users, we can also just fix up the release scripts. I don't really care and I also don't really know if someone is using this.

ggwpez:
>If that breaks downstream users, we can also just fix up the release scripts

IMHO this is better instead of just breaking stuff again.

Otherwise the only thing that comes to my mind is that the upgrades would have to applied with force through `System::set_code_without_checks`.

liamaharon:
In light of the suggestions to not make breaking changes, I have opted to not update the spec names but instead modify the release script to substitute "Statemint" and "Statemine" for "Asset Hub Polkadot" and "Asset Hub Kusama".

Tested this on my fork: https://github.com/liamaharon/runtimes/releases/tag/v1.0.0

bkchr (approved):
Ty!

muharem (approved):

bkontur (approved):

ggwpez (approved):

ordian (approved):

PR #78, opened 2023-10-25
Introduce chain spec generator

bkchr:
Until https://github.com/paritytech/polkadot-sdk/issues/25 is finished, we need a way to generate a chain spec for the runtimes. This pull request introduces a chain spec generator that can fulfill this task. It currently only supports the relay chain, but can be extended in the future to also support the system chains.

When https://github.com/paritytech/polkadot-sdk/issues/25 is delivered, we can remove the chain spec generator as it will not be required anymore.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+### Added
+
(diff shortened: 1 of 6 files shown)


pepoviola (approved):

NachoPal:
Extension for System Parachains in #81

NachoPal:
This PR and #81 are going to be merged?

bkchr:
> This PR and #81 are going to be merged?

Not really needed anymore with the new chain spec builder runtime api.

bkchr:
Yes!

Issue #20, opened 2023-08-28
Release 1.0 runtimes


bkchr:
The repository should now be prepared enough to create the first runtime release from this repo. Parity released all crates alongside releasing the 1.0 release of the node. As the 1.0 release is now already out for some time, I propose that we do the runtimes 1.0 release based on the next Parity Polkadot node release.

So, we first need to wait for this next node release to happen. After the release happened, someone can propose a pull request that updates to these crates on crates.io.

CC @ggwpez @coderobe

bkchr:
BTW, this time we don't want to copy the commits as before. From this point on this repo will work independently from the Parity repos when it comes to updating the runtimes. The runtimes in this repo are also the canonical ones.

bkchr:
> o just copying the changes instead of the full history?

Yes



> Replacing the dependencies to use crates-io instead of git? okay.

I mean I hope we get the latest release this week on crates.io, otherwise we need to use git dependencies.

acatangiu:
What’s the timeline here? When can we expect a new runtime release? 

Seeing how node [v1.1.0 is out](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.1.0) should we not release runtimes 1.1.0 directly (instead of v1.0)?

bkchr:
https://github.com/polkadot-fellows/runtimes/pull/28 it is being worked on here.

acatangiu:
#28 was merged. Can we do a release now? We need a runtime release for [BEEFY on Kusama](https://github.com/orgs/paritytech/projects/27/views/11?pane=issue&itemId=14196332).
Anything I can help with getting this over the finish line?

bkchr:
We need to wait for the staking changes runtime upgrade to be approved. After that is approved, we need to merge this fix here to make the repo canonical again. Then we can bring the release over the finish line. I think we can actually do that before the fix is enacted on chain, but we should be certain that it gets enacted on chain. 

Maybe we can then do the next release faster...

acatangiu:
Ok, thanks for the context! 👍

bkchr:
So, I would like to merge the following pr first: https://github.com/polkadot-fellows/runtimes/pull/63

After this pr is merged, we can update this pr https://github.com/polkadot-fellows/runtimes/pull/26 and then finally do the release.

joepetrowski:
Also required:

- [ ] https://github.com/polkadot-fellows/runtimes/pull/46
[...]

PR #86, opened 2023-11-11
Restore governance lock periods to 7 days in Polkadot

rphmeier:
This is a revert of https://github.com/paritytech/polkadot/commit/d73503cc2351ffec047a4912ecdb3e1eab5f46aa which has not been released.

Deploying without this change would lead to referendum participants and delegates having their tokens locked for 4 times longer than expected.

Participants in any referendums ending after the runtime upgrade is applied would be affected.
Even more of an issue is delegates, where all current delegates would be affected by 4 times longer lock up issues. The affected user-base is enormous, and there is little to no possibility of reaching all these users to get them to adapt in time.

Although a 7 Day lock multiplier wasn't the originally intended target for Polkadot, breaking user expectations with this change is an even larger issue.

### DIFF
--- a/relay/polkadot/src/governance/mod.rs
+++ b/relay/polkadot/src/governance/mod.rs
@@ -32,7 +32,7 @@ mod tracks;
 pub use tracks::TracksInfo;
 
 parameter_types! {
-	pub const VoteLockingPeriod: BlockNumber = prod_or_fast!(28 * DAYS, 1);
+	pub const VoteLockingPeriod: BlockNumber = 7 * DAYS;
 }
 
 impl pallet_conviction_voting::Config for Runtime {
--- a/relay/polkadot/src/governance/mod.rs
+++ b/relay/polkadot/src/governance/mod.rs
@@ -32,7 +32,7 @@ mod tracks;
 pub use tracks::TracksInfo;
 
 parameter_types! {
-	pub const VoteLockingPeriod: BlockNumber = 7 * DAYS;
+	pub const VoteLockingPeriod: BlockNumber = prod_or_fast!(7 * DAYS, 1);
 }
 
 impl pallet_conviction_voting::Config for Runtime {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [1.0.1] 22.10.2023
+
+### Changed
+
+- Restore governance lock periods to 7 days in Polkadot ([polkadot-fellows/runtimes#86](https://github.com/polkadot-fellows/runtimes/pull/86))
+
 ## [1.0.0] 22.10.2023
 
 ### Changed


ggwpez (approved):
>This is a revert of https://github.com/paritytech/polkadot/commit/d73503cc2351ffec047a4912ecdb3e1eab5f46aa which has not been released.

It is released, but the Polkadot runtime upgrade is already being cancelled due to XCM issues.

Prior investigation lead to this: https://github.com/polkadot-fellows/runtimes/issues/20#issuecomment-1768635824

xlc:
I see two possible migrations:
- Gradually increase the lock period from 7 days to 28 days over, say, a year. We could increase the value by, say, 3 days for every runtime upgrade, until it is upgraded to 28 days. 
- Have a migration to reduce existing convictions in such way that the conviction is reduced but the lock period remains the same or reduced. This ensures people won't unexpected have longer locked period. They will have reduced convictions in future votes but that's not too bad. They can just redelegate with the new convictions.

kianenigma:
gathering some data on this might be useful. For example, hoe many people are voting/delegating with large convictions such that this will imply months of added lock time. 

While we agreed upon releasing 1000000 runtime that extending locks is fine, I now find the second suggestion from @xlc to be the more conservative and therefore preferred choice. 

Either a user decides to lower their conviction, or extend their lock, but the default behavior should have been the former.

eskimor (approved):

bkchr:
> * Have a migration to reduce existing convictions in such way that the conviction is reduced but the lock period remains the same or reduced. This ensures people won't unexpected have longer locked period. They will have reduced convictions in future votes but that's not too bad. They can just redelegate with the new convictions.

I would do the same. I would propose that we let the runtime upgrade getting applied and then we do another one that adjusts everyones locking period that had voted before the runtime upgrade was applied, even when the referenda was finished after the runtime upgrade was applied.

rphmeier:
> then we do another one that adjusts everyones locking period that had voted before the runtime upgrade was applied

We would also need to do a migration which updates all delegation convictions for every delegation which had been put in place before the runtime upgrade was applied. Many users have already updated their convictions manually in anticipation of this.

bkchr:
> Many users have already updated their convictions manually in anticipation of this.

Reverting the change as you proposed here would then have the same result for them.

liamaharon (approved):

rphmeier:
> Reverting the change as you proposed here would then have the same result for them.

Not really, because the intention is still to increase the conviction locking periods (for example, to those outlined in https://github.com/polkadot-fellows/RFCs/pull/20#issuecomment-1673553108)

I argue that we should have clear communication about the exact changes to convictions and a plan for migrating delegations and votes before releasing code that would change the deployed status-quo

ggwpez:
For the follow up changes i propose this: https://github.com/paritytech/polkadot-sdk/issues/125#issuecomment-1807268641  
B it would still need a migration to correct the current behaviour.

relay/polkadot/src/governance/mod.rs
```diff
@@ -32,7 +32,7 @@ mod tracks;
 pub use tracks::TracksInfo;
 
 parameter_types! {
-	pub const VoteLockingPeriod: BlockNumber = prod_or_fast!(28 * DAYS, 1);
+	pub const VoteLockingPeriod: BlockNumber = 7 * DAYS;
```
bkchr, line 35:
```suggestion
	pub const VoteLockingPeriod: BlockNumber = prod_or_fast!(7 * Days, 1);
```

bkchr (approved):

ordian (approved):

xlc (approved):

PR #90, opened 2023-11-15
Enable async backing on Kusama

Sophia-Gold:
This enables async backing in the Kusama runtime. Note that for candidates to actually be backed asynchronously, a subsequent governance proposal changing two new parameters is needed.

### DIFF
--- a/relay/kusama/src/lib.rs
+++ b/relay/kusama/src/lib.rs
@@ -1910,7 +1910,7 @@ sp_api::impl_runtime_apis! {
 		}
 	}
 
-		#[api_version(7)]
+	#[api_version(7)]
 	impl primitives::runtime_api::ParachainHost<Block, Hash, BlockNumber> for Runtime {
 		fn validators() -> Vec<ValidatorId> {
 			parachains_runtime_api_impl::validators::<Runtime>()
--- a/relay/kusama/src/lib.rs
+++ b/relay/kusama/src/lib.rs
@@ -1910,6 +1910,7 @@ sp_api::impl_runtime_apis! {
 		}
 	}
 
+		#[api_version(7)]
 	impl primitives::runtime_api::ParachainHost<Block, Hash, BlockNumber> for Runtime {
 		fn validators() -> Vec<ValidatorId> {
 			parachains_runtime_api_impl::validators::<Runtime>()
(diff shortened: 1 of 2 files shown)


relay/kusama/src/lib.rs
```diff
@@ -2040,6 +2041,18 @@ sp_api::impl_runtime_apis! {
 				key_ownership_proof,
 			)
 		}
+
+			fn minimum_backing_votes() -> u32 {
+			parachains_runtime_api_impl::minimum_backing_votes::<Runtime>()
```
eskimor, line 2046: Formatting seems off.

eskimor (approved):
Looks good, although I think we should now also bump the version of the trait to (7) - at least once we have this on Polkadot as well. @tdimitrov ?

relay/kusama/src/lib.rs
Sophia-Gold: Done in 9f5cef432e41972579d394c66cee74d8da2043e1. Not sure why this wasn't caught by rustfmt.

relay/kusama/src/lib.rs
```diff
@@ -1910,6 +1910,7 @@ sp_api::impl_runtime_apis! {
 		}
 	}
 
+		#[api_version(7)]
```
tdimitrov, line 1913:
```suggestion
	#[api_version(7)]
```

tdimitrov:
> Looks good, although I think we should now also bump the version of the trait to (7) - at least once we have this on Polkadot as well. @tdimitrov ?

Yes, but it is already done?
https://github.com/polkadot-fellows/runtimes/pull/90/files#diff-efa4caeb17487ecb13d8f5eb7863c3241d84afa2e73fbf25909a2ca89df0f362R1913

Sophia-Gold:
> Looks good, although I think we should now also bump the version of the trait to (7) - at least once we have this on Polkadot as well. @tdimitrov ?

I assume you mean 8?

eskimor:
> Yes, but it is already done?

No. I meant the trait itself, but this can only be done once we also launched on Polkadot: So disregard.

ordian (approved):
Expect chaos.

joepetrowski (commented):
How about enabling system parachains to use it?

Sophia-Gold:
> How about enabling system parachains to use it?

They won't be able to make use of it until we propose an update to the new parameters after this is enacted. 

Glutton is already upgraded in Polkadot-SDK. I'll move that here. Can your team upgrade the others?

joepetrowski:
> Glutton is already upgraded in Polkadot-SDK. I'll move that here. Can your team upgrade the others?

Yes

PR #58, opened 2023-10-10
Change name of weight files for multiple `pallet_balances` instances of Kusama runtime.

bkontur:
Based on: https://github.com/polkadot-fellows/runtimes/pull/57

## Summary
This PR changes benchmarking setup for generating nicer output weight file names for multiple `pallet_balances` instance of Kusama runtime:
- `pallet_balances_balances.rs` -> `pallet_balances_native.rs`
- `pallet_balances_nis_counterpart_balances.rs` -> `pallet_balances_nis.rs`

## TODO
- [x] run benchmarking for Kusama runtime, just to be sure, that it works as expected (need to find a way how to run them)

PR #91, opened 2023-11-16
Enable `runtime-benchmarks` feature for chain spec generator

bkontur:
Based on: https://github.com/polkadot-fellows/runtimes/pull/78

Without this fix we cannot run benchmarks for kusama/polkadot.

Tested here: https://github.com/polkadot-fellows/runtimes/pull/58#issuecomment-1814123768


bkontur:
@bkchr I cannot push to `polkadot-fellows` directly, so I cannot push to your branch, but I would suggest if you could just cherry-pick this commit to your branch to avoid unnecessary begging for approves.

is it really necessary to require and follow "review" process for sub-PRs as for PRs to `main` branch?

bkchr (approved):

PR #99, opened 2023-11-21
Remove unused weight files

svyatonik:
It causes some confusion when I was working on weights for bridge pallets - e.g. there were two copies of `pallet_xcm_benchmarks_fungible.rs` and `pallet_xcm_benchmarks_generic.rs` and I wasn't sure - which one is actually used (until looking at `weights/mod.rs`).

Three's also a lot of unused weight files in `polkadot-runtime`, but they're referenced by `relay/polkadot/src/governance/old.rs`, which is also **not compiled**, so those files are unused too. But maybe there's some reason to keep this old governance code (and hence weights it is using), so I left those files. Could also remove on demand


bkontur:
there is another PR https://github.com/polkadot-fellows/runtimes/pull/57 what about rebasing this on the top of that?

svyatonik:
> there is another PR #57 what about rebasing this on the top of that?

:facepalm: Let's close this PR for now then - I can reopen later

PR #102, opened 2023-11-23
Fix ParentOrSiblings

franciscoaguirre:
We were accepting all type of sibling, not only sibling parachains.


acatangiu (approved):

xlc:
I am not going to block this PR but how do we know if this is doing the right thing? and we definitely need regression test somewhere. I won't require it to happen in this PR / this repo but someone have to make sure it happens.

muharem (approved):

muharem:
> I am not going to block this PR but how do we know if this is doing the right thing? and we definitely need regression test somewhere. I won't require it to happen in this PR / this repo but someone have to make sure it happens.

we have some tests with xcm-emulator in CI pipeline here https://github.com/paritytech/polkadot-sdk/pull/2566
and we have e2e tests which actually includes asserts for the xcm version updates via subscription (`ParentOrSiblings` type involved there). not yet ideal, but we have some and should continues to add more cases there.

bkontur (approved):

bkontur:
I would suggest to close this PR, because this will be fixed also here: https://github.com/polkadot-fellows/runtimes/pull/159/commits/44eefcca721585627ebda8dc87c7710abff24c12 as a part of 1.6.0 bump

PR #107, opened 2023-11-30
Fellowship entities get free xcm execution 

muharem:
Fellowship entities get free xcm execution.

Same as the Fellowship entities allowed to pass `UnpaidExecution` command to bypass the fees charged by `Trader` impl, we include the Fellowship locations to `WaivedLocations` to not be charged by `FeeManager` impl.


muharem:
> We really need to have good tests setup. I just don't like those PR changes important parameters and no tests to ensure it does the right thing.

For sure. I did test this with xcm-emulator in Westend env. I know @NachoPal is working on setting up an env for the runtimes in this env.

joepetrowski (approved):

acatangiu (approved):

ggwpez (approved):

bkchr (approved):

Issue #59, opened 2023-10-10
Move/copy runtime-common to fellowship repo


gpestana:
Some of the implementations under polkadot-sdk's `polkadot-runtime-common` crate are critical to Polkadot and Kusama (e.g. `fn era_payout`, which implements the inflation formula. This logic should probably be under the umbrella of the fellowship. Since the westend and rococo runtimes also need the runtime-common, another option is to keep a different runtime-common crate in each repo.

bkchr:
Yeah this requires some kind of clean-up. We probably should start by moving out the crates from `runtime-common` in the Parity repos. Then having a copy of this crate should be fine.

joepetrowski:
Should also include https://github.com/polkadot-fellows/runtimes/pull/109#discussion_r1420818182

bkontur:
relates to: https://github.com/paritytech/polkadot-sdk/issues/3054

PR #108, opened 2023-11-30
Added Polkadot <> Kusama bridge configuration

svyatonik:
This PR adds bridge configuration to allow exchanging assets between Polkadot and Kusama Asset Hubs. So far we only support sending KSM from Kusama Asset Hub to Polkadot Asset Hub as wKSM + DOT from Polkadot Asset Hub to Kusama Asset Hub as wDOT. Other assets may be added later.

The opposite direction is also supported, though it'll need dependencies upgrade to >= v1.4.

---

Similar Rococo <> Westend bridge is also active on live chains. You may read docs on how to play with it here: https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/runtimes/bridge-hubs/README.md

For Polkadot <> Kusama bridge we have a dedicated branch for local zombienet tests: https://github.com/svyatonik/runtimes/tree/sv-pk-bridge-with-zombienet-tests. You may run autotests (exchanging assets) there - see https://github.com/svyatonik/runtimes/blob/sv-pk-bridge-with-zombienet-tests/system-parachains/bridge-hubs/zombienet/README.md for details. You may do manual tests by spawning local chains using those scripts and use https://github.com/svyatonik/runtimes/blob/sv-pk-bridge-with-zombienet-tests/system-parachains/bridge-hubs/zombienet/scripts/bridges_kusama_polkadot.sh script to run various commands - i.e. send assets manually. Polkadot js apps is also supported.


system-parachains/asset-hubs/asset-hub-polkadot/src/weights/xcm/pallet_xcm_benchmarks_fungible.rs
```diff
@@ -1,44 +1,26 @@
...
+//! DATE: 2023-11-30, STEPS: `50`, REPEAT: `20`, LOW RANGE: `[]`, HIGH RANGE: `[]`
 //! WORST CASE MAP SIZE: `1000000`
-//! HOSTNAME: `runner-ynta1nyy-project-238-concurrent-0`, CPU: `Intel(R) Xeon(R) CPU @ 2.60GHz`
-//! EXECUTION: , WASM-EXECUTION: Compiled, CHAIN: Some("asset-hub-polkadot-dev"), DB CACHE: 1024
+//! HOSTNAME: `svyatonik-benchmarking`, CPU: `Intel(R) Xeon(R) CPU @ 2.80GHz`
+//! WASM-EXECUTION: Compiled, CHAIN: Some("ah-polkadot-local-raw.json"), DB CACHE: 1024
 
 // Executed Command:
-// ./target/production/polkadot-parachain
+// ../polkadot-sdk/target/release/polkadot-parachain-benchmarks
```
acatangiu, line 11: Weight differences are quite big..

I think you need to re-run benchmarks using binary built with `--profile production` - should use same `--profile production` for building WASM runtimes too

svyatonik: Yeah - I've noticed that too. Currently running it on ref hardware and will push along with updated costs

system-parachains/bridge-hubs/bridge-hub-kusama/primitives/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+	dispatch::DispatchClass,
+	sp_runtime::{MultiAddress, MultiSigner},
+};
+use sp_runtime::{FixedPointNumber, FixedU128, RuntimeDebug, Saturating};
+
+/// BridgeHubKusama parachain.
+#[derive(RuntimeDebug)]
+pub struct BridgeHubKusama;
+
+impl Chain for BridgeHubKusama {
```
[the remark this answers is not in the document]
svyatonik: This struct is also used from e.g. Polkadot BridgeHub. So we can't just add reference to Kusama BridgeHub runtime from there. That's why it is in a separate crate and uses its own definitions

system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs
```diff
@@ -658,3 +711,128 @@ where
...
+
+/// All configuration related to bridging
+pub mod bridging {
+	use super::*;
+	use assets_common::matching;
+	use sp_std::collections::btree_set::BTreeSet;
+	use xcm_builder::NetworkExportTableItem;
+
+	parameter_types! {
+		/// Base price of every byte of the Kusama -> Polkadot message. Can be adjusted via
```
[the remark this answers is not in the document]
svyatonik: That's a base price - fixed comment. Thanks

system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs
```diff
@@ -258,6 +262,17 @@ impl Contains<RuntimeCall> for SafeCallFilter {
 			}
 		}
 
+		// Allow to change dedicated storage items (called by governance-like)
+		match call {
+			RuntimeCall::System(frame_system::Call::set_storage { items })
+				if items.iter().all(|(k, _)| k.eq(&bridging::XcmBridgeHubRouterBaseFee::key())) ||
```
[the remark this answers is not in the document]
svyatonik: @bkontur Can we do that? I think we had some concerns about `.all(... || ...)` vs `.all() || .all()`? But I can't recall it now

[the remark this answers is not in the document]
svyatonik: Let's do it alltogether when everything here is ready. I've filed https://github.com/paritytech/parity-bridges-common/issues/2734 to track it

serban300 (approved):

system-parachains/asset-hubs/asset-hub-polkadot/tests/tests.rs
```diff
@@ -657,3 +670,600 @@ asset_test_utils::include_create_and_manage_foreign_assets_for_local_consensus_p
...
+			if let Some(new_value) = old_value.checked_add(1) {
+				new_value
+			} else {
+				old_value.checked_sub(1).unwrap()
+			}
+		},
+	)
+}
+
+// missing stuff from asset_test_utils::test_cases_over_bridge
```
acatangiu, line 820: nit: maybe add TODO here to remove with/when required deps update

acatangiu (approved):

joepetrowski (commented):
Looked at Asset Hub runtimes

system-parachains/asset-hubs/asset-hub-kusama/primitives/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+use codec::{Decode, Encode};
+use frame_support::weights::Weight;
+use scale_info::TypeInfo;
+use xcm::prelude::*;
+
+pub use bp_xcm_bridge_hub_router::XcmBridgeHubRouterCall;
+
+/// `AssetHubKusama` Runtime `Call` enum.
+///
+/// The enum represents a subset of possible `Call`s we can send to `AssetHubKusama` chain.
```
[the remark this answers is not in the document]
svyatonik: This `pallet-xcm-bridge-hub-router` (aka `ToPolkadotXcmRouter`) is deployed at the sending chain (Kusama AH here) and it is used to connect to the bridged chain consensus (Polkadot). That's why it is called `ToPolkadot`. So the usage of `ToPolkadot` name is imo justified.

Yet there's an ongoing work to add with-Ethereum pallet, so Kusama AH (or Polkadot AH - I'm not sure) will have another instance of this `pallet-xcm-bridge-hub-router`, which will connect it to the Ethereum consensus and will use the `ToEthereumXcmRouter` name (see eg. https://github.com/paritytech/polkadot-sdk/pull/2522/files#diff-a3b656e8095fe815cf8611c8319aff7b2f32482cdbe7c24ac37a2ce45235f32dR941-R942).

This enum is used to encode and dispatch calls **within** the same consensus. It is used by the **source** (Kusama) BH to say that the bridge is congested to the **source** (Kusama) AH. Then the `pallet-xcm-bridge-hub-router` will start increasing the fee, paid for the Kusama -> Polkadot messages. So it needs to use source chain indices: if message route is `Kusama.AH -> Kusama.BH -> Polkadot.BH -> Polkadot.AH`, it is used in `Kusama.AH <> Kusama.BH` communications.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -871,6 +902,9 @@ construct_runtime!(
 		Multisig: pallet_multisig::{Pallet, Call, Storage, Event<T>} = 41,
 		Proxy: pallet_proxy::{Pallet, Call, Storage, Event<T>} = 42,
 
+		// Bridge utilities.
+		ToPolkadotXcmRouter: pallet_xcm_bridge_hub_router::<Instance1>::{Pallet, Storage, Call} = 43,
```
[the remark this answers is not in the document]
svyatonik: Ok, changed from `43` to `34`

system-parachains/asset-hubs/asset-hub-polkadot/tests/tests.rs
svyatonik: Done

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
```diff
@@ -785,6 +816,9 @@ construct_runtime!(
 		Multisig: pallet_multisig::{Pallet, Call, Storage, Event<T>} = 41,
 		Proxy: pallet_proxy::{Pallet, Call, Storage, Event<T>} = 42,
 
+		// Bridge utilities.
+		ToKusamaXcmRouter: pallet_xcm_bridge_hub_router::<Instance1>::{Pallet, Storage, Call} = 43,
```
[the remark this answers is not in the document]
svyatonik: Ok, changed from `43` to `34`

system-parachains/asset-hubs/asset-hub-kusama/primitives/src/lib.rs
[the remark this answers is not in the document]
joepetrowski: OK I think I understand more. However, some things I still don't see w/r/t what needs to be kept in sync. E.g.:

```
/// All entries here (like pretty much in the entire file) must be kept in sync with
/// `AssetHubKusama` `construct_runtime`, so that we maintain SCALE-compatibility.
```

But you are in the AH Kusama runtime, so you can just use `PalletInfoAccess` to get the index. Do you mean that it needs to be in sync with AH Polkadot? Or that AH Polkadot relies on this index not changing? If you're relying on someone not changing an index down the road, perhaps add a test like [here](https://github.com/polkadot-fellows/runtimes/pull/109/commits/90f334bd2eb31055b3178b0acb6bd5133a7c7ec7)?

We're also in progress for moving `runtime-common` into the Fellowship (see https://github.com/polkadot-fellows/runtimes/issues/59). So perhaps declare some `const` pallet indices that runtimes can import (without of course importing the whole runtime).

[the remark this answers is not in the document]
svyatonik: > But you are in the AH Kusama runtime,

That's not true - we are not in the AH Kusama runtime, we are in a separate crate (`bp-asset-hub-kusama`), which is used to declare this call. Then this crate is used:
- in Kusama BH to encode an XCM message that is then sent to the Kusama AH;
- in Kusama AH to test that it is SCALE-compatible with the actual runtime call.

Previously we were declaring crates like that in bridges repo (which are then published to crates.io from `polkadot-sdk`). But it declares some constants (like this call and its weight) and if we need to change it (because e.g. call weight increases), then it'd be hard to do - we need to cross three repos and do a publish, before changes are actually applied. So we decided to go with a crates in this repo. So this PR adds 4 crates:
- `bp-asset-hub-*` to declare those calls and their weights;
- `bp-bridge-hub-*` to declare cost-related constants (there's also other stuff there, but it is unlikely will ever change).

joepetrowski (approved):

acatangiu:
@KiChjang @muharem please help with reviews of this PR which is mostly XCM and System Chains changes.

system-parachains/bridge-hubs/bridge-hub-kusama/primitives/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+
+/// Name of the With-BridgeHubKusama messages pallet instance that is deployed at bridged chains.
+pub const WITH_BRIDGE_HUB_KUSAMA_MESSAGES_PALLET_NAME: &str = "BridgeKusamaMessages";
+
+/// Name of the With-BridgeHubKusama bridge-relayers pallet instance that is deployed at bridged
+/// chains.
+pub const WITH_BRIDGE_HUB_KUSAMA_RELAYERS_PALLET_NAME: &str = "BridgeRelayers";
+
+/// Pallet index of `BridgePolkadotMessages: pallet_bridge_messages::<Instance1>`.
+pub const WITH_BRIDGE_KUSAMA_TO_POLKADOT_MESSAGES_PALLET_INDEX: u8 = 53;
```
muharem, line 83: these look like constants, not primitives. wont block PR because of this

system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs
```diff
@@ -258,6 +262,17 @@ impl Contains<RuntimeCall> for SafeCallFilter {
 			}
 		}
 
+		// Allow to change dedicated storage items (called by governance-like)
+		match call {
+			RuntimeCall::System(frame_system::Call::set_storage { items })
```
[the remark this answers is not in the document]
svyatonik: Only reads affect proof size and `set_storage` makes no any reads. Also we use `.iter().all()`, comparing with just two keys, so it won't make more than two writes, so this vec is kinda bounded.

system-parachains/bridge-hubs/bridge-hub-kusama/primitives/src/lib.rs
svyatonik: It declares some primitives as well - e.g. it reexports hashers, block numbers and etc. So it is both consts and primitives.

system-parachains/bridge-hubs/bridge-hub-kusama/primitives/src/lib.rs
```diff
@@ -0,0 +1,155 @@
...
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+//! Module with configuration which reflects BridgeHubKusama runtime setup (AccountId, Headers,
+//! Hashes...)
+
+#![cfg_attr(not(feature = "std"), no_std)]
```
bkchr, line 22:
```suggestion
```

Once is enough :P

system-parachains/bridge-hubs/bridge-hub-polkadot/primitives/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+	/// (initially was calculated by test `BridgeHubPolkadot::can_calculate_weight_for_paid_export_message_with_reserve_transfer` + `33%`)
+	pub const BridgeHubPolkadotBaseXcmFeeInDots: Balance = 4_858_960_000;
+
+	/// Transaction fee that is paid at the Polkadot BridgeHub for delivering single inbound message.
+	/// (initially was calculated by test `BridgeHubPolkadot::can_calculate_fee_for_complex_message_delivery_transaction` + `33%`)
+	pub const BridgeHubPolkadotBaseDeliveryFeeInDots: Balance = 16_954_899_613;
+
+	/// Transaction fee that is paid at the Polkadot BridgeHub for delivering single outbound message confirmation.
+	/// (initially was calculated by test `BridgeHubPolkadot::can_calculate_fee_for_complex_message_confirmation_transaction` + `33%`)
+	pub const BridgeHubPolkadotBaseConfirmationFeeInDots: Balance = 16_183_099_613;
```
bkchr, line 89: 1.6DOT? is that correct?

system-parachains/bridge-hubs/bridge-hub-polkadot/src/bridge_to_kusama_config.rs
```diff
@@ -0,0 +1,400 @@
...
+parameter_types! {
+	/// Number of Kusama headers to keep in the runtime storage.
+	///
+	/// Note that we are keeping only required header information, not the whole header itself. Roughly, it
+	/// is the 2 hours of real time (assuming that every header is submitted).
+	pub const RelayChainHeadersToKeep: u32 = 1_200;
+	/// Number of Kusama Bridge Hub headers to keep in the runtime storage.
+	///
+	/// Note that we are keeping only required header information, not the whole header itself. Roughly, it
+	/// is the 2 hours of real time (assuming that every header is submitted).
```
bkchr, line 103: But 1 hour are 600 blocks, so shouldn't this be 1200 for keeping 2 hours?

bkchr: Ahh, `RelayChainHeadersToKeep` is correct. I assume this still assumes that there are only a parachain block every 12 seconds? If yes, this is not correct anymore and we should update this as well to 1200.

system-parachains/bridge-hubs/bridge-hub-kusama/src/bridge_to_polkadot_config.rs
```diff
@@ -0,0 +1,404 @@
...
+	/// Number of Polkadot headers to keep in the runtime storage.
+	///
+	/// Note that we are keeping only required header information, not the whole header itself. Roughly, it
+	/// is the 2 hours of real time (assuming that every header is submitted).
+	pub const RelayChainHeadersToKeep: u32 = 1_200;
+	/// Number of Polkadot Bridge Hub headers to keep in the runtime storage.
+	///
+	/// Note that we are keeping only required header information, not the whole header itself. Roughly, it
+	/// is the 2 hours of real time (assuming that every header is submitted).
+	pub const ParachainHeadsToKeep: u32 = 600;
```
bkchr, line 104: Same comment as for Polkadot.

bkchr (approved):
Left some comments. Clearly did not read every detail of the pr, but looking good from what I can say.

bkontur:
> @svyatonik @acatangiu can you please merge master and fix the merge conflicts?

I think both are ooo, I will try

system-parachains/bridge-hubs/bridge-hub-kusama/primitives/src/lib.rs
bkontur: Just in case :D :D

system-parachains/bridge-hubs/bridge-hub-polkadot/primitives/src/lib.rs
bkchr: @bkontur any idea about this?

acatangiu: Yes this is the current cost

PR #115, opened 2023-12-08
Add kusama and polkadot SP constants

seadanda:
This PR brings the Kusama and Polkadot constants for System Parachains into the control of the Fellowship.

Following up on https://github.com/paritytech/polkadot-sdk/pull/2620 and paired with https://github.com/paritytech/polkadot-sdk/pull/2666.

Types are still used from the `parachains-common` crate, but constants are moved into a new crate `system-parachains-constants` in this repo.

The re-exports `pub use parachains_common as common` in both asset hubs and collectives have been removed. This is a breaking change for anybody using them in this way, but these should be directly imported from `parachains-common`. This should be noted on the changelog.


system-parachains/collectives/collectives-polkadot/src/lib.rs
```diff
@@ -77,15 +77,17 @@ use frame_system::{
 	limits::{BlockLength, BlockWeights},
 	EnsureRoot,
 };
+// TODO make combined mod to re-export
```
seadanda, line 80: Who are these re-exports used by (if anyone)? Is it an artifact from when the runtime code was directly used in `polkadot-parachain` or is this still used in something like RPCs?

There are two options here:
1. Make it clear in each parachain runtime exactly where the types and constants come from (as is done here) but break this API for anybody who uses e.g. `collectives_polkadot_runtime::common::polkadot`
2. Re-export the contents of `parachains-common` in `system-parachains-constants` and keep this `pub use`, but then all of the `use parachains_common::*` statements throughout the parachains will be moved too and it will be less clear whether they are defined in `system-parachains-constants` or `parachains-common`

system-parachains/constants/src/kusama.rs
```diff
@@ -0,0 +1,122 @@
...
+
+	pub const UNITS: Balance = 1_000_000_000_000;
+	pub const QUID: Balance = UNITS / 30;
+	pub const CENTS: Balance = QUID / 100;
+	pub const GRAND: Balance = QUID * 1_000;
+	pub const MILLICENTS: Balance = CENTS / 1_000;
+
+	pub const fn deposit(items: u32, bytes: u32) -> Balance {
+		// map to 1/100 of what the kusama relay chain charges (v9020)
+		(items as Balance * 2_000 * CENTS + (bytes as Balance) * 100 * MILLICENTS) / 100
```
joepetrowski, line 43: Is it possible to use `kusama_runtime_constants::currency::deposit(..) / 100` so we automatically track changes?

joepetrowski: Same for `EXISTENTIAL_DEPOSIT`

seadanda: Done for Kusama and Polkadot in [9736cb0](https://github.com/polkadot-fellows/runtimes/pull/115/commits/9736cb02b2da2034ab199557bdd7e2f4f3b27106).

bkontur (approved):

system-parachains/constants/Cargo.toml
```diff
@@ -0,0 +1,35 @@
+[package]
+name = "system-parachains-constants"
+repository.workspace = true
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+
```
ggwpez, line 8: Description please.

system-parachains/constants/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+frame-support = { default-features = false , version = "25.0.0" }
+sp-runtime = { default-features = false , version = "28.0.0" }
+
+# Polkadot
+polkadot-core-primitives = { default-features = false, version = "4.0.0"}
+polkadot-primitives = { default-features = false , version = "4.0.0" }
+polkadot-runtime-constants = { path = "../../relay/polkadot/constants", default-features = false}
+kusama-runtime-constants = { path = "../../relay/kusama/constants", default-features = false}
+
+# Cumulus
```
ggwpez, line 22: I think they can be in one list. Having these arbitrary separations is not helpful for finding thinks IMHO.

system-parachains/constants/src/kusama.rs
```diff
@@ -0,0 +1,122 @@
...
+	pub const BLOCK_PROCESSING_VELOCITY: u32 = 1;
+	/// Relay chain slot duration, in milliseconds.
+	pub const RELAY_CHAIN_SLOT_DURATION_MILLIS: u32 = 6000;
+}
+
+/// Constants relating to KSM.
+pub mod currency {
+	use polkadot_core_primitives::Balance;
+
+	/// The existential deposit.
```
ggwpez, line 32:
```suggestion
	/// The existential deposit. 1/10th of the relay deposit.
```

system-parachains/constants/src/kusama.rs
```diff
@@ -0,0 +1,122 @@
...
+	pub const EXISTENTIAL_DEPOSIT: Balance =
+		kusama_runtime_constants::currency::EXISTENTIAL_DEPOSIT / 10;
+
+	pub const UNITS: Balance = 1_000_000_000_000;
+	pub const QUID: Balance = UNITS / 30;
+	pub const CENTS: Balance = QUID / 100;
+	pub const GRAND: Balance = QUID * 1_000;
+	pub const MILLICENTS: Balance = CENTS / 1_000;
+
+	pub const fn deposit(items: u32, bytes: u32) -> Balance {
```
ggwpez, line 42:
```suggestion
	/// Deposit for stored data. 1/100th of the relay deposit.
	pub const fn deposit(items: u32, bytes: u32) -> Balance {
```

system-parachains/constants/src/lib.rs
```diff
@@ -0,0 +1,64 @@
...
+pub mod kusama;
+pub mod polkadot;
+
+pub use constants::*;
+
+/// re-export types from upstream
+pub use parachains_common::BlockNumber;
+
+/// Common constants of parachains.
+mod constants {
```
ggwpez, line 28: I think we should not name things by kind. Having a module named "enums" or "structs" would be equally weird - especially since its all re-exported anyway.

It could be ordered by topic (time, currency, fee...), however i think its not worth it since the file is so small.

system-parachains/constants/Cargo.toml
seadanda: Added in [6542acb](https://github.com/polkadot-fellows/runtimes/pull/115/commits/6542acb3e22f518882a150eac658f2618ae41f9e)

system-parachains/constants/src/lib.rs
seadanda: Happy to change, I have no attachment to how it was in `polkadot-sdk`

ggwpez (approved):

ascjones (approved):

joepetrowski (approved):

ggwpez:
@Bullrich the bot is not picking up the approvals here.

Bullrich:
> @Bullrich the bot is not picking up the approvals here.

Has something changed in the identity pallet? Some users who had an account now are null.

For example, when fetching @joepetrowski's account:
> Fetching identity of '15DCWHQknBjc5YPFoVj8Pn2KoqrqYywJJ95BYNYJ4Fj3NLqz', rank: 2
> Identity is null. Skipping

This used to work before and I haven't modified any of this code, so I first need to be aware of any changes so I can keep the code up to date.

Bullrich:
Seems that the problem is caused by subidentities. I have an issue for it: https://github.com/paritytech/review-bot/issues/107

Will put it as top priority and start working on it right now 🦾

xlc (approved):

bkchr (approved):

Issue #66, opened 2023-10-18
Improve release changelogs


bkchr:
Currently the release changelogs just consumes the entire `CHANGELOG.md` file. This isn't that great and we should only include the change of the particular release.

bkchr:
We should also mention the `srtool-cli` commands to reproduce the builds.

bkchr:
`srtool build --root --profile production --package collectives-polkadot-runtime --runtime-dir system-parachains/collectives/collectives-polkadot/ --build-opts="--features=on-chain-release-build"`

PR #117, opened 2023-12-08
Improve release changelogs

bkchr:
This improves the release by only adding the relevant parts of the CHANGELOG.md file. Besides that it also adds information on how to reproduce the build with srtool.

Closes: https://github.com/polkadot-fellows/runtimes/issues/66


.github/changelog-processor.py
```diff
@@ -26,21 +26,38 @@
     help="Should a release be made? Prints `1` or `0`.",
     action="store_true"
 )
+group.add_argument(
+    "--print-changelog-until-previous-version",
+    dest="changelog_until_previous",
+    help="Print the entire changelog until the previous version.",
```
joepetrowski, line 32: It doesn't really make sense to say "until" something in the past, but I see what you mean that the CHANGELOG file itself goes backwards in time. IMO `--print-changelog-from-last-release` expresses this better.

joepetrowski (approved):
One naming suggestion but still OK as it is.

.github/changelog-processor.py
bkchr: My naming is always shit :D

.github/changelog-processor.py
```diff
@@ -26,21 +26,38 @@
     help="Should a release be made? Prints `1` or `0`.",
     action="store_true"
 )
+group.add_argument(
+    "--print-changelog-until-previous-version",
```
bkchr, line 30:
```suggestion
    "--print-changelog-from-last-release",
```

.github/changelog-processor.py
```diff
@@ -26,21 +26,38 @@
...
+    dest="changelog_until_previous",
```
bkchr, line 31:
```suggestion
    dest="changelog_last_release",
```

.github/changelog-processor.py
bkchr, line 32:
```suggestion
    help="Print the changelog from the last release.",
```

.github/changelog-processor.py
```diff
@@ -26,21 +26,38 @@
...
     for line in lines:
         if not line.startswith("## ["):
+            changelog_until_previous_version += line
             continue
+        elif not found_last_version:
+            changelog_until_previous_version += line
+            found_last_version = True
+            version = line.strip().removeprefix("## [").split("]")[0]
+        else:
+            break
```
bkchr, line 54:
```suggestion
    changelog_last_release = ""
    found_last_version = False

    # Find the latest version
    for line in lines:
        if not line.startswith("## ["):
            changelog_last_release += line
            continue
        elif not found_last_version:
            changelog_last_release += line
            found_last_version = True
            version = line.strip().removeprefix("## [").split("]")[0]
        else:
            break
```

.github/changelog-processor.py
```diff
@@ -26,21 +26,38 @@
...
 
-        version = line.strip().removeprefix("## [").split("]")[0]
-        break
 
-    if args.print_latest_version:
+    if args.changelog_until_previous:
+        print(changelog_until_previous_version, end = "")
+        sys.exit(0)
```
bkchr, line 59:
```suggestion
    if args.changelog_last_release:
        print(changelog_last_release, end = "")
        sys.exit(0)
```

.github/workflows/release.yml
```diff
@@ -105,12 +105,21 @@ jobs:
           CONTEXT=$(find . -name '*_srtool_output.json')
           SRTOOL() { <$(<<<$CONTEXT head -n1) jq -r .$1; }
           WASM() { <${JSON} jq -r ".runtimes.compressed.subwasm.$1"; }
+
+          # Copy the relevant parts of the changelog
+          .github/changelog-processor.py CHANGELOG.md --print-changelog-until-previous-version > DRAFT
```
bkchr, line 110:
```suggestion
          .github/changelog-processor.py CHANGELOG.md --print-changelog-from-last-release > DRAFT
```

.github/changelog-processor.py
```diff
@@ -26,21 +26,38 @@
...
     for line in lines:
         if not line.startswith("## ["):
+            changelog_last_release += line
             continue
+        elif not found_last_version:
+            changelog_last_release += line
+            found_last_version = True
+            version = line.strip().removeprefix("## [").split("]")[0]
+        else:
+            break
```
muharem, line 54: how about not including a header from the changelog.
can be something like (tested locally) -

```suggestion
    # Find the latest version
    # Collect CHANGELOG from the latest version
    for line in lines:
        if found_last_version and line.startswith("## ["):
            break
        if not found_last_version and line.startswith("## ["):
            found_last_version = True
            version = line.strip().removeprefix("## [").split("]")[0]
        if found_last_version:
            changelog_last_release += line
```

muharem (approved):

.github/changelog-processor.py
muharem: probably even better to have that header in the release. just ignore this if you think so (:

bkchr: I would keep it for now and maybe change it in the future if it is not fitting.

muharem: I can not resolve the conversation for some reason

pepyakin (approved):

.github/workflows/release.yml
```diff
@@ -105,12 +105,21 @@ jobs:
           CONTEXT=$(find . -name '*_srtool_output.json')
           SRTOOL() { <$(<<<$CONTEXT head -n1) jq -r .$1; }
           WASM() { <${JSON} jq -r ".runtimes.compressed.subwasm.$1"; }
+
+          # Copy the relevant parts of the changelog
+          .github/changelog-processor.py CHANGELOG.md --print-changelog-from-last-release > DRAFT
```
ggwpez, line 110: I dont quite remember, but do we replace the `Unreleased` with the version number before calling this?  
Otherwise we have to replace it here.

.github/workflows/release.yml
```diff
@@ -105,12 +105,21 @@ jobs:
...
-          ## Runtime info
+          # Runtime info
           *These runtimes were built with **$(SRTOOL rustc)** using **[$(SRTOOL gen)](https://github.com/paritytech/srtool)***
+
+          To replicate the build, use the following command:
+          ```sh
+          srtool build \
+                 --root --profile production \
+                 --package CRATE_NAME --runtime-dir PATH_TO_CRATE \
+                 --build-opts="--features=on-chain-release-build"
```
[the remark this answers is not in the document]
bkchr: I want this to be explicit.

.github/workflows/release.yml
bkchr: Yes. When a pr is merged to master that replaces `Unreleased` with a version, this entire machinery here is started.

andresilva (approved):

PR #121, opened 2023-12-11
Fellowship Salary Budget & Period

muharem:
Fellowship Salary Budget & Period

In alignment with the Fellowship Salary [RFC](https://github.com/polkadot-fellows/RFCs/pull/50), the projected total monthly budget stands at 240,833 USDt.

With this PR, I invite the Fellowship members to refine the following constant parameters:
- RegistrationPeriod - # of blocks within a cycle which accounts have to register their intent to claim.
- PayoutPeriod - # of blocks within a cycle which accounts have to claim the payout.
- Budget - the total budget per cycle.

Please be aware that the pallet does not have a visibility into the actual balance of the account holding the salary budget. If this account lacks sufficient assets, subsequent salary claims will encounter failures.


joepetrowski (approved):

xlc (approved):

bkchr (approved):

system-parachains/collectives/collectives-polkadot/src/fellowship/mod.rs
```diff
@@ -251,7 +251,7 @@ impl pallet_salary::Config<FellowshipSalaryInstance> for Runtime {
 	// 15 days to claim the salary payment.
 	type PayoutPeriod = ConstU32<{ 15 * DAYS }>;
 	// Total monthly salary budget.
-	type Budget = ConstU128<{ 100_000 * USDT_UNITS }>;
+	type Budget = ConstU128<{ 250_000 * USDT_UNITS }>;
```
xlc, line 254: Actually, we will want to make this number configurable, as obviously we need to adjust the budget every once a while. This will be a good candidate to integrate with parameters pallet, whatever it is the one in [ORML](https://github.com/open-web3-stack/open-runtime-module-library/blob/master/parameters/src/lib.rs), or [polkadot-sdk](https://github.com/paritytech/polkadot-sdk/pull/2061)

bkontur: if we want very very quick win for this release:

```
parameter_types! {
     pub storage Budget: u128 = 250_000 * USDT_UNITS`;
}
```

```
type Budget = Budget
```

and allow it for `SafeCallFilter` [here](https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/collectives/collectives-polkadot/src/xcm_config.rs#L145)
```
                // Allow to change dedicated storage items (called by governance-like)
		match call {
			RuntimeCall::System(frame_system::Call::set_storage { items })
				if items.iter().all(|(k, _)| k.eq(&Budget::key()))) => return true,
			_ => (),
		};
```

xlc: I prefer just do the correct way. the storage parameters is not in metadata and hard to work with in frontend

eskimor (approved):

PR #130, opened 2023-12-22
Adds Snowbridge to Kusama and Polkadot Runtimes

claravanstaden:
- [x] Adds asset hub and bridge hub to Kusama and Polkadot runtimes
- [x] based off https://github.com/polkadot-fellows/runtimes/pull/108 - needs to be merged first
- [x] snowbridge crates needs to be published on crates.io - depends on `polkadot-sdk 1.7.0`
- [x] MessageQueue Pallet is not configured on Kusama and Polkadot asset hub - added in https://github.com/polkadot-fellows/runtimes/pull/137
- [x] `FeeManager` `handle_fee` misses reason, crate needs to be updated I think

cc @vgeddes @alistair-singh @yrong @musnit


bkontur:
There is bump to `polkadot@1.5.0` with MessageQueues [here](https://github.com/polkadot-fellows/runtimes/pull/137),
So, I would maybe suggest rebasing this PR on the current `main` and merging `#137` into this PR—it will bring us a little bit closer to the end :)

After #137 is merged, we just need to bump to `polkadot@1.6.0`.

claravanstaden:
Waiting on merging of:

- https://github.com/polkadot-fellows/runtimes/pull/159
- https://github.com/polkadot-fellows/runtimes/pull/187

Real diff can be viewed here: https://github.com/bkontur/runtimes/compare/bko-bump-to-1.7...Snowfork:runtimes:sv-pk-bridge-snowbridge

claravanstaden:
Fellows, may you please start with a preliminary review. This PR is dependent on https://github.com/polkadot-fellows/runtimes/pull/187 and the real diff can be viewed here: https://github.com/bkontur/runtimes/compare/bko-bump-to-1.7...Snowfork:runtimes:sv-pk-bridge-snowbridge

Outstanding:
- [ ] Merge https://github.com/polkadot-fellows/runtimes/pull/187
- [ ] Add Snowbridge integration tests once ☝🏻 the above integration tests pass

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -472,6 +508,179 @@ impl pallet_utility::Config for Runtime {
...
+		},
+		deneb: Fork {
+			version: [4, 0, 0, 0], // 0x04000000
+			epoch: 269568,
+		},
+	};
+}
+
+parameter_types! {
+	pub const MaxExecutionHeadersToKeep: u32 = prod_or_fast!(8192 * 2, 1000);
```
alistair-singh, line 660: Increase the number of headers as per
https://github.com/Snowfork/polkadot-sdk/pull/120

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
```diff
@@ -473,6 +509,179 @@ impl pallet_utility::Config for Runtime {
...
+		},
+		deneb: Fork {
+			version: [4, 0, 0, 0], // 0x04000000
+			epoch: 269568,
+		},
+	};
+}
+
+parameter_types! {
+	pub const MaxExecutionHeadersToKeep: u32 = prod_or_fast!(8192 * 2, 1000);
```
alistair-singh, line 661: Increase the number of headers as per
https://github.com/Snowfork/polkadot-sdk/pull/120

alistair-singh (changes requested):

claravanstaden:
TODO:

- [ ] Remove copied penpal-runtime once https://github.com/paritytech/polkadot-sdk/pull/3564 is merged
- [ ] Set Ethereum base fee to MAX (to disable the bridge) and set base fee to a sane amount in integration tests

☝🏻 This should not block reviewing to start. 😅 Both items are related to tests.

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -472,6 +507,179 @@ impl pallet_utility::Config for Runtime {
...
+parameter_types! {
+	pub storage EthereumGatewayAddress: H160 = H160(hex_literal::hex!("EDa338E4dC46038493b885327842fD3E301CaB39"));
+}
+
+parameter_types! {
+	pub const CreateAssetCall: [u8;2] = [53, 0];
+	pub const CreateAssetDeposit: u128 = (UNITS / 10) + EXISTENTIAL_DEPOSIT;
+	pub const InboundQueuePalletInstance: u8 = system_parachains_constants::kusama::snowbridge::INBOUND_QUEUE_PALLET_INDEX;
+	pub Parameters: PricingParameters<u128> = PricingParameters {
+		exchange_rate: FixedU128::from_rational(1, 400),
```
yrong, line 520: The price for eth-ksm could be referenced from https://www.kraken.com/convert/eth/ksm

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -472,6 +507,179 @@ impl pallet_utility::Config for Runtime {
...
+		fee_per_gas: gwei(20),
+		rewards: Rewards { local: 1 * UNITS, remote: meth(1) }
```
yrong, line 522: Is the local reward  too much here?

claravanstaden: Good catch @yrong! Updated here: https://github.com/polkadot-fellows/runtimes/pull/130/commits/d44fa3305087752131d651adf120a1390a222667

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -472,6 +507,179 @@ impl pallet_utility::Config for Runtime {
 	type WeightInfo = weights::pallet_utility::WeightInfo<Runtime>;
 }
 
+// Ethereum Bridge
+parameter_types! {
+	pub storage EthereumGatewayAddress: H160 = H160(hex_literal::hex!("EDa338E4dC46038493b885327842fD3E301CaB39"));
```
svyatonik, line 512: Perhaps very DQ, but I'm just starting here :) Gateway address is the same at Kusama and Polkadot BH. Is that correct? If we'll be using the same gateway contract, can the message be dispatched on both bridge hubs? Looking at the inbound pallet code:
```rust
			// submit message to verifier for verification
			T::Verifier::verify(&message.event_log, &message.proof)
				.map_err(|e| Error::<T>::Verification(e))?;

			// Decode event log into an Envelope
			let envelope =
				Envelope::try_from(&message.event_log).map_err(|_| Error::<T>::InvalidEnvelope)?;

			// Verify that the message was submitted from the known Gateway contract
			ensure!(T::GatewayAddress::get() == envelope.gateway, Error::<T>::InvalidGateway);

			// Retrieve the registered channel for this message
			let channel =
				T::ChannelLookup::lookup(envelope.channel_id).ok_or(Error::<T>::InvalidChannel)?;
```

To me it looks like the same message could pass all the checks if lookup will return a channel (and IIUC it willl, because some channel ids are static and others are derived based on para id, which is the same on Kusama and Polkadot).

system-parachains/bridge-hubs/bridge-hub-polkadot/tests/snowbridge.rs
svyatonik, line 1: Can we also add the `ethereum_outbound_queue_processes_messages_before_message_queue_works` test here?

system-parachains/bridge-hubs/bridge-hub-polkadot/src/xcm_config.rs
```diff
@@ -163,7 +169,11 @@ impl Contains<RuntimeCall> for SafeCallFilter {
 			RuntimeCall::System(frame_system::Call::set_storage { items })
 				if items.iter().all(|(k, _)| {
 					k.eq(&DeliveryRewardInBalance::key()) ||
-						k.eq(&RequiredStakeForStakeAndSlash::key())
+						k.eq(&RequiredStakeForStakeAndSlash::key()) ||
+						k.eq(&EthereumGatewayAddress::key()) ||
+						// Allow resetting of Ethereum nonces in Rococo only.
```
svyatonik, line 174: Rococo? Do we need it here?

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
yrong: Actually Gateway address will be different between Kusama and Polkadot. Check here:
```
// Verify that the message was submitted from the known Gateway contract
ensure!(T::GatewayAddress::get() == envelope.gateway, Error::<T>::InvalidGateway);
```

svyatonik: Ok, good to know that it should be different :) But now it is the same on both BHs. Is it going to be changed later, using governance calls? Then is it safe to set it before that in code to some `EDa338E4dC46038493b885327842fD3E301CaB39` (which looks like something copied from Rococo configuration). Maybe let's set it to some dummy value then? Or if you know gateway address in advance, then please fix it here

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -472,6 +507,179 @@ impl pallet_utility::Config for Runtime {
...
+}
+
+parameter_types! {
+	pub const CreateAssetCall: [u8;2] = [53, 0];
+	pub const CreateAssetDeposit: u128 = (UNITS / 10) + EXISTENTIAL_DEPOSIT;
+	pub const InboundQueuePalletInstance: u8 = system_parachains_constants::kusama::snowbridge::INBOUND_QUEUE_PALLET_INDEX;
+	pub Parameters: PricingParameters<u128> = PricingParameters {
```
svyatonik, line 519:
```suggestion
	pub Parameters: PricingParameters<Balance> = PricingParameters {
```

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
svyatonik: IIUC the local reward for single message is still `1 KSM + <length-dependent>`. Is it intended? I mean - `1 KSM` for every message seems too large to me. Or am I missing something?

svyatonik: Also: IIUC this value is correlated with the `BridgeHubEthereumBaseFee` at sibling asset hub. I mean - we expect sender to pay at least this reward to AH first and then sovereign account of AH to pay it to the relayer, right?

If so, can we please document it somewhere in the code? Just because someday someone may want to update pricing parameters here, at BH and forget to update base fee at the AH => AH will start paying to relayer out of his pocket. At least some code warning aside both values may add some protection against that in the future.

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -472,6 +507,179 @@ impl pallet_utility::Config for Runtime {
...
 }
 
+// Ethereum Bridge
+parameter_types! {
+	pub storage EthereumGatewayAddress: H160 = H160(hex_literal::hex!("EDa338E4dC46038493b885327842fD3E301CaB39"));
+}
+
+parameter_types! {
+	pub const CreateAssetCall: [u8;2] = [53, 0];
+	pub const CreateAssetDeposit: u128 = (UNITS / 10) + EXISTENTIAL_DEPOSIT;
```
svyatonik, line 517: Let's also have test for that. Things are changing over time and we'll definitely see issue with all those hardcoded values after they'll be deployed on-chain. Better to detect that by tests :)

svyatonik: I also suggest maybe to reuse our crates in this repo (`bp-asset-hub-*` and `bp-bridge-hub-*`) as a proper place to define constants that need to be referenced by both AH and BH. You could make a `mod snowbridge` there for the sake of cleanliness

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -472,6 +507,179 @@ impl pallet_utility::Config for Runtime {
...
+	type MessageConverter = MessageToXcm<
+		CreateAssetCall,
+		CreateAssetDeposit,
+		InboundQueuePalletInstance,
+		AccountId,
+		Balance,
+	>;
+	type WeightToFee = WeightToFee;
+	type LengthToFee = ConstantMultiplier<Balance, TransactionByteFee>;
+	type MaxMessageSize = ConstU32<2048>;
```
svyatonik, line 585: Unrelated question - is that enforced somewhere? In Rust code, I can see it is only used to compute `delivery_cost` (sent to eth side in `Command::SetPricingParameters`). I wonder if not checking it is fine - if we expect messages not much than `2k` bytes, won't it break anything (apart from mismatched payment) if eth side will send `8k` message?

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -472,6 +507,179 @@ impl pallet_utility::Config for Runtime {
...
+	type WeightInfo = weights::snowbridge_pallet_inbound_queue::WeightInfo<Runtime>;
+	type PricingParameters = EthereumSystem;
+	type AssetTransactor = <xcm_config::XcmConfig as xcm_executor::Config>::AssetTransactor;
+}
+
+impl snowbridge_pallet_outbound_queue::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type Hashing = Keccak256;
+	type MessageQueue = MessageQueue;
+	type Decimals = ConstU8<12>;
```
svyatonik, line 595: Don't know if that makes sense - maybe keeping fixed value here is better. But you could use something like that: `type Decimals = ConstU8<{ UNITS.ilog10() as  }>;`

system-parachains/bridge-hubs/bridge-hub-polkadot/tests/snowbridge.rs
claravanstaden: That test relies on helper method added in polkadot-sdk 1.8.0: https://github.com/paritytech/polkadot-sdk/blob/polkadot-v1.8.0/cumulus/parachains/runtimes/test-utils/src/lib.rs#L303 The test helper is also in 1.8.0, which I think is fine to copy over, but also copying the `run_to_block_with_finalize` method is hairy - requires importing extra crates and so on. Let me know what you think. :)

system-parachains/bridge-hubs/bridge-hub-polkadot/src/xcm_config.rs
claravanstaden: Good catch, removed in https://github.com/polkadot-fellows/runtimes/pull/130/commits/575155243e4f0d4b92af7907f231ac48baa901fb.

system-parachains/bridge-hubs/bridge-hub-polkadot/tests/snowbridge.rs
claravanstaden: Added in: https://github.com/polkadot-fellows/runtimes/pull/130/commits/cdf9e245b8c888d83a8a35150c0ab3d6d893091c

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -472,6 +507,179 @@ impl pallet_utility::Config for Runtime {
 	type WeightInfo = weights::pallet_utility::WeightInfo<Runtime>;
 }
 
+// Ethereum Bridge
+parameter_types! {
+	pub storage EthereumGatewayAddress: H160 = H160(hex_literal::hex!("EDa338E4dC46038493b885327842fD3E301CaB39"));
+}
+
+parameter_types! {
+	pub const CreateAssetCall: [u8;2] = [53, 0];
```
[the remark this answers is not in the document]
claravanstaden: Added test in https://github.com/polkadot-fellows/runtimes/pull/130/commits/6900ea26e02d20859761441f1582be8123808e0b.

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
claravanstaden: @vgeddes would it make sense for us to test the `CreateAssetDeposit` value? It is used in our fee calculation https://github.com/paritytech/polkadot-sdk/blob/master/bridges/snowbridge/primitives/router/src/inbound/mod.rs#L167

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
yrong: It's only used to estimate the average cost as basis to charge on Etherum side, no enforce when processing.

claravanstaden: I think this is OK because our message size is nearly constant (fixed message format). This might changed with `Transact` support, right @yrong?

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
claravanstaden: The value is clearer to me personally but I don't have a strong preference.

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
yrong: Exactly!

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
yrong: Maybe we need to reopen https://github.com/Snowfork/snowbridge/pull/1112?

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
vgeddes: Yeah, we should initialize the address to `H160::zero()` here.

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
vgeddes: The `PricingParameters::rewards::local` config field is used to calculate the combined refund+reward to pay relayers for delivering messages to BridgeHub from Ethereum. So its not associated with `BridgeHubEthereumBaseFee`.

On the other hand the `PricingParameters::rewards::remote` config field is used to calculate the combined refund+reward to pay relayers for delivering messages to Ethereum from BridgeHub, so it is associated with `BridgeHubEthereumBaseFee`.

Since the bridge is disabled on initial launch, none of these hardcoded pricing parameters need to be accurate. They will updated post launch via a governance proposal.

However, I suggest changing `PricingParameters::rewards::local` to `0.01 KSM`, which is probably what we will use anyway for the initial bridge activation proposal.

@svyatonik I'll add more comments in the runtime config and `PricingParameters` struct describing these fields and their relationships.

integration-tests/emulated/penpal/Cargo.toml
```diff
@@ -0,0 +1,192 @@
+[package]
+name = "penpal-runtime-dotsama"
+version = "0.14.0"
+authors = ["Anonymous"]
+description = "A parachain for communication back and forth with XCM of assets and uniques."
```
bkontur, line 5:
```suggestion
description = "TODO: remove `penpal-runtime-dotsama` with 1.3.0 (https://github.com/polkadot-fellows/runtimes/issues/186) - A parachain for communication back and forth with XCM of assets and uniques."
publish = false
```

integration-tests/emulated/tests/bridges/bridge-hub-kusama/Cargo.toml
```diff
@@ -8,8 +8,12 @@ description = "Bridge Hub Kusama runtime integration tests with xcm-emulator"
 publish = false
 
 [dependencies]
+codec = { package = "parity-scale-codec", version = "3.4.0", default-features = false }
```
bkontur, line 11:
```suggestion
codec = { package = "parity-scale-codec", version = "3.6.9" }
```

integration-tests/emulated/tests/bridges/bridge-hub-kusama/Cargo.toml
```diff
@@ -32,6 +36,17 @@ pallet-bridge-messages = { version = "0.8.0" }
...
 # Local
 bridge-hub-kusama-runtime = { path = "../../../../../system-parachains/bridge-hubs/bridge-hub-kusama" }
+asset-hub-kusama-runtime = { path = "../../../../../system-parachains/asset-hubs/asset-hub-kusama" }
 integration-tests-helpers = { path = "../../../helpers" }
 kusama-polkadot-system-emulated-network = { path = "../../../networks/kusama-polkadot-system" }
 kusama-system-emulated-network = { path = "../../../networks/kusama-system" }
+system-parachains-constants = { path = "../../../../../system-parachains/constants" }
+
+# Snowbridge
+snowbridge-beacon-primitives = { default-features = false , version = "0.1.0" }
```
bkontur, line 46:
```suggestion
snowbridge-beacon-primitives = { version = "0.1.0" }
```

integration-tests/emulated/tests/bridges/bridge-hub-polkadot/Cargo.toml
```diff
@@ -8,8 +8,12 @@ description = "Bridge Hub Polkadot runtime integration tests with xcm-emulator"
 publish = false
 
 [dependencies]
+codec = { package = "parity-scale-codec", version = "3.4.0", default-features = false }
```
bkontur, line 11:
```suggestion
codec = { package = "parity-scale-codec", version = "3.6.9" }
```

integration-tests/emulated/tests/bridges/bridge-hub-polkadot/Cargo.toml
```diff
@@ -32,5 +36,17 @@ pallet-bridge-messages = { version = "0.8.0" }
...
 # Runtimes
 bridge-hub-polkadot-runtime = { path = "../../../../../system-parachains/bridge-hubs/bridge-hub-polkadot" }
+asset-hub-polkadot-runtime = { path = "../../../../../system-parachains/asset-hubs/asset-hub-polkadot" }
 integration-tests-helpers = { path = "../../../helpers" }
 kusama-polkadot-system-emulated-network = { path = "../../../networks/kusama-polkadot-system" }
+polkadot-system-emulated-network = { path = "../../../networks/polkadot-system" }
+system-parachains-constants = { path = "../../../../../system-parachains/constants" }
+
+# Snowbridge
+snowbridge-beacon-primitives = { default-features = false , version = "0.1.0" }
```
bkontur, line 46:
```suggestion
snowbridge-beacon-primitives = { version = "0.1.0" }
```

integration-tests/emulated/tests/bridges/bridge-hub-kusama/src/tests/snowbridge.rs
```diff
@@ -0,0 +1,681 @@
...
+	EthereumInboundQueue::submit(
+		RuntimeOrigin::signed(BridgeHubKusamaSender::get()),
+		fixture.message,
+	)
+}
+
+/// Create an agent on Ethereum. An agent is a representation of an entity in the Polkadot
+/// ecosystem (like a parachain) on Ethereum.
+#[test]
+#[ignore]
```
bkontur, line 81: @claravanstaden why `#[ignore]`? (also in `polkadot-sdk`) Is there anything missing? If so, I would remove them and create separate issue to `polkadot-sdk` and fix them

integration-tests/emulated/tests/bridges/bridge-hub-kusama/src/tests/snowbridge.rs
```diff
@@ -0,0 +1,681 @@
...
+	AssetHubKusama::fund_accounts(vec![(AssetHubKusamaReceiver::get(), INITIAL_FUND)]);
+
+	BridgeHubKusama::execute_with(|| {
+		type RuntimeEvent = <BridgeHubKusama as Chain>::RuntimeEvent;
+
+		// Construct RegisterToken message and sent to inbound queue
+		send_inbound_message(make_register_token_message()).unwrap();
+
+		// Construct SendToken message and sent to inbound queue
+		send_inbound_message(make_send_token_message()).unwrap();
```
bkontur, line 380:
```suggestion
		// Construct RegisterToken message and sent to inbound queue
		assert_ok!(send_inbound_message(make_register_token_message()));

		// Construct SendToken message and sent to inbound queue
		assert_ok!(send_inbound_message(make_send_token_message()));
```

@claravanstaden please use `asset_ok!` everywhere for `send_inbound_message`

integration-tests/emulated/tests/bridges/bridge-hub-kusama/src/tests/snowbridge.rs
claravanstaden: Create agent and create channel is disabled for initial launch and just removed from the safe call filter in the runtimes. If you prefer I can remove these tests and add them again when we enable these operations.

bkontur: > ...when we enable these operations.

When this will happen? Does it depend on anything event, release or what?

I mean, "remove now and re-add when enabling those operations" is ok, but also leave them here is ok, but in that case, please add there some more info "why ignored"  + TODO-issue e.g. `#[ignore = "TODO:issue#XYZ blahblah"]`

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
claravanstaden: @svyatonik @vgeddes I decreased the local fee to `0.01` KSM: https://github.com/polkadot-fellows/runtimes/pull/130/commits/042aa1b6f3ea74bf7118b42ed86254f9acdd7f47#diff-da00a6055ce091ceb6769ed5cc0dc6527aca586de0394a60aa3d18c38ff89716R523

integration-tests/emulated/tests/bridges/bridge-hub-kusama/src/tests/snowbridge.rs
claravanstaden: I think it's more of a business decision, not a technical one since the operations are technically supported.

integration-tests/emulated/tests/bridges/bridge-hub-kusama/src/tests/snowbridge.rs
claravanstaden: Changed in https://github.com/polkadot-fellows/runtimes/pull/130/commits/fdea2730c3a36229f273e6c8f2d0393e03730206.

svyatonik (commented):
Looks good so far

system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs
```diff
@@ -676,10 +690,17 @@ pub mod bridging {
 			sp_std::vec::Vec::new().into_iter()
 			.chain(to_polkadot::BridgeTable::get())
 			.collect();
+
+		pub EthereumBridgeTable: sp_std::vec::Vec<NetworkExportTableItem> =
+			sp_std::vec::Vec::new().into_iter()
```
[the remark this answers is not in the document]
claravanstaden: Nice! Done in https://github.com/polkadot-fellows/runtimes/pull/130/commits/4a0d841746256bf7878f2c1fd64c4a2b868438f0.

system-parachains/constants/src/kusama.rs
```diff
@@ -133,3 +133,19 @@ pub mod fee {
 		<WeightToFee as frame_support::weights::WeightToFee>::weight_to_fee(weight)
 	}
 }
+
+pub mod snowbridge {
+	use frame_support::parameter_types;
+	use xcm::opaque::lts::NetworkId;
+
+	/// The pallet index of the Ethereum inbound queue pallet in the bridge hub runtime.
+	pub const INBOUND_QUEUE_PALLET_INDEX: u8 = 80;
```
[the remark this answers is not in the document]
claravanstaden: Added over here: https://github.com/polkadot-fellows/runtimes/pull/130/commits/c00fc1442d472949e0feeb1bf1c5f918f2b58d87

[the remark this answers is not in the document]
svyatonik: IMO it is better to have it in one of the `bp-*` crates, because it is used in AH <> BH communications and those crates are exactly for that. But I'm not forcing you to change that - maybe it is fine to leave it here

system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs
```diff
@@ -601,6 +611,9 @@ pub type XcmRouter = WithUniqueTopic<(
 	// Router which wraps and sends xcm to BridgeHub to be delivered to the Polkadot
 	// GlobalConsensus
 	ToPolkadotXcmRouter,
+	// Router which wraps and sends xcm to BridgeHub to be delivered to the Ethereum
+	// GlobalConsensus
+	SovereignPaidRemoteExporter<bridging::EthereumNetworkExportTable, XcmpQueue, UniversalLocation>,
```
bkontur, line 616:
```suggestion
	SovereignPaidRemoteExporter<
		xcm_builder::NetworkExportTable<to_ethereum::BridgeTable>,
		XcmpQueue,
		UniversalLocation
	>,
```

My original idea for extending more routings was to simply use `BridgeTable` and add a new `.chain()` function. However, I now realize I was mistaken (as it could potentially introduce a bug where we allow routing with the wrong router). I will prepare another PR to refactor `pub BridgeTable:`.

So, if you don't need `bridging::EthereumNetworkExportTable` elsewhere, I suggest removing it along with `pub EthereumBridgeTable` as they are not really needed.

system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs
```diff
@@ -684,10 +698,14 @@ pub mod bridging {
 			sp_std::vec::Vec::new().into_iter()
 			.chain(to_polkadot::BridgeTable::get())
 			.collect();
+
+		pub EthereumBridgeTable: sp_std::vec::Vec<NetworkExportTableItem> = to_ethereum::BridgeTable::get();
```
bkontur, line 702:
```suggestion
```

as suggested before, just to remove this (also, this location was meant to be common for all bridges and this is ETH specific)

system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs
```diff
@@ -684,10 +698,14 @@ pub mod bridging {
...
 	}
 
 	pub type NetworkExportTable = xcm_builder::NetworkExportTable<BridgeTable>;
 
+	pub type EthereumNetworkExportTable = xcm_builder::NetworkExportTable<EthereumBridgeTable>;
+
```
bkontur, line 708:
```suggestion
```

as suggested before, just to remove this

system-parachains/constants/src/kusama.rs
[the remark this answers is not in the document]
bkontur: @claravanstaden I agree with Slava, I would move `INBOUND_QUEUE_PALLET_INDEX` to the `bp-bridge-hub-kusama` (the same for BHP) and also move that sanity [test](https://github.com/polkadot-fellows/runtimes/commit/c00fc1442d472949e0feeb1bf1c5f918f2b58d87#diff-bd26e40b81351dde58a2a2ef23c118cf04731f18bcda5ce2961b8df25026a5f3R649) directly to the bridge-hub-kusama-runtime, we did this exactly on other places, e.g. [here](https://github.com/polkadot-fellows/runtimes/blob/main/relay/kusama/src/lib.rs#L2796-L2798) or [here](https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/bridge-hubs/bridge-hub-kusama/src/bridge_to_polkadot_config.rs#L427-L432)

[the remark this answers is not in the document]
bkontur: @claravanstaden again agree with Slava ;)
This `constants` module, as I see it, is primarily dedicated to very common shared items among all parachains. The ETH<>K/P bridge constants are specific to a particular use case rather than being common. If anyone needs to use them, they can import the `bp-*` primitives crate.

system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs
```diff
@@ -759,6 +777,55 @@ pub mod bridging {
...
+					PalletInstance(system_parachains_constants::kusama::snowbridge::INBOUND_QUEUE_PALLET_INDEX),
+				]
+			);
+
+			/// Set up exporters configuration.
+			/// `Option<MultiAsset>` represents static "base fee" which is used for total delivery fee calculation.
+			pub BridgeTable: sp_std::vec::Vec<NetworkExportTableItem> = sp_std::vec![
+				NetworkExportTableItem::new(
+					EthereumNetwork::get(),
+					Some(sp_std::vec![Junctions::Here]),
```
bkontur, line 802: @claravanstaden iirc, `Junctions::Here` is for `remote_location_filter`, so  e.g. for `pallet_xcm::reserve_transfer_assets` only destination `Location(2, [GlobalConsensus(Ethereum)]` will work. And if anything else is added after `GlobalConsensus` (e.g. `Location(2, [GlobalConsensus(Ethereum), AccountId(XYZ)]`) it won't be matched and allowed for routing.  
So do you expect here just ``Location(2, [GlobalConsensus(Ethereum)]`` or anything else?
If anything else, I would suggest to add here simple unit-tests (like this [one](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/xcm-builder/src/universal_exports.rs#L630)) to match several expected locations.

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -551,6 +561,9 @@ pub type XcmRouter = WithUniqueTopic<(
 	// Router which wraps and sends xcm to BridgeHub to be delivered to the Kusama
 	// GlobalConsensus
 	ToKusamaXcmRouter,
+	// Router which wraps and sends xcm to BridgeHub to be delivered to the Ethereum
+	// GlobalConsensus
+	SovereignPaidRemoteExporter<bridging::EthereumNetworkExportTable, XcmpQueue, UniversalLocation>,
```
bkontur, line 566:
```suggestion
	SovereignPaidRemoteExporter<
		xcm_builder::NetworkExportTable<to_ethereum::BridgeTable>,
		XcmpQueue,
		UniversalLocation
	>,
```

the same as for AHK

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -634,10 +648,14 @@ pub mod bridging {
 			sp_std::vec::Vec::new().into_iter()
 			.chain(to_kusama::BridgeTable::get())
 			.collect();
+
+		pub EthereumBridgeTable: sp_std::vec::Vec<NetworkExportTableItem> = to_ethereum::BridgeTable::get();
```
bkontur, line 652:
```suggestion
```

the same as for AHK

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -634,10 +648,14 @@ pub mod bridging {
...
 	}
 
 	pub type NetworkExportTable = xcm_builder::NetworkExportTable<BridgeTable>;
 
+	pub type EthereumNetworkExportTable = xcm_builder::NetworkExportTable<EthereumBridgeTable>;
+
```
bkontur, line 658:
```suggestion
```

the same as for AHK

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -709,6 +727,55 @@ pub mod bridging {
...
+					PalletInstance(system_parachains_constants::polkadot::snowbridge::INBOUND_QUEUE_PALLET_INDEX),
+				],
+			);
+
+			/// Set up exporters configuration.
+			/// `Option<MultiAsset>` represents static "base fee" which is used for total delivery fee calculation.
+			pub BridgeTable: sp_std::vec::Vec<NetworkExportTableItem> = sp_std::vec![
+				NetworkExportTableItem::new(
+					EthereumNetwork::get(),
+					Some(sp_std::vec![Junctions::Here]),
```
bkontur, line 752: @claravanstaden the same question as for AHK

system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs
claravanstaden: Removed in https://github.com/polkadot-fellows/runtimes/pull/130/commits/eed8cd9341914cb869328ea79d79cc9e52d27daf and https://github.com/polkadot-fellows/runtimes/pull/130/commits/3bc7a566730eff282f1a74ac07d370e12c0fe9e0.

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Feature for enabling debug prints in the Polkadot and Kusama runtime ([polkadot-fellows/runtimes#85](https://github.com/polkadot-fellows/runtimes/pull/85))
 - Added new "Wish for Change" track ([polkadot-fellows/runtimes#184](https://github.com/polkadot-fellows/runtimes/pull/184))
 - Enable Coretime and on-demand on Kusama ([polkadot-fellows/runtimes#159](https://github.com/polkadot-fellows/runtimes/pull/159)
+- Adds Snowbridge to Kusama and Polkadot ([polkadot-fellows/runtimes#130](https://github.com/polkadot-fellows/runtimes/pull/130)
```
joepetrowski, line 17:
```suggestion
- Adds Snowbridge to Kusama and Polkadot ([polkadot-fellows/runtimes#130](https://github.com/polkadot-fellows/runtimes/pull/130))
```

system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs
claravanstaden: > And if anything else is added after GlobalConsensus (e.g. Location(2, [GlobalConsensus(Ethereum), AccountId(XYZ)]) it won't be matched and allowed for routing.

Correct, this is the expected behaviour.

> So do you expect here just Location(2, [GlobalConsensus(Ethereum)] or anything else?

Just `Location(2, [GlobalConsensus(Ethereum)]` here. I added some tests anyway. 😄 

Tests added:
- [AssetHub Kusama](https://github.com/polkadot-fellows/runtimes/blob/c4c2995ba409ea739c6219994cb343be5bb5418d/system-parachains/asset-hubs/asset-hub-kusama/tests/snowbridge.rs)
- [AssetHub Polkadot ](https://github.com/polkadot-fellows/runtimes/blob/c4c2995ba409ea739c6219994cb343be5bb5418d/system-parachains/asset-hubs/asset-hub-polkadot/tests/snowbridge.rs)

svyatonik (approved):
LGTM, modulo remaining unresolved comments (this one in particular: https://github.com/polkadot-fellows/runtimes/pull/130#discussion_r1514466535).

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
```diff
@@ -473,6 +508,180 @@ impl pallet_utility::Config for Runtime {
...
 
+// Ethereum Bridge
+parameter_types! {
+	// The gateway address is set by governance.
+	pub storage EthereumGatewayAddress: H160 = H160::zero();
+}
+
+parameter_types! {
+	pub const CreateAssetCall: [u8;2] = [53, 0];
+	pub const CreateAssetDeposit: u128 = (UNITS / 10) + EXISTENTIAL_DEPOSIT;
```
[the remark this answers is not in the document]
vgeddes: good catch - we should make the `CreateAssetDeposit` match the ForeignAssets deposit exactly, ie. `100_000_000_000` (which looks like 10 DOT).

The deposit is some funds we teleport from BridgeHub to AssetHub to pay the cost of registering wrapped ERC20 tokens in the ForeignAssets pallet (an instance of `pallet_assets`.).

integration-tests/emulated/chains/parachains/bridges/bridge-hub-polkadot/src/genesis.rs
```diff
@@ -22,6 +22,7 @@ use emulated_integration_tests_common::{
 };
 use parachains_common::Balance;
 
+pub const ASSETHUB_PARA_ID: u32 = 1000;
```
joepetrowski, line 25: Just a preference, I would separate them:
```suggestion
pub const ASSET_HUB_PARA_ID: u32 = 1000;
```

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
joepetrowski, line 1: Just putting this in the runtime doesn't activate anything, correct? As in, nothing will happen and we can test it on Kusama for some time before making a separate motion to activate the Polkadot bridge?

system-parachains/bridge-hubs/bridge-hub-kusama/src/xcm_config.rs
```diff
@@ -342,3 +371,22 @@ fn treasury_pallet_account_not_none() {
 		LocationToAccountId::convert_location(&RelayTreasuryLocation::get()).unwrap()
 	)
 }
+
+pub struct XcmFeeManagerFromComponentsBridgeHub<WaivedLocations, HandleFee>(
```
joepetrowski, line 375: This type should have some documentation.

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
```diff
@@ -60,21 +69,29 @@ use frame_system::{
...
 
 #[cfg(any(feature = "std", test))]
 pub use sp_runtime::BuildStorage;
 
 // Polkadot imports
 use polkadot_runtime_common::{BlockHashCount, SlowAdjustingFeeUpdate};
+use polkadot_runtime_constants::{
+	currency::EXISTENTIAL_DEPOSIT,
+	system_parachain::{ASSET_HUB_ID, BRIDGE_HUB_ID},
+};
```
joepetrowski, line 86: We shouldn't be bringing this in. It's just confusing to have a `SYSTEM_PARA_EXISTENTIAL_DEPOSIT` and an `EXISTENTIAL_DEPOSIT`. Plus, it looks like you just used it because it happens to be equal to something convenient (1 DOT), which we have a type already for (`UNITS`).
```suggestion
use polkadot_runtime_constants::{
	system_parachain::{ASSET_HUB_ID, BRIDGE_HUB_ID},
};
```

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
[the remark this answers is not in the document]
joepetrowski: This has nothing to do with the Relay Chain's existential deposit though.

joepetrowski, line 519:
```suggestion
	pub const CreateAssetDeposit: u128 = 10 * UNITS;
```

system-parachains/bridge-hubs/bridge-hub-polkadot/src/xcm_config.rs
```diff
@@ -353,3 +382,22 @@ impl cumulus_pallet_xcm::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type XcmExecutor = XcmExecutor<XcmConfig>;
 }
+
+pub struct XcmFeeManagerFromComponentsBridgeHub<WaivedLocations, HandleFee>(
```
joepetrowski, line 386: docs

system-parachains/bridge-hubs/bridge-hub-polkadot/tests/snowbridge.rs
```diff
@@ -0,0 +1,399 @@
...
+fn construct_and_apply_extrinsic(
+	origin: sp_keyring::AccountKeyring,
+	call: RuntimeCall,
+) -> sp_runtime::DispatchOutcome {
+	let xt = construct_extrinsic(origin, call);
+	let r = Executive::apply_extrinsic(xt);
+	r.unwrap()
+}
+
+// TODO remove when Ethereum network ID has been extracted as a param
```
joepetrowski, line 245: Issue :)

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -60,20 +69,28 @@ use frame_system::{
...
+use xcm_config::XcmRouter;
+use xcm_config::{
+	FellowshipLocation, GovernanceLocation, TreasuryAccount, XcmOriginToTransactDispatchOrigin,
+};
 
 #[cfg(any(feature = "std", test))]
 pub use sp_runtime::BuildStorage;
 
+use kusama_runtime_constants::{
+	currency::EXISTENTIAL_DEPOSIT,
```
joepetrowski, line 82:
```suggestion
```

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -393,7 +393,10 @@ impl pallet_assets::Config<ForeignAssetsInstance> for Runtime {
 	type AssetIdParameter = xcm::v3::Location;
 	type Currency = Balances;
 	type CreateOrigin = ForeignCreators<
-		(FromSiblingParachain<parachain_info::Pallet<Runtime>, xcm::v3::Location>,),
+		(
+			FromSiblingParachain<parachain_info::Pallet<Runtime>, xcm::v3::Location>,
+			FromNetwork<xcm_config::UniversalLocation, EthereumNetwork, xcm::v3::Location>,
```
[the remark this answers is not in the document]
yrong: We change origin to `EthereumNetwork` before the transact
https://github.com/paritytech/polkadot-sdk/blob/ea458d0b95d819d31683a8a09ca7973ae10b49be/bridges/snowbridge/primitives/router/src/inbound/mod.rs#L186-L187

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
claravanstaden: Both the Kusama and Polkadot runtimes will be deployed disabled by default. From the Ethereum -> Polkadot side, the default operating mode in the Ethereum contracts will be `halted`. For the Polkadot -> Ethereum side, the transaction fee is set to max balance, effectively also disabling the bridge: https://github.com/polkadot-fellows/runtimes/blob/c4c2995ba409ea739c6219994cb343be5bb5418d/system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs#L787

On a related note, I do believe there is some discussion around the Kusama/Polkadot timing, since Kusama is much more expensive to run (due to the cost of running of the Beefy light client on Ethereum).

system-parachains/bridge-hubs/bridge-hub-kusama/tests/snowbridge.rs
```diff
@@ -0,0 +1,399 @@
...
+	snowbridge_runtime_test_common::ethereum_to_polkadot_message_extrinsics_work(
+		collator_session_keys(),
+		1013,
+		construct_and_apply_extrinsic,
+	);
+}
+
+#[test]
+fn ethereum_outbound_queue_processes_messages_before_message_queue_works() {
+	// TODO: add test after dependencies are upgraded to >= 1.8
```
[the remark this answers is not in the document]
claravanstaden: https://github.com/polkadot-fellows/runtimes/issues/234 😄

system-parachains/bridge-hubs/bridge-hub-polkadot/tests/snowbridge.rs
claravanstaden: https://github.com/polkadot-fellows/runtimes/issues/234 😄

system-parachains/bridge-hubs/bridge-hub-polkadot/src/xcm_config.rs
claravanstaden: Addressed in https://github.com/polkadot-fellows/runtimes/pull/130/commits/0c6e184afa829c55e242623fb923defff61dcf2c.

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
claravanstaden: Addressed in https://github.com/polkadot-fellows/runtimes/pull/130/commits/0c6e184afa829c55e242623fb923defff61dcf2c.

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
claravanstaden: Agreed, removed in https://github.com/polkadot-fellows/runtimes/pull/130/commits/0c6e184afa829c55e242623fb923defff61dcf2c.

integration-tests/emulated/chains/parachains/bridges/bridge-hub-polkadot/src/genesis.rs
claravanstaden: Addressed in https://github.com/polkadot-fellows/runtimes/pull/130/commits/0c6e184afa829c55e242623fb923defff61dcf2c.

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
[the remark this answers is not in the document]
claravanstaden: Agreed and removed.

system-parachains/constants/src/kusama.rs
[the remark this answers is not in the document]
claravanstaden: @svyatonik @bkontur I cleaned up the constants in https://github.com/polkadot-fellows/runtimes/pull/130/commits/0c6e184afa829c55e242623fb923defff61dcf2c. I did leave the tests in the integration tests instead of the runtime tests, because the constants are related to BH but needs to be compared to the AH runtime values. If the AH runtime constants are moved to `bp-asset-*` crates, I could move the tests to the runtime tests, but I didn't feel at liberty to start moving unrelated code around. Let me know your thoughts. :)

CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Refund any leases that are not migrated to Coretime (have holes in them/have not yet started) ([polkadot-fellows/runtimes#206](https://github.com/polkadot-fellows/runtimes/pull/206))
 - Enable Elastic Scaling node side feature for Kusama ([polkadot-fellows/runtimes#205](https://github.com/polkadot-fellows/runtimes/pull/205))
 - Cancel Parachain Auctions ([polkadot-fellows/runtimes#215](https://github.com/polkadot-fellows/runtimes/pull/215))
+- Enable Coretime and on-demand on Kusama ([polkadot-fellows/runtimes#159](https://github.com/polkadot-fellows/runtimes/pull/159)
```
joepetrowski, line 20: Error when merging main?
```suggestion
```

system-parachains/bridge-hubs/bridge-hub-polkadot/primitives/src/lib.rs
```diff
@@ -149,6 +149,29 @@ fn convert_from_uksm_to_udot(price_in_uksm: Balance) -> Balance {
...
+	use polkadot_runtime_constants::currency::UNITS;
+	use snowbridge_core::{gwei, meth, PricingParameters, Rewards};
+	use sp_runtime::FixedU128;
+
+	parameter_types! {
+		/// Should match the ForeignAssets::create index on AssetHub
+		pub const CreateAssetCall: [u8;2] = [53, 0];
+		/// Should match the AssetDeposit of the ForeignAssets pallet on AssetHub
+		pub const CreateAssetDeposit: u128 = 10 * UNITS;
+		/// The pallet index of the Ethereum inbound queue pallet in the BridgeHub runtime.
```
joepetrowski, line 164:
```suggestion
		/// Should match the `ForeignAssets::create` index on Asset Hub.
		pub const CreateAssetCall: [u8;2] = [53, 0];
		/// Should match the `AssetDeposit` of the `ForeignAssets` pallet on Asset Hub.
		pub const CreateAssetDeposit: u128 = 10 * UNITS;
		/// The pallet index of the Ethereum inbound queue pallet in the Bridge Hub runtime.
```

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
```diff
@@ -318,21 +341,31 @@ impl cumulus_pallet_parachain_system::Config for Runtime {
 impl parachain_info::Config for Runtime {}
 
 parameter_types! {
-	pub MessageQueueServiceWeight: Weight = Perbill::from_percent(35) * RuntimeBlockWeights::get().max_block;
+	/// Amount of weight that can be spent per block to service messages. This was increased
+	/// from 35% to 60% of the max block weight to accommodate the Ethereum beacon light client
+	/// extrinsics. The force_checkpoint and submit extrinsics (for submit, optionally) includes
+	/// the sync committee's pubkeys (512 x 48 bytes)
```
joepetrowski, line 347:
```suggestion
	/// extrinsics. The `force_checkpoint` and `submit` extrinsics (for submit, optionally) includes
	/// the sync committee's pubkeys (512 x 48 bytes).
```

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -317,21 +340,31 @@ impl cumulus_pallet_parachain_system::Config for Runtime {
 impl parachain_info::Config for Runtime {}
 
 parameter_types! {
-	pub MessageQueueServiceWeight: Weight = Perbill::from_percent(35) * RuntimeBlockWeights::get().max_block;
+	/// Amount of weight that can be spent per block to service messages. This was increased
+	/// from 35% to 60% of the max block weight to accommodate the Ethereum beacon light client
+	/// extrinsics. The force_checkpoint and submit extrinsics (for submit, optionally) includes
+	/// the sync committee's pubkeys (512 x 48 bytes)
```
joepetrowski, line 346:
```suggestion
	/// extrinsics. The `force_checkpoint` and `submit` extrinsics (for submit, optionally) includes
	/// the sync committee's pubkeys (512 x 48 bytes).
```

system-parachains/bridge-hubs/bridge-hub-kusama/primitives/src/lib.rs
```diff
@@ -158,6 +158,29 @@ fn convert_from_udot_to_uksm(price_in_udot: Balance) -> Balance {
...
+	use kusama_runtime_constants::currency::UNITS;
+	use snowbridge_core::{gwei, meth, PricingParameters, Rewards};
+	use sp_runtime::FixedU128;
+
+	parameter_types! {
+		/// Should match the ForeignAssets::create index on AssetHub
+		pub const CreateAssetCall: [u8;2] = [53, 0];
+		/// Should match the AssetDeposit of the ForeignAssets pallet on AssetHub
+		pub const CreateAssetDeposit: u128 = UNITS / 10;
+		/// The pallet index of the Ethereum inbound queue pallet in the BridgeHub runtime.
```
joepetrowski, line 173:
```suggestion
		/// Should match the `ForeignAssets::create` index on Asset Hub.
		pub const CreateAssetCall: [u8;2] = [53, 0];
		/// Should match the `AssetDeposit` of the `ForeignAssets` pallet on Asset Hub.
		pub const CreateAssetDeposit: u128 = UNITS / 10;
		/// The pallet index of the Ethereum inbound queue pallet in the Bridge Hub runtime.
```

joepetrowski (approved):
Just a few docs nits I noticed.

claravanstaden:
> Just a few docs nits I noticed.

Thanks for fixing! :)

integration-tests/emulated/chains/parachains/testing/penpal/Cargo.toml
```diff
@@ -17,7 +17,10 @@ frame-support = { version = "29.0.0" }
 parachains-common = { version = "8.0.0" }
 cumulus-primitives-core = { version = "0.8.0" }
 emulated-integration-tests-common = { version = "4.0.0" }
-penpal-runtime = { version = "0.15.0" }
+# TODO: remove `penpal-runtime-dotsama` with 1.3.0 (https://github.com/polkadot-fellows/runtimes/issues/186)
+# temp crate until EthereumNetwork is extracted from penpal definition
+penpal-runtime = { package="penpal-runtime-dotsama", path = "../../../../penpal" }
+#penpal-runtime = { version = "0.15.0" }
```
bkontur, line 23:
```suggestion
penpal-runtime = { version = "0.15.1" }
```

integration-tests/emulated/penpal/Cargo.toml
```diff
@@ -0,0 +1,193 @@
+[package]
+name = "penpal-runtime-dotsama"
+version = "0.14.0"
+authors = ["Anonymous"]
+description = "TODO: remove `penpal-runtime-dotsama` with 1.3.0 (https://github.com/polkadot-fellows/runtimes/issues/186) - A parachain for communication back and forth with XCM of assets and uniques."
```
bkontur, line 5: @claravanstaden 
you can remove this whole module now, and use patched `penpal-runtime@0.15.1` with `CustomizableAssetFromSystemAssetHub` as we did for roroco with `set_storage`

claravanstaden: Done. 😄

acatangiu:
> @acatangiu will you review this pr?

yes, multi-day review is in progress, will finish today or early tomorrow

system-parachains/bridge-hubs/bridge-hub-kusama/src/xcm_config.rs
```diff
@@ -342,3 +371,25 @@ fn treasury_pallet_account_not_none() {
...
+	PhantomData<(WaivedLocations, HandleFee)>,
+);
+impl<WaivedLocations: Contains<Location>, FeeHandler: HandleFee> FeeManager
+	for XcmFeeManagerFromComponentsBridgeHub<WaivedLocations, FeeHandler>
+{
+	fn is_waived(origin: Option<&Location>, fee_reason: FeeReason) -> bool {
+		let Some(loc) = origin else { return false };
+		if let Export { network, destination: Here } = fee_reason {
+			return !(network == EthereumNetwork::get())
```
acatangiu, line 387: should this not still check `WaivedLocations` when exporting to something else than Ethereum? aka, only short-circuit return false for Ethereum, but still check for all other cases.

(same for BHP)

acatangiu (approved):
modulo comments

system-parachains/bridge-hubs/bridge-hub-kusama/src/xcm_config.rs
claravanstaden: Yep, it is checked here 2 lines down: https://github.com/polkadot-fellows/runtimes/pull/130/files/bd2c69f2c31f289da11c41647deb1d23938c94a1#diff-9c700431fc38984083e2309b01d164c9b15bcbedd02a96210e73ea9265cd5f7eR389

So the behaviour is the same as `XcmFeeManagerFromComponents` except when exporting to Ethereum.

acatangiu: my point is that it's not equivalent, it shortcircuits and returns true or false without checking WaivedLocations.

e.g. `fee_reason = Export { network: Polkadot, destination: Here }` would just (incorrectly) return `true`

I was thinking changing this to smth like:

```rust
	fn is_waived(origin: Option<&Location>, fee_reason: FeeReason) -> bool {
		let Some(loc) = origin else { return false };
		if let Export { network, destination: Here } = fee_reason {
			if network == EthereumNetwork::get() {
				return false
			}
		}
		WaivedLocations::contains(loc)
	}
```

system-parachains/constants/src/kusama.rs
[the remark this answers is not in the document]
bkontur: @claravanstaden moved + small nits, please check/merge https://github.com/Snowfork/runtimes/pull/3

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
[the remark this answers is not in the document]
bkontur: also safer to use UNITS from `system_parachains_constants::kusama/polkadot::currency::UNITS` - fixed https://github.com/Snowfork/runtimes/pull/3

system-parachains/bridge-hubs/bridge-hub-kusama/src/xcm_config.rs
claravanstaden: Great catch, thanks @acatangiu. I changed it and added some unit tests to check the various permutations. https://github.com/polkadot-fellows/runtimes/pull/130/commits/336acb042e7aa6d0cb76681dfc25eb59333795b2

system-parachains/bridge-hubs/bridge-hub-polkadot/primitives/src/lib.rs
```diff
@@ -149,6 +149,29 @@ fn convert_from_uksm_to_udot(price_in_uksm: Balance) -> Balance {
...
+		pub const CreateAssetCall: [u8;2] = [53, 0];
+		/// Should match the `AssetDeposit` of the `ForeignAssets` pallet on Asset Hub.
+		pub const CreateAssetDeposit: u128 = 10 * UNITS;
+		/// The pallet index of the Ethereum inbound queue pallet in the Bridge Hub runtime.
+		pub const InboundQueuePalletInstance: u8 = 80;
+		/// Pricing parameters used for fees.
+		pub Parameters: PricingParameters<Balance> = PricingParameters {
+			exchange_rate: FixedU128::from_rational(1, 75),
+			fee_per_gas: gwei(20),
+			rewards: Rewards { local: 1 * UNITS / 100, remote: meth(1) } // 0.01 KSM
```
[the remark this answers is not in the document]
claravanstaden: Those values are actually just placeholders. The correct values will be set through governance. Made a change to better reflect this: https://github.com/polkadot-fellows/runtimes/pull/130/commits/82626eb868e90bd97721a784691f44ee6f77ace6

system-parachains/bridge-hubs/bridge-hub-kusama/src/xcm_config.rs
acatangiu: looks great, thanks for the good test coverage! 💪

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -128,6 +144,11 @@ pub type Migrations = (
 	cumulus_pallet_xcmp_queue::migration::v4::MigrationToV4<Runtime>,
 	// permanent
 	pallet_xcm::migration::MigrateToLatestXcmVersion<Runtime>,
+	snowbridge_pallet_system::migration::v0::InitializeOnUpgrade<
+		Runtime,
+		ConstU32<BRIDGE_HUB_ID>,
+		ConstU32<ASSET_HUB_ID>,
+	>,
```
bkontur, line 151: @claravanstaden Is this a one-time shot, or do we need it for every runtime upgrade? If it's the former, please move it before `// permanent`

system-parachains/bridge-hubs/bridge-hub-kusama/tests/tests.rs
```diff
@@ -326,6 +328,21 @@ pub fn can_calculate_weight_for_paid_export_message_with_reserve_transfer() {
 	)
 }
 
+#[test]
+fn change_ethereum_gateway_by_governance_works() {
```
bkontur, line 332: @claravanstaden very nit, why not to move it to the `tests/snowbridge.rs` also for BHP

bkontur (approved):
lgtm, just left few nits plus https://github.com/Snowfork/runtimes/pull/4

system-parachains/bridge-hubs/bridge-hub-kusama/tests/tests.rs
claravanstaden: https://github.com/polkadot-fellows/runtimes/pull/130/commits/756d00b6acb67908e866a258e2cc45c792ade930

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
claravanstaden: https://github.com/polkadot-fellows/runtimes/pull/130/commits/59b35027a6df62c6a6224714cd1a7f560dc96b26

system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs
```diff
@@ -780,6 +798,57 @@ pub mod bridging {
...
 
+	pub mod to_ethereum {
+		use super::*;
+		pub use bp_bridge_hub_kusama::snowbridge::EthereumNetwork;
+		use bp_bridge_hub_kusama::snowbridge::InboundQueuePalletInstance;
+
+		parameter_types! {
+			/// User fee for ERC20 token transfer back to Ethereum.
+			/// Configure the fee to max Balance so that it is disabled.
+			/// Sensible value was 2_750_872_500_000.
```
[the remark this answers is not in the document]
claravanstaden: https://github.com/polkadot-fellows/runtimes/pull/130/commits/b3d97cf5738902417c0bdecb4e926e10d593359d

acatangiu:
@claravanstaden I think this is ready to merge now, yes?

claravanstaden:
@acatangiu yes, thank you!

PR #131, opened 2023-12-31
Lower System Parachain Existential Deposits

joepetrowski:
As referenced in a few discussions (e.g. [forum](https://forum.polkadot.network/t/lowering-or-removing-existential-deposit-in-the-long-run/2167/10)), the Asset Hub Existential Deposit was set at genesis to 1/10th of the Relay Chain's. The goal however was always to lower it as the Relay Chain became more efficient at executing parachains to reflect the lower resource consumption on parachains and provide stronger incentives to prefer parachains for user-applications over the Relay Chain.

This PR lowers the system chain existential deposits by a factor of ten, to 1/100th of the Relay Chain's ED, or 0.01 DOT.

For discussion/comment: Should all system paras have a lower ED, or just Asset Hub? On one hand, it is less confusing if they are all the same. On the other, we actually _don't_ necessarily want to encourage lots of accounts on a chain like Bridge Hub. For this PR, I've adjusted all of the system paras, but in separate commits.

### DIFF
--- a/system-parachains/asset-hubs/asset-hub-kusama/src/constants.rs
+++ b/system-parachains/asset-hubs/asset-hub-kusama/src/constants.rs
@@ -17,8 +17,8 @@ pub mod currency {
 	use kusama_runtime_constants as constants;
 	use polkadot_core_primitives::Balance;
 
-	/// The existential deposit. Set to 1/10 of its parent Relay Chain.
-	pub const EXISTENTIAL_DEPOSIT: Balance = constants::currency::EXISTENTIAL_DEPOSIT / 10;
+	/// The existential deposit. Set to 1/100 of its parent Relay Chain.
+	pub const EXISTENTIAL_DEPOSIT: Balance = constants::currency::EXISTENTIAL_DEPOSIT / 100;
 
 	pub const UNITS: Balance = constants::currency::UNITS;
 	pub const CENTS: Balance = constants::currency::CENTS;
--- a/system-parachains/asset-hubs/asset-hub-kusama/src/weights/mod.rs
+++ b/system-parachains/asset-hubs/asset-hub-kusama/src/weights/mod.rs
@@ -39,5 +39,4 @@ pub mod xcm;
 
 pub use block_weights::constants::BlockExecutionWeight;
 pub use extrinsic_weights::constants::ExtrinsicBaseWeight;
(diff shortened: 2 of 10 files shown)


ggwpez:
>For discussion/comment: Should all system paras have a lower ED, or just Asset Hub? On one hand, it is less confusing if they are all the same. On the other, we actually don't necessarily want to encourage lots of accounts on a chain like Bridge Hub. For this PR, I've adjusted all of the system paras, but in separate commits.

Yea i think this is a good point. We can still reduce them if it should be a problem, but only reducing asset hub should already help.

joepetrowski:
> > For discussion/comment: Should all system paras have a lower ED, or just Asset Hub? On one hand, it is less confusing if they are all the same. On the other, we actually don't necessarily want to encourage lots of accounts on a chain like Bridge Hub. For this PR, I've adjusted all of the system paras, but in separate commits.
> 
> Yea i think this is a good point. We can still reduce them if it should be a problem, but only reducing asset hub should already help.

Yeah, I'm inclined to just do Asset Hub since that's where we want DOT, and having DOT elsewhere should come with some particular motivation. For UX, wallets/apps should be able to read the ED of the chain they are on and inform users correctly.

ggwpez (approved):

bkchr (approved):

adamsteeber:
Sorry if this isn't the best place for this or if it is too late being that it's merged already, but would it be possible to also decrease the `assets.approvalDeposit` to 0.01 DOT as part of this merge as well?

joepetrowski:
> Sorry if this isn't the best place for this or if it is too late being that it's merged already, but would it be possible to also decrease the `assets.approvalDeposit` on Asset Hub to 0.01 DOT as part of this merge as well?

The `ApprovalDeposit` is the existential deposit, so it's not only possible, it's done.
https://github.com/polkadot-fellows/runtimes/blob/v1.1.0/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L263

ggwpez:
Looks like this is missing a changelog entry and did therefore not appear in the release changelog.

joepetrowski:
Oops, adding one would trigger a new release though?

ggwpez:
I will just manually add it for now.

bkchr:
> Oops, adding one would trigger a new release though?

No. You can open a pr to add it to the CHANGELOG.

PR #56, opened 2023-10-04
upgrade to v1.2

muharem:
In the PR we bump versions for all dependacy crates from `polkadot-sdk` and apply all required changes

Changes:
- `parachains_runtime_api_impl` `v5` -> `v7`,
  - from https://github.com/paritytech/polkadot-sdk/pull/1543;
- `Consideration` setup for preimage pallet instances,
  - from  https://github.com/paritytech/polkadot-sdk/pull/1363;
- `experimental` feature removed for `pallet-society`, `pallet-xcm`, `runtime-common` crates,
  - from https://github.com/paritytech/polkadot-sdk/pull/1503;
- `pallet_election_provider_multi_phase::Config::SignedDepositBase` set to `GeometricDepositBase`,
  - from https://github.com/paritytech/polkadot-sdk/pull/1547;
- support `open_outbound_hrmp_channel_for_benchmarks_or_tests` func rename,
  - from https://github.com/paritytech/polkadot-sdk/pull/1422;
- missing weight functions added for `runtime_parachains_hrmp` and `pallet_preimage`;

TODO:
- [x] Deleted the Copy&paste para scheduler migration from https://github.com/polkadot-fellows/runtimes/pull/26 in b6666134d288374838fb4e47c7bdb0b9d92c0eed. This should be applied in the next runtime upgrade, but i did not properly check yet.

PR #133, opened 2024-01-04
CI: Enable try-runtime idempotency check

ggwpez:
Enable one more check in the CI to ensure that our migrations are idempotent. Can only merge after https://github.com/polkadot-fellows/runtimes/pull/118


bkchr (approved):

bkchr:
IMO it would cool if we could enable this check optionally based on if the pr wants to release a new version. See [here](https://github.com/polkadot-fellows/runtimes/blob/84e8bf1ba5e56a3ffd71d27479189c2d8c545bdf/.github/workflows/release.yml#L19) on how to detect if the changelog changes from unreleased to "released".

.github/workflows/check-migrations.yml
```diff
@@ -60,21 +60,43 @@ jobs:
...
+            onchain.snap
+          
+          # Check with pre-and-post
+          ./try-runtime \
+            --runtime $RUNTIME_BLOB_PATH \
+            on-runtime-upgrade --checks=pre-and-post \
+            $EXTRA_FLAGS $NO_WEIGHT_WARNINGS_FLAG \
+            snap -p onchain.snap
+        
+          # Optional: also check try-state invariants
```
bkchr, line 95: How is this optional? And why not just directly do it on the first run?

ggwpez: I am trying it out for now, could be that it fails and in that case i still wanted to see that the `pre-and-post` works.

bkchr: I think `try-runtime-cli` is already doing these checks separately.

ggwpez: yea maybe. I will remove it that we can merge. Seems like maybe we should merge your MR first since there a non-idempotent migration in polkadot.

s0me0ne-unkn0wn:
Not sure if it's the right place to raise this question, but it's close enough. Currently, #56 bumps `frame-system` to 24.0.0 in this release. But to ensure the idempotency of migrations requiring spec version checks (see paritytech/polkadot-sdk#2265) we need `frame-system` to be at least 27.0.0. I'm not familiar with this versioning, so I'm afraid to screw things up if I do that bump myself without any guidance (especially considering I've already broken everything locally when I tried). So what's the procedure for bumping FRAME and Substrate crate versions in runtimes?

@ordian do we need to push with `im-online` removal in this very release?

ggwpez:
> But to ensure the idempotency of migrations requiring spec version checks (see [paritytech/polkadot-sdk#2265](https://github.com/paritytech/polkadot-sdk/pull/2265)) we need `frame-system` to be at least 27.0.0

I guess we have to delete it manually this time?

s0me0ne-unkn0wn:
> I guess we have to delete it manually this time?

Not sure I follow, what do we delete manually?

ggwpez:
> > I guess we have to delete it manually this time?
> 
> Not sure I follow, what do we delete manually?

The migration i mean. We delete it after it executed on-chain.

s0me0ne-unkn0wn:
> The migration i mean. We delete it after it executed on-chain.

Ah yes. We can do that, sure. But that's not the only problem. Your PR enables mandatory idempotency checks. And without spec version checking API, my migration cannot be idempotent because the structure updated is not versioned. And that spec version checking API (`last_runtime_upgrade_spec_version()`) has been introduced in `frame-system` 27.0.0.

bkchr:
> @ordian do we need to push with `im-online` removal in this very release?

No. I already told you this in element. Just wait for the next release.

s0me0ne-unkn0wn:
> No. I already told you this in element. Just wait for the next release.

That's the elaboration of your answer I waited for! :smiley: Thanks, we're just postponing it then.

ggwpez:
Okay so we basically have to ignore the idempotency check until the next runtime upgrade on Polkadot... I will selectively disable it.

ordian (approved):

Issue #113, opened 2023-12-08
Convert Some Treasury Configs to `NeverEnsure` Once v1.5.0 is Released


joepetrowski:
Some of the `pallet_treasury` config items in the Polkadot Fellowship will require updating once `NeverEnsure` is available. These have been marked with links to this issue.

              we should have issues linking to todos to remind people there are future work to be done

_Originally posted by @xlc in https://github.com/polkadot-fellows/runtimes/pull/109#discussion_r1419790820_

PR #137, opened 2024-01-09
Upgrade to latest `polkadot-sdk@1.5` release

bkontur:
Attached result of `cargo upgrade -v --pinned --incompatible`  [cargo-upgrade-version-bump.log](https://github.com/polkadot-fellows/runtimes/files/13873056/cargo-upgrade-version-bump.log)

E.g.: `frame-support` from `25.0.0` to `27.0.0`

_Note: Encointer was not upgraded (because its pallet references `polkadot@1.3.0` release)._

## TODO

- [x] fix compilation
- [x] fix integration tests
- [x] fix benchmarks (also try them)
- [ ] patch for `pallet-nomination-pools` migration fix https://github.com/paritytech/polkadot-sdk/pull/3094
- [x] patch for `xcm-executor` fix https://github.com/paritytech/polkadot-sdk/pull/3174
  - [x] search for `TODO:(PR#137) - wait for xcm-executor patch` (missing `FrameTransactionalProcessor`)
- [x] check/verify `MaxExposurePageSize = 512` / `MaxNominators = 512` for Kusama/Polkadot
  - [x] check with SRLabs (Jakob) https://github.com/paritytech-secops/srlabs_findings/issues/327

Closes: https://github.com/polkadot-fellows/runtimes/issues/113


bkontur:
@bkchr @ggwpez guys, any idea what is wrong [here](https://github.com/polkadot-fellows/runtimes/actions/runs/7508298401/job/20443495670?pr=137)? 

```
error: failed to run custom build command for `asset-hub-kusama-runtime v1.0.0 (/home/runner/work/runtimes/runtimes/system-parachains/asset-hubs/asset-hub-kusama)`
Caused by:
  process didn't exit successfully: `/home/runner/work/runtimes/runtimes/target/release/build/asset-hub-kusama-runtime-c709c9b8e18d4c08/build-script-build` (exit status: 1)
  --- stderr
  Cannot compile the WASM runtime: no standard library sources found!
  You can install them with `rustup component add rust-src` if you're using `rustup`.
warning: build failed, waiting for other jobs to finish...
error: failed to run custom build command for `bridge-hub-kusama-runtime v1.0.0 (/home/runner/work/runtimes/runtimes/system-parachains/bridge-hubs/bridge-hub-kusama)`
Caused by:
  process didn't exit successfully: `/home/runner/work/runtimes/runtimes/target/release/build/bridge-hub-kusama-runtime-c9ab234e71cf4721/build-script-build` (exit status: 1)
  --- stderr
  Cannot compile the WASM runtime: no standard library sources found!
  You can install them with `rustup component add rust-src` if you're using `rustup`.
Error: Process completed with exit code 101.
```

I just  see that this error message is coming from [here](https://github.com/paritytech/polkadot-sdk/blob/polkadot-v1.5.0/substrate/utils/wasm-builder/src/prerequisites.rs#L149-L150).


I thought it is related to the [`Fetch cache`](https://github.com/polkadot-fellows/runtimes/actions/runs/7507624988/job/20441549625?pr=137#step:8:1) step error:
```
Warning: Error parsing Cargo.toml manifest, fallback to caching entire file: Error: Invalid TOML document: expected key-value, found comma

123:  serde_json = "1.0.111"
124:  remote-externalities = { package = "frame-remote-externalities" , version = "0.34.0" }
```
but it is not.

Locally it works (I know, I know...), maybe I missed something to port, will investigate Monday.

bkchr:
@bkontur change the stable toolchain installation to use the default profile. We need rust-src to be installed.

bkontur:
> @bkontur change the stable toolchain installation to use the default profile. We need rust-src to be installed.

@bkchr ok, cool, thank you, back in the game, just adding `rust-src` to stable components with minimal profile works.

ggwpez:
> I thought it is related to the [`Fetch cache`](https://github.com/polkadot-fellows/runtimes/actions/runs/7507624988/job/20441549625?pr=137#step:8:1) step error:

It should fix on the next update: https://github.com/Swatinem/rust-cache/pull/172#issuecomment-1883286183

relay/kusama/src/lib.rs
```diff
@@ -920,7 +923,8 @@ impl claims::Config for Runtime {
 parameter_types! {
 	// Minimum 100 bytes/KSM deposited (1 CENT/byte)
 	pub const BasicDeposit: Balance = 1000 * CENTS;       // 258 bytes on-chain
-	pub const FieldDeposit: Balance = 250 * CENTS;        // 66 bytes on-chain
+	// TODO:(PR#137) - check ByteDeposit?
+	pub const ByteDeposit: Balance = deposit(0, 1);
```
ggwpez, line 927: Correct by definition, but you can add a sanity check test if you want.

relay/kusama/src/lib.rs
```diff
@@ -2324,6 +2334,46 @@ sp_api::impl_runtime_apis! {
...
+						MultiAsset {
+							fun: Fungible(EXISTENTIAL_DEPOSIT),
+							id: Concrete(Here.into())
+						},
+						crate::Junction::Parachain(43211234).into(),
+					))
+				}
+
+				fn set_up_complex_asset_transfer(
+				) -> Option<(MultiAssets, u32, MultiLocation, Box<dyn FnOnce()>)> {
```
ggwpez, line 2363: Weird format?

relay/polkadot/src/lib.rs
```diff
@@ -667,7 +667,12 @@ parameter_types! {
 		"DOT_SLASH_DEFER_DURATION"
 	);
 	pub const RewardCurve: &'static PiecewiseLinear<'static> = &REWARD_CURVE;
-	pub const MaxNominatorRewardedPerValidator: u32 = 512;
+	// TODO:(PR#137) - check MaxExposurePageSize/MaxNominators 512?
+	pub const MaxExposurePageSize: u32 = 512;
+	// Note: this is not really correct as Max Nominators is (MaxExposurePageSize * page_count) but
+	// this is an unbounded number. We just set it to a reasonably high value, 1 full page
+	// of nominators.
+	pub const MaxNominators: u32 = 512;
```
ggwpez, line 675: @gpestana maybe you can check this?

system-parachains/bridge-hubs/bridge-hub-kusama/src/bridge_to_polkadot_config.rs
```diff
@@ -231,24 +242,31 @@ type FromPolkadotMessageBlobDispatcher = BridgeBlobDispatcher<
...
 	type UncongestedMessage = bp_asset_hub_kusama::UncongestedMessage;
 }
 
+/// Add support for the export and dispatch of XCM programs.
+pub type XcmOverBridgeHubPolkadotInstance = pallet_xcm_bridge_hub::Instance1;
+impl pallet_xcm_bridge_hub::Config<XcmOverBridgeHubPolkadotInstance> for Runtime {
+	type UniversalLocation = UniversalLocation;
+	type BridgedNetworkId = PolkadotGlobalConsensusNetwork;
+	type BridgeMessagesPalletInstance = WithBridgeHubPolkadotMessagesInstance;
+	type MessageExportPrice = ();
```
ggwpez, line 262: Dont know what this means, but why is the price zero?

ggwpez (approved):

relay/kusama/src/lib.rs
joepetrowski: @georgepisaltu can verify

bkontur:
@ggwpez should we add DMP migrations (`UndeployDmpQueue`, `DeleteDmpQueue`) from [here](https://github.com/paritytech/polkadot-sdk/pull/1246) also?

system-parachains/bridge-hubs/bridge-hub-kusama/src/bridge_to_polkadot_config.rs
bkontur: `MessageExportPrice` is just propagated to the [`xcm_builder::HaulBlobExporter`](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.5.0/polkadot/xcm/xcm-builder/src/universal_exports.rs#L425-L429) component [here](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.5.0/bridges/modules/xcm-bridge-hub/src/exporter.rs#L34-L37), and we don't need extra additional price here, because actual weights/price is covered by XCM `ExportMessage` instruction and  benchmark

relay/kusama/src/lib.rs
bkontur: well, I think the problem is that `cargo +nightly fmt` does not touch code inside `impl_runtime_apis! {` macro,
however, this code's format is based on the trait signature [here](https://github.com/paritytech/polkadot-sdk/blob/release-polkadot-v1.5.0/polkadot/xcm/pallet-xcm/src/benchmarking.rs#L74-L77), which was formatted by `cargo +nightly fmt`

ggwpez:
> @ggwpez should we add DMP migrations (`UndeployDmpQueue`, `DeleteDmpQueue`) from [here](https://github.com/paritytech/polkadot-sdk/pull/1246) also?

Its a two step process, first we do a runtime upgrade without them and then in the next one we add these migrations to delete the `DmpQueue`.  
It is also not critical if it happens later since the pallet is already defunc.

relay/polkadot/src/lib.rs
Ank4n: This configuration needs to be reviewed by Jakob from SR labs as well.
https://github.com/paritytech-secops/srlabs_findings/issues/327

Ank4n:
pallet-staking bump > 26 will introduce a paged exposure feature that will break some dapps if they have not integrated the changes. Is it possible to upgrade only Kusama and have it running only on Kusama for a while before upgrading Polkadot?

https://forum.polkadot.network/t/staking-update-paged-staking-reward-to-avoid-validator-oversubscription-issue-is-live-on-westend/5215

relay/kusama/src/lib.rs
georgepisaltu: It is the cost per byte stored, as can be seen [here](https://github.com/paritytech/polkadot-sdk/blob/13f2342edd70d83377d620385610bc8dc859b64b/substrate/frame/identity/src/lib.rs#L1240) and [here](https://github.com/paritytech/polkadot-sdk/blob/13f2342edd70d83377d620385610bc8dc859b64b/substrate/frame/identity/src/lib.rs#L1386).

joepetrowski (approved):

bkchr:
@bkontur what is the status of this pr? What are you waiting on?

bkontur:
> @bkontur what is the status of this pr? What are you waiting on?

@bkchr 
1. I think the only issue is this [PR](https://github.com/paritytech/polkadot-sdk/pull/2942) and backport/release it as patch version for `pallet-nomination-pools` above `polkadot@1.5.0-fix` tag.
2. Besides that, there are "problems/issues" with undecodable storage keys, but @liamaharon created separate issues.
3. And I found out that Encointer benchmarks does not work: https://github.com/polkadot-fellows/runtimes/issues/155  - but this is not a problem with this release, this is also problem for actual `main` branch.

So, overall, I am just waiting for  that `1.` .


One question, when should we bump runtime versions `1_001_000` -> `1_002_000` ? When we start working on new release or just at the end right before we do release? Should I bump it now?

relay/polkadot/src/lib.rs
bkontur: plus somebody should check  `MaxExposurePageSize` also

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -460,7 +493,9 @@ construct_runtime!(
 		XcmpQueue: cumulus_pallet_xcmp_queue = 30,
 		PolkadotXcm: pallet_xcm = 31,
 		CumulusXcm: cumulus_pallet_xcm = 32,
+		// Temporary to migrate the remaining DMP messages:
 		DmpQueue: cumulus_pallet_dmp_queue = 33,
+		MessageQueue: pallet_message_queue = 34,
```
claravanstaden, line 498: Thanks for adding the MessageQueue pallet, @bkontur! Can you please move this pallet to be at index 250? We rely on the MessageQueue pallet to be after the Snowbridge pallets. More details here: https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/runtimes/bridge-hubs/bridge-hub-rococo/src/lib.rs#L731

joepetrowski: This has [come up before](https://github.com/paritytech/polkadot-sdk/pull/1378#issuecomment-1733250552), there are really two indices here, the `= 34` index, which specifies the pallet index in call encoding, and the order of the variants in `construct_runtime`, which specifies the order of `on_initialize` running.

With Construct Runtime v2, I think anything that depends on this behavior needs to have tests to avoid things breaking. See e.g. [Rococo's construct runtime](https://github.com/paritytech/polkadot-sdk/blob/polkadot-v1.6.0/polkadot/runtime/rococo/src/lib.rs#L1338) where MMR's index is higher than Session's but it is "before" it.

All this is to say, Message Queue could stay at 34 but move to the end of construct runtime. Personally this seems like a footgun and I'd like a single order instead of two orders, but I think in the Relay Chains all of the low-number indices are claimed so there's not a good way around it without a breaking change. I think anything that depends on orders like this should have clear documentation and tests to guarantee that things like underlying changes to the `construct_runtime` macro don't inadvertently break things.

bkontur: thank you @joepetrowski, so I will move it at the end with index 34.

@claravanstaden could you:
1. confirm that `MessageQueue: pallet_message_queue = 34,` moved to the end is ok?
2. create an issue to the `polkadot-sdk` for adding a test-case to the `bridge-hub-test-utils` ensuring this ordering/indexing feature works as expected? I would like to reuse that test-case here and also for Rococo<>Eth

joepetrowski: Sorry it wasn't clear, I actually meant the opposite. We should try to make the indices match the order that things are in `construct_runtime`. So 250 is technically fine, but it also leaves very little room for things that would need to come after later.

claravanstaden: @bkontur 

1. I think the MessageQueue pallet needs to be larger than the EthereumOutboundQueue pallet index for this to work correct.
2. I logged issue https://github.com/paritytech/polkadot-sdk/issues/3208.

bkchr: The index doesn't change the order of `on_initialize`. The order of `on_initialize` is determined based on the order of the pallet in `construct_runtime!`. There is also no way that you can get this order right now in the code.

acatangiu: > There is also no way that you can get this order right now in the code.

I mean, there is.. the literal order of the lines of code in `construct_runtime!` 😆  hard codes the order in which hooks are called for the pallets.

But yes, the pallet index number does not matter here and there is no way to specify a number or some other ordering mechanism than the literal order of declaration in the macro.

claravanstaden: Yip, the order of the pallets does what we need. 🤷🏻‍♀️

joepetrowski: Yes but please also change the index to be greater than whatever it is after. What I meant in my first comment is that by not doing so, we are setting ourselves up for something to break when there is a change to the `construct_runtime` macro.

gupnik: FWIW, in [CR v2](https://github.com/paritytech/polkadot-sdk/pull/1378), I am planning to make the order of hooks as per the pallet index, but support an attribute `legacy_ordering` that would follow the current approach. This would ensure that the older ones continue to work but the newer chains use the correct paradigm.

joepetrowski: Yeah but let's just order them both on the chains where we can, like here. I know there are some places on the Relay it's not possible, but I don't see a reason to add something here that we know we'll need to patch around with this `legacy_ordering`. We know about CRv2 so let's just make something that works.

bkontur: I hope I understood it well :D:
- if so, please, resolve this conversation
- if not, please, correct me

https://github.com/polkadot-fellows/runtimes/pull/137/commits/7fc05c1671f1f5b01ca26ad57c350c0dc87892be

moved to the end:
```
MessageQueue: pallet_message_queue = 175,
```

claravanstaden: Thanks @bkontur! Is it fine that the MessageQueue index is at `250` on Rococo and `175` here? I mean, it won't break anything but just wanted to note that it is different now. Is there a max pallet index?

bkontur: I would say that it is totally fine. If your pallet's common code (reused for Rococo and Polkadot) expects a hard-coded 250., then it is not fine and that code should be changed :)

Do you use MessageQueue and/or its index directly somewhere, e.g. as an encoded calls for Transact?

joepetrowski: I would not recommend 250, just because it means that in the future there are only 5 pallets that could possible go after it without a breaking change. Since (a) pallet order can be important, and (b) we have many subsystems that are compositions of several pallets (governance, bridges, staking, etc.) then putting it at 250 seems like we are setting ourselves up for problems later.

claravanstaden: > Do you use MessageQueue and/or its index directly somewhere, e.g. as an encoded calls for Transact?

No, we don't use it anywhere. So it should be fine.

claravanstaden: @joepetrowski agreed. The MessageQueue pallet index is just unfortunately already 250 on Rococo, but agreed 175 is better for Kusama and Polkadot.

bkontur:
Fellows, please review this bump.

The only thing missing is the failed `check-migrations (Polkadot)` job, but this will be automatically resolved when `Polkadot@1.1.2` is enacted on February 19: [link to referendum](https://polkadot.polkassembly.io/referenda/458).

chain-spec-generator/src/system_parachains_specs.rs
```diff
@@ -131,33 +124,28 @@ pub fn bridge_hub_kusama_session_keys(keys: AuraId) -> bridge_hub_kusama_runtime
...
-		},
-		balances: asset_hub_polkadot_runtime::BalancesConfig {
+) -> serde_json::Value {
+	serde_json::json!({
+		"balances": asset_hub_polkadot_runtime::BalancesConfig {
 			balances: endowed_accounts
 				.iter()
 				.cloned()
-				.map(|k| (k, ASSET_HUB_POLKADOT_ED * 4096))
+				.map(|k| (k, ASSET_HUB_POLKADOT_ED * 4096 * 4096))
```
muharem, line 136: intentional increase?

chain-spec-generator/src/system_parachains_specs.rs
```diff
@@ -574,127 +470,90 @@ pub fn bridge_hub_kusama_local_testnet_config() -> Result<Box<dyn ChainSpec>, St
...
-	)))
+	Ok(Box::new(
+		GluttonKusamaChainSpec::builder(
+			glutton_kusama_runtime::WASM_BINARY.expect("GluttonKusama wasm not available!"),
+			Extensions { relay_chain: "kusama-local".into(), para_id: 1300 },
+		)
+		.with_name("Kusama Glutton Local")
+		.with_id("glutton-kusama-local")
+		.with_chain_type(ChainType::Local)
+		.with_genesis_config_patch(glutton_kusama_local_genesis(1300.into()))
```
muharem, line 513: changing it from 1002 to 1300?

system-parachains/encointer/src/lib.rs
```diff
@@ -1023,3 +1008,237 @@ fn test_ed_is_one_tenth_of_relay() {
...
+		::system_parachains_constants::kusama::fee::calculate_weight_to_fee(
+			&::system_parachains_constants::MAXIMUM_BLOCK_WEIGHT
+		),
+		system_parachains_constants::kusama::fee::calculate_weight_to_fee(
+			&system_parachains_constants::MAXIMUM_BLOCK_WEIGHT
+		)
+	);
+}
+
+// TODO: Encointer pallets does not have compatible `polkadot-sdk` versions,
```
muharem, line 1071: it can be very easy to forget to update this when something updated in constants. may be we should try to drop some dependancies from constant crates. Should we have at least an issue for thus?

muharem (approved):

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
```diff
@@ -810,8 +840,10 @@ construct_runtime!(
 		XcmpQueue: cumulus_pallet_xcmp_queue = 30,
 		PolkadotXcm: pallet_xcm = 31,
 		CumulusXcm: cumulus_pallet_xcm = 32,
+		// Temporary to migrate the remaining DMP messages:
```
bkchr, line 843: We should probably have some issue to remove them in the next upgrade.

bkchr (approved):

chain-spec-generator/src/system_parachains_specs.rs
bkchr: These are just genesis configs for testing, so changing them is fine.

chain-spec-generator/src/system_parachains_specs.rs
bkontur: yes, 1002 is BridgeHubKusama

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
bkontur: good point, [issue](https://github.com/polkadot-fellows/runtimes/issues/186) created and TODOs updated

https://github.com/polkadot-fellows/runtimes/pull/137/commits/97f81f9b9d759bbb5f7f64431ba71fa3718b2348

andresilva (approved):

acatangiu (approved):

system-parachains/bridge-hubs/bridge-hub-kusama/src/bridge_to_polkadot_config.rs
acatangiu: Let's add a code comment explaining this

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -940,7 +973,10 @@ pub type SignedExtra = (
 pub type UncheckedExtrinsic =
 	generic::UncheckedExtrinsic<Address, RuntimeCall, Signature, SignedExtra>;
 /// Migrations to apply on runtime upgrade.
-pub type Migrations = ();
+pub type Migrations = (
```
acatangiu, line 976: something else to check:

I see Polkadot AH has [a migration for "Uniques" v0->v1](https://github.com/polkadot-fellows/runtimes/blob/97f81f9b9d759bbb5f7f64431ba71fa3718b2348/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L887) while Kusama AH doesn't (also couldn't find it ever having it in `git log -p`).

Maybe it's not needed, but better to ask.

system-parachains/bridge-hubs/bridge-hub-kusama/src/bridge_to_polkadot_config.rs
bkontur: added here: https://github.com/polkadot-fellows/runtimes/pull/137/commits/f82cc0536f5b4c272cea185f3e31d2ca9d9537df

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
bkontur: yes, not needed, KusamaAssetHub has already version 1:
image

chain-spec-generator/src/system_parachains_specs.rs
muharem: yes, but still might affect tests, wanted to point in case of accidental change

Issue #134, opened 2024-01-04
CI: fix rust caching


ggwpez:
Eg [here](https://github.com/polkadot-fellows/runtimes/actions/runs/7409368464/job/20159489326?pr=132) it prints `No cache found` although there should be some?!  
Also seems to happen on other branches. Maybe @alvicsam knows how to fix it? Can we maybe even use this new Parity caching tool?

The `check-migrations` job could also need some caching for its build 😅

bkchr:
Will be fixed by: https://github.com/polkadot-fellows/runtimes/pull/135

ggwpez:
I still dont get whats going on.  

In [this](https://github.com/polkadot-fellows/runtimes/actions/runs/7461967014/job/20303211183) job, it restored fine from key `v0-rust-fellowship-cache-35cfef99`, now in [another job](https://github.com/polkadot-fellows/runtimes/actions/runs/7462837903/job/20306115920) it fails to restore from the same key.  
The second job even ran after the first one 🤔

bkchr:
Hmm weird :see_no_evil:

alvicsam:
I removed `shared-key` in this PR https://github.com/polkadot-fellows/runtimes/pull/142 and now it runs faster because each workflow has its own cache. Usually the first run in a PR doesn't use cache (it might download it from the `main` branch but cargo will not use it). But then the cache action will upload the updated version and use it in the next runs. Also GitHub cache is only [10GB for repo](https://github.com/polkadot-fellows/runtimes/actions/caches) and now one PR uses ~3GB of it. The good news is that GitHub doesn't erase the oldest cache immediately so it can be used for some time.

> In [this](https://github.com/polkadot-fellows/runtimes/actions/runs/7461967014/job/20303211183#step:8:67) job, it restored fine from key v0-rust-fellowship-cache-35cfef99, now in [this one](https://github.com/polkadot-fellows/runtimes/actions/runs/7462837903/job/20306115920#step:8:66) it fails to restore from the same key. The second job even ran after the first one 🤔

In this case [enabling debug for runners](https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging) can help to understand what went wrong.

bkchr:
Wait, what? But what @ggwpez was for downloading the cache? Then we should really move to only generate the cache in the `main` branch and only download them in prs. I added the `shared-key` exactly to not always rebuild everything the first time a pr is opened.

[...]

PR #142, opened 2024-01-11
[ci] Remove shared key for cache and add cache for migrations

alvicsam:
PR removes shared key for rust-cache action. When 2 different workflows use the same shared-key they try to use the same destination for cache which makes it unusable for them (they have race condition who will be the first to write there). Also PR adds cache for migration jobs and cancelling jobs on previous runs.

cc https://github.com/polkadot-fellows/runtimes/issues/134


bkchr (approved):

alvicsam:
fyi, I don't have the `Merge` button

ggwpez:
> Error: unable to get latest version

Now it fails

ordian (approved):

PR #149, opened 2024-01-18
Deprecate `xcm::body::TREASURER_INDEX` constant, use standard `Treasury` varian instead

muharem:
Deprecate the `xcm::body::TREASURER_INDEX` constant and use the standard `Treasury` variant from the `xcm::BodyId` type instead.

The custom body index is redundant here since the standard `Treasury` body variant exists. It was overlooked previously, and the index was added by mistake.
Removing the index should not affect anything. It's only used as one of the origins that guards permissioned operations for the fellowship treasury instance and asset rate pallet instance, and not used to derive any account ids.


relay/polkadot/constants/src/lib.rs
```diff
@@ -109,6 +109,7 @@ pub mod xcm {
 		// The bodies corresponding to the Polkadot OpenGov Origins.
 		pub const FELLOWSHIP_ADMIN_INDEX: u32 = 1;
 		// The body corresponding to the Treasurer OpenGov track.
+		#[deprecated = "use `xcm::latest::BodyId::Treasury` instead"]
```
ggwpez, line 112: Can you please add a removal date? We are trying to that in the SDK at least like this: https://github.com/paritytech/polkadot-sdk/blob/a5370fb1877cc9672cb47fea36a27cfed3f4da63/substrate/frame/support/src/storage/migration.rs#L40

muharem: done

ggwpez (approved):

acatangiu (approved):

PR #150, opened 2024-01-18
Clean up undecodable proxy

### LINKED ISSUE #145: `Proxy::Proxies` undecodable storage on Polkadot
https://github.com/polkadot-fellows/runtimes/actions/runs/7553599023/job/20564716569?pr=137

Looking into it

liamaharon:
Closes #145 

The key is undecodable because it contains a Proxy with type `SudoBalances` which then also failed to migrate in https://github.com/paritytech/polkadot-sdk/commit/cf4c744eb3413d9ef4755c34312ba65d5431a899 due to the type being removed from the runtime at the same time as the migration was added to the runtime. See the linked issue for details.

As the offending proxy belongs being a dormant, old sudo account and is almost certainly not needed, in the interests of keeping this PR simple and minimising time spent on this issue I opted to just delete the undecodable key. If deemed worth the time, I could instead properly migrate the proxy / unreserve the locked funds from it but I doubt this is necessary.


bkontur (approved):

liamaharon:
Removing instead via referenda https://polkadot.polkassembly.io/referenda/440

PR #152, opened 2024-01-20
Remove Old Construct Runtime Syntax

joepetrowski:
Test


system-parachains/encointer/src/migrations_fix.rs
```diff
@@ -22,7 +22,6 @@ pub mod balances {
 	use pallet_balances::*;
 	#[cfg(feature = "try-runtime")]
 	use sp_runtime::TryRuntimeError;
-	use sp_std::vec::Vec;
```
ggwpez, line 25: Looks like this is needed to make it compile.

ggwpez (approved):
I think its fine without Changelog.

system-parachains/encointer/src/migrations_fix.rs
joepetrowski: Must be due to using `production` profile, with `dev` it gives an unused import warning. Added it back.

bkontur: `Vec` is used here only in the functions marked with `#[cfg(feature = "try-runtime")]`, so there are two solutions:

1. add feature before import:
```
use sp_std::vec::Vec;
```
2.  or remove import and use it directly:
```
fn pre_upgrade() -> Result<sp_std::vec::Vec<u8>
...
fn post_upgrade(_state: sp_std::vec::Vec<<u8>)
```

bkontur (approved):

kianenigma (approved):
Thanks, I hope we move to the beautiful https://github.com/paritytech/polkadot-sdk/pull/1378 soon.

bkchr (approved):

xlc (approved):

liamaharon (approved):

PR #164, opened 2024-01-26
Enable idempotency check

ggwpez:
Trying to enable this check that was disabled because of a non-idempotent migration.

cc @liamaharon 


ggwpez:
Yes was just testing if it goes green 😄

ggwpez:
@brenzi there is an error for decoding the `polkadotXcm::versionNotifyTargets`, looks like we discussed it here:  https://github.com/polkadot-fellows/runtimes/pull/80#discussion_r1448449405

brenzi:
Yes. I think this can be safely ignored because the storage has been written by the new pallet version already. Only issue that the storage-version wasn't bumped with the proper migration

bkchr (approved):

andresilva (approved):

PR #168, opened 2024-01-29
Review Bot version 2.4.0 - Fellow Scores

### LINKED ISSUE #93: Introduce more sophisticated approval rights
Currently the approval rights allow to only specify a minimum rank and the number of approvals required to approve a pull request. While this is working, it is not "very inviting" as lower ranks are not incentivized to even review/approve. However on the other side it is sometimes hard to "cat herd" to get everyone together for reviewing/approval. Thus, we should improve the approval bot to have some sort of "voting power". This voting power should be similar to how on chain voting is done. Every rank gets a certain voting power and we will then require X amount of voting power to approve a pull request. Thus, we could for example have that 3 rank II members together have the same voting power as one rank III member.

CC @Bullrich

Bullrich:
Updated Review Bot to version 2.4.0 which contains paritytech/review-bot#110 (solved in paritytech/review-bot#113)

This resolves #93 

Updated review bot to latest version which brings new changes to the fellow object:

Added new optional field to the config named `scores` that has the score of each fellow per dan (added up to dan IX).

When the rule of type `fellows` has the optional field of `minTotalScore` it will check that the score of the fellows who approved the PR sums up to that number. If it doesn't sum to that number it will fail with a custom error listing the score of each fellow for reference (and saying how many points are missing).

Find a complete explanation of the new configuration here: [Fellows Rule](https://github.com/paritytech/review-bot/blob/v2.4.0/README.md#fellows-rule)

I also added some scores to all the dans (feel free to modify them) and updated the `Relay and system files` rule to have a minimum score of 6.

### DIFF
--- a/.github/review-bot.yml
+++ b/.github/review-bot.yml
@@ -14,7 +14,8 @@ rules:
         - ^CHANGELOG.md$
     type: fellows
     minRank: 2
-    minApprovals: 3
+    minApprovals: 2
+    minTotalScore: 6
   - name: General Files
     condition:
       include:
@@ -26,3 +27,13 @@ rules:
         - ^CHANGELOG.md$
     type: fellows
     minRank: 1
+scores:
+  dan1: 1
+  dan2: 2
+  dan3: 5
+  dan4: 9
+  dan5: 10
+  dan6: 15
+  dan7: 20
+  dan8: 25
+  dan9: 30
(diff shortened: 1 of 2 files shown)


ggwpez (approved):
Thanks!

bkchr (approved):
Nice! Ty!

xlc (approved):

PR #170, opened 2024-01-30
Remove state trie migration from kusama, add it to polkadot

cheme:
This PR add the state trie migration to V1 to polkadot runtime.
The configuration is the same as the one that was run on kusama.
As in kusama this PR changes the state-trie-version to v1, so the migration declared in unreleased Have to run with the next upgrade. In case the migration is not include, the line `state_version: 1` have to be roll back to `state_version: 0` as warpsync will stop working until the end of the migration.


ggwpez:
We can dry-run this with chopsticks since it exposes a dev-rpc to mint new blocks.

cheme:
> We can dry-run this with chopsticks since it exposes a dev-rpc to mint new blocks.

Would be nice. For the kusama one my main worry was that the migration would consume too much weight on the block, is there a way to get the remaining weight in each block with chopstick (I do not know chopstick at all)?

ggwpez:
No i dont think it can print anything. But you can subscribe on `System::BlockWeights` in PJS or so.

relay/kusama/src/lib.rs
```diff
@@ -1626,9 +1606,6 @@ construct_runtime! {
 		Auctions: auctions::{Pallet, Call, Storage, Event<T>} = 72,
 		Crowdloan: crowdloan::{Pallet, Call, Storage, Event<T>} = 73,
 
-		// State trie migration pallet, only temporary.
-		StateTrieMigration: pallet_state_trie_migration = 98,
```
bkchr, line 1630: We probably need to write some migration to remove all the old pallet storage items? There already exists the `RemovePallet` migration that we can use for this.

relay/polkadot/src/lib.rs
```diff
@@ -1703,6 +1726,8 @@ pub mod migrations {
 	pub type Unreleased = (
 		// Upgrade SessionKeys to include BEEFY key
 		UpgradeSessionKeys,
+		init_state_migration::InitMigrate,
+		pallet_nomination_pools::migration::versioned_migrations::V5toV6<Runtime>,
```
bkchr, line 1730:
```suggestion
```

bkchr:
> For the kusama one my main worry was that the migration would consume too much weight on the block,

For the relay chain this isn't really a big problem, because we can also build a block that takes slightly longer.

relay/kusama/src/lib.rs
cheme: yes I will try to use the code we ran on westend

relay/kusama/src/lib.rs
```diff
@@ -2714,63 +2692,45 @@ mod remote_tests {
 	}
 }
 
-mod init_state_migration {
+mod clean_state_migration {
```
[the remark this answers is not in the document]
cheme: whoa looks amazing, thx

bkchr (approved):

eskimor (approved):

PR #175, opened 2024-02-05
upgraded keep-pr-up-to-date action to 0.2.0

Bullrich:
Upgraded the action [paritytech/up-to-date-action](https://github.com/paritytech/up-to-date-action) to version `0.2.0`.

This version brings paritytech/up-to-date-action#9. A feature that, when the action fails with a PR, it comments in it so the user can try to fix the problem instead of silently failing.

### DIFF
--- a/.github/workflows/up-to-date.yml
+++ b/.github/workflows/up-to-date.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   updatePullRequests:
+    name: Keep PRs up to date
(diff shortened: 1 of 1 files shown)


Bullrich:
Hi @bkchr!

When you have a moment could you please look at this PR?

It should help notifying about problems with the [action failing because of lack of permission or merge problems](https://github.com/polkadot-fellows/runtimes/actions/runs/7726039358).

Issue #140, opened 2024-01-10
1.2.0 Release


bkontur:
Please everyone leave a comment with a link to a pr or issue, so we have it tracked.

## TODO
- [x] PR with bump to `polkadot@1.5` https://github.com/polkadot-fellows/runtimes/pull/137 (minimal configurations, [just to compile](https://github.com/polkadot-fellows/runtimes/issues/101))
- [x] PR with bump to `polkadot-sdk@1.6.0` https://github.com/polkadot-fellows/runtimes/pull/159
- [x] PR with bump to `polkadot-sdk@1.7.0` https://github.com/polkadot-fellows/runtimes/pull/187
- [x] check/fix migrations when 1.1.0 / 1.1.2 deployed https://github.com/polkadot-fellows/runtimes/pull/195
- [x] regenerate weights
  - [x] regenerate weights for all pallets https://github.com/polkadot-fellows/runtimes/pull/223
    - [x] ping https://github.com/polkadot-fellows/runtimes/pull/187#issuecomment-1969253891
  - [x] check bridge transfer fee constants
- [x] bump spec_versions to `1_002_000` - done in https://github.com/polkadot-fellows/runtimes/pull/187
- [x] check and bump transaction versions https://github.com/polkadot-fellows/runtimes/issues/250
- [x] check Nacho's integration tests
  - [x] check/verify e2e tests: OpenGov -> relay to AssetHub - see [comment](https://github.com/polkadot-fellows/runtimes/pull/187/files#r1507284250) @muharem 
- [x] fix `#[api_version(7)] impl primitives::runtime_api::ParachainHost - fn node_features` + verify what is missing https://github.com/polkadot-fellows/runtimes/pull/194
- [x] add benchmark `cumulus_pallet_parachain_system` to SP runtimes
- [x] address comment [here](https://github.com/polkadot-fellows/runtimes/pull/57#issuecomment-1885008379)
- [x] integration tests improvement - adapt [comment](https://github.com/polkadot-fellows/runtimes/pull/114#discussion_r1420432244) fixed by https://github.com/polkadot-fellows/runtimes/pull/185
  ```
  # TODO: replace with `emulated-integration-tests-common@X.Y.Z` from `polkadot-sdk`
  integration-tests-common = { path = "../common" }
  ```
- [x] remove unneeded dependencies (`cargo machete`)
- [x] fix `pallet_asset_conversion benchmarks` https://github.com/paritytech/polkadot-sdk/issues/3440
[...]

PR #178, opened 2024-02-09
Remove `im-online` pallet from Kusama and Polkadot runtimes

s0me0ne-unkn0wn:
# Summary

This PR aims to remove `im-online` pallet, its session keys, and its on-chain storage from both Kusama and Polkadot relay chain runtimes, thus giving up liveness slashing.

# Motivation

* Missing out on rewards because of being offline is enough disincentive for validators. Slashing them for being offline is redundant.
* Disabling liveness slashing is a prerequisite for validator disabling.

# See also

paritytech/polkadot-sdk#1964
paritytech/polkadot-sdk#784


s0me0ne-unkn0wn:
@bkchr @ordian I think it's ready for review now. `try-runtime` is still failing in CI, but that'll be fixed with paritytech/try-runtime-cli#79. I've tested locally using a state from a block not containing any `im-online` events, and it passes OK.

@bkchr WDYT about adding offchain storage cleanup into the same PR? Seems unsound to me, as some previous offchain workers may still be there, but I'm not sure if there's any synchronization of their access to the offchain db. Could it make sense?

relay/polkadot/src/lib.rs
```diff
@@ -1689,27 +1698,111 @@ pub type Migrations = migrations::Unreleased;
...
+			old_state: sp_std::vec::Vec<u8>,
+		) -> Result<(), sp_runtime::TryRuntimeError> {
+			if System::last_runtime_upgrade_spec_version() > UPGRADE_SESSION_KEYS_FROM_SPEC {
+				log::warn!(target: "runtime::session_keys", "Skipping session keys migration post-upgrade check due to spec version (already applied?)");
+				return Ok(())
+			}
+
+			let key_ids: Vec<_> = SessionKeys::key_ids()
+				.into_iter()
+				.filter(|&k| *k != sp_core::crypto::key_types::BEEFY)
```
bkchr, line 1772: Why do you filter out the beefy keys?

relay/polkadot/src/lib.rs
```diff
@@ -1689,27 +1698,111 @@ pub type Migrations = migrations::Unreleased;
...
+				.collect();
+			frame_support::ensure!(
+				key_ids
+					.clone()
+					.into_iter()
+					.find(|&k| *k == sp_core::crypto::key_types::IM_ONLINE) ==
+					None,
```
bkchr, line 1731:
```suggestion
					.find(|&k| *k == sp_core::crypto::key_types::IM_ONLINE)
					.is_none(),
```

relay/polkadot/src/lib.rs
```diff
@@ -457,7 +482,6 @@ fn transform_session_keys(v: AccountId, old: OldSessionKeys) -> SessionKeys {
 	SessionKeys {
 		grandpa: old.grandpa,
 		babe: old.babe,
-		im_online: old.im_online,
 		para_validator: old.para_validator,
 		para_assignment: old.para_assignment,
 		authority_discovery: old.authority_discovery,
```
bkchr, line 487: You should copy the beefy key.

relay/polkadot/src/lib.rs
```diff
@@ -1689,27 +1698,111 @@ pub type Migrations = migrations::Unreleased;
...
+			log::info!(target: "runtime::session_keys", "Session keys migrated successfully");
+			Ok(())
+		}
 	}
 
 	/// Unreleased migrations. Add new ones here:
 	pub type Unreleased = (
-		// Upgrade SessionKeys to include BEEFY key
+		// Upgrade SessionKeys to exclude ImOnline key and include BEEFY
 		UpgradeSessionKeys,
```
ordian, line 1794: beefy upgrade has been already applied in 1.1.0: https://github.com/polkadot-fellows/runtimes/releases/tag/v1.1.0
beefy-polkadot
we need to remove it from unreleased

ordian: i hope we didn't add it to subsequent releases again 🙈

ordian: looks like we skipped some releases and applied 1.1.2 directly

bkontur: > looks like we skipped some releases and applied 1.1.2 directly
> 
> but given how easy it is to overlook this `UpgradeSessionKeys` slipping into the next release, we should be more careful, not forget removing this in the next release and do something about it [paritytech/polkadot-sdk#2421](https://github.com/paritytech/polkadot-sdk/issues/2421)

We won't forget, it is one of todo points for the next release: https://github.com/polkadot-fellows/runtimes/issues/140

relay/polkadot/src/lib.rs
s0me0ne-unkn0wn: I'm not very familiar with how it works. This PR is against `main` branch, and there's no `beefy` in `impl_opaque_keys!` in `main`. Instead, there's a migration that introduces beefy keys, which I kept. Should I rebase on some other branch?

acatangiu: Let's merge https://github.com/polkadot-fellows/runtimes/pull/195 then all is good and clean

relay/polkadot/src/lib.rs
s0me0ne-unkn0wn: And even if we forget, the migration is guarded by the spec version: https://github.com/polkadot-fellows/runtimes/pull/178/files#diff-efa4caeb17487ecb13d8f5eb7863c3241d84afa2e73fbf25909a2ca89df0f362R1751

relay/polkadot/src/lib.rs
```diff
@@ -1652,13 +1688,104 @@ pub type Migrations = migrations::Unreleased;
...
+	use sp_core::crypto::ByteArray;
+
+	parameter_types! {
+		pub const ImOnlinePalletName: &'static str = "ImOnline";
+	}
+
+	/// Upgrade Session keys to exclude `ImOnline` key.
+	/// When this is removed, should also remove `OldSessionKeys`.
+	pub struct UpgradeSessionKeys;
+	const UPGRADE_SESSION_KEYS_FROM_SPEC: u32 = 1001002;
```
ordian, line 1701: let's hope we include it in the next release 🤞

ordian (approved):

relay/polkadot/src/lib.rs
```diff
@@ -1652,13 +1688,104 @@ pub type Migrations = migrations::Unreleased;
...
+			Ok(())
+		}
+	}
 
 	/// Unreleased migrations. Add new ones here:
 	pub type Unreleased = (
+		// Upgrade SessionKeys to exclude ImOnline key and include BEEFY
+		UpgradeSessionKeys,
+		pallet_nomination_pools::migration::versioned::V5toV6<Runtime>,
+		pallet_nomination_pools::migration::versioned::V6ToV7<Runtime>,
```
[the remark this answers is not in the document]
s0me0ne-unkn0wn: Oops. Good catch! Seems like a merge went awry. Fixed.

relay/polkadot/src/lib.rs
s0me0ne-unkn0wn: I'll keep an eye on it. If it's not in the next release somehow, I'll update the version.

bkchr (approved):

bkchr:
@s0me0ne-unkn0wn can you merge master? Than we can merge this.

s0me0ne-unkn0wn:
@bkchr it's at the latest `master` rn. But we still have required migration tests failing...

integration-tests/emulated/chains/relays/kusama/src/genesis.rs
```diff
@@ -18,7 +18,6 @@ use authority_discovery_primitives::AuthorityId as AuthorityDiscoveryId;
 use babe_primitives::AuthorityId as BabeId;
 use beefy_primitives::ecdsa_crypto::AuthorityId as BeefyId;
 use grandpa::AuthorityId as GrandpaId;
-use pallet_im_online::sr25519::AuthorityId as ImOnlineId;
```
acatangiu, line 21: please also remove [`pallet-im-online` dep from Cargo.toml](https://github.com/polkadot-fellows/runtimes/blob/59172d5fb8feea2646490618bead3b2b4fb06220/integration-tests/emulated/chains/relays/kusama/Cargo.toml#L20)

integration-tests/emulated/chains/relays/polkadot/src/genesis.rs
```diff
@@ -18,7 +18,6 @@ use authority_discovery_primitives::AuthorityId as AuthorityDiscoveryId;
 use babe_primitives::AuthorityId as BabeId;
 use beefy_primitives::ecdsa_crypto::AuthorityId as BeefyId;
 use grandpa::AuthorityId as GrandpaId;
-use pallet_im_online::sr25519::AuthorityId as ImOnlineId;
```
acatangiu, line 21: same, remove from `Cargo.toml` as well

acatangiu (approved):

integration-tests/emulated/chains/relays/kusama/src/genesis.rs
s0me0ne-unkn0wn: Thanks for the catch! Removed here and from chainspec generator as well.

s0me0ne-unkn0wn:
@bkchr okay, I bumped the Kusama spec version in this PR to let it pass the idempotent try-runtime check, and now the only thing failing is the encointer runtime migrations that IIUC have nothing to do with this PR. So it's good to go, but I'm not sure how to sort out those encointer problems, and that CI job is required :man_shrugging:

bkontur:
> @bkchr okay, I bumped the Kusama spec version in this PR to let it pass the idempotent try-runtime check, and now the only thing failing is the encointer runtime migrations that IIUC have nothing to do with this PR. So it's good to go, but I'm not sure how to sort out those encointer problems, and that CI job is required 🤷‍♂️

@s0me0ne-unkn0wn Encointer is known issue, please check comment: https://github.com/polkadot-fellows/runtimes/pull/159#issuecomment-1966836443 

@bkchr I think we should temporary skip `check-migrations (Encointer)` as required, because until it is enacted, you will need to use [superpower](https://github.com/polkadot-fellows/runtimes/pull/204#issuecomment-1970495926) for every PR, wdyt?

relay/kusama/src/lib.rs
```diff
@@ -156,7 +156,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("kusama"),
 	impl_name: create_runtime_str!("parity-kusama"),
 	authoring_version: 2,
-	spec_version: 1_001_000,
+	spec_version: 1_001_003,
```
bkontur, line 159:
```suggestion
	spec_version: 1_002_000,
```

@s0me0ne-unkn0wn why `1_001_003`? Are we going to hotfix this for `1.1.2`?
If this PR is going to the `main` and is expected to be release as a next `1.2.0` release, then I would suggest to change it to the `1_002_000` and also you could maybe change it for other runtimes as it is part of TODOs here: https://github.com/polkadot-fellows/runtimes/issues/140:
```
- [ ]  check and bump runtime/transaction versions
 ```

s0me0ne-unkn0wn: I'm not familiar with how the versioning goes; I just needed it to be higher than `1.1.2` to pass [the idempotency check](https://github.com/polkadot-fellows/runtimes/pull/178/files#diff-efa4caeb17487ecb13d8f5eb7863c3241d84afa2e73fbf25909a2ca89df0f362R1782). Otherwise, the migration applies twice. I'll change it to `1.2.0` for both runtimes in this PR then. Thanks for pointing out!

bkchr:
Yes, I also already thought of disabling the try runtime check for encointer. Can you please make this happen?

bkchr:
@s0me0ne-unkn0wn I have thought again about this pr, in the context of: https://github.com/paritytech/polkadot-sdk/issues/64

I think we need to revert this PR for now or do some other manual work. The problem being that when we call [`candidate_events`](https://github.com/paritytech/polkadot-sdk/blob/12e5e19c2033383fc914bcc286ad16ae565d4734/polkadot/primitives/src/runtime_api.rs#L183) it will iterate over the events. Now, when we remove the `im-online`, they would be undecodable in the block that enacted the runtime upgrade. `im-online` is sending these [events](https://github.com/paritytech/polkadot-sdk/blob/12e5e19c2033383fc914bcc286ad16ae565d4734/substrate/frame/im-online/src/lib.rs#L801-L811) in `on-initialize` (aka before the parachain inherent is executed).

So, the problem could be that we miss backed candidates (not such a big issue I think). However when we miss the candidate included events we could run into a finality stall until the training wheels force approve the candidates. Given this, we should fix this.

A possible fix could be that we filter out these broken events in `candidate_events` before we do the actual call to the implementation.

s0me0ne-unkn0wn:
@bkchr Hmmm :thinking: 
Any ideas why didn't we hit it on Rococo and Westend?
Anyway, I'll research that and will try to push on the fix to make it into the current release.

bkchr:
You need to be "lucky". The problem only appears when the block we enact the runtime upgrade, also contains one of these events. As they don't appear that often, triggering this bug is quite unlikely, but it can happen. At least from my understanding. I didn't looked that deep into the approval voting and if this is maybe also triggered in a different way than using the candidate events it should be fine as is.

s0me0ne-unkn0wn:
> `im-online` is sending these [events](https://github.com/paritytech/polkadot-sdk/blob/12e5e19c2033383fc914bcc286ad16ae565d4734/substrate/frame/im-online/src/lib.rs#L801-L811) in `on-initialize` (aka before the parachain inherent is executed).

Looking at that closer, the code you're pointing to is executed in `on_before_session_ending`, that is, not in _every_ `on_initialize` but at the session boundary only. As we're scheduling a new runtime enactment at a specific block, IIUC, it wouldn't be a big problem for us to schedule it in such a way that it's guaranteed to miss the session boundary. Does that make sense?

bkchr:
We don't schedule the upgrade on a specific block. It depends on OpenGov voting etc when the upgrade is enacted.

s0me0ne-unkn0wn:
> We don't schedule the upgrade on a specific block

Then I probably don't fully understand what's going on here: https://polkadot.polkassembly.io/referenda/458

bkchr:
You got me :P I also didn't know of this :D 

However, if a block number is a session change or not is still something that is not 100% predictable. Sessions end based on time. Thus, skipped blocks let a session end at an earlier block number.

PR #184, opened 2024-02-14
Add a Root Signaller Track

joepetrowski:
Posting this PR for discussion, but perhaps people have other solutions in mind.

The Root track was designed to handle only one referendum at a time, because calls requiring Root are usually sensitive and should be evaluated one at a time. However, we have also used the Root track to make remarks to signal desires/wishes of the network to various bodies in the network (e.g. RFC-12, instructing the Fellowship to add a collective).

These statements do not execute any stateful logic that would affect the network, and in my opinion more than one could be evaluated at a time. These referenda should not hold up voting on proposals that actually do require Root, nor be forced to queue (for example, it should be possible to propose two new collectives in parallel).

The approval/support requirements are the same as Root, but the origin does not map to any privilege. Passing something on the track is merely a signal.

### DIFF
--- a/relay/kusama/src/governance/origins.rs
+++ b/relay/kusama/src/governance/origins.rs
@@ -86,6 +86,8 @@ pub mod pallet_custom_origins {
 		Fellowship8Dan,
 		/// Origin commanded by rank 9 of the Polkadot Fellowship and with a success of 9.
 		Fellowship9Dan,
+		/// Origin for signaling wishes of the network.
+		RootSignaller,
 	}
 
 	macro_rules! decl_unit_ensures {
@@ -128,6 +130,7 @@ pub mod pallet_custom_origins {
 		ReferendumCanceller,
 		ReferendumKiller,
 		WhitelistedCaller,
+		RootSignaller,
 		FellowshipInitiates: u16 = 0,
 		Fellows: u16 = 3,
 		FellowshipExperts: u16 = 5,
--- a/relay/kusama/src/governance/tracks.rs
+++ b/relay/kusama/src/governance/tracks.rs
@@ -65,7 +65,7 @@ const APP_WHITELISTED_CALLER: Curve =
 const SUP_WHITELISTED_CALLER: Curve =
 	Curve::make_reciprocal(1, 28, percent(20), percent(5), percent(50));
 
-const TRACKS_DATA: [(u16, pallet_referenda::TrackInfo<Balance, BlockNumber>); 15] = [
+const TRACKS_DATA: [(u16, pallet_referenda::TrackInfo<Balance, BlockNumber>); 16] = [
 	(
 		0,
 		pallet_referenda::TrackInfo {
@@ -92,6 +92,20 @@ const TRACKS_DATA: [(u16, pallet_referenda::TrackInfo<Balance, BlockNumber>); 15
 			min_enactment_period: 10 * MINUTES,
 			min_approval: APP_WHITELISTED_CALLER,
 			min_support: SUP_WHITELISTED_CALLER,
+		},
+	),
+	(
+		2,
+		pallet_referenda::TrackInfo {
+			name: "root_signaller",
+			max_deciding: 10,
+			decision_deposit: 20 * GRAND,
+			prepare_period: 2 * HOURS,
+			decision_period: 28 * DAYS,
+			confirm_period: 24 * HOURS,
+			min_enactment_period: 10 * MINUTES,
+			min_approval: APP_ROOT,
+			min_support: SUP_ROOT,
 		},
 	),
 	(
@@ -294,6 +308,7 @@ impl pallet_referenda::TracksInfo<Balance, BlockNumber> for TracksInfo {
 		} else if let Ok(custom_origin) = origins::Origin::try_from(id.clone()) {
 			match custom_origin {
 				origins::Origin::WhitelistedCaller => Ok(1),
+				origins::Origin::RootSignaller => Ok(2),
 				// General admin
 				origins::Origin::StakingAdmin => Ok(10),
 				origins::Origin::Treasurer => Ok(11),
--- a/relay/polkadot/src/governance/origins.rs
+++ b/relay/polkadot/src/governance/origins.rs
@@ -61,6 +61,8 @@ pub mod pallet_custom_origins {
 		BigSpender,
 		/// Origin able to dispatch a whitelisted call.
 		WhitelistedCaller,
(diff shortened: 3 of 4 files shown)


xlc:
Looks reasonable, however few questions:
- Do we need RFC for this?
- Do we want to call it RootSignaller or just Signaller? I don't think we will ever have more than one signing track?

joepetrowski:
> Do we need RFC for this?

IMO no, although we should have a way to amend RFCs (like 12). And perhaps an RFC that is a generalized version of RFC-12, like "how to signal network desire to do something".

> Do we want to call it RootSignaller or just Signaller? I don't think we will ever have more than one signing track?

Fair enough. I don't have a strong opinion, either one is fine with me.

ggwpez:
> Do we want to call it RootSignaller or just Signaller? I don't think we will ever have more than one signing track?

I think we can extend the concept. For example when the Ambassador wants to make changes to their code - but cannot do it themselves - they may add an `AmbassadorSignaller`.

bkchr (approved):

xlc (approved):

sourabhniyogi:
I could be missing something, but if the "Wish for Change" is merely a signal, it shouldn't need a ... 20,000 DOT decision deposit?  And if it is _non-binding_, its like a Polkassembly poll that tallies like other but not meaningful until coupled with the _behavior_ that the Wish for Change demands in said wish?

A genie that pops out of the lamp and says "you get 3 wishes" but then isn't bound to fulfilling any of them is hardly a genie.  

On the other hand, if the will of the DOT Tokenholder is expressed in the WFC, and the governing body (Fellows, or some other  collective) MUST fulfill the WFC in a _binding_ way, then we have something real.  This does not demand anything real yet.

xlc:
It is not possible to make this binding. I can wish for world peace and there is good chance other DOT holders supports my wish but then what?

But yeah I agree decision deposit could be reduced.

joepetrowski:
> On the other hand, if the will of the DOT Tokenholder is expressed in the WFC, and the governing body (Fellows, or some other collective) MUST fulfill the WFC in a binding way, then we have something real. This does not demand anything real yet.

Well, the point of having a high decision deposit and the same support/approval criteria as the Root track is to show the seriousness of the desire. No, the "wish" cannot _force_ the Fellowship to implement/integrate something, but the real Root track can disband the Fellowship and just remove it entirely from the runtime. Fellowship RFCs are also not binding, they just say, "we should do this".

The high deposit is meant to stop a single person from spamming the track with a bunch of considerations. Although the track can consider more than one at a time, these are still major undertakings (e.g. "add an EVM system chain"). The track should not be used for casually dropping your shower thoughts.

The goal is to (a) override the Fellowship RFC process if the Fellowship doesn't want to vote for something (which may be for any number of reasons) and (b) aid in the chicken vs. egg scenario; for example, should you pay developers to build a system EVM chain with no signal at all as to whether it would be included.

sourabhniyogi:
> > On the other hand, if the will of the DOT Tokenholder is expressed in the WFC, and the governing body (Fellows, or some other collective) MUST fulfill the WFC in a binding way, then we have something real. This does not demand anything real yet.
> 
> Well, the point of having a high decision deposit and the same support/approval criteria as the Root track is to show the seriousness of the desire. No, the "wish" cannot _force_ the Fellowship to implement/integrate something, but the real Root track can disband the Fellowship and just remove it entirely from the runtime. Fellowship RFCs are also not binding, they just say, "we should do this".
> 
> The high deposit is meant to stop a single person from spamming the track with a bunch of considerations. Although the track can consider more than one at a time, these are still major undertakings (e.g. "add an EVM system chain"). The track should not be used for casually dropping one's shower thoughts.

Can you make the 20K DOT into 1K DOT so that poor people's non-shower thoughts can put to the WFC track -- 1K DOT is at the same level as _TREASURER_, and 2.5x the 400 DOT of _Big Spender_:

  https://wiki.polkadot.network/docs/learn-polkadot-opengov-origins#origins-and-tracks-info

Question: Why are the "ranks" not being used to tally votes of the Fellowship according to some formula?  I want to see "This opinion is not widely shared in the Fellowship" mapped into numbers like OpenGov on the Collectives chain publicly:

 https://twitter.com/rphmeier/status/1757840554488930712

If the "we should do this" is IGNORED by the Fellowship [(a)'s "override the Fellowship RFC process"], then everyone knows that ultimate power is in the Fellowship rather than OpenGov.

I didn't see this tallying on #32 and am not aware of any efforts to improve Fellowship tallying despite Collectives chain having this capability.  What's the bottleneck or what did I miss?

Happy to put "add EVM to the Relay Chain" WFC as the test case (a very worthy one).  This is not a shower thought, or if it is, please explain what has to happen for this particular shower thought to be developed correctly.  For this specific concept, a whole feasibility report is very much required, but that standard wasn't applied to RFC #32 Minimal Relay Chain.  So having clarity on the standard required for non-binding WFC "We should do this" is essential.

ggwpez:
>Can you make the 20K DOT into 1K DOT so that poor people's non-shower thoughts can put to the WFC track -- 1K DOT is at the same level as TREASURER, and 2.5x the 400 DOT of Big Spender:

It is not about the proposer being poor or not. Poor people can also put theirs thoughts on the track. Just need to find someone with 20K to also think that its a good idea and place the deposit for them (should be easy for good ideas).

sourabhniyogi:
> > Can you make the 20K DOT into 1K DOT so that poor people's non-shower thoughts can put to the WFC track -- 1K DOT is at the same level as TREASURER, and 2.5x the 400 DOT of Big Spender:
> 
> It is not about the proposer being poor or not. Poor people can also put theirs thoughts on the track. Just need to find someone with 20K to also think that its a good idea and place the deposit for them (should be easy for good ideas).

This is true of all OpenGov proposals.  This WFC is on the same order of seriousness as Treasurer + Big Spender, not 20x or 50x more serious.  If not, please explain why non-binding wishes need 20x-50x more seriousness in decision deposit.

ggwpez:
> This WFC is on the same order of seriousness as Treasurer + Big Spender, not 20x or 50x more serious. If not, please explain why non-binding wishes need 20x-50x more seriousness in decision deposit.

The treasurer is definitely much less powerful than root. It can only do a very limited set of things: delegate funds.  
Root can just arbitrarily change the balance of any account, or fork the network by placing an invalid Runtime Blob.  
These two dont play in the same league.

WFC is for essential network changes - not delegating some funds.

kianenigma (approved):

PR #192, opened 2024-02-19
Unified tests into final step ✅ 

Bullrich:
Unified all the tests into having one final step at the end.

image

This step will only run if all the previous steps have ran. If any of these tests fail, it will fail.

All the status checks will still appear in the PR, so the user can see any failed case:
image

But, by adding this last test as the only requirement, the list to block the branch becomes significantly smaller.

### DIFF
--- a/.github/workflows/check-migrations.yml
+++ b/.github/workflows/check-migrations.yml
@@ -80,7 +80,7 @@ jobs:
 
   confirmMigrationsPassed:
     runs-on: ubuntu-latest
-    name: Confirms that all migrations passed
+    name: All migrations passed
     # If any new job gets added, be sure to add it to this list
     needs: [check-migrations]
     steps:
--- a/.github/workflows/check-migrations.yml
+++ b/.github/workflows/check-migrations.yml
@@ -16,6 +16,7 @@ concurrency:
 permissions: {}
 
 jobs:
+  # This generates a matrix with all the required jobs which will be run in the next step
   runtime-matrix:
     runs-on: ubuntu-latest
     outputs:
@@ -36,6 +37,8 @@ jobs:
           TASKS=$(echo $TASKS | jq -c .)
           echo "runtime=$TASKS" >> $GITHUB_OUTPUT
 
+  # This runs all the jobs in the matrix. It is required by the "confirmMigrationsPassed" job, so
+  # if they all pass, that job will pass too.
   check-migrations:
     needs: [runtime-matrix]
     continue-on-error: true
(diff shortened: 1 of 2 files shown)


bkontur (approved):

Bullrich:
Here we can see a failed test:
image

It will cause the final action to be skipped, not allowing the user to merge it.

Bullrich:
Only the following action should be required:

image

https://github.com/polkadot-fellows/runtimes/actions/runs/7962468177/job/21737338698?pr=192

acatangiu (approved):

Bullrich:
I just added the same process to migrations:
image

muharem (approved):

skunert (approved):

bkchr:
The problem with this is that we currently have removed the encointer migration check as requirement for merging. How could we achieve the same with this job?

Bullrich:
> The problem with this is that we currently have removed the encointer migration check as requirement for merging. How could we achieve the same with this job?

Uh that will be a problem.

This job would ensure that all the tests pass first, _including the encointer one._

We would have to [remove it from the required tests](https://github.com/polkadot-fellows/runtimes/blob/6334588c9db0115f6b3445ce11d02a457275b34b/.github/workflows/runtimes-matrix.json#L58).

A solution would be to remove it, and then make a new PR trying to fix it (I assume that you don't want to have broken tests in your pipeline) but that's a subpar suggestion from me.

If you prefer I can also disable the required check for migrations and **leave only the one for tests**

bkchr:
> __
> 
> We would have to [remove it from the required tests](https://github.com/polkadot-fellows/runtimes/blob/6334588c9db0115f6b3445ce11d02a457275b34b/.github/workflows/runtimes-matrix.json#L58).

We kept the test for now to remember us about the test. Currently it is just removed from the list of required tests. I assume fetching the list of required tests for the "all X passed" test would be too complicated?

Bullrich:
> We kept the test for now to remember us about the test. Currently it is just removed from the list of required tests. I assume fetching the list of required tests for the "all X passed" test would be too complicated?

Yeah, I'm afraid so.

It can be done, but would require a lot of overhead on the test, and once it is fixed you would have to manually remove such overhead, which I believe it would not be worth the time.

---

On a side topic, this PR can be merged for now, and, _once the tests are fixed, you can set it to have the new `All tests passed` status to be required._

There is no need to require the final step for now.

bkchr:
> It can be done, but would require a lot of overhead on the test, and once it is fixed you would have to manually remove such overhead, which I believe it would not be worth the time.

Could we not parse the required tests from some file?

Bullrich:
> > It can be done, but would require a lot of overhead on the test, and once it is fixed you would have to manually remove such overhead, which I believe it would not be worth the time.
> 
> Could we not parse the required tests from some file?

No, as it works as a matrix and it needs *the whole matrix* to finish.

Unless that we make a new matrix with required tests vs non required tests, but that may be naughty.

To be honest, I would recommend that you remove this failing test, and, then make a PR where you add it again, and work there in an isolated environment.

What advantages brings having a failing test in the repo **versus having a issue to track it**?

bkchr:
@Bullrich I think we can merge this now :see_no_evil:

.github/workflows/check-migrations.yml
```diff
@@ -77,3 +77,11 @@ jobs:
           node-uri: ${{ matrix.runtime.uri }}
           checks: "pre-and-post"
           extra-args: ${{ env.EXTRA_ARGS }}
+
+  confirmMigrationsPassed:
+    runs-on: ubuntu-latest
+    name: All migrations passed
+    # If any new job gets added, be sure to add it to this list
```
bkchr, line 84: You should put some docs at the top of the file to make this more obvious!

Bullrich: Good idea. I properly documented all the steps.

Let me know if there is anything else missing 🙂

pepoviola (approved):

bkchr (approved):

ggwpez (approved):

PR #201, opened 2024-02-26
Lift external deps to the workspace

ggwpez:
Pulling up all external deps to the workspace. Can be reproduced with the latest version of Zepter (1.3.1):

```sh
cargo upgrade -p pallet-xcm@8.0.4
zepter transpose dependency lift-to-workspace "regex:.*" --fix --version-resolver=unambiguous --source-location=remote
```

This is a No-OP change and can be verified with:
```sh
git checkout -q main
cargo tree -e features > main.out

git checkout -q oty-lift-deps-to-workspace
cargo tree -e features | diff main.out -

# No print - empty diff
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8425,9 +8425,9 @@ dependencies = [
 
 [[package]]
 name = "pallet-xcm"
-version = "8.0.3"
+version = "8.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0493c09d29fefc62579284b596618fcceb68fdb3dd38de79fe3c922a19916249"
+checksum = "c13f5c598737e84294880333170d1df3868a11ad7ee79d0b1d1af37365e1c277"
 dependencies = [
  "bounded-collections",
  "frame-benchmarking",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8425,9 +8425,9 @@ dependencies = [
 
 [[package]]
 name = "pallet-xcm"
-version = "8.0.4"
+version = "8.0.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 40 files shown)


Cargo.lock
```diff
@@ -8425,9 +8425,9 @@ dependencies = [
 
 [[package]]
 name = "pallet-xcm"
-version = "8.0.4"
+version = "8.0.3"
```
acatangiu, line 8428: why downgrade `pallet-xcm`? 8.0.4 was a needed patch fix

Cargo.toml
```diff
@@ -5,6 +5,199 @@ edition = "2021"
...
+pallet-sudo = { version = "29.0.0", default-features = false }
+pallet-timestamp = { version = "28.0.0", default-features = false }
+pallet-transaction-payment = { version = "29.0.0", default-features = false }
+pallet-transaction-payment-rpc-runtime-api = { version = "29.0.0", default-features = false }
+pallet-treasury = { version = "28.0.0", default-features = false }
+pallet-uniques = { version = "29.0.0", default-features = false }
+pallet-utility = { version = "29.0.0", default-features = false }
+pallet-vesting = { version = "29.0.0", default-features = false }
+pallet-whitelist = { version = "28.0.0", default-features = false }
+pallet-xcm = { version = "8.0.3", default-features = false }
```
acatangiu, line 132:
```suggestion
pallet-xcm = { version = "8.0.4", default-features = false }
```

xlc (approved):

Cargo.lock
ggwpez: Ah, thanks!  
Looks like i ran the old command, one crate is on that version to i got mixed up.

ordian (approved):

acatangiu (approved):

PR #206, opened 2024-02-29
Refund leases that have not started.

eskimor:
Needed for a smooth migration to Coretime.


seadanda:
If we go for this approach we should be careful to ensure that we have configured enough `MaxLeasedCores` in https://github.com/seadanda/runtimes/blob/donal-coretime-kusama-1.6/system-parachains/coretime/coretime-kusama/src/coretime.rs#L228 as it may push us over what we expect, depending on the auctions.

relay/kusama/src/lib.rs
```diff
@@ -1725,9 +1725,18 @@ pub mod migrations {
 			if lease.is_empty() {
 				return None
 			}
-			// Lease not yet started, ignore:
+			// Lease not yet started/or having holes, refund (coretime can't handle this):
 			if lease.iter().any(Option::is_none) {
-				return None
+				if let Err(err) = slots::Pallet::<Runtime>::clear_all_leases(
```
seadanda, line 1730: If they currently have an active lease which ends in say three months, then also won an auction which gives them a lease starting in three months, then this will kick them off their current lease as well as the one that has not yet started. Surely the intended functionality is that they are refunded for only the lease that has not yet started.
Or can they only swap at the point their current lease ends, i.e. the second lease that has not started is always under a separate `para_id`?

eskimor: In your particular scenario there would be no `None`s .. .you have a lease for the next three months and one an auction for the ones afterwards ... all good. 

In general, what happens here is you either get your lease to coretime or not. If not, you get all your leases refunded - even an ongoing one as in fact you just lost it. So this looks correct to me.

seadanda: Sorry, bad example. Slightly different to get a `None` in there but keeping my question the same:
If they currently have an active lease which ends three months, then also won an auction which gives them a lease starting in four and half months, they have a `None` in there, they'll be refunded their future lease but also booted off their current lease three months earlier than they expected.

Maybe that's a bit contrived, but I don't think a current lease should not be migrated, even if it is refunded, just because they have a gap before the lease they have won starts.

eskimor: But in that case they already have a hole and thus downtime. I would assume if that edge case even exists that some more downtime won't kill them. 

Not saying that this is ideal, but:

(1) I would really assume that this is an edge case that will very likely not even exist. And if so, a solution might be similar to whatever we are going to do about swaps.
(2) We could trim the leases to the first `None`, but then we would have another problem: The leases after the `None`s would keep their token locked, despite the fact that we would have ended their lease on coretime.

seadanda: Missed this the other day, that sounds sensible to me

CHANGELOG.md
```diff
@@ -13,7 +13,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 0022-adopt-encointer-runtime.md) ([polkadot-fellows/runtimes#80](https://github.com/polkadot-fellows/runtimes/pull/80))
 - Feature for enabling debug prints in the Polkadot and Kusama runtime ([polkadot-fellows/runtimes#85](https://github.com/polkadot-fellows/runtimes/pull/85))
 - Added new "Wish for Change" track ([polkadot-fellows/runtimes#184](https://github.com/polkadot-fellows/runtimes/pull/184))
-- Enable Coretime and on-demand on Kusama ([polkadot-fellows/runtimes#159](https://github.com/polkadot-fellows/runtimes/pull/159)
+- Enable Coretime and on-demand on Kusama ([polkadot-fellows/runtimes#159](https://github.com/polkadot-fellows/runtimes/pull/159))
+- Refund any leases, that are not migrated to Coretime (have holes in them/have not yet started) ([polkadot-fellows/runtimes#206](https://github.com/polkadot-fellows/runtimes/pull/206))
```
joepetrowski, line 17:
```suggestion
- Refund any leases that are not migrated to Coretime (have holes in them/have not yet started) ([polkadot-fellows/runtimes#206](https://github.com/polkadot-fellows/runtimes/pull/206))
```

joepetrowski (approved):

acatangiu (approved):

seadanda (approved):

bkchr (approved):

PR #207, opened 2024-02-29
Temporary skip Encointer `check-migrations` (TODO: https://github.com/polkadot-fellows/runtimes/issues/186)

bkontur:

### DIFF
--- a/.github/workflows/check-migrations.yml
+++ b/.github/workflows/check-migrations.yml
@@ -112,7 +112,7 @@ jobs:
               on-runtime-upgrade --checks=pre-and-post \
               $EXTRA_FLAGS \
               live --uri ${{ matrix.runtime.uri }} || \
-              echo "TODO:(https://github.com/polkadot-fellows/runtimes/issues/186): skipping failed `check-migrations` (remove this IF, when the Encointer 1.2.0 is enacted live)"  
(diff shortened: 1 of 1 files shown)


.github/workflows/check-migrations.yml
```diff
@@ -106,8 +106,17 @@ jobs:
 
           echo "Flags: $EXTRA_FLAGS"
 
+          if [[ "${{ matrix.runtime.name }}" == "encointer-kusama" ]]; then
```
ggwpez, line 109: We can just un-require it for merge.

ggwpez:
@bkchr could you please un-require the check-migrations encointer for merge?

bkontur:
> @bkchr could you please un-require the check-migrations encointer for merge?

Haha :D, that's actually what I suggested [before](https://github.com/polkadot-fellows/runtimes/pull/159#issuecomment-1966836443). However, after Basti's [comment](https://github.com/polkadot-fellows/runtimes/pull/178#issuecomment-1971121588), I got the impression that it needs to be done here and not in the GitHub configuration.

Anyway, looks like this hack works: https://github.com/polkadot-fellows/runtimes/actions/runs/8097358985/job/22128262518?pr=207 :D

ggwpez:
Hm i see. I think having it red to still remind of of fixing is still better than having it display green while not working.

bkchr:
> Hm i see. I think having it red to still remind us of fixing is still better than having it display green while not working.

Yeah good idea. Removed the check from the required checks.

PR #208, opened 2024-02-29
Set safe on demand base fee

antonva:
Sets the base fee to a minimum of  0.01 KSM per order. This can be lowered later via governance if deemed too high.


bkchr:
The call is already available on chain. IMO, we should directly do some referendum right now to set it.

PR #187, opened 2024-02-14
Upgrade to latest polkadot-sdk@1.7 release

bkontur:
Based on bump to [`polkadot-sdk@1.6.0`](https://github.com/polkadot-fellows/runtimes/pull/159).

Attached result of `cargo upgrade -v --pinned --incompatible` [cargo-upgrade-version-bump.log](https://github.com/polkadot-fellows/runtimes/files/14281072/cargo-upgrade-version-bump.log)

~~## For reviewers~~

~~This PR is against `polkadot-fellows`'s main to bring it to the fellows repo, but if you want to see a real diff relevant to the `polkadot-sdk@1.7.0` update please check: https://github.com/bkontur/runtimes/compare/bko-bump-to-1.6...bkontur:runtimes:bko-bump-to-1.7.~~

## TODO

- [x] fix compilation
  - [x] apply patch for AssetHub tests: https://github.com/paritytech/polkadot-sdk/pull/3404
- [x] apply patch for `coretime` https://github.com/paritytech/polkadot-sdk/pull/3497 see [comment](https://github.com/polkadot-fellows/runtimes/pull/187#issuecomment-1966287320)
- [x] fix integration tests - fix after https://github.com/polkadot-fellows/runtimes/pull/185
- [x] fix benchmarks (also try them)
  - [x] apply patch for `pallet-xcm` benchmarks  https://github.com/paritytech/polkadot-sdk/pull/3465
- [x] check ranked collectives:
  - [x] `ExchangeOrigin` see [comment](https://github.com/polkadot-fellows/runtimes/pull/187#discussion_r1494171425)
  - [x] `MemberSwappedHandler` / `BenchmarkSetup` see [comment](https://github.com/polkadot-fellows/runtimes/pull/187#discussion_r1494173538)

PR #212, opened 2024-03-01
Add the Kusama Coretime Chain runtime

seadanda:
This is similar to the two chains we deployed to Rococo and Westend, with a different `PriceAdapter` implementation. This is likely to be the most contentious addition here and is obviously not final, but a start until the correct function and parametrisation can be decided upon. Discussion is welcome.

I have also considered adding a minimum factor in `adapt_price` to ensure that we never actually hit 0. Due to the price corrections being purely multiplicative, if the price ever hits zero there is no going back, which I think should be configurable to avoid. I would be interested to hear people's thoughts on this. I can also add this minimum value to the broker pallet upstream (which I think is probably the best place for that).

Due to the Transact XCMs in the `CoretimeInterface`, there is a need for hardcoded weights for several relay chain calls. These should be checked after benchmarking the relay chain. There are hardcoded weights on the Relay Chain which should also be checked when the Coretime Chain is benchmarked. XCM emulator tests are being put together to ensure this happens automatically.

TODO:
- [x] Refine hardcoded relay weights once they are available (likely in the merge of #187). These calls and the hardcoded weights in the relay should be checked after benchmarks are rerun before release
- [x] Add proxy pallet
- [x] Define new `OnRevenue` to burn revenue from Coretime ([RFC-010](https://polkadot-fellows.github.io/RFCs/approved/0010-burn-coretime-revenue.html)) WIP: [here](https://github.com/seadanda/runtimes/tree/donal-coretime-kusama-burn).

---

## Pricing model

The `AdaptPrice` implementation in this PR is designed together with a guideline configuration to provide conservative limits in the early sales to avoid high volatility. I have listed the suggested configuration and outlined the reasoning below.

#### Price adapter behaviour:
`leadin_factor_at`: linear with a max of 5x
`adapt_price`: linear with conservative limits, maximum factor not relevant for our ideal bulk proportion settings, but set to 1.2 as a conservative limit in case we want to change config without a runtime upgrade.

#### Start sales parameters:
`initialPrice`: ~5 KSM 
`coreCount`: 3

#### Configuration parameters:
`advanceNotice`: 10 (1 min)
`interludeLength`: 50400 (7 days)
`leadinLength`: 50400 (7 days)
`regionLength`: 5040 (28 days)
`idealBulkProportion`: 1000000000 (100% for the first few sales)
`limitCoresOffered`: 3
`renewalBump`: 30000000 (3% per sale / ~47% annually)
`contributionTimeout`: 5040 (28 days)

---

### Price adapter reasoning
For the price adapter the main goals are to avoid frontrunning, while also minimising volatility in the early sales. The assumption is that the price will take time to stabilise, and with large corrections we end up flip flopping between values far above and far below the true market price, which will take longer to stabilise the higher those values are between sales. We assume few cores per sale and take into consideration the fact that most cores are occupied by a lease.

The decision to go for a higher lead in factor and a very conservative price adapter are due to the expectation that the early sales will be set such that all cores are sold. A high lead-in max factor allows people to buy at the price they feel comfortable, which affects the market price. This in combination with a conservative price adapter (at the start with no upward correction) means that the price that people are willing to pay is used in the next sale when the cores are sold out, and is adjusted downwards if not all cores are sold. In the case where no cores are sold a minimum factor of 0.5 is applied to gradually bring the price down until a price is found, rather than having something symmetrical which snaps up then can rebound right back down in the extreme case.

### Start sales reasoning
This is purely for the first sale, after which the price adapter takes over.
We suggest a starting value between ~4.4 KSM and ~36.8 KSM depending on the figure we assume for the utilisation. These values represent those for the average (mean) of 13% and for maximum utilisation of 100% respectively. We estimate that the true market price lies between these two values, and so we suggest starting from the average (mean) with a lead-in factor of 5.
This lets the market decide between guardrails of ~13%-65% utilisation in the first sale. We then round up the suggestion to 5 KSM to acknowledge the uncertainty around this value.

This is based on the model set out in [Erin's proposal](https://forum.polkadot.network/t/initial-coretime-pricing/5187/7) and [(also Erin's) Kusama tweak to reflect utilisation](https://forum.polkadot.network/t/initial-coretime-pricing/5187/24).

### Configuration reasoning
```rust
/// The number of Relay-chain blocks in advance which scheduling should be fixed and the
/// `Coretime::assign` API used to inform the Relay-chain.
pub advance_notice: RelayBlockNumber,
```
`advanceNotice`: 10 (1 min) - We require ~10 blocks to send notifyCore messages for 100 cores if the max is 10 XCM messages per block. Note 6s blocks

```rust
/// The length in blocks of the Interlude Period for forthcoming sales.
pub interlude_length: BlockNumber,
```
`interludeLength`: 50400 (7 days) - gives people ample time to renew without reducing sale too much. Note 12s blocks

```rust
/// The length in blocks of the Leadin Period for forthcoming sales.
pub leadin_length: BlockNumber,
```
`leadinLength`: 50400 (7 days) - price drops for 7 days before becoming fixed, then is fixed for just 14 days before we hit the next region. Note 12s blocks

```rust
/// The length in timeslices of Regions which are up for sale in forthcoming sales.
pub region_length: Timeslice,
```
`regionLength`: 5040 (28 days) - this is in timeslices, which are set to 80 blocks in the coretime-kusama runtime. These blocks also are relay chain blocks (6s)

```rust
/// The proportion of cores available for sale which should be sold in order for the price
/// to remain the same in the next sale.
pub ideal_bulk_proportion: Perbill,
```
`idealBulkProportion`: 1000000000 or 400000000 - 100% paired with a small core offering or 40-60% when more people are off their leases and there's a more fluid market with a higher core limit later on - this needs to be modeled before being updated after the first few sales

```rust
/// An artificial limit to the number of cores which are allowed to be sold. If `Some` then
/// no more cores will be sold than this.
pub limit_cores_offered: Option<CoreIndex>,
```
`limitCoresOffered`: 3 initially - This should replace the auctions that were cancelled (coupled with 100% target means we don't cause price spikes)

```rust
/// The amount by which the renewal price increases each sale period.
pub renewal_bump: Perbill,
```
`renewalBump`: 30000000 - 3% every 28 days equates to around 47% annually

```rust
/// The duration by which rewards for contributions to the InstaPool must be collected.
contribution_timeout: Timeslice,
```
`contributionTimeout`: 5040 (28 days) - not much thought behind this one, 1 region to collect seems like a good starting point. This is not relevant until the Coretime Credits are implemented on the relay chain.

### DIFF
--- a/chain-spec-generator/Cargo.toml
+++ b/chain-spec-generator/Cargo.toml
@@ -55,9 +55,6 @@ runtime-benchmarks = [
 	"polkadot-runtime/runtime-benchmarks",
 	"encointer-kusama-runtime/runtime-benchmarks",
 	"glutton-kusama-runtime/runtime-benchmarks",
-<<<<<<< variant A
 	"coretime-kusama-runtime/runtime-benchmarks",
->>>>>>> variant B
 	"people-kusama-runtime/runtime-benchmarks",
-======= end
 ]
--- a/chain-spec-generator/src/main.rs
+++ b/chain-spec-generator/src/main.rs
@@ -90,8 +90,8 @@ fn main() -> Result<(), String> {
 				"coretime-kusama-local",
 				Box::new(|| system_parachains_specs::coretime_kusama_local_testnet_config())
 					as Box<_>,
-            ),
-            (
+			),
+			(
 				"people-kusama-local",
 				Box::new(|| system_parachains_specs::people_kusama_local_testnet_config())
 					as Box<_>,


system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,268 @@
...
+
+/// Implements the [`AdaptPrice`] trait to control the price changes of bulk coretime. This tweaks
+/// the [`pallet_broker::Linear`] implementation which hard-corrects to 0 if cores are offered but
+/// not sold for just one sale. The monotonic lead-in behaviour is unchanged, while the change in
+/// base price between sales has a lower limit of 1/4 to match the upper limit of 4.
+pub struct LinearPlusC;
+impl AdaptPrice for LinearPlusC {
+	fn leadin_factor_at(when: FixedU64) -> FixedU64 {
+		// Start at 2x the base price and decrease to the base price at the end of leadin.
+		FixedU64::from(2).saturating_sub(when)
```
eskimor, line 234: Shall we increase this to 8, as suggested [here](https://forum.polkadot.network/t/initial-coretime-pricing/5187) .. .any downsides?

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,268 @@
...
+	#[cfg(feature = "runtime-benchmarks")]
+	fn ensure_notify_revenue_info(when: RCBlockNumberOf<Self>, revenue: Self::Balance) {
+		CoretimeRevenue::set(&Some((when, revenue)));
+	}
+}
+
+/// Implements the [`AdaptPrice`] trait to control the price changes of bulk coretime. This tweaks
+/// the [`pallet_broker::Linear`] implementation which hard-corrects to 0 if cores are offered but
+/// not sold for just one sale. The monotonic lead-in behaviour is unchanged, while the change in
+/// base price between sales has a lower limit of 1/4 to match the upper limit of 4.
```
eskimor, line 229: Seems like the upper limit is 2?

eskimor (commented):
So we start with an uninitialized [broker configuration](https://github.com/paritytech/polkadot-sdk/blob/2e70dd3bbe2deff61eabfb574a077a2ece04f7c4/substrate/frame/broker/src/types.rs#L256) and then set it via governance?

system-parachains/coretime/coretime-kusama/src/coretime.rs
seadanda: Maximally sold == all cores sold at the start of the lead-in period. The new base price is calculated using the last price, which is double the previous base price when the cores sell out immediately like this, then that last price is doubled again by the `PriceAdapter`.

system-parachains/coretime/coretime-kusama/src/coretime.rs
seadanda: This will lead to a much faster correction in price upwards, but IMO if not properly projected and if we don't have enough cores/a reasonable enough bulk ratio target, a factor that is too high could very easily lead to runaway prices within a few sales. If we were to go for a higher factor like this (a factor of 8 here means that cores being maximally sold leads to a 16x jump in prices), I would maybe consider some sort of exponential decay towards the fixed price for the lead-in calculation.

I think that the number 8 is as arbitrary as the number 2, but I suppose that if we increase the downward correction as well, then the prices will self correct.

I just think that if we were to choose a number that is too high we could end up overshooting and undershooting the "true" market price many times and by large amounts before it stabilises, meaning that it would flip flop between expensive and cheap sales.
The longer term implications would be quite different too, not just when we are approaching the true market price. With a low number of cores offered per sale, if we oversell the target by even one core, the price would be kicked far from the true market price and again potentially take a long time to stabilise again.

I think that trying to set the initial price to something as close as possible to what we expect the true price to be and then keeping the correction factors relatively small would be the ideal scenario.

As I mentioned previously, I think that with higher correction factors (or even if we are uncertain about the true value) we should use some sort of exponential price adapter (as well or instead of the lead-in factor function), which allows small deviations from the ideal to happen without sending us into a long adjustment period, while adjusting the price hard if we sell a number which is far from the target. This needs to be carefully modelled with the number of cores we plan to offer per sale, as the inputs are discrete (number of cores). e.g. if we just threw an exponential at it with large maxima and minima but only offered 2 cores with a 50% ideal, then we would have a huge correction either way if we didn't hit the exact ideal.

A bit dense for a GitHub review response, sorry 😅

eskimor: I am not sure I am following. `adapt_price` seems to be independent of the leadin.

eskimor: Like, sure if people are actually buying all the cores at 8x the price, then it will go up, but well that is in fact the market price then, no?

We seem to be taking the lowest sale price [here](https://github.com/paritytech/polkadot-sdk/blob/2ea6bcf1958b6427cd67eb63968b61782f2becc4/substrate/frame/broker/src/tick_impls.rs#L174) or in case of under performing, the last sale price (which should then be the lowest of the last sale.

system-parachains/coretime/coretime-kusama/src/coretime.rs
eskimor: :exploding_head: :+1:

system-parachains/coretime/coretime-kusama/src/coretime.rs
seadanda: Yes, but then the price that the last core sold at from the previous sale is multiplied by the factor from `adapt_price` - what I'm saying is that the combination of these factors could lead to a spike in price for the next sale to a price that nobody paid in the previous sale.

If it gets to the point that the next sale undersells, it is then corrected downwards. If the corrections are too big up and down it takes longer to get to the point where the "true" market price is found.

I find it difficult to disentangle the lead-in price and price adaption factor (as well as the ideal bulk proportion and number of cores offered) because they are so tightly coupled in how they define the next sale price, I'm not sure that the magnitude of any single one can be discussed in isolation.

system-parachains/coretime/coretime-kusama/src/coretime.rs
seadanda: This might be worded in a way that is not clear, I'll think of a better way to convey that

eskimor: All good, got it.

system-parachains/coretime/coretime-kusama/src/coretime.rs
eskimor: From reading the code, I really don't get how the lead_in factor would lead to that described situation.

If cores get sold out at 8x the set price - overshooting the target, leading to a 2x increase next time ... Sure maybe nobody is willing to pay twice that price, but the situation is exactly the same as if leadin_factor was 1 and we sold out - and arrive at twice the price. In both cases, we doubled what the market was willing to pay last time.

If we think 2x is too much, we should adjust that, but I think it is too early to say that. In any case, if anything a higher lead_in factor would help us finding the real market price _earlier_ if it is higher than what we think, if it is lower, it should change nothing.

On the flipside, the front running concern is valid. If lead_in is high, then it is more likely that you will get a core if you really want it, as long as you are willing to pay enough.

eskimor: Basically a lead_in of only 2 assumes that we have a _real_ good guess on what the price should be. A higher lead_in makes our guess quality less important, which is a good property.

seadanda: Yea I know what you're saying, but why 8x specifically? What drives this specific number?

In your example, the price at the start of the next lead-in period would be 128 times higher than the starting price when we start sales, that's a huge amount in the space of 28 days, and even dropping the price adaption to 1 it's 64 times higher, falling to 8 times. With an analogy to buying a house - just because one person decides to bid 30% over the asking price for a house on the market, it does not mean that the house is worth that much.

The number of cores released in each sale will be small, especially in the beginning. This leaves the prices highly sensitive to what 3 or 4 teams are willing to pay for the first cores. In an extreme but absolutely possible example: if they are wrong by a lot, and then no cores are sold in the next sale due to that, we get a hard correction back to the original base price for the following sale and are no closer to finding the true price.

Increasing these two factors will not find the true market price sooner, but it definitely makes it more volatile. We are just as likely to undershoot and overshoot repeatedly as we correct strongly above and below the true price until we reach the equilibrium point. Taking an initial price, if we assume we are off in our estimate by 5%, 10%, 25%, 50%, 100% and 200% of the true market price, it can actually take longer to converge with higher factors the closer we are to being right with the initial price.

On the other hand if we go for more conservative numbers, the price in a given sale is not as sensitive to what went on only in the previous sale, and we overshoot by less when we pass the true value.

IMO ideally we make the best possible prediction of what the market value will be, set conservative parameters and adapt over time when we have historical data. If we introduce pricing patterns that are too volatile, we may end up having data that is hard to interpret and learn from due to its volatility.

If we really are totally uncertain about the starting price, then setting a higher lead-in price while still maintaining a relatively conservative price adaption might be the way to go.

eskimor: Sounds like we are bikeshedding parameters for a model that is not that great to begin with? In any case, you convinced me that 2 is good enough to give it a shot.

eskimor: Actually, you raised a very good point your e-mail: If we set target to the maximum, then we can not actually overshoot and the factor of 2 can never happen? This would allow for higher lead-in factors?

seadanda: Yea that occurred to me just after replying with my last message when I was thinking about good starting values for the config. I think that will work well at the start and avoid the issue, while still adjusting downwards in the second sale if the first people are overenthusiastic about how much they bid for the first sale.

I think it would be very interesting to explore a factor of 4 for the lead-in with an ideal bulk ratio of 100% and the same lower limit as I have in this impl. I just personally think that 8 seems extreme inside one sale.

eskimor: What about adjusting [`adapt_price`](https://github.com/paritytech/polkadot-sdk/blob/46bd466e48b03d4d5b4ee81c81a0d745c674c8d3/substrate/frame/broker/src/adapt_price.rs#L35) ... I don't actually understand the reason why this needs to move fast at all, especially upwards. This is in essence the minimum price we want ... why should this ever double from one month to the other? Sure if DOT dropped in price by 50%, but still it is just the mimimum price. If DOT is cheap, people are likely to bid more in the lead in, no?  (also in one month time, the price of DOT could be totally different again)

system-parachains/coretime/coretime-kusama/src/lib.rs
```diff
@@ -0,0 +1,880 @@
...
+
+/// Unchecked extrinsic type as expected by this runtime.
+pub type UncheckedExtrinsic =
+	generic::UncheckedExtrinsic<Address, RuntimeCall, Signature, SignedExtra>;
+
+/// Migrations to apply on runtime upgrade.
+pub type Migrations = (
+	cumulus_pallet_xcmp_queue::migration::v4::MigrationToV4<Runtime>,
+	pallet_xcm::migration::MigrateToLatestXcmVersion<Runtime>,
+);
```
joepetrowski, line 115: We shouldn't need any migration at genesis.

system-parachains/coretime/coretime-kusama/src/lib.rs
```diff
@@ -0,0 +1,880 @@
...
+		pub aura: Aura,
+	}
+}
+
+#[sp_version::runtime_version]
+pub const VERSION: RuntimeVersion = RuntimeVersion {
+	spec_name: create_runtime_str!("coretime-kusama"),
+	impl_name: create_runtime_str!("coretime-kusama"),
+	authoring_version: 1,
+	spec_version: 1_006_001,
```
joepetrowski, line 138:
```suggestion
	spec_version: 1_002_000,
```

system-parachains/coretime/coretime-kusama/src/lib.rs
```diff
@@ -0,0 +1,880 @@
...
+			weights.max_total = Some(MAXIMUM_BLOCK_WEIGHT);
+			// Operational transactions have some extra reserved space, so that they
+			// are included even if block reached `MAXIMUM_BLOCK_WEIGHT`.
+			weights.reserved = Some(
+				MAXIMUM_BLOCK_WEIGHT - NORMAL_DISPATCH_RATIO * MAXIMUM_BLOCK_WEIGHT
+			);
+		})
+		.avg_block_initialization(AVERAGE_ON_INITIALIZE_RATIO)
+		.build_or_panic();
+	pub const SS58Prefix: u8 = 42;
```
joepetrowski, line 173:
```suggestion
	pub const SS58Prefix: u8 = 2;
```

system-parachains/coretime/coretime-kusama/src/lib.rs
```diff
@@ -0,0 +1,880 @@
...
+	type ExistentialDeposit = ExistentialDeposit;
+	type AccountStore = System;
+	type WeightInfo = weights::pallet_balances::WeightInfo<Runtime>;
+	type MaxLocks = ConstU32<50>;
+	type MaxReserves = ConstU32<50>;
+	type ReserveIdentifier = [u8; 8];
+	type RuntimeHoldReason = RuntimeHoldReason;
+	type RuntimeFreezeReason = RuntimeFreezeReason;
+	type FreezeIdentifier = ();
+	type MaxFreezes = ConstU32<0>;
```
joepetrowski, line 237: Do any Coretime actions necessitate freezes? Guessing not if it's worked on Rococo/Westend, but just double checking that there are no edge cases or untested features.

system-parachains/coretime/coretime-kusama/src/lib.rs
```diff
@@ -0,0 +1,880 @@
...
+
+		// XCM & related
+		XcmpQueue: cumulus_pallet_xcmp_queue = 30,
+		PolkadotXcm: pallet_xcm = 31,
+		CumulusXcm: cumulus_pallet_xcm = 32,
+		MessageQueue: pallet_message_queue = 34,
+
+		// Handy utilities.
+		Utility: pallet_utility = 40,
+		Multisig: pallet_multisig = 41,
```
joepetrowski, line 460: Worth having a Coretime proxy?

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,280 @@
...
+	AdaptPrice, CoreAssignment, CoreIndex, CoretimeInterface, PartsOf57600, RCBlockNumberOf,
+};
+use parachains_common::{AccountId, Balance, BlockNumber};
+use sp_runtime::{
+	traits::{One, Saturating},
+	FixedU64,
+};
+use xcm::latest::prelude::*;
+
+pub struct CreditToCollatorPot;
```
joepetrowski, line 38: `pub` item should have docs

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,280 @@
...
+
+	fn request_core_count(count: CoreIndex) {
+		use crate::coretime::CoretimeProviderCalls::RequestCoreCount;
+		let request_core_count_call = RelayRuntimePallets::Coretime(RequestCoreCount(count));
+
+		// Weight for `request_core_count` from Kusama runtime benchmarks:
+		// `ref_time` = 7889000 + (3 * 25000000) + (1 * 100000000) = 182889000
+		// `proof_size` = 1636
+		// Add 5% to each component and round to 2 significant figures.
+		let call_weight = Weight::from_parts(190_000_000, 1700);
```
joepetrowski, line 99: Don't we have weight info for the same pallet on this chain? Can we make some assumption like the Relay Chain weight would be no more than 10% of the local weight? And add a test for it :).

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,280 @@
...
+				check_origin: None,
+			},
+			Instruction::Transact {
+				origin_kind: OriginKind::Native,
+				require_weight_at_most: call_weight,
+				call: assign_core_call.encode().into(),
+			},
+		]);
+
+		match PolkadotXcm::send_xcm(Here, Location::parent(), message.clone()) {
```
joepetrowski, line 213: Do you need the `clone` here? It's not needed anymore in this scope.

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,280 @@
...
+		// Add 5% to each component and round to 2 significant figures.
+		let call_weight = Weight::from_parts(190_000_000, 1700);
+
+		let message = Xcm(vec![
+			Instruction::UnpaidExecution {
+				weight_limit: WeightLimit::Unlimited,
+				check_origin: None,
+			},
+			Instruction::Transact {
+				origin_kind: OriginKind::Native,
```
joepetrowski, line 107: I'm surprised this is `Native`, I would have expected `Xcm` with a mapping to some origin or `BodyId`. So just sanity checking, this is what we use on Rococo?

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,280 @@
...
+	fn leadin_factor_at(when: FixedU64) -> FixedU64 {
+		// Start at 2x the base price and decrease to the base price at the end of leadin.
+		FixedU64::from(2).saturating_sub(when)
+	}
+
+	fn adapt_price(sold: CoreIndex, target: CoreIndex, limit: CoreIndex) -> FixedU64 {
+		if sold <= target {
+			// Range of [0.25, 1.0].
+			FixedU64::from_rational(1, 4)
+				.saturating_add(FixedU64::from_rational((3 * sold).into(), (4 * target).into()))
```
joepetrowski, line 253:
```suggestion
				.saturating_add(FixedU64::from_rational((sold.saturating_mul(3)).into(), (target.saturating_mul(4)).into()))
```

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,280 @@
...
+		} else {
+			// Range of (1.0, 2.0].
+			FixedU64::one().saturating_add(FixedU64::from_rational(
+				(sold - target).into(),
+				(limit - target).into(),
```
joepetrowski, line 258: We are certain that `limit < target`? I mean it makes sense but especially as we are dividing by this difference, if it is zero then this could cause problems.

system-parachains/coretime/coretime-kusama/src/xcm_config.rs
```diff
@@ -0,0 +1,310 @@
...
+	ParentIsPreset<AccountId>,
+	// Sibling parachain origins convert to AccountId via the `ParaId::into`.
+	SiblingParachainConvertsVia<Sibling, AccountId>,
+	// Straight up local `AccountId32` origins just alias directly to `AccountId`.
+	AccountId32Aliases<RelayNetwork, AccountId>,
+);
+
+/// Means for transacting the native currency on this chain.
+#[allow(deprecated)]
+pub type CurrencyTransactor = CurrencyAdapter<
```
joepetrowski, line 82: BTW I updated People to use `FungiblesTransactor`. Should do the same and remove deprecated code here?

system-parachains/coretime/coretime-kusama/src/coretime.rs
seadanda: It's a [different pallet](https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/runtime/parachains/src/coretime/mod.rs#L17)

I'm nearly finished the emulated tests for the hardcoded weights (that's the middle TODO in the PR). I originally suggested adding this pallet specifically so we could use the weight with a safety factor but it was pointed out that it just starts sounding like hardcoded weights when you need to increase the safety factor if a test fails.

system-parachains/coretime/coretime-kusama/src/coretime.rs
seadanda: Yes this is what we used on Rococo, these calls need to have the origin of the coretime chain (which is what `Native` converts to in this case)

system-parachains/coretime/coretime-kusama/src/lib.rs
seadanda: The second is a sort of "perennial" migration that we leave in. I'll remove the first, that shouldn't be there

system-parachains/coretime/coretime-kusama/src/coretime.rs
seadanda: In this branch of the `if` we know that sold < target.
The `limit` must be `>= sold` by construction, and thus `target` must be `> limit`.

If you were to use this anywhere else except as a PriceAdapter for the broker it wouldn't be certain I suppose, but it's a pretty tightly coupled interface.

system-parachains/coretime/coretime-kusama/src/lib.rs
seadanda: No, there are no freezes

.cargo/config.toml
```diff
@@ -10,7 +10,7 @@ rustflags = [
   "-Aclippy::all",
   "-Dclippy::correctness",
   "-Aclippy::if-same-then-else",
-  "-Aclippy::clone-double-ref",
+  "-Asuspicious_double_ref_op",
```
seadanda, line 13: Just a note: this is a drive-by update unrelated to the rest of the PR. Without this change a useless warning is emitted for many packages.

system-parachains/coretime/coretime-kusama/src/weights/pallet_broker.rs
```diff
@@ -0,0 +1,518 @@
...
+			.saturating_add(T::DbWeight::get().reads(1))
+			.saturating_add(T::DbWeight::get().writes(1))
+	}
+	/// Storage: `Broker::Leases` (r:1 w:1)
+	/// Proof: `Broker::Leases` (`max_values`: Some(1), `max_size`: Some(401), added: 896, mode: `MaxEncodedLen`)
+	/// Storage: `ParachainSystem::ValidationData` (r:1 w:0)
+	/// Proof: `ParachainSystem::ValidationData` (`max_values`: Some(1), `max_size`: None, mode: `Measured`)
+	/// Storage: `ParachainSystem::LastRelayChainBlockNumber` (r:1 w:0)
+	/// Proof: `ParachainSystem::LastRelayChainBlockNumber` (`max_values`: Some(1), `max_size`: None, mode: `Measured`)
+	fn set_lease() -> Weight {
```
eskimor, line 94: Weights look good, Relay chain configuration:
```
		require_weight_at_most: Weight::from_parts(170_000_000, 20_000),
```
worst case call is reserve with 13k ... still fits.

seadanda: These are copied over from Rococo, so there may be some changes when I run bench bot.

I've been putting together some emulated tests for the relay calls in the coretime chain in case things change drastically any time the weights are regenerated, might be worth having a weight sanity check for these too.

system-parachains/coretime/coretime-kusama/build.rs
```diff
@@ -0,0 +1,34 @@
...
+	substrate_wasm_builder::WasmBuilder::new()
+		.with_current_project()
+		.export_heap_base()
+		.import_memory()
+		.build();
+
+	substrate_wasm_builder::WasmBuilder::new()
+		.with_current_project()
+		.set_file_name("fast_runtime_binary.rs")
+		.enable_feature("fast-runtime")
```
muharem, line 27: Can you tell me when this is useful? When you can pass the env var

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,288 @@
...
+	AdaptPrice, CoreAssignment, CoreIndex, CoretimeInterface, PartsOf57600, RCBlockNumberOf,
+};
+use parachains_common::{AccountId, Balance, BlockNumber};
+use sp_runtime::{
+	traits::{One, Saturating},
+	FixedU64,
+};
+use xcm::latest::prelude::*;
+
+pub struct CreditToCollatorPot;
```
muharem, line 38: There is a generic impl doing the same - https://github.com/paritytech/polkadot-sdk/blob/3ff9a102f2a4075c94d6ed04fb2ca63a4edbb05f/substrate/frame/support/src/traits/tokens/imbalance/on_unbalanced.rs#L76

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,288 @@
...
+		Option<relay_chain::BlockNumber>,
+	),
+}
+
+parameter_types! {
+	pub const BrokerPalletId: PalletId = PalletId(*b"py/broke");
+}
+
+parameter_types! {
+	pub storage CoreCount: Option<CoreIndex> = None;
```
muharem, line 78: not used anywhere?

muharem (approved):

system-parachains/coretime/coretime-kusama/build.rs
seadanda: This is used in zombienet tests. It is used in the Rococo/Westend relay and parachain runtimes (not just coretime). I don't know which env var you're suggesting would do the same thing, maybe there's another way to do it that isn't used in `polkadot-sdk`

system-parachains/coretime/coretime-kusama/src/lib.rs
seadanda: I've thought more about this and actually think that the coretime ownership model means that there is no real meaningful scope to give someone on the coretime chain which is useful but not potentially destructive.

Excluding the ones that need sudo or don't check the origin, the broker calls are:
```rust
`purchase`
`assign`
`renew`
`transfer`
`pool`
`partition`
`interlace`
`claim_revenue`
`purchase_credit`
```

Taking the potentially destructive ones first:
- `assign`: the caller specifies the parachain ID.
- `renew`: the caller specifies the core, so any core could be renewed.
- `transfer`: self explanatory.
- `pool`: the caller defines the account that gets paid for the contribution.
- `purchase_credit`: the caller specifies the beneficiary of the credit.
So none of these can be trustlessly delegated.

- `purchase`: establishes the caller as the owner, but if they can't then assign/transfer/pool then there is no point in them owning the region.
- `partition`/`interlace`: can only be done to the caller's owned regions, and even if that was allowed, then we fall back to the argument for purchasing.

Which leaves the only operation that could be trustlessly delegated: `claim_revenue`, but this can come from any signed origin for a small transaction fee.

seadanda: I'll push what I have since it could still be useful for a (probably multisig) proxy to have access to all Broker calls without division.

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,287 @@
...
+/// not sold for just one sale. The monotonic lead-in is increased in magnitude to enable faster
+/// price finding. The change in base price between sales has a lower limit of 0.5 to allow downward
+/// pressure to be applied, while keeping a conservative upper limit of 1.2 (movements capped at 20%
+/// if cores sell out) to avoid runaway prices in the early sales. The intention is that this will
+/// be coupled with a low number of cores per sale and a 100% ideal bulk ratio for the first sales.
+pub struct LinearPlusC;
+impl AdaptPrice for LinearPlusC {
+	fn leadin_factor_at(when: FixedU64) -> FixedU64 {
+		// Start at 5x the base price and decrease to the base price at the end of leadin.
+		FixedU64::from(5).saturating_sub(when)
```
eskimor, line 248: I think this should be `saturating_sub(4*when)`

eskimor: I think we should have tests here.

system-parachains/coretime/coretime-kusama/src/lib.rs
```diff
@@ -0,0 +1,1013 @@
...
+	fn default() -> Self {
+		Self::Any
+	}
+}
+
+impl InstanceFilter<RuntimeCall> for ProxyType {
+	fn filter(&self, c: &RuntimeCall) -> bool {
+		match self {
+			ProxyType::Any => true,
+			ProxyType::NonTransfer => !matches!(
```
eskimor, line 463: blacklisting instead of whitelisting is extremely brittle.

eskimor (changes requested):
Leadin seems off.

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,287 @@
...
+			),
+			Err(e) => log::error!(
+				target: "runtime::coretime",
+				"Failed to send request to update schedulable cores: {:?}",
+				e
+			),
+		}
+	}
+
+	fn request_revenue_info_at(when: RCBlockNumberOf<Self>) {
```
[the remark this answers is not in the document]
seadanda: Yes I had assumed these would be done for Kusama, I'll just log and not send the XCM

system-parachains/coretime/coretime-kusama/src/coretime.rs
seadanda: 🤦‍♂️

seadanda: Fixed and added tests

system-parachains/coretime/coretime-kusama/src/lib.rs
seadanda: It's implemented this way in `asset-hub`s and `collectives` as the concept itself is a blacklist. I see what you're saying though, that adding a new pallet which contains a transfer-like extrinsic without updating this would be an issue. The same issue applies the other way though, if you whitelist a whole pallet (as is commonly done) and a new extrinsic is added it must be changed here too.
For me I don't mind either way, this is just less code. WDYT?

eskimor: Just flagging, that I think it is a bad idea. Should probably be a ticket on its own.

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,249 @@
...
+		end_hint: Option<RCBlockNumberOf<Self>>,
+	) {
+		use crate::coretime::CoretimeProviderCalls::AssignCore;
+		let assign_core_call =
+			RelayRuntimePallets::Coretime(AssignCore(core, begin, assignment, end_hint));
+
+		// Weight for `assign_core` from Kusama runtime benchmarks:
+		// `ref_time` = 10177115 + (1 * 25000000) + (2 * 100000000) + (80 * 13932) = 236291675
+		// `proof_size` = 3612
+		// Add 5% to each component and round to 2 significant figures.
```
eskimor, line 159: Do we need to be so cheap? Going to high here only limits how many messages we can send - right? I think a safety margin of e.g. 50% would add significantly more robustness against code upgrades.

eskimor (approved):
Looks good from my perspective.

system-parachains/coretime/coretime-kusama/src/coretime.rs
seadanda: I had wondered about this, my reasoning was the following: for `assign_core` we send one of these messages for each core every timeslice, so it's relatively often - but yes only affects the number we can send each time. I didn't work out how many would fit in a block, but since the worst case is 80 assignments per call this is already quite a lot more than the median expected (obviously we don't have numbers for this, but I think it's reasonable to expect that quite a lot of these will be a maximum of a handful of assignments). So really the 5% buffer is 5% on top of the worst case benchmark for the worst case core, but over 100% buffer in the best case of one assignment per core.

That’s for the ref time only, the proof size really is just 5% as it doesn't depend on the number of assignments per core.

I guess we could factor out the part that depends on the number of assignments and multiply that by the length of the actual assignments `Vec` that comes in and slap 50% on that?

eskimor: > core every timeslice

Is this true? If assignments don't change, this is actually not required.

> That’s for the ref time only, the proof size really is just 5% as it doesn't depend on the number of assignments per core.

Why is that? It should, shouldn't it?

Anyhow, your argument that this is already 5% on the absolute worst case, and that we might end up having a lot of these messages makes sense. Any hardcoded value will be brittle anyway.

system-parachains/coretime/coretime-kusama/build.rs
```diff
@@ -0,0 +1,34 @@
...
+		.import_memory()
+		.export_heap_base()
+		.build();
```
bkchr, line 30:
```suggestion
```

bkchr: We don't need this here. In polkadot-sdk we need this because we generate these local network chain specs. Which isn't the case here.

seadanda: Ah didn't consider this. Even with fast-runtime we just use the actual WASM. Removing, thanks.

bkontur:
@seadanda please add it also to the chain-spec-builder

system-parachains/coretime/coretime-kusama/build.rs
```diff
@@ -0,0 +1,26 @@
+// Copyright (C) Parity Technologies (UK) Ltd.
```
kianenigma, line 1: not sure if this is the right license?

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,249 @@
...
+		let staking_pot = CollatorSelection::account_id();
+		let _ = <Balances as Balanced<_>>::resolve(&staking_pot, credit);
+	}
+}
+
+/// A type containing the encoding of the coretime pallet in the Relay chain runtime. Used to
+/// construct any remote calls. The codec index must correspond to the index of `Coretime` in the
+/// `construct_runtime` of the Relay chain.
+#[derive(Encode, Decode)]
+enum RelayRuntimePallets {
```
kianenigma, line 50: I am a little confused why we do this while we are in the same repository, and we have access to the specific `RuntimeCall` of the relay chain as well? There should be a way to fetch the index of a specific variant based on the name, without needing to re-hardcode number 74 here. 

Although truthfully a pallet index will really hardly change.

kianenigma: I mean, I am not saying kusama-coretime runtime should have a kusama runtime nested into it, but there should be a way to bring over some of the constants and types of the kusama relay into individual parachains?

system-parachains/coretime/coretime-kusama/src/coretime.rs
kianenigma: similar comment to `RelayRuntimePallets`, why is the real values not used?

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,249 @@
...
+			Err(e) => log::error!(
+				target: "runtime::coretime",
+				"Core assignment failed to send: {:?}",
+				e
+			),
+		}
+	}
+
+	fn check_notify_revenue_info() -> Option<(RCBlockNumberOf<Self>, Self::Balance)> {
+		let revenue = CoretimeRevenue::get();
```
kianenigma, line 188: Suprised there is no `::take` on this

system-parachains/coretime/coretime-kusama/src/xcm_config.rs
```diff
@@ -0,0 +1,329 @@
...
+
+pub struct ParentOrParentsPlurality;
+impl Contains<Location> for ParentOrParentsPlurality {
+	fn contains(location: &Location) -> bool {
+		matches!(location.unpack(), (1, []) | (1, [Plurality { .. }]))
+	}
+}
+
+/// A call filter for the XCM Transact instruction. This is a temporary measure until we properly
+/// account for proof size weights.
```
kianenigma, line 150: I've seen this comment long ago and am worried that it is something that we are just ignoring. We have proof weights now, is this needed? 

is there a tracking issue for this?

system-parachains/coretime/coretime-kusama/build.rs
seadanda: Ah I hadn't seen the new licenses. I'll fix.

system-parachains/coretime/coretime-kusama/src/xcm_config.rs
seadanda: AFAIK we still need this. CC @joepetrowski @franciscoaguirre

seadanda:
@kianenigma I have been watching that develop with interest. I'm planning to use it as soon as I get back from holiday

system-parachains/coretime/coretime-kusama/src/coretime.rs
seadanda: It's defined with `parameter_types` macro which [doesn't seem to implement take](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/support/src/lib.rs#L336), only get and set (and key)

system-parachains/coretime/coretime-kusama/src/xcm_config.rs
joepetrowski: Yeah I think we need this until XCM v5?

joepetrowski (approved):

joepetrowski:
AFAIK there is just one more thing to address here, viz. handling the burning of Coretime revenue. There is a [branch here](https://github.com/seadanda/runtimes/compare/donal-coretime-kusama...seadanda:runtimes:donal-coretime-kusama-burn?expand=1) that puts the revenue in a temp pot that we can handle "properly" in the next runtime upgrade.

The reason we don't burn funds directly on each sale is that the Relay Chain's `CheckingAccount` tracks all DOT that has been teleported out, so if we burn it on the Coretime chain then it won't actually update the total issuance on the Relay.

IMO the correct solution is to accumulate funds in some temp pot and every period (session, day, sales period, etc.) we send an XCM program to the Relay Chain that will teleport the assets (to withdraw them from `Checking`) and then burn them. We probably don't want to send this program on every single Coretime purchase.

Donal suggested the above branch as a way to hold funds until the next upgrade handles this XCM program.

system-parachains/coretime/coretime-kusama/Cargo.toml
```diff
@@ -0,0 +1,202 @@
...
+description = "Kusama's Coretime parachain runtime"
+repository.workspace = true
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+build = "build.rs"
+
+[dependencies]
+codec = { package = "parity-scale-codec", version = "3.6.1", default-features = false, features = ["derive"] }
```
bkontur, line 12:
```suggestion
codec = { package = "parity-scale-codec", version = "3.6.9", default-features = false, features = ["derive"] }
```

system-parachains/coretime/coretime-kusama/Cargo.toml
```diff
@@ -0,0 +1,202 @@
...
+hex-literal = "0.4.1"
+log = { version = "0.4.20", default-features = false }
+scale-info = { version = "2.9.0", default-features = false, features = ["derive"] }
+serde = { version = "1.0.195", optional = true, features = ["derive"] }
+smallvec = "1.11.1"
```
bkontur, line 17:
```suggestion
scale-info = { version = "2.10.0", default-features = false, features = ["derive"] }
serde = { version = "1.0.196", optional = true, features = ["derive"] }
```

system-parachains/coretime/coretime-kusama/Cargo.toml
```diff
@@ -0,0 +1,202 @@
...
+
+# Substrate
+frame-benchmarking = { default-features = false, optional = true, version = "29.0.0" }
+frame-executive = { default-features = false , version = "29.0.0" }
+frame-support = { default-features = false , version = "29.0.0" }
+frame-system = { default-features = false , version = "29.0.0" }
+frame-system-benchmarking = { default-features = false, optional = true , version = "29.0.0" }
+frame-system-rpc-runtime-api = { default-features = false , version = "27.0.0" }
+frame-try-runtime = { default-features = false, optional = true , version = "0.35.0" }
+pallet-aura = { default-features = false, version = "28.0.0" }
```
bkontur, line 31:
```suggestion
pallet-aura = { default-features = false, version = "28.0.0", features = ["experimental"] }
```
this should make it compilable

system-parachains/coretime/coretime-kusama/Cargo.toml
```diff
@@ -0,0 +1,202 @@
...
+sp-offchain = { default-features = false, version = "27.0.0" }
+sp-runtime = { default-features = false, version = "32.0.0" }
+sp-session = { default-features = false, version = "28.0.0" }
+sp-std = { default-features = false, version = "14.0.0" }
+sp-storage = { default-features = false, version = "20.0.0" }
+sp-transaction-pool = { default-features = false, version = "27.0.0" }
+sp-version = { default-features = false, version = "30.0.0" }
+
+# Polkadot
+pallet-xcm = { default-features = false, version = "8.0.1" }
```
bkontur, line 58:
```suggestion
pallet-xcm = { default-features = false, version = "8.0.3" }
```

system-parachains/coretime/coretime-kusama/Cargo.toml
```diff
@@ -0,0 +1,202 @@
...
+pallet-xcm-benchmarks = { default-features = false, optional = true , version = "8.0.2" }
+polkadot-core-primitives = { default-features = false, version = "8.0.0" }
+polkadot-parachain-primitives = { default-features = false, version = "7.0.0" }
+polkadot-runtime-common = { default-features = false, version = "8.0.1" }
+xcm = { package = "staging-xcm", default-features = false, version = "8.0.0" }
```
bkontur, line 63:
```suggestion
xcm = { package = "staging-xcm", default-features = false, version = "8.0.1" }
```

system-parachains/coretime/coretime-kusama/Cargo.toml
```diff
@@ -0,0 +1,202 @@
...
+xcm-builder = { package = "staging-xcm-builder", default-features = false, version = "8.0.1" }
+xcm-executor = { package = "staging-xcm-executor", default-features = false, version = "8.0.1" }
+
+# Cumulus
+cumulus-pallet-aura-ext = { default-features = false , version = "0.8.0" }
+cumulus-pallet-parachain-system = { default-features = false, features = ["parameterized-consensus-hook",] , version = "0.8.1" }
+cumulus-pallet-session-benchmarking = { default-features = false, version = "10.0.0" }
+cumulus-pallet-xcm = { default-features = false , version = "0.8.0" }
+cumulus-pallet-xcmp-queue = { default-features = false , features = ["bridging"] , version = "0.8.0" }
```
bkontur, line 72:
```suggestion
cumulus-pallet-xcmp-queue = { default-features = false , version = "0.8.0" }
```

bkontur (commented):
please apply Cargo.toml suggestions which should make it compilable

bkontur:
> @seadanda please add it also to the chain-spec-builder

> @seadanda and also it should be added to the `runtimes-matrix.json` (because of release pipeline uses that): https://github.com/polkadot-fellows/runtimes/pull/245/files#diff-ad68d7934f89ae4eaf252d564bbe2e1892b7e1810eddc3dbe53d533a1e07a958R57-R62 (without `uri`)

Fixed here: https://github.com/seadanda/runtimes/pull/2

bkontur:
also I am investigating one failed benchmarks with new weights will be fixed here: https://github.com/seadanda/runtimes/pull/2
```
2024-03-21 11:14:20.632  INFO main frame::benchmark::pallet: Starting benchmark: pallet_broker::start_sales    
2024-03-21 11:14:20.633 ERROR main runtime: panicked at /home/bparity/.cargo/registry/src/index.crates.io-6f17d22bba15001f/frame-system-29.0.0/src/lib.rs:1894:9:
assertion `left == right` failed: expected event <wasm:stripped> is not equal to the last event <wasm:stripped>
  left: <wasm:stripped>
 right: <wasm:stripped>    
Error: 
   0: Invalid input: Error executing and verifying runtime benchmark: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed
      WASM backtrace:
      error while executing at wasm backtrace:
          0: 0x4cda - <unknown>!rust_begin_unwind
          1: 0x2481 - <unknown>!core::panicking::panic_fmt::hbb0c5dd9c7ebab99
          2: 0x5a9d - <unknown>!core::panicking::assert_failed_inner::h96e1bd7bd3a6b9b7
          3: 0x284e70 - <unknown>!core::panicking::assert_failed::hb0c4fae7bcf657cd
          4: 0x281904 - <unknown>!frame_system::<impl frame_system::pallet::Pallet<T>>::assert_last_event::heed0b4ca0fe4794b
          5: 0x2bac76 - <unknown>!core::ops::function::FnOnce::call_once{{vtable.shim}}::h403ba4fe8f41ce76
          6: 0x2383be - <unknown>!<coretime_kusama_runtime::Runtime as frame_benchmarking::utils::runtime_decl_for_benchmark::BenchmarkV1<sp_runtime::generic::block::Block<sp_runtime::generic::header::Header<u32,sp_runtime::traits::BlakeTwo256>,sp_runtime::generic::unchecked_extrinsic::UncheckedExtrinsic<sp_runtime::multiaddress::MultiAddress<<<sp_runtime::MultiSignature as sp_runtime::traits::Verify>::Signer as sp_runtime::traits::IdentifyAccount>::AccountId,()>,coretime_kusama_runtime::RuntimeCall,sp_runtime::MultiSignature,(frame_system::extensions::check_non_zero_sender::CheckNonZeroSender<coretime_kusama_runtime::Runtime>,frame_system::extensions::check_spec_version::CheckSpecVersion<coretime_kusama_runtime::Runtime>,frame_system::extensions::check_tx_version::CheckTxVersion<coretime_kusama_runtime::Runtime>,frame_system::extensions::check_genesis::CheckGenesis<coretime_kusama_runtime::Runtime>,frame_system::extensions::check_mortality::CheckMortality<coretime_kusama_runtime::Runtime>,frame_system::extensions::check_nonce::CheckNonce<coretime_kusama_runtime::Runtime>,frame_system::extensions::check_weight::CheckWeight<coretime_kusama_runtime::Runtime>,pallet_transaction_payment::ChargeTransactionPayment<coretime_kusama_runtime::Runtime>)>>>>::dispatch_benchmark::h356a9fad89f80485
          7: 0x2f1aeb - <unknown>!Benchmark_dispatch_benchmark

Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
Run with RUST_BACKTRACE=full to include source snippets.
```

bkontur:
> also I am investigating one failed benchmarks with new weights will be fixed here: [seadanda#2](https://github.com/seadanda/runtimes/pull/2)
> 
> ```
> 2024-03-21 11:14:20.632  INFO main frame::benchmark::pallet: Starting benchmark: pallet_broker::start_sales    
> 2024-03-21 11:14:20.633 ERROR main runtime: panicked at /home/bparity/.cargo/registry/src/index.crates.io-6f17d22bba15001f/frame-system-29.0.0/src/lib.rs:1894:9:
> assertion `left == right` failed: expected event <wasm:stripped> is not equal to the last event <wasm:stripped>
>   left: <wasm:stripped>
>  right: <wasm:stripped>    
> Error: 
>    0: Invalid input: Error executing and verifying runtime benchmark: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed
>       WASM backtrace:
>       error while executing at wasm backtrace:
>           0: 0x4cda - <unknown>!rust_begin_unwind
>           1: 0x2481 - <unknown>!core::panicking::panic_fmt::hbb0c5dd9c7ebab99
>           2: 0x5a9d - <unknown>!core::panicking::assert_failed_inner::h96e1bd7bd3a6b9b7
>           3: 0x284e70 - <unknown>!core::panicking::assert_failed::hb0c4fae7bcf657cd
>           4: 0x281904 - <unknown>!frame_system::<impl frame_system::pallet::Pallet<T>>::assert_last_event::heed0b4ca0fe4794b
>           5: 0x2bac76 - <unknown>!core::ops::function::FnOnce::call_once{{vtable.shim}}::h403ba4fe8f41ce76
>           6: 0x2383be - <unknown>!<coretime_kusama_runtime::Runtime as frame_benchmarking::utils::runtime_decl_for_benchmark::BenchmarkV1<sp_runtime::generic::block::Block<sp_runtime::generic::header::Header<u32,sp_runtime::traits::BlakeTwo256>,sp_runtime::generic::unchecked_extrinsic::UncheckedExtrinsic<sp_runtime::multiaddress::MultiAddress<<<sp_runtime::MultiSignature as sp_runtime::traits::Verify>::Signer as sp_runtime::traits::IdentifyAccount>::AccountId,()>,coretime_kusama_runtime::RuntimeCall,sp_runtime::MultiSignature,(frame_system::extensions::check_non_zero_sender::CheckNonZeroSender<coretime_kusama_runtime::Runtime>,frame_system::extensions::check_spec_version::CheckSpecVersion<coretime_kusama_runtime::Runtime>,frame_system::extensions::check_tx_version::CheckTxVersion<coretime_kusama_runtime::Runtime>,frame_system::extensions::check_genesis::CheckGenesis<coretime_kusama_runtime::Runtime>,frame_system::extensions::check_mortality::CheckMortality<coretime_kusama_runtime::Runtime>,frame_system::extensions::check_nonce::CheckNonce<coretime_kusama_runtime::Runtime>,frame_system::extensions::check_weight::CheckWeight<coretime_kusama_runtime::Runtime>,pallet_transaction_payment::ChargeTransactionPayment<coretime_kusama_runtime::Runtime>)>>>>::dispatch_benchmark::h356a9fad89f80485
>           7: 0x2f1aeb - <unknown>!Benchmark_dispatch_benchmark
> 
> Backtrace omitted. Run with RUST_BACKTRACE=1 environment variable to display it.
> Run with RUST_BACKTRACE=full to include source snippets.
> ```

fixed by https://github.com/seadanda/runtimes/pull/2/commits/648eddeec89fed6c1e005aaf63dc3aa3cbb58897

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,252 @@
...
+
+/// Implements the [`AdaptPrice`] trait to control the price changes of bulk coretime. This tweaks
+/// the [`pallet_broker::Linear`] implementation which hard-corrects to 0 if cores are offered but
+/// not sold for just one sale. The monotonic lead-in is increased in magnitude to enable faster
+/// price finding. The change in base price between sales has a lower limit of 0.5 to allow downward
+/// pressure to be applied, while keeping a conservative upper limit of 1.2 (movements capped at 20%
+/// if cores sell out) to avoid runaway prices in the early sales. The intention is that this will
+/// be coupled with a low number of cores per sale and a 100% ideal bulk ratio for the first sales.
+pub struct PriceAdapter;
+impl AdaptPrice for PriceAdapter {
```
Szegoo, line 207: Given that the implementation of the price adapter id different from the ones utilized in Rococo and Westend, it becomes even more important to expose a runtime api for querying the price. This will prevent mistakes in offchain code that is connecting to the Coretime chain: https://github.com/paritytech/polkadot-sdk/pull/3485

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -0,0 +1,252 @@
...
+				(limit - target).saturating_mul(5).into(),
+			))
+		}
+	}
+}
+
+impl pallet_broker::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type Currency = Balances;
+	type OnRevenue = CreditToCollatorPot;
```
bkchr, line 236: We had a RFC for burning the fees. Questionable if we should/want to enable this without a public referendum. Maybe we should do this and highlight in the upgrade that this will burn the fees. Then the community could reject it if not wanted.

bkchr: Another question, don't we need to register the coretime chain as a location that supports teleporting the native token? Or is this automatically done by it being a system chain? CC @joepetrowski @bkontur

system-parachains/coretime/coretime-kusama/src/xcm_config.rs
bkchr: That should be resolved by using the new mq pallet or @ggwpez?

system-parachains/coretime/coretime-kusama/src/coretime.rs
joepetrowski: It should be a teleporter, it wouldn't make a difference in burning though AFAICT.

But Donal already has a branch that burns it, we should just merge that into this (see https://github.com/polkadot-fellows/runtimes/pull/212#issuecomment-2011307981)

seadanda: The branch doesn't actually burn the funds yet, just sets it aside in a holding account.
This way it can be burnt by a future version, so we'd have a runtime upgrade which is more directly about burning.

joepetrowski: Ah right, but we don't want to merge this with crediting revenue to the collator pot. Holding account "to be burned" is OK IMO.

bkontur: > Another question, don't we need to register the coretime chain as a location that supports teleporting the native token? Or is this automatically done by it being a system chain? CC @joepetrowski @bkontur


coretime is correctly configured to trust other system parachains + relaychain for KSM teleports (`type IsTeleporter = ConcreteAssetFromSystem<KsmRelayLocation>;`), and also other system parachains trust coretime for KSM teleports (they also use `ConcreteAssetFromSystem`) and relaychain is also [configured](https://github.com/polkadot-fellows/runtimes/blob/main/relay/kusama/src/xcm_config.rs#L148) to trust coretime with KSM teleports

seadanda: Holding account merged in [e373dec](https://github.com/polkadot-fellows/runtimes/commit/1a1b2ea9bd3efc21fb517b5cd22b053c8e373dec).

bkchr (approved):

Issue #216, opened 2024-03-05
Zombienet: KusamaAssetHub and KusamaBridgeHub are not producing blocks


serban300:
KusamaAssetHub and KusamaBridgeHub are not producing blocks when started in zombienet

Using a `git blame` like approach, I reached this commit as the one who introduced the issue: https://github.com/polkadot-fellows/runtimes/commit/59172d5fb8feea2646490618bead3b2b4fb06220

The zombienet test that I'm using: https://github.com/polkadot-fellows/runtimes/blob/acf1797c84cc14935ccfd8d34043ef5e20a3de08/integration-tests/bridges/environments/polkadot-kusama/bridge_hub_kusama_local_network.toml

bkontur:
I don't see anything suspicious e.g. for AssetHubKusama [lib.rs](https://github.com/polkadot-fellows/runtimes/commit/59172d5fb8feea2646490618bead3b2b4fb06220#diff-80668ff8e793b64f36a9a3ec512df5cbca4ad448c157a5d81abda1b15f35f1da) and also for chain-spec-generator there is just encointer-related [change](https://github.com/polkadot-fellows/runtimes/commit/59172d5fb8feea2646490618bead3b2b4fb06220#diff-7a4c15e83e4faeffcd759ccc4f2d9d874386bc0103cb898823db093c1ff5c05e).

just a blind guess, try add more collators (3-4)?

bkontur:
Relay chain produces blocks? I don't know, isn't there anything in the logs?

serban300:
Yes, relay chain produces blocks. I don't see anything worth mentioning in the logs

bkontur:
so I would guess that it could be related to the coretime stuff maybe:
https://github.com/polkadot-fellows/runtimes/commit/59172d5fb8feea2646490618bead3b2b4fb06220#diff-efa4caeb17487ecb13d8f5eb7863c3241d84afa2e73fbf25909a2ca89df0f362

because this was introduced in that commit just for Kusama, so maybe some coretime magic should happen?
 @seadanda @eskimor any ideas?

serban300:
Yes, it works if I undo the `coretime`-related changes

@seadanda @eskimor can you PTAL ?

eskimor:
Can we make these Zombienet tests part of the CI?

serban300:
[...]

PR #221, opened 2024-03-07
Fix Kusama coretime config

serban300:

Closes https://github.com/polkadot-fellows/runtimes/issues/216

Credits to @bkchr for the suggested fix

### DIFF
--- a/relay/kusama/src/lib.rs
+++ b/relay/kusama/src/lib.rs
@@ -1265,7 +1265,7 @@ impl parachains_paras::Config for Runtime {
 	type QueueFootprinter = ParaInclusion;
 	type NextSessionRotation = Babe;
 	type OnNewHead = Registrar;
-	type AssignCoretime = ();
+	type AssignCoretime = CoretimeAssignmentProvider;
 }
 
 parameter_types! {


eskimor (approved):

bkontur (approved):

acatangiu (approved):

acatangiu:
@eskimor @sandreim I want to point you to https://github.com/polkadot-fellows/runtimes/pull/198 which is the only existing test at this point (indirectly) validating block production.

Without ☝️ we might have shipped this bug to Kusama and stall parachains.

Asking for a bit of love in the shape of a review of above 😄

eskimor:
> @eskimor @sandreim I want to point you to #198 which is the only existing test at this point (indirectly) validating block production.
> 
> Without ☝️ we might have shipped this bug to Kusama and stalled parachains.
> 
> Asking for a bit of love in the shape of a review of above 😄

If I am not completely mistaken, the above fix only fixes Zombienet tests. It should have no effect on production. But indeed, running the runtime in Zombienet should be the bare minimum we should do for a release!

PR #228, opened 2024-03-09
Enable async backing for system parachains: Phase 1

s0me0ne-unkn0wn:
This PR aims to prepare system parachains' runtimes for the async backing enabling. Namely, it:
* Implements `AuraUnincludedSegmentApi` for every runtime;
* Allows relay block numbers to increase monotonically (as opposed to increasing strictly) to enable producing more than one parablock on top of every relay block;
* Enables constant Aura slot durations.


joepetrowski (approved):

joepetrowski:
@georgepisaltu could you take a look too?

alexggh:
Are there any node equivalent changes ? And are there any dependencies between them, do they need to be deployed in a certain order ?

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
```diff
@@ -417,7 +419,7 @@ impl pallet_aura::Config for Runtime {
 	type MaxAuthorities = ConstU32<100_000>;
 	type AllowMultipleBlocksPerSlot = ConstBool<false>;
 	#[cfg(feature = "experimental")]
```
bkchr, line 421: This feature is not enabled. 

We should really remove this feature, here and in polkadot-sdk. For now we can remove it here and enable for the crates directly.

s0me0ne-unkn0wn:
@alexggh Yes, for both questions.

This is phase 1; that is, it doesn't change any behavior of the runtime or node, but it exposes new interfaces (namely `AuraUnincludedSegmentApi`) and otherwise prepares the runtime for async backing, relaxing the relay block number incrementation rules, etc.

Phase 2 will involve updating the node to use those interfaces (that is done in paritytech/polkadot-sdk#3630; testing and description are in progress). After phase 1 is enacted, we will merge phase 2 and start upgrading the collators. After that is done, everything will be async backing ready but still running at 12-sec blocktimes.

After that, at phase 3, one more update to the parachain runtimes will be made, bumping constants (slot duration, capacity, and velocity) to enable async backing with 6-sec blocktimes.

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
s0me0ne-unkn0wn: Do you mean to remove its usage from the code or get rid of it completely in manifests?

bkchr: For this repo you should remove the feature from the code entirely. Enable the feature by default for the crate. 

In Polkadot-sdk, get rid off the feature completely. It was added randomly to guard something now, it is there while all the other code changes of async backing just got merged as is.

s0me0ne-unkn0wn: Please check the last commit, hope I got it right.

bkchr (approved):

eskimor (approved):

s0me0ne-unkn0wn:
Seems like it requires a re-review after manually merging `main` although the only change was in Changelog :shrug:

ordian (approved):

joepetrowski (approved):

eskimor (approved):

PR #235, opened 2024-03-12
Keep `im-online` events decodable after pallet removal

s0me0ne-unkn0wn:
This is a follow-up to #178, which aims to keep the `im-online` pallet events decodable after the pallet itself is removed.

As [discussed](https://github.com/polkadot-fellows/runtimes/pull/178#issuecomment-1977608363), the inability to decode events may result in missing a candidate in approval voting and a parachain finality stall. Although the circumstances of such an event are unlikely, it's better to safeguard them.

The transient code introduced in this PR should be removed after the upgrade is enacted.


eskimor (approved):
Looks sensible to me, assuming it was tested to actually work. (E.g. does the pallet path matter?)

s0me0ne-unkn0wn:
> assuming it was tested to actually work

Yes, I reproduced the problem in a local synthetic test and made sure it was resolved with this patch. Paths do not matter, TypeInfo only includes names, I'm not sure it's even critical in this case, but I still named the thing `pallet_im_online` to be on the safe side.

ordian (approved):
This is a reasonable workaround. I assume we can safely remove this mock pallet after the runtime upgrade is applied (let's create an issue to not forget)?

s0me0ne-unkn0wn:
> I assume we can safely remove this mock pallet after the runtime upgrade is applied (let's create an issue to not forget)?

Fair enough, raised #242

bkchr:
Thank you! Good approach!

PR #246, opened 2024-03-18
Phase 2 of People Chain Launch

joepetrowski:
Do not merge until after the 1.2.0 release. These runtimes should be applied first to the Kusama Relay Chain (after People genesis) and then to the People Chain (after migration is complete).

See details: https://forum.polkadot.network/t/people-chain-launch-and-identity-migration-plan/5930


eskimor (approved):

acatangiu (approved):

ggwpez (changes requested):
Blocking until we are sure that this does not need to be done again: https://github.com/polkadot-fellows/runtimes/pull/263  
@s0me0ne-unkn0wn do we have to bump the versions again?

s0me0ne-unkn0wn:
> @s0me0ne-unkn0wn do we have to bump the versions again?

We're limiting the migration by the version FROM which we're upgrading, so it doesn't matter if we upgrade 1.1.3 -> 1.2.0 or 1.1.3 -> 1.2.1, it will work fine. And it will not trigger the migration again if we upgrade 1.2.0 -> 1.2.1. So everything is sound IIUC.

joepetrowski:
> We're limiting the migration by the version FROM which we're upgrading, so it doesn't matter if we upgrade 1.1.3 -> 1.2.0 or 1.1.3 -> 1.2.1, it will work fine. And it will not trigger the migration again if we upgrade 1.2.0 -> 1.2.1. So everything is sound IIUC.

These changes need to be applied after 1.2.0, so it will definitely be 1.2.0 -> 1.2.1. So should we branch off from v1.2.0 or is it OK to release this from master with 263 in?

ggwpez (approved):
So it should be fine as long as we dont have to retract one of the `1.1.3 -> *` runtime upgrades.

>These changes need to be applied after 1.2.0, so it will definitely be 1.2.0 -> 1.2.1. So should we branch off from v1.2.0 or is it OK to release this from master with 263 in?

263 is in 1.2.0 already, should be as intended.

joepetrowski:
> 263 is in 1.2.0 already, should be as intended.

Ah I lazily did not check, I just saw it was after the "Triggers 1.2.0 Release" commit.

Issue #240, opened 2024-03-15
[CI] Install version locked Rust


ggwpez:
https://github.com/polkadot-fellows/runtimes/actions/runs/8295745077/job/22703405791

CI suddenly fails on old release branch that we want to patch up, probably since it is installing a too new nightly.  
Should use a hard-coded version. cc @alvicsam

PR #247, opened 2024-03-19
ci: install rust with pinned version

alvicsam:
PR installs rust in CI with pinned version

close https://github.com/polkadot-fellows/runtimes/issues/240


.github/workflows/test.yml
```diff
@@ -177,19 +175,16 @@ jobs:
...
 
       - name: Install nightly toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: nightly
-          target: wasm32-unknown-unknown
+        run: |
+          rustup toolchain install "nightly-${{ env.RUST_NIGHTLY_VERSION }}" --profile minimal --component rust-src
+          rustup target add wasm32-unknown-unknown --toolchain "nightly-${{ env.RUST_NIGHTLY_VERSION }}"
```
ggwpez, line 187: Not sure if we even need nightly for this job as it only uses `cargo test`?

ggwpez (approved):

.github/workflows/test.yml
alvicsam: It seems we don't, removed in https://github.com/polkadot-fellows/runtimes/pull/247/commits/9373fde97864b53aff55bee83d01d7e47e7160c3

.github/workflows/fmt.yml
```diff
@@ -6,21 +6,19 @@ on:
   pull_request:
   workflow_dispatch:
 
+env:
+  RUST_NIGHTLY_VERSION: 2024-03-14 # 1.78.0-nightly
```
bkchr, line 10: Could we please put these env variables into one file? So we don't always need to touch all the files. We can then also relax the review requirements on this file to the same level as the normal source files.

bkchr (approved):

.github/workflows/fmt.yml
bkchr: Ty!

andresilva (approved):

PR #251, opened 2024-03-21
Collectives tests: Teleports and Whitelist Call

muharem:
Introduce teleport and whitelist call tests for Polkadot Collectives.

Based on https://github.com/polkadot-fellows/runtimes/pull/233


integration-tests/emulated/helpers/src/lib.rs
```diff
@@ -123,6 +123,184 @@ macro_rules! test_parachain_is_trusted_teleporter {
 	};
 }
 
+#[macro_export]
+macro_rules! test_relay_is_trusted_teleporter {
```
muharem, line 127: In the analogy with `test_parachain_is_trusted_teleporter` macro from above. 
There are two different ways now to test teleports, using the macro from above and using xcm-emulator::Test. I find the macro option more clear and easy to read compered to the `Test`.
I propose to rename the existing `test_parachain_is_trusted_teleporter` to `test_sibling_is_trusted_teleporter`, and have two additional `test_relay_is_trusted_teleporter` and `test_parachain_is_trusted_teleporter`

bkontur: Actually, iirc, the plan is/was to remove this crate from the fellows and re-use macros from `polkadot-sdk` to avoid unnecessary duplications.
I don't remember exactly was was missing in 1.7.0, but @acatangiu just temporary added it [here](https://github.com/polkadot-fellows/runtimes/pull/187/commits/111f5c97f9b91e053842e576048bd1ba662d328b#diff-7346706278168c97dfec9e73bec1a115274a2be89a3147b914dd2eedec9f6269R31-R32).
See the comment https://github.com/polkadot-fellows/runtimes/pull/251/files#diff-7346706278168c97dfec9e73bec1a115274a2be89a3147b914dd2eedec9f6269R31-R32,
```
/// TODO: when bumping to polkadot-sdk v1.8.0,
/// remove this crate altogether and get the macros from `emulated-integration-tests-common`.
```


also I can see `test_chain_can_claim_assets` was added here also... @franciscoaguirre Cisco, do we need/want to move it to `polkadot-sdk` and setup those tests there or maybe do we have an issue for that?

bkontur: From what I see, `test_parachain_is_trusted_teleporter` and `test_sibling_is_trusted_teleporter` do almost exactly the same. Besides multi-receiver support, the only significant difference I can see is just asset construction: `let assets: Assets = (Parent, $amount).into();`. I like your multi-receiver support and assets as input parameters, it makes it much more generic.

So, I would vote here for keeping just one macro (temporarily, one nice generic macro here and port it asynchronously to `polkadot-sdk` for later removal from fellows) with multi-receiver support + the possibility to set assets as input parameters.

muharem: Yes, I agree. But until we have it in polkadot-sdk we can keep this, and remove when we have macro to replace it.

bkontur: Yes, so at least TODO+issue maybe?
Well, I suggest to pimp that macro to its final version here, and then backport it to `polkadot-sdk`. This way, when bumping with `polkadot-sdk` later, we'll only need to just change imports without causing much disruption. However, please feel free to disregard this suggestion.

integration-tests/emulated/tests/collectives/collectives-polkadot/src/tests/teleport.rs
```diff
@@ -0,0 +1,70 @@
...
+use collectives_polkadot_runtime::xcm_config::XcmConfig as CollectivesPolkadotXcmConfig;
+use frame_support::assert_ok;
+use integration_tests_helpers::{
+	test_parachain_is_trusted_teleporter, test_relay_is_trusted_teleporter,
+	test_sibling_is_trusted_teleporter,
+};
+use polkadot_runtime::xcm_config::XcmConfig as PolkadotXcmConfig;
+
+#[test]
+fn teleport_from_relay_to_relay() {
```
joepetrowski, line 27:
```suggestion
fn teleport_from_and_to_relay() {
```

integration-tests/emulated/tests/collectives/collectives-polkadot/src/tests/teleport.rs
```diff
@@ -0,0 +1,70 @@
...
+	test_parachain_is_trusted_teleporter!(
+		CollectivesPolkadot,          // Origin
+		CollectivesPolkadotXcmConfig, // XCM Configuration
+		Polkadot,                     // Destination
+		amount
+	);
+}
+
+#[test]
+fn teleport_to_asset_hub_from_collectives() {
```
joepetrowski, line 47:
```suggestion
fn teleport_from_collectives_to_asset_hub() {
```

integration-tests/emulated/tests/collectives/collectives-polkadot/src/tests/teleport.rs
```diff
@@ -0,0 +1,70 @@
...
+	test_sibling_is_trusted_teleporter!(
+		CollectivesPolkadot,          // Origin
+		CollectivesPolkadotXcmConfig, // XCM Configuration
+		vec![AssetHubPolkadot],       // Destinations
+		(native_asset, amount)
+	);
+}
+
+#[test]
+fn teleport_to_collectives_from_asset_hub() {
```
joepetrowski, line 60:
```suggestion
fn teleport_from_asset_hub_to_collectives() {
```

joepetrowski (approved):

integration-tests/emulated/helpers/src/lib.rs
muharem: I tried and I do not like the result. I have to remove some event asserts and leave only the ones which relevant for all cases. Also I have to pass some argument and resolve the xcm pallet's name from it, because they are different on Relay and Paras. Probably it can be still improved, but I think it's good enough for now. Having separate macro also leaves some flexibility for later changes.

joepetrowski (approved):

integration-tests/emulated/tests/collectives/collectives-polkadot/src/tests/fellowship.rs
```diff
@@ -0,0 +1,72 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::*;
+use codec::Encode;
+use collectives_polkadot_runtime::fellowship::pallet_fellowship_origins::Origin::Fellows as FellowsOrigin;
+use frame_support::{assert_ok, sp_runtime::traits::Dispatchable};
+
+#[test]
+fn fellows_whitelist_call() {
```
acatangiu, line 22: nice! 👍

acatangiu (approved):

acatangiu:
needs a clippy fix and a CHANGELOG entry and is good to go

joepetrowski (approved):

PR #237, opened 2024-03-14
Change NFT Deposits According to RFC-45

joepetrowski:
As specified in [RFC 45](https://github.com/polkadot-fellows/RFCs/blob/main/text/0045-nft-deposits-asset-hub.md#enhanced-approach-to-further-lower-barriers-for-entry)

PR #259, opened 2024-03-28
Change Asset Class Creation Deposits

joepetrowski:
Change the asset class creation deposits to use standard storage deposit functionality.

Similar to https://github.com/polkadot-fellows/runtimes/pull/237

### DIFF
--- a/system-parachains/asset-hubs/asset-hub-polkadot/primitives/src/lib.rs
+++ b/system-parachains/asset-hubs/asset-hub-polkadot/primitives/src/lib.rs
@@ -24,7 +24,7 @@ use xcm::prelude::*;
 
 pub use bp_xcm_bridge_hub_router::XcmBridgeHubRouterCall;
 
-use system_parachains_constants::polkadot::currency::UNITS;
+use system_parachains_constants::polkadot::currency::*;
 
 /// `AssetHubPolkadot` Runtime `Call` enum.
 ///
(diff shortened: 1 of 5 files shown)


system-parachains/asset-hubs/asset-hub-polkadot/primitives/src/lib.rs
```diff
@@ -52,7 +52,7 @@ frame_support::parameter_types! {
 	pub UncongestedMessage: Xcm<()> = build_congestion_message(false).into();
 
 	/// Should match the `AssetDeposit` of the `ForeignAssets` pallet on Asset Hub.
-	pub const CreateForeignAssetDeposit: u128 = 10 * UNITS;
+	pub const CreateForeignAssetDeposit: u128 = system_para_deposit(1, 190);
```
joepetrowski, line 55: cc @claravanstaden @acatangiu

shawntabrizi: 190 is based on the bytes? there should be some way to get that number without hardcoding no?

with `max_encoded_len`?

shawntabrizi (approved):

muharem (approved):

claravanstaden (approved):

system-parachains/asset-hubs/asset-hub-polkadot/primitives/src/lib.rs
joepetrowski: It is the encoded length of [`AssetDetails`](https://github.com/paritytech/polkadot-sdk/blob/bc2e5e1/substrate/frame/assets/src/types.rs#L52) in this configuration.

I'm not sure if we can use `max_encoded_len` here? I calculated it by writing a test and asserting `encoded_len`.

bkchr: If you remove the const it should work.

bkchr (approved):

PR #261, opened 2024-03-28
kusama chains: allow arbitrary XCM execution

acatangiu:
Kusama relay is the only chain here allowing arbitrary XCM execution.

This was enabled in https://github.com/paritytech/polkadot/pull/4809 for (afaict) allowing crafting and executing XCM DOT transfers. Now, these are done through dedicated extrinsics and manual `pallet_xcm::execute` is not required (for transfers at least).

AFAICT, allowing `Everything` on Kusama should also be safe at this point (XCM is pretty mature) and fosters experimentation and more flexibility on Kusama. But in that case, maybe we should also allow it on Kusama system chains?

Thoughts?


bkchr:
AFAIK xcm execution on the relay chain was always safe? On the parachains the problem was just around proof size, which should now be fixed and we could actually enable arbitrary xcm on the parachains.

Szegoo:
> On the parachains the problem was just around proof size

We also have `SafeCall` filter on parachains so there probably shouldn't be any issue with allowing generic execution.

ggwpez (approved):
cc @franciscoaguirre

muharem (approved):

PR #263, opened 2024-03-29
Bump versions for sessions keys migration

### LINKED ISSUE #262: `im-online` session keys migration will not run on Kusama
We have a Kusama runtime upgrade to 1.1.3 on track that will be enacted soon: https://kusama.subsquare.io/referenda/359

That will prevent the execution of the session keys migration in 1.2.0 release: https://github.com/polkadot-fellows/runtimes/blob/2dd830a4e30538836198a4d4eaf68da686a0ef3e/relay/kusama/src/lib.rs#L1863

CC @bkchr

s0me0ne-unkn0wn:
Fixes #262

CC @bkchr 

### DIFF
--- a/relay/kusama/src/lib.rs
+++ b/relay/kusama/src/lib.rs
@@ -1860,7 +1860,7 @@ pub mod migrations {
 	/// Upgrade Session keys to exclude `ImOnline` key.
 	/// When this is removed, should also remove `OldSessionKeys`.
 	pub struct UpgradeSessionKeys;
-	const UPGRADE_SESSION_KEYS_FROM_SPEC: u32 = 1001002;
+	const UPGRADE_SESSION_KEYS_FROM_SPEC: u32 = 1001003;
 
 	impl OnRuntimeUpgrade for UpgradeSessionKeys {
 		#[cfg(feature = "try-runtime")]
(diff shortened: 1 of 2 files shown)


kianenigma (approved):
It is an interesting tie to safeguard a migration to a specific spec version. But the downside is that an emergency upgrade like 1.1.2 can disrupt the flow :) 

Somewhat more ideally, a migration would be smart enough to know if it is executed or not. In some migrations, this is possible, in some it is harder but can be achieved if we attach a unique identifier to all impls of `OnRuntimeUpgrade` and then we will never have to worry about duplicate or stale migrations again.

bkchr (approved):
Ty!

bkchr:
@kianenigma https://github.com/paritytech/polkadot-sdk/issues/2421

s0me0ne-unkn0wn:
@kianenigma yes, exactly. The root cause of the problem is that some runtime structures, as `SessionKeys`, are not versioned. We should either make them versioned or implement something like what you're talking about (we have paritytech/polkadot-sdk#2421 for that), but until then we have no options but use this hacky and brittle way :shrug:

acatangiu (approved):

relay/kusama/src/lib.rs
```diff
@@ -1860,7 +1860,7 @@ pub mod migrations {
 	/// Upgrade Session keys to exclude `ImOnline` key.
 	/// When this is removed, should also remove `OldSessionKeys`.
 	pub struct UpgradeSessionKeys;
-	const UPGRADE_SESSION_KEYS_FROM_SPEC: u32 = 1001002;
+	const UPGRADE_SESSION_KEYS_FROM_SPEC: u32 = 1001003;
```
ggwpez, line 1863: It would be safer to create a storage key like `:upgrade_session_keys_migration_executed` and then manually cleaning it up via governance.  
But i think we are past that point.

PR #270, opened 2024-04-10
[encointer] introduce collator selection and send fees to authors instead of treasury

### LINKED ISSUE #146: encointer chain-spec-generator genesis state config
A chain spec generator was added here https://github.com/polkadot-fellows/runtimes/pull/127

Specs for all chains except Encointer can successfully generate benchmarks. Need to update the encointer spec [here](https://github.com/polkadot-fellows/runtimes/blob/7515fc0385f8d48f2d08b2182a24174148e70135/chain-spec-generator/src/system_parachains_specs.rs#L640) so it can be used to generate benchmarks.

See https://github.com/encointer/encointer-parachain/blob/master/polkadot-parachains/src/chain_spec.rs for reference.

### LINKED ISSUE #176: remove treasury from Encointer runtime
Since #158, Encointer's treasury will no longer receive fees. While this is more the result of an oversight AFAIU, it may actually make sense to get rid of the treasury on the Encointer system chain altogether.

The new multi-asset treasury is way too complex for just collecting KSM fees. Yet, it is not powerful enough to be used for our plans of [per-community treasuries](https://forum.polkadot.network/t/towards-encointer-self-sustainability/4195)

I suggest to simply remove the treasury pallet and ignore the negligible funds currently in the treasury (<<1KSM)

### LINKED ISSUE #183: implement collator-selection for Encointer
Encointer still operates with hard-coded collator keys which is very rigid and hinders decentralization of infrastructure.
This should finally be changed and aligned with other system parachains

There may be synergy when implementing this before #176 

draft strategy:
1. instantiate the necessary pallets for release 1.3.0 but don't change collation logic yet (as the pallet is initialized with zero invulnerables)
2. resolve #176 and replace Treasury with ToAuthor as fee beneficiary
3. deploy runtime upgrade 1.3.0
4. establish invulnerables = existing hard-coded keys
5. activate collator-selection, deprecate hard-coded collator keys

Questions:
* we may be able to use state migration to initialize the set of invulnerables and do all this in a single upgrade. Check how Asset Hub did it

brenzi:
closes #183 
closes #176
closes #146 (should be closed already, but this PR adds more)

This PR aligns collator selection and fee distribution logic of Encointer with other system chains and thereby is a small step towards [self-sustainability of infrastructure](https://forum.polkadot.network/t/towards-encointer-self-sustainability/4195)

* [x] remove treasury (and leave negligible funds on trsry account)
* [x] add collator-selection, session and authorship pallets
* [x] instead of burning fees in community currency , send them to block authors like native KSM
* [x] remove DMP queue pallet (pure cleanup, no change in functionality)
* [x] add migration to inject existing collators as invulnerables
* [x] remove applied migrations from release 1.2.0

steps to reproduce the new fee logic: https://gist.github.com/brenzi/ca9e0c326198c73ff2dd0a2e74468053


system-parachains/encointer/src/migrations_fix.rs
```diff
@@ -17,173 +17,115 @@
...
+			hex!("0cecb8d1c2c744ca4c5cea57f5d6c40238f4dad17afa213672b8b7d43b80a659");
+		const INVULNERABLE_C: [u8; 32] =
+			hex!("ca1951a3c4e100fb5a899e7bae3ea124491930a72000c5e4b2775fea27ecf05d");
+		const INVULNERABLE_D: [u8; 32] =
+			hex!("484b443bd95068b860c92b0f66487b78f58234eca0f88e2adbe80bae4807b809");
+		const INVULNERABLE_E: [u8; 32] =
+			hex!("6c642fb4b571a5685a869cd291fafd575be47a918b231ba28165e5c0cd0cfa15");
+
+		pub struct InitInvulnerables<T>(sp_std::marker::PhantomData<T>);
+		impl<T> OnRuntimeUpgrade for InitInvulnerables<T>
```
brenzi, line 51: @joepetrowski may I ask you for  a sanity-check?
All we want to do is keep the same collators we have now as invulnerables and be sure to produce blocks after the runtime upgrade ;-)

system-parachains/encointer/src/lib.rs
```diff
@@ -624,16 +570,74 @@ impl pallet_membership::Config<pallet_membership::Instance1> for Runtime {
...
+impl<R> HandleCredit<AccountIdOf<R>, pallet_encointer_balances::Pallet<R>>
+	for AssetsToBlockAuthor<R>
+where
+	R: pallet_authorship::Config + pallet_encointer_balances::Config,
+	AccountIdOf<R>: From<polkadot_primitives::AccountId> + Into<polkadot_primitives::AccountId>,
+{
+	fn handle_credit(credit: Credit<AccountIdOf<R>, pallet_encointer_balances::Pallet<R>>) {
+		if let Some(author) = pallet_authorship::Pallet::<R>::author() {
+			// In case of error: Will drop the result triggering the `OnDrop` of the imbalance.
+			let _ = pallet_encointer_balances::Pallet::<R>::resolve(&author, credit);
```
clangenb, line 586: It is not obvious to me how this handles native transfer fees aka KSM.

system-parachains/encointer/src/lib.rs
```diff
@@ -122,7 +125,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: create_runtime_str!("encointer-parachain"),
 	impl_name: create_runtime_str!("encointer-parachain"),
 	authoring_version: 1,
-	spec_version: 1_002_000,
+	spec_version: 1_002_010,
```
joepetrowski, line 128: Why this bump? I think we just bump everything to `1_003_000` on the next release, unless you want to make an intermediate release?

system-parachains/encointer/src/lib.rs
```diff
@@ -624,16 +570,74 @@ impl pallet_membership::Config<pallet_membership::Instance1> for Runtime {
 	type WeightInfo = weights::pallet_membership::WeightInfo<Runtime>;
 }
 
+/// A `HandleCredit` implementation that naively transfers the fees to the block author.
+/// Will drop and burn the assets in case the transfer fails.
+pub struct AssetsToBlockAuthor<R>(PhantomData<R>);
```
joepetrowski, line 575: In other system paras we send only 50% to the author and 50% to a pot to share with other collators. If you send 100% to the author, then the author can attempt to increase the fee rate by making transactions up to the block limit at zero cost (since they pocket the fees). Sharing amongst others would make them at least collude to drive up the fee rate.

Impl here: https://github.com/paritytech/polkadot-sdk/blob/4e10d3b0a6ec2eccf58c471e7739948c1a867acf/cumulus/parachains/common/src/impls.rs#L62

The pot actually takes care of the 50/50 part.

system-parachains/encointer/src/migrations_fix.rs
```diff
@@ -17,173 +17,114 @@
...
+			<T as pallet_balances::Config>::Balance: EncodeLike<
+				<<T as pallet_collator_selection::Config>::Currency as Currency<
+					<T as frame_system::Config>::AccountId,
+				>>::Balance,
+			>,
+		{
 			#[cfg(feature = "try-runtime")]
 			fn pre_upgrade() -> Result<sp_std::vec::Vec<u8>, TryRuntimeError> {
 				Ok(().encode())
 			}
```
joepetrowski, line 69: Using try-runtime would be nice here

system-parachains/encointer/src/migrations_fix.rs
```diff
@@ -17,173 +17,114 @@
...
+
+				for pk in raw_keys.clone() {
+					pallet_session::NextKeys::<T>::insert::<
+						<T as pallet_session::Config>::ValidatorId,
+						<T as pallet_session::Config>::Keys,
+					>(pk.into(), SessionKeys { aura: sr25519::Public::from_raw(pk).into() }.into());
+					pallet_session::KeyOwner::<T>::insert::<
+						_,
+						<T as pallet_session::Config>::ValidatorId,
+					>((AURA, pk.encode()), pk.into());
```
joepetrowski, line 110: Are the aura keys the same as the validator account IDs?

system-parachains/encointer/src/migrations_fix.rs
```diff
@@ -17,173 +17,114 @@
...
-		#[derive(Clone, RuntimeDebug, Encode, Decode)]
-		pub(crate) struct ScheduledV1<Call, BlockNumber> {
-			maybe_id: Option<Vec<u8>>,
-			priority: schedule::Priority,
-			call: Call,
-			maybe_periodic: Option<schedule::Period<BlockNumber>>,
-		}
+				pallet_collator_selection::CandidacyBond::<T>::put::<BalanceOf<T>>(
+					533_333_328u128.into(),
+				);
```
joepetrowski, line 121: IIRC it's 50 KSM for other system chains. Perhaps put the same value there?

system-parachains/encointer/src/migrations_fix.rs
joepetrowski: It looks like this sets everything needed. It's weird to me that the collators have the same Aura key and Account ID. @georgepisaltu could also do a sanity check on this.

system-parachains/encointer/src/lib.rs
brenzi: Yeah, I can revert that

system-parachains/encointer/src/migrations_fix.rs
brenzi: Yes

system-parachains/encointer/src/migrations_fix.rs
brenzi: Ok. Seems I copy-pasted from the wrong place. Will check

system-parachains/encointer/src/migrations_fix.rs
joepetrowski: For account ID I mean their on-chain address.

system-parachains/encointer/src/migrations_fix.rs
georgepisaltu: The `ValidatorId` happens to be the same type as `AccountId`. The key type is built with `SessionKeys { aura: sr25519::Public::from_raw(pk).into() }.into()`.

> All we want to do is keep the same collators we have now as invulnerables and be sure to produce blocks after the runtime upgrade

About this, I looked at the logic below and it seems you're setting all the storage values to what you want. I didn't check the hardcoded addresses though 😁 .

system-parachains/encointer/src/migrations_fix.rs
brenzi: As we'd like individual Encointer communities to run collators, 50 KSM seems too much and will likely prevent self-sustained collators running in the global south. I set it to 5 KSM instead for starters.

system-parachains/encointer/src/lib.rs
brenzi: I think I addressed your valid concern by burning 50% of fees if they are paid in community currency. In the future, we'll direct a fraction to community treasuries instead of burning. Left a comment in the respective code

system-parachains/encointer/src/migrations_fix.rs
joepetrowski: Since it uses an election mechanism now, we can probably reduce the other chains to 5 as well and just let the market find what it's worth to be a collator.

system-parachains/encointer/src/migrations_fix.rs
brenzi: according to @georgepisaltu 's comment above, I guess this should work as-is

system-parachains/encointer/src/lib.rs
brenzi: native is handled elsewhere with DealWithFees

system-parachains/encointer/src/migrations_fix.rs
georgepisaltu: The `ValidatorId` type is actually the same as the `AccountId`, as seen [here](https://github.com/encointer/runtimes/blob/bbf92f696ad8b253ca9f8901ef65022d887fe451/system-parachains/encointer/src/lib.rs#L616).

The aura session key is a wrapper over an `AuraId`, which is defined [here](https://github.com/paritytech/polkadot-sdk/blob/8eb1f5592d15c7c748fbe21c2ea951d3bfe890e2/cumulus/parachains/common/src/lib.rs#L56), being a [`sr25519::Public`](https://github.com/paritytech/polkadot-sdk/blob/8eb1f5592d15c7c748fbe21c2ea951d3bfe890e2/substrate/primitives/consensus/aura/src/lib.rs#L46).

It seems to me everything should match, WDYT @joepetrowski ?

joepetrowski: Yes the key types match, but my confusion comes from this:

- This block of code is setting the `ValidatorId`s, these are the session keys.
- Just below this, it puts this `BoundedVec` of keys into storage as the `Invulnerable`s. That means you are using the session key for the on-chain account.

For sure it is possible, it's just highly not recommended because you can only change the Invulnerables via governance, but anyone running a collator node should be able to change their session keys. So usually people would have different levels of security on these two keys.

georgepisaltu: You can just ask each of those collators to generate a session key pair and give you the public so you can use them alongside the account public keys.

system-parachains/encointer/src/lib.rs
```diff
@@ -617,16 +570,82 @@ impl pallet_membership::Config<pallet_membership::Instance1> for Runtime {
...
+	fn handle_credit(credit: Credit<AccountIdOf<R>, pallet_encointer_balances::Pallet<R>>) {
+		if let Some(author) = pallet_authorship::Pallet::<R>::author() {
+			// We will burn 50% of community currency fees and send 50% to the block author.
+			// reasoning: If you send 100% to the author, then the author can attempt to increase
+			// the fee rate by making transactions up to the block limit at zero cost
+			// (since they pocket the fees).
+			// In the future, fees might be collected in community treasuries instead of being
+			// burned (https://forum.polkadot.network/t/towards-encointer-self-sustainability/4195)
+			let half_amount = credit.peek() / 2;
+			let (author_credit, _burnable) = credit.split(half_amount);
```
joepetrowski, line 592: I don't see how this treats community token fees different from KSM. If you want to burn KSM, it needs to go into some dedicated "burn account" and periodically sent back to the Relay Chain to update the Checking account.

system-parachains/encointer/src/migrations_fix.rs
```diff
@@ -17,173 +17,131 @@
...
-				if onchain_version >= 1 {
-					log::warn!(
-						target: TARGET,
-						"skipping bruteforce to v1 migration: executed on wrong storage version."
-					);
-					return T::DbWeight::get().reads(1)
+				let invulnerables_len = pallet_collator_selection::Invulnerables::<T>::get().len();
+				if invulnerables_len > 0 {
+					info!(target: TARGET, "no need to initialize invulnerables");
+					return T::DbWeight::get().reads_writes(0, 0)
```
joepetrowski, line 76: It had to read `Invulnerables`
```suggestion
					return T::DbWeight::get().reads_writes(1, 0)
```

system-parachains/encointer/src/migrations_fix.rs
```diff
@@ -17,173 +17,131 @@
...
-		#[cfg(feature = "try-runtime")]
-		#[frame_support::storage_alias]
-		pub(crate) type Lookup<T: Config> =
-			StorageMap<Pallet<T>, Twox64Concat, TaskName, TaskAddress<BlockNumberFor<T>>>;
+				let mut invulnerables: Vec<<T as frame_system::Config>::AccountId> =
+					raw_keys.iter().map(|&pk| pk.into()).collect();
+				invulnerables.sort();
+				let invulnerables: BoundedVec<_, T::MaxInvulnerables> =
+					invulnerables.try_into().unwrap();
+				pallet_collator_selection::Invulnerables::<T>::put(invulnerables);
```
joepetrowski, line 124: This is not technically wrong but I still think it is bad practice for the on-chain `AccountId` to be the same key as the Aura key.

The on-chain account should be a long-standing key held in a wallet with backups. They can only be changed via governance motion.

The session key is a "hot" in-node key used for signing blocks. It is not meant to be long-standing. For example, if one of the invulnerables wants to start a new collator machine, standard practice would be to start the second machine, call `rotate_keys`, and set the new session keys to correspond to the unchanged on-chain account. The old ones would disappear when killing the old machine.

This pattern prevents this, either the keys in the migration are session keys and potentially not properly backed up or they are wallet keys that collators will need to inject into online collator nodes to get started before rotating.

joepetrowski (commented):
One question about fee burning, will approve after that.

Left my opinion about the session/account keys, but I won't stop you.

system-parachains/encointer/src/lib.rs
brenzi: I'll run a test locally to be sure....

joepetrowski: > I don't think Encointer needs or should have the authority to burn KSM

It is a trusted teleporter though, so it does have the authority. So, it needs to be accounted for.

system-parachains/encointer/src/migrations_fix.rs
joepetrowski: It depends what you mean by "collator keys" here.

- Whatever values you `put` into `pallet_collator_selection::Invulnerables::<T>::put(invulnerables);` should be keys that are in a wallet/cold storage, as in keys you want to live for a long time and would expect to hold funds and make transactions.
- `Validators` should be a vec of the account keys (wallets).
- `QueuedKeys` and `NextKeys` should be the map of `ValidatorId => SessionKey`.
- `KeyOwner` should be the map of `(AURA, SessionKey) => ValidatorId`.

Maybe @georgepisaltu can double check this from pallet-session?

brenzi: this sounds coherent. Did it like this in my most recent commits

system-parachains/encointer/src/lib.rs
brenzi: I tested the fee handling and these are my observations:
* native fees are handled correctly: they are smeared over several blocks, nothing gets burned. total issuance is constant
* CC fees are credited to the current author to 100%, so my halving logic has no effect

I guess the problem is the default OnDrop behavior which will credit the author with the remainder ???

brenzi: I added commit 6a75792911a39da420e7fd69a75aeec245845067 which demonstrates what I assumed:

* native fees are untouched by this HandleCredit implementation
* CC fees will now be split 50/50 between the current block author and a "dead" account with pubkey `0x00`

Not sure I want to leave it like that, but this way I could verify that it now works as intended

brenzi: I suggest to leave the "dead" account approach for now and revisit this when we have per-community treasuries which could collect fees depending on their asset id / CommunityIdentifier

liamaharon: > Yeah I would expect `HandleImbalanceDrop` to decrease the issuance. @liamaharon could you take a look?

@joepetrowski yes, the credit should decrease the issuance when it is dropped 
- https://github.com/paritytech/polkadot-sdk/blob/6b0bc052e93b742e0813f79bc331bfa0375bb304/substrate/frame/balances/src/impl_fungible.rs#L347
- https://github.com/paritytech/polkadot-sdk/blob/4c09a0631334c3e021a30aa49a503815aa047b29/substrate/frame/support/src/traits/tokens/fungibles/imbalance.rs#L66

@brenzi can you confirm if there is an issue here with dropped amounts not changing the TI and if so provide some instructions to reproduce? Thanks.

brenzi: @liamaharon I assembled steps to reproduce and verify the current 50/50 fee split:
https://gist.github.com/brenzi/ca9e0c326198c73ff2dd0a2e74468053

I consider this done, as the question if the drop gets burned or not is no longer relevant. If there is a problem it may also be rooted in our custom fee handling elsewhere, so I'm not sure it's worth digging deeper. However, If you 'd like to make sure, just remove this line: 
https://github.com/encointer/runtimes/blob/1a655369951601c0dfb3b107c0aba721fb935081/system-parachains/encointer/src/lib.rs#L603-L604

then you could verify if 50% gets burned or not

joepetrowski: > the question if the drop gets burned or not is no longer relevant for the current implementation

How is this not relevant? We need to track issuance of KSM.

liamaharon: Tracking this, will investigate when I get time https://github.com/paritytech/polkadot-sdk/issues/4397

joepetrowski (approved):
KSM does go to the Staking Pot as expected:
https://gist.github.com/brenzi/ca9e0c326198c73ff2dd0a2e74468053#file-gistfile1-txt-L50-L64

CHANGELOG.md
```diff
@@ -60,7 +61,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Remove state-trie-migration pallet from kusama, add state trie migration to V1 on polkadot ([polkadot-fellows/runtimes#170](https://github.com/polkadot-fellows/runtimes/pull/170))
 - Introduce chain spec generator ([polkadot-fellows/runtimes#127](https://github.com/polkadot-fellows/runtimes/pull/127))
 - Add [Encointer](https://encointer.org) system parachain runtime, completing [RFC22](https://github.com/polkadot-fellows/RFCs/blob/main/text/
-0022-adopt-encointer-runtime.md) ([polkadot-fellows/runtimes#80](https://github.com/polkadot-fellows/runtimes/pull/80))
+  0022-adopt-encointer-runtime.md) ([polkadot-fellows/runtimes#80](https://github.com/polkadot-fellows/runtimes/pull/80))
```
acatangiu, line 64: The RFC link is broken (by the newline)

acatangiu (approved):

CHANGELOG.md
brenzi: fixed. thx

acatangiu (approved):

brenzi:
happy to announce the migration worked smoothly on rococo and the session pallet works as expected:
image

notice the author's appearing after the upgrade

PR #284, opened 2024-04-23
updated auto-merge to v1.1.2

Bullrich:
Updated the action to version `1.1.2` which includes paritytech/get-fellows-action#23 which is a fix for the breaking change in V1.2.0 which causes the `Identity` object to become a tuple.

The updated code gets the first element of the tuple.

This also updates PAPI to version `0.4.0`.


Bullrich:
Unfortunately this PR won't be mergeable until #282 is merged, and I can't trigger the `/merge` command as the `get-fellows-action` is failing.

Issue #273, opened 2024-04-18
Remove XCM SafeCallFilter for chains using Weights::v2 (also accounting PoV weight)


acatangiu:
e.g. https://github.com/polkadot-fellows/runtimes/blob/31ba26287ec752574244f0d690167f7ae8430c8b/system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs#L274C1-L282C48

acatangiu:
Fixed in #285

PR #285, opened 2024-04-24
Remove XCM SafeCallFilter for chains using Weights::v2

acatangiu:
Remove `SafeCallFilter` from system chains as they all now use Weights::v2 which already accounts for call PoV size. This effectively removes artificial limitations on what users can`XCM::Transact` on these chains (blockspace limitations are still upheld).

Fixes https://github.com/polkadot-fellows/runtimes/issues/273


system-parachains/bridge-hubs/bridge-hub-kusama/src/xcm_config.rs
```diff
@@ -382,7 +291,7 @@ impl<WaivedLocations: Contains<Location>, FeeHandler: HandleFee> FeeManager
 		let Some(loc) = origin else { return false };
 		if let Export { network, destination: Here } = fee_reason {
 			if network == EthereumNetwork::get() {
-				return false
+				return false;
```
joepetrowski, line 294:
```suggestion
				return false
```

ggwpez (approved):

muharem (approved):

PR #298, opened 2024-04-30
PoC for short benchmarking in CI with `frame-omni-bencher`

bkontur:
Closes: https://github.com/polkadot-fellows/runtimes/issues/197

- [ ] bump `sp-genesis-builder` and add `get_preset` feature to the runtimes
   ```
   [2024-04-30T11:14:22Z WARN  frame_omni_bencher] The FRAME omni-bencher is not yet battle tested - double check the results.
   [2024-04-30T11:14:22Z INFO  frame_benchmarking_cli::pallet::command] Loading WASM from ./target/production/wbuild/asset-hub-kusama-runtime/asset_hub_kusama_runtime.compact.compressed.wasm
   Error: Input("Could not call runtime API to build the genesis spec: Other: Exported method GenesisBuilder_get_preset is not found")
   ```
- [ ] find out how to use `frame-omni-bencher` - GHA? or download released binary? or whatever?
- [x] Does not require a CHANGELOG entry

PR #305, opened 2024-05-08
Fix people chain spec parsing in chainspec generator

s0me0ne-unkn0wn:
I think it was just forgotten when people chain was added.

### DIFF
--- a/chain-spec-generator/src/common.rs
+++ b/chain-spec-generator/src/common.rs
@@ -20,7 +20,7 @@ use crate::{
 	system_parachains_specs::{
 		AssetHubKusamaChainSpec, AssetHubPolkadotChainSpec, BridgeHubKusamaChainSpec,
 		BridgeHubPolkadotChainSpec, CollectivesPolkadotChainSpec, EncointerKusamaChainSpec,
-		GluttonKusamaChainSpec,
+		GluttonKusamaChainSpec, PeopleKusamaChainSpec,
 	},
 	ChainSpec,
 };
@@ -90,6 +90,8 @@ pub fn from_json_file(filepath: &str, supported: String) -> Result<Box<dyn Chain
 			Ok(Box::new(GluttonKusamaChainSpec::from_json_file(path)?)),
 		x if x.starts_with("encointer-kusama") =>
 			Ok(Box::new(EncointerKusamaChainSpec::from_json_file(path)?)),
+		x if x.starts_with("people-kusama") =>
+			Ok(Box::new(PeopleKusamaChainSpec::from_json_file(path)?)),
 		_ => Err(format!("Unknown chain 'id' in json file. Only supported: {supported}'")),
 	}
 }


chain-spec-generator/src/common.rs
```diff
@@ -90,6 +90,8 @@ pub fn from_json_file(filepath: &str, supported: String) -> Result<Box<dyn Chain
 			Ok(Box::new(GluttonKusamaChainSpec::from_json_file(path)?)),
 		x if x.starts_with("encointer-kusama") =>
 			Ok(Box::new(EncointerKusamaChainSpec::from_json_file(path)?)),
+		x if x.starts_with("people-kusama") =>
```
bkontur, line 93: Hmm, I just wonder why coretime-kusama is also missing, could you add it also?

bkontur (approved):

chain-spec-generator/src/common.rs
bkontur: I think we could extract these IFs for chain-id matching out of the common.rs (it is not really common) to the main.rs and e.g. join it with supported networks cfg here: https://github.com/polkadot-fellows/runtimes/blob/main/chain-spec-generator/src/main.rs#L39

xlc:
slightly unrelated, why do we even need to maintain the chain spec generator in the runtime repo? it is only going to be used once (for production runtimes). for dev/testing purpose, we will use the westend/rococo runtime in the polkadot-sdk repo

besides, why do we need another chain spec generator at all? to my understanding, the polkadot-sdk already provides a chain spec generator cli and it should work. there will be additional chain spec editing needed but it can be just a simple script

https://github.com/open-web3-stack/XCQ/blob/dc572745ffb54edfb0de2ce038121b11855e2c6e/Makefile#L24-L25

bkontur:
> slightly unrelated, why do we even need to maintain the chain spec generator in the runtime repo? it is only going to be used once (for production runtimes). for dev/testing purpose, we will use the westend/rococo runtime in the polkadot-sdk repo
> 
> besides, why do we need another chain spec generator at all? to my understanding, the polkadot-sdk already provides a chain spec generator cli and it should work. there will be additional chain spec editing needed but it can be just a simple script
> 
> https://github.com/open-web3-stack/XCQ/blob/dc572745ffb54edfb0de2ce038121b11855e2c6e/Makefile#L24-L25

Maybe I am missing something, but I think we won't need it, but we need to bump sp-genesis-builder with 'get_preset' feature first, which I also need here: https://github.com/polkadot-fellows/runtimes/pull/298

xlc:
I only spend 2 minutes reading the related code so my understanding could be wrong. But the preset feature appears to be overengineered to me. Why the runtime needs to know some network preset? If you check my poc runtime, it is using a version of polkadot-sdk without the preset feature. All I need to do is use jq or whatever tool to edit the chainspec json to do the modification needed and that's it. Yeah we don't get nice Rust type check but does it matter? again, it is one off thing and the chainspec generator can validate the chainspec json.

s0me0ne-unkn0wn:
> for dev/testing purpose, we will use the westend/rococo runtime in the polkadot-sdk repo

That's not a universal solution. I'm currently doing experiments with `people` chain on `kusama-local` relay (and that's how I came across this bug), and we also have guys in the ecosystem whose Rococo and Kusama runtimes significantly diverged over time, and they need a tool to test against the proper relay. In that sense, the chainspec generator is just a convenient tool to drop into a zombienet config and let it do the job.

xlc:
we have a chain spec generator in polkadot-sdk and I see no reason why it wouldn't be enough

s0me0ne-unkn0wn:
@bkontur you mean something like this? (see the last changes)

bkontur:
> @bkontur you mean something like this? (see the last changes)

Yes, nice, exactly. With actual chain-spec-generator setup, at least we won't forget to add this stuff when adding new chains.

PR #307, opened 2024-05-10
Add zombienet integration

pepoviola:

Hi 👋 , this pr add a very small smoke test as PoC of using [zombienet-sdk](https://github.com/paritytech/zombienet-sdk) with the _runtimes_ repo. The idea is to explore the integration and add more assertions (metrics, logs, etc) and complex scenarios in the future (e.g: test upgrades, mix node's versions, etc). 

This test use a set of _env vars_ that allow you to specify the images and the provider to use:
`POLKADOT_IMAGE` (_default_: docker.io/parity/polkadot:latest)
`CUMULUS_IMAGE` (_default_: docker.io/parity/polkadot-parachain:latest)
`ZOMBIE_PROVIDER`(_default_: Docker)

And needs to have the `chain-spec-generator` binary as part of your `PATH`, any feedback is welcome :)

--- 
Also worth notice that the `chain-spec-generator` tool is compiling all the runtimes, and would be good to only compile the needed ones (polkadot in this case), we can add some feature-gate logic or we can also move _ presets_ to runtimes (ping @michalkucharczyk ).

Thanks!


michalkucharczyk:
> or we can also move _ presets_ to runtimes

With this we could use generic `chain-spec-builder` to build chain-specs.
But there is still open question how to handle the rest of chain-spec (e.g. chain-name, id, tokenDecimals). Providing them in command line every time the tool is used will be inconvenient. As trivial/naive solution bunch of utility scripts could be provided. Any thoughts on this? Unfortunately I don't have good ideas on this.

Anyway I think that as an intermediate step we could move presets to the runtimes, and keep `chain-spec-builder` in this rep:
-  without any dependencies on runtime crates 
-  and maybe reusing this little [lib](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/bin/utils/chain-spec-builder/src/lib.rs) which is part of generic tool.

As reference, some example work of moving presets from binary to runtime is provided here: https://github.com/paritytech/polkadot-sdk/pull/3996

Cargo.toml
```diff
@@ -243,6 +243,10 @@ members = [
 	"integration-tests/emulated/tests/people/people-kusama",
 ]
 
+exclude = [
+	"integration-tests/zombienet",
+]
+
```
bkchr, line 249:
```suggestion
```

.github/workflows/test.yml
```diff
@@ -189,3 +189,54 @@ jobs:
...
+        run: |
+          rustup install $RUST_STABLE_VERSION
+          rustup default $RUST_STABLE_VERSION
+          rustup target add wasm32-unknown-unknown
+          rustup component add rust-src
+
+      - name: Fetch cache
+        uses: Swatinem/rust-cache@a95ba195448af2da9b00fb742d14ffaaf3c21f43 # v2.7.0
+        with:
+          shared-key: "fellowship-cache-tests"
```
bkchr, line 232:
```suggestion
          shared-key: "fellowship-cache-integration-tests"
```

integration-tests/zombienet/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "zombienet-sdk-tests"
+version = "0.1.0"
+edition = "2021"
+
+[workspace]
```
bkchr, line 6:
```suggestion
```

integration-tests/zombienet/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+subxt = { version = "0.35.0", default-features = false, features = ["native"] }
+subxt-signer = { version = "0.35.0" }
+tokio = "1.35.0"
+tracing-subscriber = "0.3.18"
+zombienet-sdk = { git = "https://github.com/paritytech/zombienet-sdk", tag = "v0.2.1" }
+log = { version = "0.4" }
+parity-scale-codec = { version = "3.6.9", features = ["derive", "parity-scale-codec-derive"] }
+anyhow = "1.0.82"
+
+toml_datetime = { version = "=0.6.5" }
```
bkchr, line 19:
```suggestion
subxt-signer = "0.35.0"
tokio = "1.35.0"
tracing-subscriber = "0.3.18"
zombienet-sdk = { git = "https://github.com/paritytech/zombienet-sdk", tag = "v0.2.1" }
log = "0.4"
parity-scale-codec = { version = "3.6.9", features = ["derive", "parity-scale-codec-derive"] }
anyhow = "1.0.82"

toml_datetime = "=0.6.5"
```

integration-tests/zombienet/tests/smoke.rs
```diff
@@ -0,0 +1,29 @@
...
+	let spawn_fn = get_spawn_fn();
+	let config = small_network().unwrap();
+
+	// spawn
+	let now = Instant::now();
+	let network = spawn_fn(config).await.unwrap();
+	let elapsed = now.elapsed();
+	log::info!("🚀🚀🚀🚀 network deployed in {:.2?}", elapsed);
+
+	let alice = network.get_node("alice").unwrap();
```
bkchr, line 19: Where can I find the names of the nodes? How do I know if a node belongs to a parachain or the relay chain?

integration-tests/zombienet/tests/smoke.rs
```diff
@@ -0,0 +1,29 @@
...
+	// wait until the subxt client is ready
+	let client = wait_subxt_client(alice).await.unwrap();
+
+	// wait 3 blocks
```
bkchr, line 23: We actually wait for 10 :D 

But also this test does not ensure that parachains are making any progress. So, yeah. Really just some "can build blocks smoke test".

bkchr (approved):
Left only nitpicks.

For a first integration it is good, but also doesn't do that much :D

integration-tests/zombienet/tests/smoke.rs
pepoviola: The `config` came from [small_network](https://github.com/polkadot-fellows/runtimes/pull/307/files#diff-d645fab731f2bf1bbd7e52d36bcf30a59649d7bda90e16a7ddd0719da1982d2dR11) and I placed there for reuse in other tests, but could be better to each test have it's own config in the same file to easily find the nodes/paras. I can change if you are agree.

Thx!

pepoviola: Ahh oks, in the `sdk` you have to set the `name` of the node with a like like this

```rust
.with_node(|node| node.with_name("alice"))
```
but we can make more explicit as `const` to be easily find with a quick look. wdyt?

integration-tests/zombienet/src/lib.rs
```diff
@@ -0,0 +1,66 @@
...
+			match node.client::<subxt::PolkadotConfig>().await {
+				Ok(cli) => {
+					break Ok(cli);
+				},
+				Err(e) => {
+					let cause = e.to_string();
+					log::trace!("{:?}", e);
+					if let subxt::Error::Rpc(subxt::error::RpcError::ClientError(inner)) = e {
+						log::trace!("inner: {}", inner.to_string());
+						if inner.to_string().contains("i/o error") {
```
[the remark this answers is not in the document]
pepoviola: Yes, I'm not super happy with this. Maybe @jsdw can help with a better approach to check if the inner error is related to i/o.

Thx!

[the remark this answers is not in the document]
jsdw: Because somebody can plug in an arbitrary RPC client, we end up with this `Box<dyn std::error::Error + Send + Sync + 'static>` type to store errors from them. If you know which RPC client is used, you can try to cast that into the relevant error.

In the usual case, the error will be a [`jsonrpsee_core::ClientError`](https://docs.rs/jsonrpsee-core/0.22.5/jsonrpsee_core/client/error/enum.Error.html), so in that case you could do something like:

```rust
if let Ok(client_error) = inner.downcast::<jsonrpsee_core::ClientError> {
    if matches!(client_error, ClientError::Transport(_)) {
        // ...
    }
}
```

It's not much less ugly though in the end!

Not sure if this applies for you, but in Subxt the way that we wait for the node to be ready for connections is to wait for it to log that it's listening for them, ie https://github.com/paritytech/subxt/blob/ec7cea0acb59cdffb281aacd86a62c33514331e6/testing/substrate-runner/src/lib.rs#L191-L198 (that waits for a couple of "legacy" strings that nodes don't emit now too, which we can prob remove at this point).

integration-tests/zombienet/tests/smoke.rs
bkchr:
```
.with_node(|node| node.with_name("alice"))
```

But you don't do this here anywhere in the code? Or am I blind?

pepoviola: Yes, we do as part of the `small_network` fn here https://github.com/polkadot-fellows/runtimes/pull/307/files#diff-d645fab731f2bf1bbd7e52d36bcf30a59649d7bda90e16a7ddd0719da1982d2dR27-R28

I changed to use `consts`, e.g:

```rust
// Relaychain nodes
const ALICE: &str = "alice";
const BOB: &str = "bob";

... 

.with_node(|node| node.with_name(ALICE))
.with_node(|node| node.with_name(BOB))
```

Thx!

bkchr: Fuck I'm blind :facepalm: :facepalm:

pepoviola: Thanks for the feedback! 🙌 

Yes, the _naming_ could be confusing. Both method `with_node` and `with_collator` return a `NodeConfigBuilder` that by default is _validator_ (this is also confusing because for nodes of a parachain means that we pass the flag `--collator`). 
Maybe is better to just have `with_node` with the default `_role_` set to `ROLE::AUTHORITY` and you can change to `full` with `node.role(ROLE::FULL)`

So, this way is more clear and use the same wording you see in the logs of the client. wdyt?

Thx!

bkchr: > So, this way is more clear and use the same wording you see in the logs of the client. wdyt?

Or just call them `with_validator`, `with_node` and `with_collator`. Probably the least confusing one. Otherwise I would go with `with_node` and require that you make it a validator/collator (but the argument against this being that you almost never need a normal full node in the tests)

pepoviola: Yes, I think is more explicit and will cause less confusion have those 3 methods 👍.
Thanks again for the feedback!

integration-tests/zombienet/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "zombienet-sdk-tests"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+subxt = { version = "0.35.0", default-features = false, features = ["native"] }
+tokio = "1.35.0"
+tracing-subscriber = "0.3.18"
+zombienet-sdk = { git = "https://github.com/paritytech/zombienet-sdk", tag = "v0.2.1" }
```
davxy, line 10: Why we're not pointing to `crates.io`? I saw that 0.2.2 is already there

integration-tests/zombienet/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+log = "0.4"
+anyhow = "1.0.82"
+toml_datetime = "=0.6.5"
```
davxy, line 13: is this dependency used?

davxy (approved):

integration-tests/zombienet/Cargo.toml
pepoviola: Yes, I will change it to use crates.io 👍

integration-tests/zombienet/Cargo.toml
pepoviola: No, I pinned to workaround on a transitive dependency issue but is already fixed.
Thanks!!

pepoviola:
ping @eskimor / @ordian, can you also please review this one (we need 3 `rank 3` reviews).
Thanks!!

Cargo.lock
```diff
@@ -562,7 +595,7 @@ dependencies = [
  "frame-support",
  "kusama-emulated-chain",
  "parachains-common",
- "sp-core",
+ "sp-core 29.0.0",
```
ordian, line 598: @pepoviola is it possible to deduplicate the dependencies so that we don't have 2 versions of polkadot-sdk crates?

ordian: or is it coming from subxt?

ordian (approved):

Cargo.lock
pepoviola: Yes, we can align the version of `sp-core`. There is any reason why we are using `29.0.0` here since the latest release is `32.0.0`?

Thanks!!

ordian: it's probably fine to leave it as is for now. the versions will be upgraded when we update the runtumes to the next polkadot-sdk release

pepoviola:
Ping @bkchr, can we merge this one? Thx!

integration-tests/zombienet/Cargo.lock
bkchr, line 1: Why do we still have one extra `Cargo.lock` file?

pepoviola: We shouldn't, let  me delete this one. Thanks for the feedback.

bkchr (approved):

Cargo.lock
ggwpez: Yea its fine as long as we dont have duplicated versions in the runtimes themselves.

PR #330, opened 2024-05-27
fix Encointer CI

brenzi:
@ggwpez unfortunately, I couldn't find CI logs showing the error you want me to resolve.
However, I assume the unused/deprecated warning would certainly disturb CI, so I started by fixing that one. Please let me know if you need more fixes, pointing me to the error you observe


brenzi:
thanks for double-checking. KSM fees should now be treated exactly like for AH. So I guess it's fine if it works for AH

I suggest to merge this anyway as it will remove a warning and a TODO comment

ggwpez:
Okay thanks!

PR #322, opened 2024-05-22
Update to SDK 1.13

ggwpez:
Updates the runtimes from 1.7 to 1.13.

TODO:
- [x] Wait for 1.13 crates release and integrate them.
- [x] Fix all `FAIL-CI @name`.
- [x] Enable Encointer runtimes and pallets once they are on 1.13 as well @brenzi.
- [x] Unrevert [#337](https://github.com/polkadot-fellows/runtimes/pull/337)

PR #340, opened 2024-06-03
Enable elastic scaling node side feature on Polkadot 

sandreim:
This feature is needed on Polkadot to ensure parachains can still make progress if they have multiple cores assigned. Relay chain support for elastic scaling is implemented but requires bumping `polkadot-runtime-parachains` crate version to at least `11.0.0` which will happen later.

### DIFF
--- a/relay/polkadot/src/lib.rs
+++ b/relay/polkadot/src/lib.rs
@@ -1783,6 +1783,7 @@ pub type SignedExtra = (
 	frame_system::CheckWeight<Runtime>,
 	pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
 	claims::PrevalidateAttests<Runtime>,
+	frame_metadata_hash_extension::CheckMetadataHash<Runtime>,
 );
 
 pub struct NominationPoolsMigrationV4OldPallet;
@@ -1807,6 +1808,7 @@ pub mod migrations {
 	use runtime_parachains::configuration::WeightInfo;
 	#[cfg(feature = "try-runtime")]
 	use sp_core::crypto::ByteArray;
+
 	parameter_types! {
 		pub const ImOnlinePalletName: &'static str = "ImOnline";
 	}
--- a/relay/polkadot/src/lib.rs
+++ b/relay/polkadot/src/lib.rs
@@ -1783,7 +1783,6 @@ pub type SignedExtra = (
 	frame_system::CheckWeight<Runtime>,
 	pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
 	claims::PrevalidateAttests<Runtime>,
-	frame_metadata_hash_extension::CheckMetadataHash<Runtime>,
 );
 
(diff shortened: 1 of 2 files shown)


alindima (approved):

relay/polkadot/src/lib.rs
```diff
@@ -1886,6 +1889,18 @@ pub mod migrations {
...
 	}
 
+	/// Enable the elastic scaling node side feature.
+	///
+	/// This is required for Coretime to ensure the relay chain processes parachains that are
+	/// assigned to multiple cores.
+	pub struct EnableElasticScalingNodeFeature;
+	impl OnRuntimeUpgrade for EnableElasticScalingNodeFeature {
+		fn on_runtime_upgrade() -> Weight {
+			let _ = Configuration::set_node_feature(RawOrigin::Root.into(), 1, true);
```
eskimor, line 1899: Why are we not using the [constant](https://github.com/paritytech/polkadot-sdk/blob/523e62560eb5d9a36ea75851f2fb15b9d7993f01/polkadot/primitives/src/v7/mod.rs#L2036)?

sandreim: It's currently in staging, felt that we don't want to use any of the staging stuff here.

eskimor: Hmm - don't we have v7 already in the current release? In any case, seems like we would be using staging stuff anyways, just hiding it by using a magic number? Also I think this can wait for the next polkadot-sdk release, where this should be v7 then.

sandreim: That makes sense, but this needs to go in same release with core time.

eskimor:
@sandreim Let's rebase on [this](https://github.com/polkadot-fellows/runtimes/pull/322) one.

eskimor:
> @sandreim Let's rebase on [this](https://github.com/polkadot-fellows/runtimes/pull/322) one.

Actually, ignore. I can just make this part of my coretime enabling PR while I am already at it.

bkchr (changes requested):
Do we again try to set some configuration option, that could just be a separate referendum? :P 

Please close and just open a referendum on chain.

alindima:
> Do we again try to set some configuration option, that could just be a separate referendum? :P
> 
> Please close and just open a referendum on chain.

I think the reasoning of doing this as a migration was that it's more of a bugfix than a new feature. Once coretime is enabled on polkadot, if this feature bit is not set, one para that gets scheduled on two cores will brick itself.

bkchr:
> I think the reasoning of doing this as a migration was that it's more of a bugfix than a new feature.

This can already be opened today and ready before Coretime lands on Polkadot. Not really an argument to put this into a runtime upgrade.

We have done it [here](https://github.com/polkadot-fellows/runtimes/pull/205) using a runtime upgrade, because the feature was added in the same release AFAIR. However, this is not the case anymore.

sandreim:
https://polkadot.subsquare.io/referenda/1007

PR #345, opened 2024-06-07
polkadot chains: allow arbitrary XCM execution

acatangiu:
Following #261 where arbitrary XCM execution was enabled on Kusama chains, this commit enables the same for Polkadot chains.

Allowing `XcmExecuteFilter = Everything` is considered safe at this point, and doing so greatly increases the usability of these chains since many custom actions can be done through XCM without needing specialized calls for them anymore.

This is also a major step in improving dApps integration as they can now interact uniformly with multiple chains, directly using XCM programs.


eskimor (approved):

IkerAlus (approved):
Compared against https://github.com/polkadot-fellows/runtimes/pull/261/files and it LGTM.

seadanda (approved):

bkchr (approved):

Cryptopod13 (commented):
#374

Cryptopod13 (commented):
#322

PR #346, opened 2024-06-07
Max asset ID restriction for new trusted assets

muharem:
Set the maximum asset ID for the creation of trusted assets.

This change will enable us to migrate the trusted assets instance to start auto-incrementing IDs from 50,000,000 with the release following the next one..


system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
```diff
@@ -296,11 +296,31 @@ parameter_types! {
...
+pub struct EnsureLessThanAutoIncrement;
+impl EnsureOriginWithArg<RuntimeOrigin, AssetIdForTrustBackedAssets>
+	for EnsureLessThanAutoIncrement
+{
+	type Success = AccountId;
+	fn try_origin(
+		o: RuntimeOrigin,
+		a: &AssetIdForTrustBackedAssets,
+	) -> Result<Self::Success, RuntimeOrigin> {
+		if a >= &TrustAssetIdAutoIncrement::get() {
```
bkchr, line 317:
```suggestion
}

/// We allow root to execute privileged asset operations.
pub type AssetsForceOrigin = EnsureRoot<AccountId>;

/// Ensure that the proposed asset id is less than the [`TrustAssetIdAutoIncrement`] and origin is
/// signed.
pub struct EnsureLessThanAutoIncrement;
impl EnsureOriginWithArg<RuntimeOrigin, AssetIdForTrustBackedAssets>
	for EnsureLessThanAutoIncrement
{
	type Success = AccountId;
	fn try_origin(
		o: RuntimeOrigin,
		a: &AssetIdForTrustBackedAssets,
	) -> Result<Self::Success, RuntimeOrigin> {
		if *a >= 50_000_000 {
```

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -278,11 +278,31 @@ parameter_types! {
...
+pub struct EnsureLessThanAutoIncrement;
+impl EnsureOriginWithArg<RuntimeOrigin, AssetIdForTrustBackedAssets>
+	for EnsureLessThanAutoIncrement
+{
+	type Success = AccountId;
+	fn try_origin(
+		o: RuntimeOrigin,
+		a: &AssetIdForTrustBackedAssets,
+	) -> Result<Self::Success, RuntimeOrigin> {
+		if a >= &TrustAssetIdAutoIncrement::get() {
```
bkchr, line 299:
```suggestion
}

/// We allow root to execute privileged asset operations.
pub type AssetsForceOrigin = EnsureRoot<AccountId>;

/// Ensure that the proposed asset id is less than the [`TrustAssetIdAutoIncrement`] and origin is
/// signed.
pub struct EnsureLessThanAutoIncrement;
impl EnsureOriginWithArg<RuntimeOrigin, AssetIdForTrustBackedAssets>
	for EnsureLessThanAutoIncrement
{
	type Success = AccountId;
	fn try_origin(
		o: RuntimeOrigin,
		a: &AssetIdForTrustBackedAssets,
	) -> Result<Self::Success, RuntimeOrigin> {
		if *a >= 50_000_000 {
```

bkchr (approved):

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Transaction payments work via new `fungible` trait implementation ([polkadot-fellows/runtimes#332](https://github.com/polkadot-fellows/runtimes/pull/332))
 - Block `request_judgement` calls on the Relay Chain ([polkadot-fellows/runtimes#338](https://github.com/polkadot-fellows/runtimes/pull/338))
+- Set max asset id restriction for the creation of trusted assets ([polkadot-fellows/runtimes#346](https://github.com/polkadot-fellows/runtimes/pull/346))
```
joepetrowski, line 20:
```suggestion
- Set max asset ID restriction for the creation of trusted assets ([polkadot-fellows/runtimes#346](https://github.com/polkadot-fellows/runtimes/pull/346))
```

joepetrowski (approved):

xlc:
It is probably too late to make such refactor but the asset id type should really be an enum

```rs
enum AssetId {
  Trusted(u32),
   Untrusted(u32),
   Foreign(Location),
}
```

bkchr:
What is the difference between a trusted and untrusted asset? And what would be a remote asset? Token by another chain?

joepetrowski:
> It is probably too late to make such refactor but the asset id type should really be an enum

Yeah it probably would have been better than separate pallet instances for each type. But talk about a big migration. We just didn't know that far ahead when we made the first instance with type `u32`, and had to do some refactoring to the Assets pallet to support `Location` at all. Anyway, major breaking change plus migration but if the current structure becomes a problem then yeah we'd have to consider it.

> What is the difference between a trusted and untrusted asset? And what would be a remote asset?

Not sure what Bryan meant or if it was just an example, but we have `TrustBacked` for assets like USDT that anyone can register, because the user "trusts" that they are backed by some off-chain claim. `Foreign` assets are assets backed by other locations, which could be protocols, e.g. the asset corresponding to `(1, Parachain(9000))`.

zjb0807 (approved):

PR #356, opened 2024-06-24
Fast Fellowship Promote Tracks

ggwpez:
Add fast promotion tracks with 7 days (instead of 30 days) voting period.  
The tracks have a 50% turnout and 66% approval criteria. To be used in combination with https://github.com/paritytech/polkadot-sdk/pull/4778.

It could be that the 50% turnout is too much for tracks I and II, but for III there are not so many fellows so we can expect them to all vote IMHO.


system-parachains/collectives/collectives-polkadot/src/fellowship/tracks.rs
```diff
@@ -110,13 +119,32 @@ const PROMOTE_MIN_SUPPORT: pallet_referenda::Curve = pallet_referenda::Curve::Li
 	ceil: Perbill::from_percent(100),
 };
 
+const FAST_PROMOTE_MAX_DECIDING: u32 = 10;
+const FAST_PROMOTE_DECISION_DEPOSIT: Balance = 5 * DOLLARS;
+const FAST_PROMOTE_PREPARE_PERIOD: BlockNumber = 0;
+const FAST_PROMOTE_DECISION_PERIOD: BlockNumber = 7 * DAYS;
```
ggwpez, line 125: Dunno if this is too short? The origin is already "faster" since it can bypass the promotion period.  
On the other hand, with the high turnout it will rather error on the side of rejecting so maybe its fine.

acatangiu: The main issue is having to wait for years to get to a correct level.

If this can bypass the promotion min period, I don't see the need to have it faster than the "normal" 30 days, I don't expect members to find 30 days too slow..

joepetrowski: > The main issue is having to wait for years to get to a correct level.

Yeah that needs to be fixed by making a call in Core Fellowship that will respect this origin.

> Dunno if this is too short?

I think it is OK, could also go in the middle with something like 14 days.

ggwpez: No strong opinions, so changed to 30 days after internal discussion.

joepetrowski (approved):

acatangiu (approved):

PR #357, opened 2024-06-24
remove delegated staking from prod runtimes

Ank4n:
pallet-delegated-staking and associated changes for nomination pool is still under audit and needs some issues to be fixed before prod runtimes. Setting the following is functionally no-op.
```rust
type StakeAdapter = pallet_nomination_pools::adapter::TransferStake<Self, Staking>
```


ggwpez:
Thanks!

PR #361, opened 2024-06-26
Add PoV-reclaim to kusama system chains

skunert:
Enabled PoV-reclaim on kusama system parachains.

Based on https://github.com/polkadot-fellows/runtimes/pull/322, since reclaim was introduced in 1.9.0.


ggwpez:
Thanks!

PR #364, opened 2024-06-27
Kusama: Make the current inflation formula adjustable. 

kianenigma:
This PR does not alter the inflation of Kusama, but instead make the current parameters of the system fully configurable. 

## Parameters

The parameters are as follows: 

```rust
/// Minimum inflation rate used to calculate era payouts.
pub static MinInflation: Perquintill = Perquintill::from_rational(25u64, 1000);

/// Maximum inflation rate used to calculate era payouts.
pub static MaxInflation: Perquintill = Perquintill::from_percent(10);

/// Ideal stake ratio used to calculate era payouts.
pub static IdealStake: Perquintill = Perquintill::from_percent(75);

/// Falloff used to calculate era payouts.
pub static Falloff: Perquintill = Perquintill::from_percent(5);

/// Whether to use auction slots or not in the calculation of era payouts, then we subtract
/// `num_auctioned_slots.min(60) / 200` from `ideal_stake`.
///
/// That is, we assume up to 60 parachains that are leased can reduce the ideal stake by a
/// maximum of 30%.
///
/// With the move to agile-coretime, this parameter does not make much sense and should
/// generally be set to false.
pub static UseAuctionSlots: bool = true;
```

All of the above are exactly the current values in Kusama, and leave everything unchanged, leading to roughly `1kKSM` minted for validators, and around `90KSM` for treasury. 

All of the above can be changed via the `Root` track only. 

Given that it is hard to come to consensus, I highly advise this PR (and a similar PR to Polkadot) to NOT alter any parameter and leave that to the token holders. 

Just to provide one example, if we set `UseAuctionSlots = false` and leave everything else un-changed, the amount minted per era would be: 

* `807KSM` for staking
* `238KSM` for treasury

> This, in my opinion, is the most non-controversial change as the incorporation of auctions in the inflation rate is already meaningless with agile-coretime. 

## Consideration for UIs 

This PR is shipped with a new runtime API that is only added to the Kusama runtime: `Inflation_experimental_inflation_info`. I hope this API is used by the UIs to show the inflation parameters of the network, rather than re-creating the Rust logic in the front-end.


kianenigma:
Some updates: 

1. Depends on https://github.com/polkadot-fellows/runtimes/pull/381
2. Would also preferably depend on sdk 1.15 to get the new `relay_inflation` helper fn. 
3. Based on the last fellowship call, for now we will only add this to Kusama https://www.youtube.com/live/oWppBMrgU7M?si=b4O-G_mIEBp1tnV4&t=2137

kianenigma:
Once https://github.com/polkadot-fellows/runtimes/pull/381 is merged, I shall merge https://github.com/kianenigma/runtimes/tree/kiz-inflation-with-update here.

kianenigma:
Well, I called this simulation more or less in the middle of an era, so the era duration is assumed to be half :)) So the code adds up.

relay/kusama/src/lib.rs
```diff
@@ -1774,14 +1936,86 @@ mod benches {
...
+
+use apis::*;
+
+impl Runtime {
+	fn impl_experimental_inflation_info() -> InflationInfo {
+		use pallet_staking::{ActiveEra, EraPayout, ErasTotalStake};
+		let (staked, _start) = ActiveEra::<Runtime>::get()
+			.map(|ae| (ErasTotalStake::<Runtime>::get(ae.index), ae.start.unwrap_or(0)))
+			.unwrap_or((0, 0));
+		let stake_able_issuance = Balances::total_issuance();
```
ggwpez, line 1984: Why is it using balances here and not excluding the disabled issuance with `Nis::issuance().other`?

relay/kusama/src/lib.rs
```diff
@@ -2825,6 +3059,75 @@ mod remote_tests {
...
+		use hex_literal::hex;
+		sp_tracing::try_init_simple();
+		let transport: Transport =
+			var("WS").unwrap_or("wss://rpc.dotters.network/kusama".to_string()).into();
+		let mut ext = Builder::<Block>::default()
+			.mode(Mode::Online(OnlineConfig {
+				transport,
+				hashed_prefixes: vec![
+					// entire nis pallet
+					hex!("928fa8b8d92aa31f47ed74f188a43f70").to_vec(),
```
ggwpez, line 3073: Downloading the state lazily would be much smarter -  no need to declare upfront what keys are required.

ggwpez (approved):

relay/kusama/src/lib.rs
kianenigma: yeah, in this case it is fine as nis pallet is mostly empty.

remote-ext still doesn't have lazy download, right? or are you suggesting it because it now has it :o?

relay/kusama/src/lib.rs
kianenigma, line 1984:
```suggestion
		let stake_able_issuance = Nis::issuance().other;
```

kianenigma: great catch!

relay/kusama/src/lib.rs
ggwpez: >or are you suggesting it because it now has it :o?

Nah i think it still has not. But could be nice 🙃

relay/kusama/Cargo.toml
```diff
@@ -113,6 +115,7 @@ separator = { workspace = true }
 remote-externalities = { workspace = true }
 tokio = { features = ["macros"], workspace = true }
 sp-tracing = { workspace = true }
+ss58-registry = "1.47.0"
```
ggwpez, line 118: Zepter not happy here since all deps need to be declared in the workspace.

relay/kusama/src/lib.rs
```diff
@@ -1774,14 +1936,86 @@ mod benches {
...
+
+use apis::*;
+
+impl Runtime {
+	fn impl_experimental_inflation_info() -> InflationInfo {
+		use pallet_staking::{ActiveEra, EraPayout, ErasTotalStake};
+		let (staked, _start) = ActiveEra::<Runtime>::get()
+			.map(|ae| (ErasTotalStake::<Runtime>::get(ae.index), ae.start.unwrap_or(0)))
+			.unwrap_or((0, 0));
+		let stake_able_issuance = Nis::issuance().other;
```
gpestana, line 1984:
```suggestion
		let stakable_issuance = Nis::issuance().other;
```

nit but for consistency with the rest of the code.

relay/kusama/src/lib.rs
```diff
@@ -1774,14 +1936,86 @@ mod benches {
...
+
+		let ideal_staking_rate = dynamic_params::inflation::IdealStake::get();
+		let inflation = if dynamic_params::inflation::UseAuctionSlots::get() {
+			let auctioned_slots = parachains_paras::Parachains::<Runtime>::get()
+				.into_iter()
+				// all active para-ids that do not belong to a system chain is the number of
+				// parachains that we should take into account for inflation.
+				.filter(|i| *i >= 2000.into())
```
gpestana, line 1992:
```suggestion
				.filter(|i| *i >= LOWEST_PUBLIC_ID.into())
```

relay/kusama/src/lib.rs
```diff
@@ -1774,14 +1936,86 @@ mod benches {
...
+impl Runtime {
+	fn impl_experimental_inflation_info() -> InflationInfo {
+		use pallet_staking::{ActiveEra, EraPayout, ErasTotalStake};
+		let (staked, _start) = ActiveEra::<Runtime>::get()
+			.map(|ae| (ErasTotalStake::<Runtime>::get(ae.index), ae.start.unwrap_or(0)))
+			.unwrap_or((0, 0));
+		let stake_able_issuance = Nis::issuance().other;
+
+		let ideal_staking_rate = dynamic_params::inflation::IdealStake::get();
+		let inflation = if dynamic_params::inflation::UseAuctionSlots::get() {
```
gpestana, line 1987: I'd probably refactor the body of this if statement to its own method. It is used in the `fn era_payout` as well and future changes maybe lead to inconsistencies between the two `inflation` values in the era payout and in the runtime API, if devs are not careful. Especially since there are some magic numbers (e.g. `Perquintil (auctioned_slots.min(60), 200u64)`) in it.

relay/kusama/src/lib.rs
```diff
@@ -1774,14 +1936,86 @@ mod benches {
...
+				// parachains that we should take into account for inflation.
+				.filter(|i| *i >= 2000.into())
+				.count() as u64;
+			ideal_staking_rate
+				.saturating_sub(Perquintill::from_rational(auctioned_slots.min(60), 200u64))
+		} else {
+			ideal_staking_rate
+		};
+
+		// we assume un-delayed 6h eras.
```
gpestana, line 2000: Due to this assumption, maybe we should name the runtime API as `exp_inflation_predicted_info`?

relay/kusama/src/lib.rs
```diff
@@ -623,34 +630,188 @@ impl pallet_bags_list::Config<VoterBagsListInstance> for Runtime {
...
+		// Provide the origin for the parameter returned by `Default`:
+		Ok(RuntimeOrigin::root())
+	}
+}
+
+impl pallet_parameters::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type RuntimeParameters = RuntimeParameters;
+	type AdminOrigin = DynamicParameterOrigin;
+	// TODO: add benchmarking and update weight info
```
gpestana, line 710: ping to remember before merging.

gpestana (approved):
LGTM modulo a few comments and suggestions and the benchmarks for the pallet parameters.

relay/kusama/src/lib.rs
kianenigma: Let's discuss this once I merge this and move on to Polkadot, which should follow soon :)

relay/kusama/src/lib.rs
```diff
@@ -623,34 +630,186 @@ impl pallet_bags_list::Config<VoterBagsListInstance> for Runtime {
...
+			} else {
+				None
+			},
+		};
+		log::debug!(target: "runtime::kusama", "params: {:?}", params);
+		relay_era_payout(params)
 	}
 }
 
+// ---- TODO: Below is copy pasted from sdk, remove once we pull the version containing
```
kianenigma, line 747: As a final step, moving this to a shared crate as it will be needed for Polkadot as well.

kianenigma:
bot fmt

franciscoaguirre (approved):

relay/common/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+	if total_staked > other_issuance {
+		let _cap_rest = Perquintill::from_rational(other_issuance, total_staked) * staking_payout;
+		// We don't do anything with this, but if we wanted to, we could introduce a cap on the
+		// treasury amount with: `rest = rest.min(cap_rest);`
+	}
+	(staking_payout, rest)
+}
+
+// ---- TODO: Above is copy pasted from sdk, remove once we pull the version containing
+// https://github.com/paritytech/polkadot-sdk/pull/4938
```
acatangiu, line 115: would be better if you referenced the crate name and (min) version required, instead of this PR, otherwise it becomes a non-trivial chore to figure out when this can be removed and no one will do it :)

(same above)

acatangiu (approved):
high-level looks good, did not double-check the actual staking initial values coded in

relay/common/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+	pub ideal_stake: Perquintill,
+	/// Maximum inflation rate.
+	pub max_annual_inflation: Perquintill,
+	/// Minimum inflation rate.
+	pub min_annual_inflation: Perquintill,
+	/// Falloff used to calculate era payouts.
+	pub falloff: Perquintill,
+	/// Fraction of the era period used to calculate era payouts.
+	pub period_fraction: Perquintill,
+	/// Legacy auction proportion, which substracts from `ideal_stake` if not `None`.
```
joepetrowski, line 74:
```suggestion
	/// Legacy auction proportion, which, if not `None`, is subtracted from `ideal_stake`.
```

relay/common/src/lib.rs
```diff
@@ -0,0 +1,115 @@
...
+	pub legacy_auction_proportion: Option<Perquintill>,
+}
+
+/// A specialized function to compute the inflation of the staking system, tailored for polkadot
+/// relay chains, such as Polkadot, Kusama and Westend.
```
joepetrowski, line 79:
```suggestion
/// A specialized function to compute the inflation of the staking system, tailored for Polkadot
/// Relay Chains, such as Polkadot, Kusama, and Westend.
```

relay/kusama/src/lib.rs
```diff
@@ -623,31 +631,117 @@ impl pallet_bags_list::Config<VoterBagsListInstance> for Runtime {
...
+		/// Ideal stake ratio used to calculate era payouts.
+		#[codec(index = 2)]
+		pub static IdealStake: Perquintill = Perquintill::from_percent(75);
+
+		/// Falloff used to calculate era payouts.
+		#[codec(index = 3)]
+		pub static Falloff: Perquintill = Perquintill::from_percent(5);
+
+		/// Whether to use auction slots or not in the calculation of era payouts, then we subtract
+		/// `num_auctioned_slots.min(60) / 200` from `ideal_stake`.
```
joepetrowski, line 661:
```suggestion
		/// Whether to use auction slots or not in the calculation of era payouts. If true, then we
		/// subtract `num_auctioned_slots.min(60) / 200` from `ideal_stake`.
```

relay/kusama/src/lib.rs
```diff
@@ -623,31 +631,117 @@ impl pallet_bags_list::Config<VoterBagsListInstance> for Runtime {
...
+		///
+		/// That is, we assume up to 60 parachains that are leased can reduce the ideal stake by a
+		/// maximum of 30%.
+		///
+		/// With the move to agile-coretime, this parameter does not make much sense and should
```
joepetrowski, line 666:
```suggestion
		/// With the move to Agile Coretime, this parameter does not make much sense and should
```

PR #366, opened 2024-06-28
Update bridges integration-tests instructions

bkontur:
This PR adjusts the instructions on how to run bridges integration tests.

Future improvements: https://github.com/paritytech/parity-bridges-common/issues/3016

### DIFF
--- a/integration-tests/bridges/README.md
+++ b/integration-tests/bridges/README.md
@@ -5,26 +5,30 @@ The tests are designed to be run manually.
 
 To start a test, you need to:
 
-- download latest [zombienet release](https://github.com/paritytech/zombienet/releases) to 
-`~/local_bridge_testing/bin/zombienet`.
-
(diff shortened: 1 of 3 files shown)


acatangiu (approved):

bkontur:
> Looks quite complicated and also raises the question why you need the sudo pallet with fast runtime? Let's merge this, but would be nice to improve this.

Yes, you're right. That's why I created [issue #3016](https://github.com/paritytech/parity-bridges-common/issues/3016). When we started with the zombienet bridges tests, `preopen_hrmp` stopped working at some point ([more info](https://github.com/paritytech/polkadot-sdk/pull/1616)). As a workaround, I used `force_open_hrmp_channel` because `establish_system_channel` didn't exist at that time. Sudo is needed for the `init-bridge` governance call from the relay, but that could probably be replaced with a genesis setup or with Chopsticks and governance call.

PR #371, opened 2024-07-02
Adds MetadataHash calculation on Build to Kusama People chain

SBalaguer:
This adds the calculation of the MetadataHash upon build in the Kusama People Chains Runtime. I don't think this needs a CHANGELOG entry, but I could be mistaken.


bkchr:
I think you should add a changelog entry.

SBalaguer:
Joe suggested the same, under ###FIXED. Just added it.

CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ Note: This release only affects the following runtimes and is not a full system
 ### Fixed
 
 - Kusama People: clear requested judgements that do not have corresponding deposits reserved ([polkadot-fellows/runtimes#339](https://github.com/polkadot-fellows/runtimes/pull/339))
+- Kusama People: adds the calculation of the MetadataHash upon build to be used on CheckMetadata SignedExtension ([polkadot-fellows/runtimes#371](https://github.com/polkadot-fellows/runtimes/pull/371))
```
bkchr, line 46:
```suggestion
- Kusama People: Build the metadata has at build time, so that `CheckMetadata` can use it at runtime ([polkadot-fellows/runtimes#371](https://github.com/polkadot-fellows/runtimes/pull/371))
```

bkchr (approved):

skunert (approved):

CHANGELOG.md
SBalaguer: Thanks for the suggestion, I'll change it shortly. Now this will be moved back to 1.2.6 or should it be part of 1.2.7's CHANGELOG?

bkchr: Should be part of the `Unreleased` section only.

SBalaguer: Yeah added that on my last commit, Joe pointed me on that direction too. Thanks!

CHANGELOG.md
```diff
@@ -43,6 +47,7 @@ Note: This release only affects the following runtimes and is not a full system
...
```
bkchr, line 50:
```suggestion
```

xlc (approved):

Issue #358, opened 2024-06-25
v1.2.8 Release


claravanstaden:
Bump BridgeHub mainly to upgrade Snowbridge to sync [headers on demand](https://github.com/paritytech/polkadot-sdk/pull/4829) instead of all the headers, reducing cost.

- [x] https://github.com/polkadot-fellows/runtimes/pull/365

PR #374, opened 2024-07-03
Trigger Release v1.2.8

claravanstaden:
Triggers the release of https://github.com/polkadot-fellows/runtimes/issues/358 to reduce costs to run the Ethereum client.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,8 +17,6 @@ Note: This release only affects the following runtimes and is not a full system 
 - Polkadot Asset Hub
 - Polkadot Bridge Hub
 - Polkadot Collectives
-- Kusama Relay Chain
-- Kusama Asset Hub
 - Kusama Bridge Hub
 
 ## [1.2.7] 14.06.2024
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,22 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## [Unreleased]
+## [1.2.8] 03.07.2024
 
 ### Changed
 
 - Polkadot chains: allow arbitrary XCM execution ([polkadot-fellows/runtimes#345](https://github.com/polkadot-fellows/runtimes/pull/345))
 - Snowbridge: Sync headers on demand ([polkadot-fellows/runtimes#345](https://github.com/polkadot-fellows/runtimes/pull/365))
+
+Note: This release only affects the following runtimes and is not a full system release:
+
+- Polkadot Relay Chain
+- Polkadot Asset Hub
+- Polkadot Bridge Hub
+- Polkadot Collectives
+- Kusama Relay Chain
+- Kusama Asset Hub
+- Kusama Bridge Hub
 
 ## [1.2.7] 14.06.2024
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,13 +4,23 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
...
 
+Note: This release only affects the following runtimes and is not a full system release:
+
+- Polkadot Relay Chain
+- Polkadot Asset Hub
+- Polkadot Bridge Hub
+- Polkadot Collectives
+- Kusama Relay Chain
+- Kusama Asset Hub
+- Kusama Bridge Hub
```
claravanstaden, line 22: @acatangiu Maybe you please check the affected chains list? I bumped the Collectives spec version in the PR. The rest have been updated to `1_002_008` already.

acatangiu: Everything looks good except Kusama Relay and Kusama Asset Hub.

Looking at their history since last release(s), nothing has changed except a code comment and the bumped spec version you did.

I think we should actually go back/revert to `spec_version: 1_002_006` for `Kusama Relay Chain` and `Kusama Asset Hub`.

All others have received changes worth releasing (Snowbridge improvement and/or XCM execution filters removed).

claravanstaden: @acatangiu I removed the Kusama AH and relay bump. However, CI wants the Kusama Relay bump: https://github.com/polkadot-fellows/runtimes/actions/runs/9778803977/job/26996484789?pr=374

acatangiu (approved):
@rzadp why is the kusama-migrations CI check expecting spec_version bump when there have been no changes to the runtime (except a code comment AFAICT)? Any way we can bypass it?

@bkchr @joepetrowski what is the rule we actually want here? bump independent runtimes only when there is actually something to release or should we bump and release for any small/noop change?

joepetrowski:
> what is the rule we actually want here? bump independent runtimes only when there is actually something to release or should we bump and release for any small/noop change?

In general I think it makes sense to release all runtimes as one system version, but of course for some patches / bug fixes it will only be some subset that needs to be upgraded. IMO in these cases it makes sense only to release what needs to be released.

bkchr:
> why is the kusama-migrations CI check expecting spec_version bump when there have been no changes to the runtime (except a code comment AFAICT)? Any way we can bypass it?

The check doesn't know what you changed. It just compares the on chain and the current version. Both are the same and thus, it complains. There is no other fix than bumping it right now.

acatangiu (approved):

bkontur (approved):

bkontur (approved):

bkchr (approved):

ggwpez (approved):

joepetrowski:
Since this includes the Relay Chain, we need to be careful in how we apply it. It should happen after the People Chain is producing blocks. I am planning on a rough timeline like this:

- 8 July: Polkadot Relay v1.2.6 enacts and freezes Identity
- 8 July: Make a proposal for Polkadot People genesis to enact as soon as it passes
- 11 July: Make a proposal for Polkadot Relay v1.2.7 (allows migration) with enactment for 25 July

For the Relay Chain, these need to be applied in order, and v1.2.8 can enact after. If you only need to upgrade Bridge Hub, go ahead.

acatangiu:
yeah, we can upgrade just BH for now

acatangiu:
@joepetrowski how do we actually _trigger_ the release? merging this PR didn't do it by itself..

joepetrowski:
Wait patiently:
image

PR #381, opened 2024-07-14
Update to SDK 1.14

ggwpez:
Updating the runtimes to SDK version 1.14. CHANGELOG mentions all relevant changes for UI and Wallet builders.

PR #384, opened 2024-07-15
Kusama coretime revenue integration

s0me0ne-unkn0wn:
Based on #381 

CC @ggwpez

### DIFF
--- a/system-parachains/coretime/coretime-kusama/src/coretime.rs
+++ b/system-parachains/coretime/coretime-kusama/src/coretime.rs
@@ -21,14 +21,18 @@ use cumulus_primitives_core::relay_chain;
 use frame_support::{
 	parameter_types,
 	traits::{
-		fungible::{Balanced, Credit},
-		OnUnbalanced,
+		fungible::{Balanced, Credit, Inspect},
+		tokens::{Fortitude, Preservation},
+		DefensiveResult, OnUnbalanced,
 	},
 };
+use frame_system::Pallet as System;
 use kusama_runtime_constants::system_parachain::coretime;
 use pallet_broker::{CoreAssignment, CoreIndex, CoretimeInterface, PartsOf57600, RCBlockNumberOf};
+use parachains_common::{AccountId, Balance};
 use sp_runtime::traits::AccountIdConversion;
 use xcm::latest::prelude::*;
+use xcm_executor::traits::TransactAsset;
 
 /// A type containing the encoding of the coretime pallet in the Relay chain runtime. Used to
 /// construct any remote calls. The codec index must correspond to the index of `Coretime` in the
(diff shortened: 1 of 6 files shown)


system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -184,9 +227,22 @@ impl CoretimeInterface for CoretimeAllocator {
 		}
 	}
 
-	#[cfg(feature = "runtime-benchmarks")]
-	fn ensure_notify_revenue_info(when: RCBlockNumberOf<Self>, revenue: Self::Balance) {
-		CoretimeRevenue::set(&Some((when, revenue)));
+	fn on_new_timeslice(_t: pallet_broker::Timeslice) {
+		let stash = CoretimeBurnAccount::get();
```
seadanda, line 231: Every timeslice (8 mins) still seems quite often to be sending an XCM just to burn accumulated revenue - my initial thought was to do this when a sale is rotated. Every N timeslices might be better, even if just once per day. WDYT?

```suggestion
	fn on_new_timeslice(t: pallet_broker::Timeslice) {
		// Burn roughly once per day. Unchecked math: RHS hardcoded as non-zero.
		if t % 180 != 0 {
			return
		}
```

seadanda (commented):
Looks good, just a query about the period of the burn. Looks like cargo.lock needs an update too.

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -184,9 +227,22 @@ impl CoretimeInterface for CoretimeAllocator {
...
+		let value =
+			Balances::reducible_balance(&stash, Preservation::Expendable, Fortitude::Polite);
```
seadanda, line 233: Ah also this - we don't cover the ED before payments go in, so I think this should take the whole balance to ensure that we are actually burning all revenue as the RFC states.

s0me0ne-unkn0wn: If I just replace `reducible_balance` with `total_balance`, will teleport logic succeed in checking the funds out and withdrawing them, considering that the account balance falls under the ED?

s0me0ne-unkn0wn: Going through the logic now to understand how it works. The parachain side doesn't have `CheckingAccount`, so `can_check_out` will always succeed. Then, `AssetTransactor::withdraw_asset` will call `Fungible::burn_from(..., Expendable, Exact, Polite)`, which, in turn, will call for `Fungible::reducible_balance(..., Expendable, Polite)`. That is, my check exactly matches the check `withdraw_asset` will perform. Given that the account has an additional provider ref from `OnUnbalanced` implementation, it should still return the total free balance of the account not reserving the ED: https://github.com/paritytech/polkadot-sdk/blob/926c1b6adcaa767f4887b5774ef1fdde75156dd9/substrate/frame/balances/src/impl_fungible.rs#L65-L66

To me, it seems sensible to keep it as is to match this `reducible_balance` check to the checks performed by the `withdraw_asset` implementation. WDYT?

seadanda: Sorry, missed your response in my notifications.
Maybe worth having this discussion on the main PR to include others - I'm happy to merge as is to not block, since it is a legit solution anyway

seadanda (approved):

s0me0ne-unkn0wn:
@ggwpez will you just merge it into your branch? I didn't commit `Cargo.lock` checks as there are still some version inconsistencies, and I don't think going through the CI here makes sense.

PR #386, opened 2024-07-16
Reduce Bounty Payout Delay

joepetrowski:
The bounty payout delay originally existed to provide time for the Council to remove curators of a bounty if they were not acting in accordance with its existence. The Council motion time was 7 days, and thus the bounty delay was 8 days to give the Council time to remove curators and void payouts prior to a bounty being claimable.

With OpenGov tracks being 28 days, I don't see this as anything more than a PITA for curators/awardees. Does anyone have motivation for this delay to be in effect in the current setup? (cc @muharem , @shawntabrizi )

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+### Changed
+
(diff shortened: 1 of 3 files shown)


shawntabrizi:
Your logic to remove it makes sense to me.

Ideally we could still prevent bounties from being raided by curators, but I guess this will happen through denying "top ups" rather than through the spend.

acatangiu (approved):

muddlebee:
@joepetrowski which Runtimes version this will be part of? I dont see it in changelog of upcoming ones 👀

joepetrowski:
image

PR #393, opened 2024-07-18
Re-enable `request_revenue_info_at` in CoretimeInterface

seadanda:
In the update to [sdk 1.14](https://github.com/polkadot-fellows/runtimes/pull/381) the `request_revenue_info` is now implemented but this call had been switched off when it was added in https://github.com/polkadot-fellows/runtimes/pull/212.

This PR just enables this again in the coretime interface.

cc @ggwpez @s0me0ne-unkn0wn


s0me0ne-unkn0wn (approved):
:+1:

Issue #396, opened 2024-07-23
Check `merge bot` and required pipelines


bkontur:
E.g. this [PR](https://github.com/polkadot-fellows/runtimes/pull/380) was merged but one pipeline `integration-test (bridge-hub-kusama, bridge-hub-kusama-integration-tests)` failed, how is that possible?

![image](https://github.com/user-attachments/assets/5817208b-8123-45fc-89bf-271817d26c8e)

bkontur:
cc: @Bullrich

Bullrich:
The method `continue-on-error: true` seems to be the main offender.

We should remove it in both places:
- https://github.com/polkadot-fellows/runtimes/blob/7723274a2c5cbb10213379271094d5180716ca7d/.github/workflows/test.yml#L48
- https://github.com/polkadot-fellows/runtimes/blob/7723274a2c5cbb10213379271094d5180716ca7d/.github/workflows/test.yml#L100

bkontur:
I removed `continue-on-error: true` here: https://github.com/polkadot-fellows/runtimes/pull/397 and I will also add a fix.

Maybe `fail-fast: false` looks better: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

bkontur:
definitely, `continue-on-error: true` is not good, because tests failed and this pipeline passed :)
![image](https://github.com/user-attachments/assets/b9a0eb8f-248f-4644-ac56-518ce405511d)

bkontur:
@Bullrich I removed `continue-on-error: true` and now that pipeline [`confirmTestPassed`](https://github.com/polkadot-fellows/runtimes/blob/main/.github/workflows/test.yml#L251C3-L251C20) is marked as a skipped:
`Test all features / All tests passed (pull_request) Skipped `

![image](https://github.com/user-attachments/assets/78c5ca8c-90c4-4b9a-8a8d-7ef435e9fe47)


So, is that `confirmTestPassed` marked as a required to block merge in the github configuration?

bkchr:
Marked all the "All ****" as required.

bkontur:
Fixed by https://github.com/polkadot-fellows/runtimes/pull/397

PR #380, opened 2024-07-12
Add XcmPaymentApi and DryRunApi to all runtimes

franciscoaguirre:
Adds the https://github.com/paritytech/polkadot-sdk/pull/3607 and https://github.com/paritytech/polkadot-sdk/pull/3872 to all runtimes.
These APIs work together to allow dry running and estimating execution and delivery fees for XCMs.

This PR doesn't allow querying the price in assets different than the relay token for asset hubs. Can be done in a following PR.
Also tracking https://github.com/polkadot-fellows/runtimes/issues/388 and https://github.com/polkadot-fellows/runtimes/issues/389 for future improvemens.

Old PR was https://github.com/polkadot-fellows/runtimes/pull/359, this one targets main already updated to Polkadot SDK 1.13.

## Dear reviewer

Although there are a lot of changes, the main one is the addition of the two aforementioned APIs to all relays and system parachains. The rest of the files are tests, which all have the same format.

PR #397, opened 2024-07-23
Fix CI + one integration test for `bridge-hub-kusama`

bkontur:
Closes: https://github.com/polkadot-fellows/runtimes/issues/396

This PR:
- fixes CI test pipelines to block PRs when tests fail. See more details [here](https://github.com/polkadot-fellows/runtimes/issues/396).
- Fixes an integration test for `bridge-hub-kusama-integration-tests` that was not caught due to a previous CI bug.
- Includes some minor cleanup, changing `Versioned*::V4` to `Versioned*::from` to use the latest version (for easier migration to `xcm:v5`).

## TODO

- [x] fix the integration tests

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -51,7 +51,7 @@ jobs:
   # If you add more jobs, remember to add them to the "needs" array.
   confirmChangelogChecksPassed:
     runs-on: ubuntu-latest
-    name: All checks passed
+    name: All changelog checks passed
     # If any new job gets added, be sure to add it to this list
     needs:
       - verify-changelog-updated
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -57,4 +57,4 @@ jobs:
       - verify-changelog-updated
       - verify-changelog-valid
     steps:
-      - run: echo '### Good job! All the checks passed 🚀' >> $GITHUB_STEP_SUMMARY
+      - run: echo '### Good job! All the changelog checks passed 🚀' >> $GITHUB_STEP_SUMMARY
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -50,10 +50,10 @@ jobs:
   # If you add more jobs, remember to add them to the "needs" array.
   confirmChangelogChecksPassed:
     runs-on: ubuntu-latest
-    name: All tests passed
+    name: All checks passed
     # If any new job gets added, be sure to add it to this list
     needs:
       - verify-changelog-updated
       - verify-changelog-valid
     steps:
-      - run: echo '### Good job! All the tests passed 🚀' >> $GITHUB_STEP_SUMMARY
+      - run: echo '### Good job! All the checks passed 🚀' >> $GITHUB_STEP_SUMMARY
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -1,6 +1,6 @@
 name: Verify Changelog
 
-# If you modify more test jobs, ensure that you add them as required to the job "confirmTestPassed"
+# If you modify more test jobs, ensure that you add them as required to the job "confirmChangelogChecksPassed"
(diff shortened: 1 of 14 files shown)


Bullrich (approved):

.github/workflows/changelog.yml
```diff
@@ -50,10 +51,10 @@ jobs:
   # If you add more jobs, remember to add them to the "needs" array.
   confirmChangelogChecksPassed:
     runs-on: ubuntu-latest
-    name: All tests passed
+    name: All checks passed
```
Bullrich, line 54: Good catch on changing this name.

Though it will require an admin to update the `required CI checks` as the system is waiting for the old name.

I would also suggest to name it a bit more specific

```suggestion
    name: All changelog checks passed
```

.github/workflows/changelog.yml
```diff
@@ -50,10 +51,10 @@ jobs:
...
     # If any new job gets added, be sure to add it to this list
     needs:
       - verify-changelog-updated
       - verify-changelog-valid
     steps:
-      - run: echo '### Good job! All the tests passed 🚀' >> $GITHUB_STEP_SUMMARY
+      - run: echo '### Good job! All the checks passed 🚀' >> $GITHUB_STEP_SUMMARY
```
Bullrich, line 60:
```suggestion
      - run: echo '### Good job! All the changelog checks passed 🚀' >> $GITHUB_STEP_SUMMARY
```

.github/workflows/changelog.yml
bkontur: originally I had it exactly like that `All changelog checks passed` but I removed `changelog` word before commit :D

xlc (approved):

integration-tests/emulated/tests/bridges/bridge-hub-kusama/src/tests/asset_transfers.rs
```diff
@@ -206,15 +165,8 @@ fn send_ksms_from_asset_hub_kusama_to_asset_hub_polkadot() {
...
 
-	dbg!(&sender_ksms_after);
-	dbg!(&sender_ksms_before);
-	dbg!(&amount);
-	dbg!(&delivery_fees);
-	dbg!(&receiver_ksms_after);
-	dbg!(&receiver_ksms_before);
-
 	// Sender's balance is reduced
-	assert_eq!(sender_ksms_after, sender_ksms_before - amount - delivery_fees);
```
seadanda, line 217: Is this no longer the case or are you just making the test less brittle?

seadanda (approved):

muharem (approved):

integration-tests/emulated/tests/bridges/bridge-hub-kusama/src/tests/asset_transfers.rs
bkontur: It was not correct, at least, it missed the subtraction of `execution_fees`. I tried to add them as well, but [`let result = Runtime::dry_run_call(origin, call).unwrap();`](https://github.com/polkadot-fellows/runtimes/pull/397/files#diff-44fd5c77737282dbb2dfbba20d499bbc921d1d00943b304fbacc885c82c7edcdR60) did not return `actual_weight`, so it is not possible to calculate `WeightToFee`. Additionally, there was some calculation of `intermediate_delivery_fees_amount` from BHK to BHP, which is not relevant. So yes, keep it simple. At least I added asserts to check that the dry-run succeeded and the message is ready to be relayed over bridge to the BHP.

acatangiu: This was added by @franciscoaguirre, fee estimation doesn't yet work with BridgeHub exporters: https://github.com/polkadot-fellows/runtimes/pull/380#issuecomment-2233024669

The check will be added back when that is fixed as well.

acatangiu (approved):

bkchr (approved):

PR #398, opened 2024-07-23
Reduce the base fee on Polkadot System Chains

vgantchev:
This PR reduces the base fee on all Polkadot system chains by half.  As a result, the base fee will be roughly 1/20 of the base fee on the Polkadot relay chain.

This will make all transactions on (non-congested) system chains cheaper and stimulate usage.

Among others, the change will push the transaction fees for transfers on Asset Hub below $0.01 which is an important psychological threshold.

An initiative by ecosystem agents.


Tomen:
Thank you @vgantchev for making this PR happen!

bkchr:
Generally I think we can do this. However, do we have any kind of calculation on the economics what a block/tx should cost? @joepetrowski any idea?

joepetrowski:
> Generally I think we can do this. However, do we have any kind of calculation on the economics what a block/tx should cost? @joepetrowski any idea?

Nothing quantitative, the rationale for 10x lower at first was that it's better for the network if someone uses a parachain for a feature instead of the Relay Chain, so it should be cheaper to nudge people there. IMO having a low base fee is fine as the fee-adjustment mechanism will bring fees up to previous levels if there are a lot of transactions. It's just a lot more adaptable autonomously than storage deposits are. So, I'm OK with this change.

bkchr:
@vgantchev could you then please fix all the tests? Then we can go ahead with this pr.

bkontur:
> @vgantchev could you then please fix all the tests? Then we can go ahead with this pr.

a fix for `bridge-hub-kusama` integration tests is coming here: https://github.com/polkadot-fellows/runtimes/pull/397

vgantchev:
Reducing the base fee caused [this integration test to fail](https://github.com/polkadot-fellows/runtimes/actions/runs/10078167663/job/27862464333?pr=398).  

It's a but cryptic but it looks like the reason is ED, [lowering it also by half](https://github.com/polkadot-fellows/runtimes/pull/398/commits/c7457b1e81d93b5876d953adade8171342a5ae2a) fixes the test.

Are you fine with lowering also the ED, or do you have another suggestion?

acatangiu:
> Are you fine with lowering also the ED, or do you have another suggestion?

No, lowering ED has DoS implications outside the scope of this PR. That test should work without needing to change ED. Will take a look.

acatangiu:
@vgantchev can't push to your fork, opened PR instead: https://github.com/vgantchev/polkadot-runtimes/pull/1

bkchr (approved):

xlc (approved):

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - Bounties: Remove payout delay ([polkadot-fellows/runtimes#386](https://github.com/polkadot-fellows/runtimes/pull/386))
+- Polkadot System Chains: Reduce the base transaction fee by half
```
ggwpez, line 17: Should please follow the standard convention of how we mention them here. Will fix it in my MR now.

bkchr: Good point. Another thing we should add to the CI check.

PR #400, opened 2024-07-24
CI: updated all deprecated actions

Bullrich:
I have noticed there are a lot of outdated actions with deprecated warnings:
image

This PR updates all of GitHub actions to latest version to remove such warnings.

I have also added some`TODO` to actions that have been archived by GitHub and that require an alternative action/solution. We can do that in a different PR once we identify the correct solution.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -150,6 +150,7 @@ jobs:
 
       - name: Create release
         id: create-release
+        # TODO: Replace as it has been deprecated
         uses: actions/create-release@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -183,6 +184,7 @@ jobs:
           >>$GITHUB_ENV echo SPEC=$(<${JSON} jq -r .runtimes.compact.subwasm.core_version.specVersion)
 
       - name: Upload compressed ${{ matrix.runtime.name }} v${{ env.SPEC }} wasm
+        # TODO: Replace as it has been deprecated
         uses: actions/upload-release-asset@v1
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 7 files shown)


.github/workflows/release.yml
```diff
@@ -150,6 +150,7 @@ jobs:
 
       - name: Create release
         id: create-release
+        # TODO: Replace as it has been deprecated
```
Bullrich, line 153: It seems that this should be replaced by `gh cli`.

Example solution:
```yml
      - name: Create release
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          tag: ${{ github.ref_name }}
        run: |
          gh release create "$tag" \
              --repo="$GITHUB_REPOSITORY" \
              --title="${GITHUB_REPOSITORY#*/} ${tag#v}" \
              --generate-notes
```

.github/workflows/release.yml
```diff
@@ -183,6 +184,7 @@ jobs:
           >>$GITHUB_ENV echo SPEC=$(<${JSON} jq -r .runtimes.compact.subwasm.core_version.specVersion)
 
       - name: Upload compressed ${{ matrix.runtime.name }} v${{ env.SPEC }} wasm
+        # TODO: Replace as it has been deprecated
         uses: actions/upload-release-asset@v1
         env:
```
Bullrich, line 189: From the [action's readme](https://github.com/actions/upload-release-asset?tab=readme-ov-file#github-action---releases-api) the new action that seems to replace this one is [softprops/action-gh-release](https://github.com/softprops/action-gh-release)

.github/workflows/release.yml
Bullrich: GitHub also showed some [alternatives actions in the readme](https://github.com/actions/create-release?tab=readme-ov-file#github-action---releases-api)

bkchr (approved):

ggwpez (approved):

Issue #128, opened 2023-12-14
Script to automate weight generation


liamaharon:
Now that the weight generation has been documented (https://github.com/polkadot-fellows/runtimes/pull/127), a script should be written to automate the process.

bkontur:
Can somebody please close this? We have `cmd bench` bot

mordamax:
@bkontur it wasn't tested so far yet, there is a chance that it might be giving worse numbers, compared to a machine that has been used so far, so i'd keep it for now

PR #406, opened 2024-07-26
cmd actions bench & fmt

mordamax:
Closes 
- https://github.com/polkadot-fellows/runtimes/issues/128
- https://github.com/paritytech/infrastructure/issues/41

Adds ability to run some standard helper commands, for now only for benchmarks and fmt (with taplo)
It will commit the result of operation back into PR with some reports and/or links to logs

Temporary all commands are working on `ubuntu-22.04`, but later, when the self-hosted runner is ready, bench runner should be replaced with it. Thus it's not recommended to generate weights using ubuntu-*, as it won't use correct compute resources.

<details>
<summary>Examples (screenshots):</summary>

### Benchmarks
Google Chrome 2024-07-26 13 33 05

### Bench results with Subweight report
Google Chrome 2024-07-26 13 35 36

### /cmd [command] --clean - removing bot's and author's command runs to clean up PR
Google Chrome 2024-07-26 13 33 56

### /cmd [command] --help - outputs the usage instructions
Google Chrome 2024-07-26 13 33 25

</details>


.github/scripts/cmd/cmd.py
```diff
@@ -0,0 +1,188 @@
...
+        print(f'-- listing pallets for benchmark for {runtime["name"]}')
+        wasm_file = f"target/{profile}/wbuild/{runtime['package']}/{runtime['package'].replace('-', '_')}.wasm"
+        output = os.popen(
+            f"frame-omni-bencher v1 benchmark pallet --no-csv-header --all --list --runtime={wasm_file}").read()
+        raw_pallets = output.split('\n')
+
+        all_pallets = set()
+        for pallet in raw_pallets:
+            if pallet:
+                all_pallets.add(pallet.split(',')[0])
```
ggwpez, line 102:
```suggestion
                all_pallets.add(pallet.split(',')[0].trim())
```
Just in case, i remember there was one version that would add spaces.

ggwpez (approved):
Good work, thanks! I mostly skimmed it but i think this is one of the things where we just need to experiment a bit.

Only point is that there should be a `commands.md` or `bot.md` with some overview of how the bot works, what commands are available and how the components make it work.  
The weight-generation file seems to be specific already.

.github/workflows/cmd.yml
```diff
@@ -0,0 +1,323 @@
...
+
+permissions: # allow the action to comment on the PR
+  contents: write
+  issues: write
+  pull-requests: write
+  actions: read
+
+jobs:
+  acknowledge:
+    if: ${{ startsWith(github.event.comment.body, '/cmd') }}
```
xlc, line 15: anyone can trigger this?

xlc (commented):
best to be able to only allow fellowship members to trigger the bot to prevent abuses especially with self hosted runners

xlc:
I don't know what's the plan for the self hosted runners but for ORML we are using github webhook to create new AWS EC2 instance upon request and autoscale it down when idle. In that way we don't need to run idle instance and able to scale up as many runners as needed.

.github/workflows/cmd.yml
```diff
@@ -0,0 +1,323 @@
...
+          text: ${{ github.event.comment.body }}
+          regex: '^(\/cmd )(.*)$'
+
+      - name: Save output of help
+        id: help
+        env:
+          CMD: ${{ steps.get-pr-comment.outputs.group2 }} # to avoid "" around the command
+        run: |
+          echo 'help<<EOF' >> $GITHUB_OUTPUT
+          python3 .github/scripts/cmd/cmd.py $CMD >> $GITHUB_OUTPUT
```
pavelsupr, line 85: This line doesn't look safe, I'll write in Element.

.github/workflows/cmd.yml
mordamax: yep good point, will add a similar check as for /merge
and cmd.md guide
thanks

.github/workflows/cmd.yml
mordamax: Improved regex which prevents passing adding arbitrary commands + check for fellows

image

kianenigma:
While this is being prepared, what is the standard way to generate weight files? I suppose I need access to a 'standard machine'? 

Asking for #364

mordamax:
> While this is being prepared, what is the standard way to generate weight files? I suppose I need access to a 'standard machine'?
> 
> Asking for #364

if it waits until early next week, i'd recommend to wait a bit. @tugytur was going to spawn a runner as soon as @bkchr provides details, hope today or early next week 🤞 

If not, then you can follow this guidance https://github.com/polkadot-fellows/runtimes/blob/main/docs/weight-generation.md and find `standard machine` with [similar specs](https://wiki.polkadot.network/docs/maintain-guides-how-to-validate-polkadot#standard-hardware)

kianenigma:
Not super rushed from my side, but thanks for the context :)

.github/workflows/cmd.yml
```diff
@@ -0,0 +1,364 @@
...
+              repo: context.repo.repo,
+              content: '+1'
+            })
+
+  cmd:
+    needs: [clean, fellows]
+    env:
+      JOB_NAME: 'cmd'
+    if: ${{ startsWith(github.event.comment.body, '/cmd') && !contains(github.event.comment.body, '--help') && contains(needs.fellows.outputs.github-handles, github.event.sender.login) }}
+    runs-on: ${{ startsWith(github.event.comment.body, '/cmd bench') && 'gitrun-001' || 'ubuntu-22.04' }}
```
ggwpez, line 174: It is using `ubuntu-latest` in the other cases?

mordamax: yes, the other cases are now only FMT..

mordamax: yes, the other cases are not only FMT..

ggwpez (approved):

kianenigma (approved):

bkchr (approved):

xlc (approved):

PR #407, opened 2024-07-26
Allow signed origins to send arbitrary XCMs from some system chains

acatangiu:
Relaxing the `SendXcmOrigin` filter for some of the System chains:
- Relay Kusama: already open to signed origins and pluralities
- Relay Polkadot: relax to same level as Kusama ^

- Asset Hubs (P and K): open up to signed origins
- People (P and K): allow signed origins
- Encointer: allow signed origins

- Collectives Polkadot: no change, only Fellows collective allowed
- Bridge Hubs: no change, no arbitrary XCM sends allowed
- Coretime Kusama: no change, no arbitrary XCM sends allowed

We could allow it on all chains but on some of them there are no usecases afaict, so starting out conservative, opening it up only on chains where it's wanted.

### DIFF
--- a/relay/polkadot/src/xcm_config.rs
+++ b/relay/polkadot/src/xcm_config.rs
@@ -281,7 +281,7 @@ pub type TreasurerToPlurality = OriginToPluralityVoice<RuntimeOrigin, Treasurer,
 
 /// Type to convert a pallet `Origin` type value into a `Location` value which represents an
 /// interior location of this chain for a destination chain.
-pub type LocalPalletOriginToLocation = (
+pub type LocalPalletOrSignedOriginToLocation = (
 	// GeneralAdmin origin to be used in XCM as a corresponding Plurality `Location` value.
 	GeneralAdminToPlurality,
 	// StakingAdmin origin to be used in XCM as a corresponding Plurality `Location` value.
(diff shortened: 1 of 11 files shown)


acatangiu:
> Coretime Kusama: no change, no arbitrary XCM sends allowed

Maybe from Coretime chain there would be users that want to `Transact` to other chains from their signed accounts?

acatangiu:
@joepetrowski WDYT about current list?

acatangiu:
> > Coretime Kusama: no change, no arbitrary XCM sends allowed
> 
> Maybe from Coretime chain there would be users that want to `Transact` to other chains from their signed accounts?

cc @seadanda

joepetrowski:
The list looks reasonable to me. Will let @seadanda chime in.

seadanda:
List looks fine to me too, I don't see any immediate usecase for arbitrary Transacts.
It might be useful to check with Lastic and RegionX to see if they plan to make use of functionality like that.
@poppyseedDev @Szegoo

Szegoo:
No use cases from our side so far, but there could potentially be some, so I wouldn't be against it.

poppyseedDev:
Same opinion as @Szegoo here.

relay/polkadot/src/xcm_config.rs
```diff
@@ -293,13 +293,16 @@ pub type LocalPalletOriginToLocation = (
 	FellowshipAdminToPlurality,
 	// `Treasurer` origin to be used in XCM as a corresponding Plurality `Location` value.
 	TreasurerToPlurality,
+	// And a usual Signed origin to be used in XCM as a corresponding AccountId32
```
joepetrowski, line 296:
```suggestion
	// And a usual Signed origin to be used in XCM as a corresponding `AccountId32`.
```

system-parachains/coretime/coretime-kusama/src/xcm_config.rs
joepetrowski, line 1: Make the same change on `coretime-polkadot`?

joepetrowski (approved):

seadanda (approved):

acatangiu:
@kianenigma @shawntabrizi @xlc need another review here please

bkchr (approved):

PR #401, opened 2024-07-25
Enable Coretime on Polkadot Relay

tdimitrov:
Redo of https://github.com/polkadot-fellows/runtimes/pull/354 on top of https://github.com/polkadot-fellows/runtimes/pull/381

TODOs:
- [ ] Check migrations with chopsticks
- [ ] Regenerate weights

PR #410, opened 2024-07-30
Add the Polkadot Coretime runtime

seadanda:
Add the Polkadot Coretime chain in advance of the 1.3.0 release.

This uses the new Price adapter which has been running on Kusama now for two sales cycles, and includes the mechanism to burn revenue.

TODO:
- [x] Add Transact tests for hardcoded weights after https://github.com/polkadot-fellows/runtimes/pull/401 is merged
- [ ] Rerun benchmarks and check hardcoded weights after merge

The genesis chain-spec is developing on https://github.com/seadanda/runtimes/pull/4. This can be used as a merge target for any community boot nodes who would like to be included at genesis and will be separately merged to `main`.


system-parachains/coretime/coretime-polkadot/src/lib.rs
```diff
@@ -0,0 +1,1102 @@
...
+/// The version information used to identify this runtime when compiled natively.
+#[cfg(feature = "std")]
+pub fn native_version() -> NativeVersion {
+	NativeVersion { runtime_version: VERSION, can_author_with: Default::default() }
+}
+
+parameter_types! {
+	pub const Version: RuntimeVersion = VERSION;
+	pub RuntimeBlockLength: BlockLength =
+		BlockLength::max_with_normal_ratio(5 * 1024 * 1024, NORMAL_DISPATCH_RATIO);
```
eskimor, line 157: Is block length same as PoV size?

system-parachains/coretime/coretime-polkadot/src/lib.rs
```diff
@@ -0,0 +1,1102 @@
...
+	type Keys = SessionKeys;
+	type WeightInfo = weights::pallet_session::WeightInfo<Runtime>;
+}
+
+impl pallet_aura::Config for Runtime {
+	type AuthorityId = AuraId;
+	type DisabledValidators = ();
+	type MaxAuthorities = ConstU32<100_000>;
+	type AllowMultipleBlocksPerSlot = ConstBool<false>;
+	type SlotDuration = ConstU64<SLOT_DURATION>;
```
eskimor, line 405: Hmm. SLOT_DURATION is still 12s.

eskimor: Aren't we launching with async backing directly?

eskimor (approved):
Looks good. Don't we launch with async backing?

eskimor:
Also don't we need to apply [these changes](https://github.com/paritytech/polkadot-sdk/pull/3768/files) to allow unpaid execution from the coretime chain?

system-parachains/coretime/coretime-polkadot/src/lib.rs
seadanda: We can't have async backing enabled for Coretime until we time the sales from the relay chain block, work being done in: https://github.com/paritytech/polkadot-sdk/pull/3331 which seems to have been abandoned by the author. We'll need to test this on Kusama first too

system-parachains/coretime/coretime-polkadot/src/lib.rs
seadanda: The block length here is the limit set for the total length of each extrinsic class (normal/operational/mandatory). Here we explicitly set it to the same values as `BlockLength::default()` - the first argument sets the max for operational and mandatory, the normal is a `Perbill` which is multiplied by the max, in this case Normal extrinsics can only take up 75% of the block.
I think of it as the PoV limit and respective ratio, but that's maybe an oversimplification.

eskimor: I am asking, because if this is really only the block (without state proof), then this is actually too large. Total PoV size is limited to 5Meg (compressed, but still).

system-parachains/coretime/coretime-polkadot/src/lib.rs
eskimor: I see. While I don't think we need timing on the relay chain, as we don't intend to change the block times afterwards, it still makes sense: If we did it right away, we could avoid the migration, but on the other hand we need the migration for Kusama already and copying that one over should be straight forward.

system-parachains/coretime/coretime-polkadot/src/lib.rs
seadanda: I think it is fine and I think this is that 5 Meg limit, this is what we have set for all system chains, and is also the default value.
But maybe @skunert could help here to confirm that it is correctly set

skunert: Yes that should be the correct setting. It does not matter so much however, since the `BlockLength` is mostly legacy. It will never reach the 5MB limit because `BlockWeight` tracks proof size and length combined. So it will hit the limit long before `BlockLength`.

system-parachains/coretime/coretime-polkadot/src/lib.rs
```diff
@@ -0,0 +1,1114 @@
...
+	/// Proxy able to purchase on-demand coretime credits.
+	OnDemandPurchaser,
+	/// Collator selection proxy. Can execute calls related to collator selection mechanism.
+	Collator,
+}
+impl Default for ProxyType {
+	fn default() -> Self {
+		Self::Any
+	}
+}
```
bkchr, line 488:
```suggestion
```

system-parachains/coretime/coretime-polkadot/src/lib.rs
```diff
@@ -0,0 +1,1114 @@
...
+	PartialOrd,
+	Encode,
+	Decode,
+	RuntimeDebug,
+	MaxEncodedLen,
+	scale_info::TypeInfo,
+)]
+pub enum ProxyType {
+	/// Fully permissioned proxy. Can execute any call on behalf of _proxied_.
+	Any,
```
bkchr, line 470:
```suggestion
	scale_info::TypeInfo,
	Default,
)]
pub enum ProxyType {
	/// Fully permissioned proxy. Can execute any call on behalf of _proxied_.
	#[default]j
	Any,
```

system-parachains/coretime/coretime-polkadot/src/lib.rs
```diff
@@ -0,0 +1,1114 @@
...
+	type AuthorityId = AuraId;
+	type DisabledValidators = ();
+	type MaxAuthorities = ConstU32<100_000>;
+	type AllowMultipleBlocksPerSlot = ConstBool<false>;
+	type SlotDuration = ConstU64<SLOT_DURATION>;
+}
+
+parameter_types! {
+	pub const PotId: PalletId = PalletId(*b"PotStake");
+	pub const SessionLength: BlockNumber = 6 * HOURS;
```
bkchr, line 410: Not used?
```suggestion
```

system-parachains/coretime/coretime-polkadot/src/genesis_config_presets.rs
```diff
@@ -0,0 +1,87 @@
...
+use system_parachains_constants::genesis_presets::*;
+
+const CORETIME_POLKADOT_ED: Balance = ExistentialDeposit::get();
+
+fn coretime_polkadot_genesis(
+	invulnerables: Vec<(AccountId, AuraId)>,
+	endowed_accounts: Vec<AccountId>,
+	id: ParaId,
+) -> serde_json::Value {
+	serde_json::json!({
```
bkchr, line 30: Not sure why we not use the `GenesisConfig` type from the runtime here.

system-parachains/coretime/coretime-polkadot/src/coretime.rs
```diff
@@ -0,0 +1,307 @@
...
+
+	fn request_core_count(count: CoreIndex) {
+		use crate::coretime::CoretimeProviderCalls::RequestCoreCount;
+		let request_core_count_call = RelayRuntimePallets::Coretime(RequestCoreCount(count));
+
+		// Weight for `request_core_count` from Polkadot runtime benchmarks:
+		// `ref_time` = 7889000 + (3 * 25000000) + (1 * 100000000) = 182889000
+		// `proof_size` = 1636
+		// Add 5% to each component and round to 2 significant figures.
+		// TODO check when benchmarks are rerun
```
bkchr, line 143: Who is checking these TODOs?

system-parachains/coretime/coretime-polkadot/src/lib.rs
```diff
@@ -0,0 +1,1114 @@
...
+	frame_system::CheckNonZeroSender<Runtime>,
+	frame_system::CheckSpecVersion<Runtime>,
+	frame_system::CheckTxVersion<Runtime>,
+	frame_system::CheckGenesis<Runtime>,
+	frame_system::CheckEra<Runtime>,
+	frame_system::CheckNonce<Runtime>,
+	frame_system::CheckWeight<Runtime>,
+	pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
+	frame_metadata_hash_extension::CheckMetadataHash<Runtime>,
+	cumulus_primitives_storage_weight_reclaim::StorageWeightReclaim<Runtime>,
```
bkchr, line 110:
```suggestion
```

Should not be added yet.

system-parachains/coretime/coretime-polkadot/src/coretime.rs
seadanda: I will check them, but if the hardcoded weights are not sufficient for the new weights they will make the emulated tests fail anyway.

system-parachains/coretime/coretime-polkadot/src/lib.rs
```diff
@@ -0,0 +1,1114 @@
...
+
+		fn get_preset(id: &Option<sp_genesis_builder::PresetId>) -> Option<Vec<u8>> {
+			get_preset::<RuntimeGenesisConfig>(id, &genesis_config_presets::get_preset)
+		}
+
+		fn preset_names() -> Vec<sp_genesis_builder::PresetId> {
+			vec![
+				sp_genesis_builder::PresetId::from("local_testnet"),
+				sp_genesis_builder::PresetId::from("development"),
+			]
```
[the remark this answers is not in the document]
seadanda: Ah I copied the way it was done in the other runtimes, I'll clean this up and make a PR for the others

system-parachains/coretime/coretime-polkadot/src/genesis_config_presets.rs
seadanda: Not sure what you mean here

system-parachains/coretime/coretime-polkadot/src/lib.rs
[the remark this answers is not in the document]
seadanda: Fixed in https://github.com/polkadot-fellows/runtimes/pull/410/commits/c0cd2bbe90c837937908b918f8f2a3440e377569

system-parachains/coretime/coretime-polkadot/src/genesis_config_presets.rs
bkchr: Instead of writing here a `json`, you import `super::GenesisConfig`.

system-parachains/coretime/coretime-polkadot/src/lib.rs
```diff
@@ -0,0 +1,1108 @@
...
+	pallet_transaction_payment::ChargeTransactionPayment<Runtime>,
+	frame_metadata_hash_extension::CheckMetadataHash<Runtime>,
+);
+
+/// Unchecked extrinsic type as expected by this runtime.
+pub type UncheckedExtrinsic =
+	generic::UncheckedExtrinsic<Address, RuntimeCall, Signature, SignedExtra>;
+
+/// Migrations to apply on runtime upgrade.
+pub type Migrations = (pallet_xcm::migration::MigrateToLatestXcmVersion<Runtime>,);
```
bkontur, line 117: just a formatting nit and added `// permanent` to be aligned with other runtimes

```suggestion
pub type Migrations = (
	// permanent
	pallet_xcm::migration::MigrateToLatestXcmVersion<Runtime>,
);
```

system-parachains/coretime/coretime-polkadot/src/genesis_config_presets.rs
seadanda: Ah found it: `RuntimeGenesisConfig`, had to cargo expand to find it. Thanks for the tip! Maybe it's [this comment](https://github.com/seadanda/runtimes/blob/donal-coretime-polkadot/system-parachains/coretime/coretime-polkadot/src/genesis_config_presets.rs#L63) in the `genesis_config_presets.rs` for all runtimes that says it will panic if we pass anything as the `aura` key, and using the struct means we'll have to pass something for all keys. Maybe @michalkucharczyk has more context

seadanda: Naively added on a branch https://github.com/seadanda/runtimes/commit/b552377d66d7f0a5d466cc843d0ee6edba880016, but would like to hear about this aura panic.

system-parachains/coretime/coretime-polkadot/src/lib.rs
joepetrowski: Why? It needs this
https://github.com/paritytech/polkadot-sdk/blob/839ead3/substrate/frame/proxy/src/lib.rs#L128

system-parachains/coretime/coretime-polkadot/src/genesis_config_presets.rs
michalkucharczyk: If you use rust struct here, then the function will return the _json blob_ containing all the fields of the `RuntimeGenesisConfig`, some of them set to default.

Having json here will limit the number of keys in the preset (patch) to only those which actually are changed.

I used _json_ everywhere because I wanted preset to be as compact as possible.

On the other hand using rust struct should guarantee that json is correct. In case of using json this verification can only be done during execution (when json is deserialized). But this shall be covered in tests.

seadanda: Conversation can continue on https://github.com/polkadot-fellows/runtimes/issues/431

PR #423, opened 2024-08-08
Remove pov-reclaim from kusama system-chains

skunert:
I propose to move PoV-reclaim to the next release.

It can happen that the amount to reclaim is overestimated https://github.com/paritytech/polkadot-sdk/issues/5229. 

Improvements (https://github.com/paritytech/polkadot-sdk/pull/5281, https://github.com/paritytech/polkadot-sdk/pull/5273) and a proper fix (https://github.com/paritytech/polkadot-sdk/pull/5234) are in the works, but should be tested again.

I expect the chances that the system-chains run into this scenario to be low, but it does not look like activity is so high that the inclusion of reclaim is urgent.


s0me0ne-unkn0wn (approved):
We're currently nowhere near the situation where we could benefit from having the PoV reclaim on system chains, so it makes sense to remove it for now just to be on the safe side.

bkchr (approved):

sandreim (approved):

acatangiu (approved):

ggwpez (approved):

PR #426, opened 2024-08-12
Proper handling for legacy parachain leases with gaps in coretime migration

tdimitrov:
While testing https://github.com/polkadot-fellows/runtimes/pull/401 I noticed that there are two parachains with gaps in their leases which are not migrated to coretime correctly. These are para ids 3359 and 3388.

The problem is that the migration obtains the list of parachain ids which should be migrated from `paras::Parachains` ([here](https://github.com/paritytech/polkadot-sdk/blob/1f49358db0033e57a790eac6daccc45beba81863/polkadot/runtime/parachains/src/coretime/migration.rs#L114)) and there are no entries for para ids which are not active at the moment.

Paras 3359 and 3388 are not active in the current lease period (excerpt from `slots::leases()`:
```
[
      3,359
    ]
    [
      null
      [
        14xQXJdUDC1pzyt8y3z27ANiUBgP7zTSaYutaLELJoyQrdLP
        2,010,000,000,000
      ]
      [
        14xQXJdUDC1pzyt8y3z27ANiUBgP7zTSaYutaLELJoyQrdLP
        2,010,000,000,000
      ]
      <skipped>
```
and:
```
[
      3,388
    ]
    [
      null
      [
        1TpMimWf8NvrusAQ36hVdgMpYhhtojg5Nw41CuVxG2zHPDg
        31,000,000,000,000
      ]
      [
        1TpMimWf8NvrusAQ36hVdgMpYhhtojg5Nw41CuVxG2zHPDg
        31,000,000,000,000
      ]
      <skipped>
```
And they have got no entries in `paras::parachains()`:

```
[ 1,000 1,001 1,002 1,004 2,000 2,002 2,004 2,006 2,008 2,012 2,013 2,025 2,026 2,030 2,031 2,032 2,034 2,035 2,037 2,040 2,043 2,046 2,048 2,051 2,053 2,056 2,086 2,090 2,091 2,092 2,093 2,094 2,101 2,104 2,106 3,333 3,338 3,340 3,344 3,345 3,346 3,354 3,358 3,366 3,367 3,369 3,370 3,373 3,375 3,377 ]
```

As a result the migration skips them which is wrong.

A proper fix for this issue will require a new polkadot-sdk release and a bump in the runtimes repo which requires time and effort. To avoid this the PR moves the coretime migration from polkadot-sdk to the fellowship repo.


tdimitrov:
This is a draft because I am not happy with `already_migrated` function in the PR.

tdimitrov:
As discussed offline with @eskimor we'll move the whole coretime migration to the fellowship repo so that a proper fix can be made.
The PR is tested with this semi-automated test: https://github.com/tdimitrov/polkadot-fellows-runtimes/
I'll extend the repo with some more asserts but meanwhile the PR is ready for review.

relay/polkadot/src/coretime_migration.rs
```diff
@@ -0,0 +1,319 @@
...
+		mk_coretime_call::<T>(CoretimeCalls::Reserve(schedule))
+	});
+
+	let message_content = iter::once(Instruction::UnpaidExecution {
+		weight_limit: WeightLimit::Unlimited,
+		check_origin: None,
+	});
+
+	let reservation_content = message_content.clone().chain(reservations).collect();
+	let pool_content = message_content.clone().chain(pool).collect();
```
eskimor, line 290: Pool will be empty. Do we send an empty message here?

eskimor (approved):

relay/polkadot/src/coretime_migration.rs
tdimitrov: Hmm I missed that. I guess we send an empty message. I'll fix it.

relay/polkadot/src/coretime_migration.rs
```diff
@@ -0,0 +1,319 @@
...
+			};
+			// We assume the coretime chain set this parameter to the recommended value in RFC-1:
+			const TIME_SLICE_PERIOD: u32 = 80;
+			let round_up = if valid_until % TIME_SLICE_PERIOD > 0 { 1 } else { 0 };
+			let time_slice = valid_until / TIME_SLICE_PERIOD + TIME_SLICE_PERIOD * round_up;
+			log::trace!(target: "coretime-migration", "Sending of lease holding para {:?}, valid_until: {:?}, time_slice: {:?}", p, valid_until, time_slice);
+			Some(mk_coretime_call::<T>(CoretimeCalls::SetLease(p.into(), time_slice)))
+		});
+
+	let core_count: u16 = configuration::ActiveConfig::<T>::get()
```
eskimor, line 266: Hmm. on another call in a minute, but please check that this core handling is actually correct. This core_count value should read 0 on Polkadot at the time of the migration.

tdimitrov: It's not 0. We update it on line 211-214:
https://github.com/polkadot-fellows/runtimes/pull/426/files#diff-6a311f9b6c4cd2a09f397e5e0c90ea9193ba103bf277bb88de6bd223148af52eR211-R214

The actual value is 52:
```
coretime-migration  TRACE: Set core count to 52. legacy paras count is  52
```

The reserved cores are 0 though because `legacy_paras_count == core_count`. I have got an assert for this in the test:
https://github.com/tdimitrov/polkadot-coretime-launch-test/blob/a8e7aba7d6a741b10bd5330ae3119ef15a018711/coretime-migration-check/main.js#L158-L160

Is this correct?

relay/polkadot/src/coretime_migration.rs
```diff
@@ -0,0 +1,319 @@
...
+				log::error!("Lease holding chain with no lease information?!");
+				return None
+			};
+			let valid_until: u32 = match valid_until.try_into() {
+				Ok(val) => val,
+				Err(_) => {
+					log::error!("Converting block number to u32 failed!");
+					return None
+				},
+			};
```
bkchr, line 257:
```suggestion
```

block number is u32

relay/polkadot/src/coretime_migration.rs
```diff
@@ -0,0 +1,319 @@
...
+			// We assume the coretime chain set this parameter to the recommended value in RFC-1:
+			const TIME_SLICE_PERIOD: u32 = 80;
+			let round_up = if valid_until % TIME_SLICE_PERIOD > 0 { 1 } else { 0 };
+			let time_slice = valid_until / TIME_SLICE_PERIOD + TIME_SLICE_PERIOD * round_up;
```
bkchr, line 261:
```suggestion
			let time_slice = (valid_until + TIME_SLICE_PERID - 1) / TIME_SLICE_PERIOD;
```

relay/polkadot/src/coretime_migration.rs
```diff
@@ -0,0 +1,319 @@
...
+	let reservation_content = message_content.clone().chain(reservations).collect();
+	let pool_content = message_content.clone().chain(pool).collect();
+	let leases_content_1 = message_content
+		.clone()
+		.chain(leases.by_ref().take(legacy_paras_count / 2)) // split in two messages to avoid overweighted XCM
+		.collect();
+	let leases_content_2 = message_content.clone().chain(leases).collect();
+	let set_core_count_content = message_content.clone().chain(set_core_count).collect();
+
+	let messages = vec![
```
bkchr, line 298:
```suggestion
	let messages = [
```

bkchr (approved):

relay/polkadot/src/coretime_migration.rs
bkchr: > The reserved cores are 0 though

What you mean by reserved cores? The system chains? 

Generally your code is correct.

tdimitrov: Yes, the system chains.

relay/polkadot/src/lib.rs
```diff
@@ -1947,6 +1949,15 @@ pub mod migrations {
 				<slots::Pallet<Runtime> as Leaser<BlockNumber>>::lease_period_index(now)?;
 			Some(index.saturating_add(lease.len() as u32).saturating_mul(LeasePeriod::get()))
 		}
+
+		fn get_all_parachains_with_leases() -> Vec<ParaId> {
+			let mut leases = slots::Leases::<Runtime>::iter()
+				.filter(|(_, lease)| !lease.is_empty())
+				.map(|(para, _)| para)
+				.collect::<Vec<_>>();
+			leases.sort();
```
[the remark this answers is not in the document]
tdimitrov: No need to be sorted. This is a debugging leftover. I'll remove it.

relay/polkadot/src/coretime_migration.rs
tdimitrov: No, wait. These are reserved for on demand, if I am not mistaken.
We reserve cores for system chains, then for lease cores and what's left is for on demand. Right @eskimor ?

relay/polkadot/src/coretime_migration.rs
tdimitrov: I think we do a different kind of round up here. `timeslice` is not number of periods but a lifetime in blocks (divided by 80). Your formula works only for the 'number of periods' case.

tdimitrov: I'm confused... why we multiply the round up with `TIME_SLICE_PERIOD`?

Shouldn't it be `let time_slice = valid_until / TIME_SLICE_PERIOD + round_up;`) which is equivalent to your formula?

relay/polkadot/src/coretime_migration.rs
tdimitrov: It doesn't compile due to expected `u32` found associated type errors. Let's keep it as it is.

relay/polkadot/src/coretime_migration.rs
bkchr: `round_up` is just `0` or `1`. It leads to the same result as what I have written there.

tdimitrov: It doesn't. Round up is multiplied by TIME_SLICE_PERIOD. Have a look: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=a5ce70639f2ee8a169e5d618e707ed74

relay/polkadot/src/coretime_migration.rs
```diff
@@ -0,0 +1,318 @@
...
+				"Creating assignment for existing para failed: {:?}, error: {:?}",
+				para_id,
+				err
+			);
+		}
+	}
+
+	let config = configuration::ActiveConfig::<T>::get();
+	// num_cores was on_demand_cores until now:
+	for on_demand in 0..config.scheduler_params.num_cores {
```
seadanda, line 196: On Polkadot this is not the path of this config entry, checking the chain state it's at `config.coretime_cores`. Is there a migration ahead of this one to change the configuration in storage?

relay/polkadot/src/coretime_migration.rs
tdimitrov: Fixed

relay/polkadot/src/coretime_migration.rs
tdimitrov: `parachains_configuration::migration::v12::MigrateToV12<Runtime>` does this change:
https://github.com/paritytech/polkadot-sdk/blob/ebf4f8d2d590f41817d5d38b2d9b5812a46f2342/polkadot/runtime/parachains/src/configuration/migration/v12.rs#L146-L158

relay/polkadot/src/coretime_migration.rs
```diff
@@ -0,0 +1,318 @@
...
+			let valid_until: u32 = match valid_until.try_into() {
+				Ok(val) => val,
+				Err(_) => {
+					log::error!("Converting block number to u32 failed!");
+					return None
+				},
+			};
+
+			// We assume the coretime chain set this parameter to the recommended value in RFC-1:
+			const TIME_SLICE_PERIOD: u32 = 80;
```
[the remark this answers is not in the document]
tdimitrov: Fixed

relay/polkadot/src/coretime_migration.rs
```diff
@@ -0,0 +1,318 @@
...
+	let message_content = iter::once(Instruction::UnpaidExecution {
+		weight_limit: WeightLimit::Unlimited,
+		check_origin: None,
+	});
+
+	let reservation_content = message_content.clone().chain(reservations).collect();
+	let pool_content = message_content.clone().chain(pool).collect();
+	let leases_content_1 = message_content
+		.clone()
+		.chain(leases.by_ref().take(legacy_paras_count / 2)) // split in two messages to avoid overweighted XCM
```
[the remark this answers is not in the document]
tdimitrov: Yes, but `leases` is an iterator and it's a bit involved to get its length. Let's keep it that way. It's not a big deal that the messages aren't equally distributed as long as they are sent in two passes (it's just one `ParaId` which causes the overweight).

relay/polkadot/src/coretime_migration.rs
tdimitrov: Fixed

relay/polkadot/src/coretime_migration.rs
```diff
@@ -0,0 +1,327 @@
...
+
+			let valid_until: u32 = match valid_until.try_into() {
+				Ok(val) => val,
+				Err(_) => {
+					log::error!("Converting block number to u32 failed!");
+					return None
+				},
+			};
+
+			// We assume the coretime chain set this parameter to the recommended value in RFC-1:
```
seadanda, line 258:
```suggestion
```

seadanda (approved):
Looks good

relay/polkadot/src/coretime_migration.rs
tdimitrov: :facepalm:

eskimor (approved):

relay/polkadot/src/coretime_migration.rs
tdimitrov: I've removed the comment.

seadanda (approved):
Re-approving to trigger a CI run with the broken runner disabled

tdimitrov:
There was a `try-runtime` check verifying if the relay chain is sending XMP messages to the broker chain. It was using `dmq_contents` which is not accessible from the runtime crate. Since I have tested the migration with chopsticks I think it's safe to drop it.

Please see this commit for details: https://github.com/polkadot-fellows/runtimes/pull/426/commits/467518345fd4f7f20a59acc0bdbcd38d1a84ba4a

seadanda:
LGTM, zombienet is failing in CI but seems like it's unrelated

PR #430, opened 2024-08-13
Add "Spokesperson" proxy type to Kusama

olanod:
A "Spokesperson" is a proxy type only allowed to make remarks in the name of the proxied account. Useful for DAOs with an account in the relay chain that need to make a statement. This will also allow the representative to act on behalf of the DAO in websites like subsquare/polkassembly to edit proposals for example.


bkchr (approved):

ggwpez (approved):

bkchr:
@olanod please merge master and then we can merge this pr

PR #434, opened 2024-08-14
Coretime Chain: mitigate behaviour with many assignments on one core

seadanda:
We can handle a maximum of 28 assignments inside one XCM, while it's possible to have 80 (if a region is interlaced 79 times).
This can be chunked on the coretime chain side but currently the scheduler on the relay makes assumptions that means we can't send more than one chunk for a given core.

I've made a mitigation here that requires no sdk release, but it just truncates the additional assignments until we can extend the relay to support this. This means that the first 27 assignments are taken, the final 28th is used to pad with idle to complete the mask (the relay also assumes that every schedule is complete). Any other assignments are dropped.


eskimor (approved):

eskimor (approved):

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -223,37 +223,75 @@ impl CoretimeInterface for CoretimeAllocator {
...
+			// Set the parts of the `Idle` assignment we injected at the start of the vec above.
+			assignment_truncated[0].1 = 57_600u16.saturating_sub(total_parts);
+			assignment_truncated
+		} else {
+			assignment
+		};
+
+		// The maximum number of assignments for a core in a given timeslice is 80, but only 28
+		// assignments fit inside the max size of an XCM, so we need to send in chunks.
+		for assignment_chunk in assignment.chunks(28) {
```
bkchr, line 264:
```suggestion
```

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -223,37 +223,75 @@ impl CoretimeInterface for CoretimeAllocator {
...
+			let assign_core_call = RelayRuntimePallets::Coretime(AssignCore(
+				core,
+				begin,
+				assignment_chunk.to_vec(),
```
bkchr, line 268:
```suggestion
				assignment,
```

system-parachains/coretime/coretime-kusama/src/coretime.rs
```diff
@@ -223,37 +223,75 @@ impl CoretimeInterface for CoretimeAllocator {
...
+					target: "runtime::coretime",
+					"Core assignment sent successfully."
+				),
+				Err(e) => log::error!(
+					target: "runtime::coretime",
+					"Core assignment failed to send: {:?}",
+					e
+				),
+			}
 		}
```
bkchr, line 295:
```suggestion
```

system-parachains/coretime/coretime-polkadot/src/coretime.rs
```diff
@@ -234,28 +232,68 @@ impl CoretimeInterface for CoretimeAllocator {
...
+			// Set the parts of the `Idle` assignment we injected at the start of the vec above.
+			assignment_truncated[0].1 = 57_600u16.saturating_sub(total_parts);
+			assignment_truncated
+		} else {
+			assignment
+		};
+
+		// The maximum number of assignments for a core in a given timeslice is 80, but only 28
+		// assignments fit inside the max size of an XCM, so we need to send in chunks.
+		for assignment_chunk in assignment.chunks(28) {
```
bkchr, line 266:
```suggestion
```

system-parachains/coretime/coretime-polkadot/src/coretime.rs
```diff
@@ -234,28 +232,68 @@ impl CoretimeInterface for CoretimeAllocator {
...
+			let assign_core_call = RelayRuntimePallets::Coretime(AssignCore(
+				core,
+				begin,
+				assignment_chunk.to_vec(),
```
bkchr, line 270:
```suggestion
				assignment,
```

system-parachains/coretime/coretime-polkadot/src/coretime.rs
```diff
@@ -234,28 +232,68 @@ impl CoretimeInterface for CoretimeAllocator {
...
+					target: "runtime::coretime",
+					"Core assignment sent successfully."
+				),
+				Err(e) => log::error!(
+					target: "runtime::coretime",
+					"Core assignment failed to send: {:?}",
+					e
+				),
+			}
 		}
```
bkchr, line 297:
```suggestion
```

bkchr (approved):
When the relay chain accepts multiple assignments, we should make the XCM call generation more flexible and not just assume that `28` the magic number will always work. Thus, I see no need to keep the chunking right now.

PR #448, opened 2024-08-30
Fixes for cmd bot

mordamax:

the issue_comment works from the refs/main branch by default, and doesn't bring the PR context anyhow. Since i was testing it from the default branch in paritytech-stg - it was working fine, but from real PR non-default branch it didn't push result to PR branch.


.github/workflows/cmd.yml
```diff
@@ -165,11 +171,52 @@ jobs:
...
+          echo "The PR branch is ${{ steps.get-pr.outputs.pr_branch }}"
+          echo "The repository is ${{ steps.get-pr.outputs.repo }}"
+
 
   cmd:
-    needs: [clean, fellows]
+    needs: [get-pr-branch, fellows]
     env:
       JOB_NAME: 'cmd'
+      PR_BRANCH: ${{ needs.get-pr-branch.outputs.pr-branch }}
```
bkchr, line 219: Where is this being used?

bkchr (approved):

.github/workflows/cmd.yml
mordamax: nowhere, though to use it but then didn't )) will remove, thanks

mordamax, line 219:
```suggestion
```

Issue #395, opened 2024-07-23
Remove `chain-spec-generator`


bkontur:
Where possible, we should use `chain-spec-builder` from Substrate, which can be downloaded from the `polkadot-sdk` [release page assets](https://github.com/paritytech/polkadot-sdk/releases/tag/polkadot-v1.14.0) or from [crates.io](https://crates.io/crates/staging-chain-spec-builder).

Relates to: https://github.com/paritytech/polkadot-sdk/pull/4518

## TODO
- [ ] fix bridges zombienet tests - https://github.com/polkadot-fellows/runtimes/blob/main/integration-tests/bridges/environments/polkadot-kusama/bridge_hub_kusama_local_network.toml#L8
- [ ] zombienet-sdk-tests - https://github.com/polkadot-fellows/runtimes/blob/main/integration-tests/zombienet/src/lib.rs#L10
- [ ] anything else?

kianenigma:
Seems like a "clear" task that can be mentored. Should we have a mentor tag here as well?

bkontur:
> Seems like a "clear" task that can be mentored. Should we have a mentor tag here as well?

yes, sure, but I cannot update tags/labels here. Do we even use tags/labels for the fellows repo?

kianenigma:
@anaelleltd what would you suggest? would be good to give label altering rights to existing fellowship members. 

Perhaps you can create github teams from different rank. Based on this, then you can give more proper merge/label and such permissions to members.

anaelleltd:
> @anaelleltd what would you suggest? would be good to give label altering rights to existing fellowship members.

Yes, Bryan also mentioned something along these lines for the RFCs repo; so as to update the status of each RFC.

> Perhaps you can create github teams from different rank. Based on this, then you can give more proper merge/label and such permissions to members.

I only have "Member" status in the Fellowship repo, so I will have to speak to Raul or Rob (the current repo owners) about this new setup. I will also ask them to give me team maintainer right for future admin duties.

From what I read, teams would be useful for ping-ing members of higher ranks to do review, at the very least. 
When it comes to giving members **[triage/write](https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization#permissions-for-each-role)** rights on the Fellowship repo, this will probably have to be done on a case-by-case basis. I wouldn't recommend adding all Fellowship members to teams at once. 

Would this work from your end or do you have any other requirements in mind?

[...]

PR #450, opened 2024-08-30
Clean up genesis config presets for all runtimes

seadanda:
This PR does three things related to chain-specs and genesis config presets:
- Refactor `preset_names` to `genesis_config_presets` in all runtimes as suggested for the Polkadot Coretime runtime in #410.
- Make the chain-spec generator use the preset from the WASM directly for each runtime.
- Propagate the `on_chain_release_build` feature to the chain-spec generator. Without this we cannot generate live/genesis chain-specs with the chain-spec generator without having to post-process it to replace the WASM compiled with this flag.


muharem (approved):

bkchr (approved):

kianenigma:
Why does the fellowship repo have this `chain-spec generator` and not use OG chain-spec-builder? 

Sorry of this is duplicate, I think I have heard an explanation of this recently, but I can't recall it.

bkchr:
> Why does the fellowship repo have this `chain-spec generator` and not use OG chain-spec-builder?

Historical reasons. We don't really need it anymore. Someone just needs to remove it.

bkontur:
> Why does the fellowship repo have this `chain-spec generator` and not use OG chain-spec-builder?
> 
> Sorry of this is duplicate, I think I have heard an explanation of this recently, but I can't recall it.

> Historical reasons. We don't really need it anymore. Someone just needs to remove it.


@bkchr @kianenigma https://github.com/polkadot-fellows/runtimes/issues/395

PR #468, opened 2024-10-03
change max_deciding of the FellowshipAdmin track to 30

Tomen:
This PR changes the max number of refs that can be in the `deciding` state in the `FellowshipAdmin` track from 10 to 30.

This is a defensive measure to allow a potential submission of 21 rank retention refs for Head Ambassadors to go into simultaneous decision.

## Background and Rationale
The OpenGov community is currently discussing updates to the Ambassador Program social contract. One potential path is to begin to make use of the `demotionPeriod` parameter in the `ambassadorCore` to introduce a demotion period. This would then force the existing 20 collective members of rank 3 (the highest rank) to submit retention referenda to the OpenGon `FellowshipAdmin` track. If the track would fill up and not allow sufficient decisions within the demotionPeriod, it could lead to unwanted demotions that are not in the intention of the DAO.

To avoid this situation, this PR raises the limit to 30. This gives room for 21 retention referenda + additional room for any other potential collectives chain related refs.

### DIFF
--- a/relay/polkadot/src/governance/tracks.rs
+++ b/relay/polkadot/src/governance/tracks.rs
@@ -154,7 +154,7 @@ const TRACKS_DATA: [(u16, pallet_referenda::TrackInfo<Balance, BlockNumber>); 16
 		13,
 		pallet_referenda::TrackInfo {
 			name: "fellowship_admin",
-			max_deciding: 10,
+			max_deciding: 30,
 			decision_deposit: 5 * GRAND,
 			prepare_period: 2 * HOURS,
 			decision_period: 28 * DAYS,


ggwpez (commented):
It could also make sense to add a specific track for the ambassadors, like `AmbassadorAdmin` or so if it is to be used so regularly.

wirednkod:
I think that @Tomen needs this for the ambassadors collective. I would suggest to create that only there. 
I am not sure why we need this in the fellowship collective.

Tomen:
I agree that it would be better to have a dedicated track. But right now, as it is written in the code, the Fellowship admin track is not exclusive to the Fellowship, but also serves the Ambassador Collective. 

There is a potential risk of an operational bottleneck if we don't change this parameter now that could also affect the Fellowship if a Fellowship related ref would be blocked from entering confirmation because the track is full of Ambassador Collective refs.

I suggest to temporarily raise the limit with this PR until a proper track for the other collective could be established. Then this value could be set down to 10 again.

bkchr:
Looks like we don't need it anymore?

PR #471, opened 2024-10-07
Polkadot: Constant yearly emission

ggwpez:
Sets the inflation to ~120M DOT per year as specified in [Ref 1139](https://polkadot.subsquare.io/referenda/1139). 15% of that goes to the treasury and the rest to stakers.

## Details

This MR sets the Polkadot inflation to a fixed amount per year. The yearly increase is ~120,063,259 DOT. The amount was set to 8% of the Total Issuance at block [22810263](https://polkadot.statescan.io/#/blocks/22810263) in which referendum [1139](https://polkadot.subsquare.io/referenda/1139) was executed. You can check this for yourself:

Screenshot 2024-10-08 at 16 51 07

Multiplying this by 8% and converting to DOT is about **`120,093,259` DOT per year**.  
This results in an emissions of `120,093,259 / 365.25` about **`328,797` DOT per day**.

The Total Issuance and yearly inflation look like this over the next 25 years:
inflation

<details><summary>Concrete numbers</summary>
<p>

```pre
Initial 1.501 BDOT
Year 1: 1.621 BDOT +7.407%
Year 2: 1.741 BDOT +6.897%
Year 3: 1.861 BDOT +6.452%
Year 4: 1.982 BDOT +6.061%
Year 5: 2.102 BDOT +5.714%
Year 6: 2.222 BDOT +5.405%
Year 7: 2.342 BDOT +5.128%
Year 8: 2.462 BDOT +4.878%
Year 9: 2.582 BDOT +4.651%
Year 10: 2.702 BDOT +4.444%
Year 11: 2.822 BDOT +4.255%
Year 12: 2.942 BDOT +4.082%
Year 13: 3.062 BDOT +3.922%
Year 14: 3.182 BDOT +3.774%
Year 15: 3.303 BDOT +3.636%
Year 16: 3.423 BDOT +3.509%
Year 17: 3.543 BDOT +3.390%
Year 18: 3.663 BDOT +3.279%
Year 19: 3.783 BDOT +3.175%
Year 20: 3.903 BDOT +3.077%
Year 21: 4.023 BDOT +2.985%
Year 22: 4.143 BDOT +2.899%
Year 23: 4.263 BDOT +2.817%
Year 24: 4.383 BDOT +2.740%
Year 25: 4.503 BDOT +2.667%
```

</p>
</details> 

## Implications

15% of the inflation goes to the treasury which results in a yearly treasury inflow of ~18M DOT (1.5M DOT per month).  
[...]

PR #486, opened 2024-10-17
Release 1.3.4

ggwpez:
Release for the updated inflation formula from https://github.com/polkadot-fellows/runtimes/pull/471

This release only affects the Polkadot runtime.


ggwpez:
> So why bump and release all (unaffected) runtimes? We could bump and release only Polkadot Relay.

Yes true. We did partial bumps in the past, and it then becomes difficult to know what runtimes need to be bumped to what version.  
Bumping all to 1.3.4 in the 1.3.4 release keeps the rule simple - just bump all to the release version.

acatangiu:
Releasing same runtime again but with a different spec version is not cool IMO. It adds unnecessary operational overhead to do useless runtime upgrades.

I suggest we achieve what you describe by bumping all **relevant** runtimes to latest spec version.

In this case, only Polkadot Relay gets 1.3.4. E.g. later say we want to release AH runtime, it would get 1.3.5 (and so will any other runtime that gets changes between now and then).

> difficult to know what runtimes need to be bumped to what version on the next release

"what runtimes" -> check the changelog or git history
"what version" -> always use next "globally unique" version (1.3.4 now, next 1.3.5 or 1.4.0)

ggwpez:
> It adds unnecessary operational overhead to do useless runtime upgrades.

We dont upgrade them. They are just deleted from the release page. But it makes it simpler to ctrl+f update them all.

>"what version" -> always use next "globally unique" version (1.3.4 now, next 1.3.5 or 1.4.0)

Yea we can also do that. Will change now to get this merged.

ordian (approved):

seadanda (approved):

gpestana (approved):

muharem (approved):

kianenigma (approved):

seadanda (approved):
Do we need to re-run benchmarks for MQ bump? AHK and Encointer bumps LGTM

Cargo.lock
```diff
@@ -8559,9 +8559,9 @@ dependencies = [
 
 [[package]]
 name = "pallet-message-queue"
-version = "39.0.0"
+version = "39.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "20e65a37881d1998546254a5e50a1f768b3f82deabe774e750f4ea95aba8030c"
+checksum = "8bb779dbfd58b3a4e6a84245392fe9d8c31a746ff7c6db414e91c012663859c3"
```
acatangiu, line 8564: To pick up this fix for other runtimes we need to bump them too. Not sure if the fix is urgent, if the other system chains have any overweight stuck messages in the queue.
But if they don't, I don't personally think it's worth the runtime upgrade, and we can go without bumping them.

ggwpez: Good Q. I think it should be fine to only bump PAH. Last time we checked it was only Hydra affected (Astar was false positive).

CHANGELOG.md
```diff
@@ -4,7 +4,7 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## [Unreleased]
+## [1.3.4] 31.10.2024
```
seadanda, line 7:
```suggestion
## [1.3.4] 01.11.2024
```
WDYT about today? Or are there outstanding things to fix?

ggwpez: I was just a bit paranoid about the inflation change since its so significant. Checked it twice now again locally, should be good.

seadanda: yea fair enough!

PR #508, opened 2024-12-09
[encointer] fix missing hook for enactment of proposals

brenzi:
The first 4 Encointer democracy proposals have been up for vote and two proposals have been approved. However, we realized that enactment did not happen. This is due to a missing hook in the runtime. 

This PR should fix the hook and enact the already approved after the next runtime upgrade when the cycle phase changes to "assigning"


pandres95 (approved):

seadanda (approved):
This seems to be worth an e2e test, won't block though

joepetrowski (approved):

bkchr (approved):

PR #437, opened 2024-08-15
Update hardcoded coretime chain weights

seadanda:
- [ ] TODO double check relay hardcoded weights against pallet-broker benchmarks

PR #511, opened 2024-12-17
Add dynamic `BurnParameters` for Kusama Treasury

pandres95:
This Pull Request resolves the signal raised by the Kusama [Ref. #437](https://kusama.subsquare.io/referenda/437), setting both `Burn` and `BurnDestination` parameters in Kusama's `pallet-treasury` as a structure (both of them need to be set to prevent some unwanted behaviour) that can be parametrized via OpenGov.

Default (when set to `None`) configuration behaves like what's already out there: 0%, no account.


acatangiu (commented):
@pandres95 can you please add some tests?

pandres95:
@acatangiu tests are ready

relay/kusama/src/lib.rs
```diff
@@ -874,9 +867,11 @@ pub struct TreasuryBurnHandler;
...
-		if let Some(TreasuryBurnParameters { account, .. }) =
-			dynamic_params::treasury::BurnParameters::get()
-		{
+		let portion = dynamic_params::treasury::BurnPortion::get();
+		let account = dynamic_params::treasury::BurnDestination::get();
+
+		if !portion.is_zero() && account.0.is_some() {
+			let account = account.0.expect("given `account.0.is_some`; qed");
 			// Must resolve into existing but better to be safe.
 			Balances::resolve_creating(&account, amount);
```
acatangiu: you can refactor to avoid the `expect`:

```suggestion
			if let Some(account) = account.0.as_ref() {
				// Must resolve into existing but better to be safe.
				Balances::resolve_creating(account, amount);
			}
```

no need for else branch since you know it will always be `Some`

acatangiu (approved):

relay/kusama/src/lib.rs
```diff
@@ -678,6 +687,17 @@ pub mod dynamic_params {
...
+
+	/// Parameters used by `pallet-treasury` to handle the burn process.
+	#[dynamic_pallet_params]
+	#[codec(index = 1)]
+	pub mod treasury {
+		#[codec(index = 0)]
+		pub static BurnPortion: Permill = Permill::from_percent(0);
+
+		#[codec(index = 1)]
+		pub static BurnDestination: BurnDestinationAccount = Default::default();
```
xlc, line 699:
```suggestion
		pub static BurnDestination: Option<AccountId> = None;
```

Any reason why we need a new struct?

xlc (approved):

relay/kusama/src/lib.rs
pandres95: Yes. Mentioning `AccountId` explicitly, as well as `Option<AccountId>` causes `#[dynamic_pallet_params]` macro parsing to fail.

Still trying to figure out the reasons behind, so I'll raise an issue about it today. Meanwhile, the new struct is a good, codec-compatible workaround to that issue.

PR #519, opened 2024-12-23
Bridges patch applied for stable2409

bkontur:
Relates to: https://github.com/paritytech/polkadot-sdk/pull/6781
Relates to: https://github.com/paritytech/polkadot-sdk/pull/6536
Closes: https://github.com/paritytech/parity-bridges-common/issues/3017

Basically, this PR reverts back the congestion mechanism for Bridges, which was removed by https://github.com/polkadot-fellows/runtimes/pull/490 and later reverted by https://github.com/paritytech/polkadot-sdk/pull/6781.

- [X] Does not require a CHANGELOG entry

## TODO/Follow-ups
- [x] run bridges zombienet tests
- [x] fresh weights for relevant pallets

PR #523, opened 2025-01-02
Asset hubs - Add all assets in pool with native to XcmPaymentApi

franciscoaguirre:
The `XcmPaymentApi` has two functions, `query_acceptable_payment_assets` and `query_weight_to_asset_fee`, which allow clients to know which assets can be used to pay for fees.

On the asset hubs fees can be paid both with the native asset (DOT, KSM) and with any asset in a liquidity pool with the native asset. However, these functions were only returning the native asset.

This PR adds all other assets that can be used for fee payment to this API.

Also fixed an issue where these APIs wouldn't accept any version other than latest.


acatangiu (approved):
Please add entry to Changelog.

acatangiu (commented):
@franciscoaguirre please add [these tests](https://github.com/paritytech/polkadot-sdk/blob/cccefdd965c39498825f34e105979c447b315359/cumulus/parachains/runtimes/assets/asset-hub-westend/tests/tests.rs#L1500C1-L1514C2) for all runtimes and verify the new APIs work properly

franciscoaguirre:
@acatangiu 

Bumping `asset-test-utils` in order to have these tests also bumps `pallet-xcm-bridge-hub-router` to a patch version that doesn't follow semver because it added new associated types to the pallet's config.

It would be best if we get https://github.com/polkadot-fellows/runtimes/pull/519 merged before this.

I'll still add it to try to have as few conflicts as possible.

acatangiu (approved):

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
```diff
@@ -1230,19 +1232,39 @@ impl_runtime_apis! {
...
+			match latest_asset_id {
+				Ok(asset_id) if asset_id.0 == native_asset => {
+					// for native asset
+					Ok(fee_in_native)
 				},
 				Ok(asset_id) => {
-					log::trace!(target: "xcm::xcm_runtime_apis", "query_weight_to_asset_fee - unhandled asset_id: {asset_id:?}!");
-					Err(XcmPaymentApiError::AssetNotFound)
+					// Try to get current price of `asset_id` in `native_asset`.
+					if let Ok(Some(swapped_in_native)) = assets_common::PoolAdapter::<Runtime>::quote_price_tokens_for_exact_tokens(
```
eskimor, line 1257: This function has a pretty suboptimal signature. Way too easy to mix up arguments (not introduced here, just noticed that it makes it unnecessary hard to review correctness).

eskimor: Note: Not even function docs nor naming of parameters helps here in any way. What is asset1? What is asset2? I need to read the code of the trait implementation to figure this out.

eskimor: In what asset is the provided amount? Based on the implementation of the called get_amount_in, I am inferring that the amount is the amount we want to receive in the native token and that the first parameter is indeed the asset_in, but that seems to contradict the other code path. I am sure I am just holding it wrong, but the point is, I should not need to read a particular implementation of a trait function to check that the parameters have been passed correctly (because I see that they are too generic for the type checker to be able to catch that).

Note: This is by no means unique to this code, we have plenty of similar examples in the parachains/polkadot code base. I am just using this opportunity to raise awareness in a broader scope: If we improved here, our code quality and also our speed of reviews could be greatly enhanced as everything starts to become locally reasonable: I should not need to study implementation details of code that is not part of that PR to be able to infer that it is correct.

eskimor (approved):
Looks correct.

karolk91 (approved):

bkchr (approved):

franciscoaguirre:
Any idea why CI is failing?

acatangiu:
@franciscoaguirre try merging master for latest CI changes

PR #525, opened 2025-01-03
Fix whitelisted keys in benchmarks

gui1117:
Currently whitelisted keys are manually set in `dispatch_benchmark`. Actually pallet already declare the whitelisted keys, and it is missing many whitelist.

in asset hub the key `relay_well_known_keys::ACTIVE_CONFIG` is no longer whitelisted. It is actually only used in the relay chain pallet `configuration`, and this pallet would whitelist the key if it was used. But asset hub doesn't use this pallet, so it doesn't need to whitelist the key.

This PR will improve the benchmark precision, lower the cost of some calls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
 - Fix missing Encointer democracy pallet hook needed for enactment ([polkadot-fellows/runtimes/pull/508](https://github.com/polkadot-fellows/runtimes/pull/508))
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
```
bkchr, line 8:
```suggestion

### Fixed

```

bkchr (approved):

acatangiu (approved):

skunert (approved):

acatangiu (approved):

bkchr (approved):

Issue #441, opened 2024-08-22
Highlight breaking changes and disruptions more prominently in the CHANGELOG


anaelleltd:
*This is part of the overall strategy to standardise and improve [runtime releases management](https://github.com/polkadot-fellows/runtimes/issues/304).*

Over the past 3 months, I have been pushing targeted notifications into the community to advise network builders and teams about upcoming **breaking changes** (e.g PR [337](https://github.com/polkadot-fellows/runtimes/pull/337)) and **disruptions** (e.g PR [414](https://github.com/polkadot-fellows/runtimes/pull/414), PR [350](https://github.com/polkadot-fellows/runtimes/pull/350) and PR [344](https://github.com/polkadot-fellows/runtimes/issues/344)). These notifications aim to provide the following information: 1) what kind of changes the PR introduces, 2) who will be affected by theses changes in practice, and 3) how the team/builders affected need to follow up. 

I have recently proposed to update the docs and PR template in the runtimes repo (PR [439](https://github.com/polkadot-fellows/runtimes/pull/439)) as the first step for formally onboarding all code contributors to this new process. The second step would be to update the current structure of the CHANGELOG to highlight reported breaking changes and disruptions more prominently.  

This could be done by creating 2 new sections *ad hoc* in the CHANGELOG:
🚨 Breaking changes 🚨 
🚧 Disruptions 🚧

More options/alternatives for getting this done need to be proposed/discussed so that we choose the most convenient and workable solution. 

cc: @bkchr @joepetrowski @kianenigma

ggwpez:
I added some note like this now, but it is not really machine readable. Eventually we may also want to use prdocs here if it does not cause too much churn. They can then define `major` for breaking changes, just like we do in the SDK.

<img width="948" alt="Screenshot 2024-08-26 at 13 10 43" src="https://github.com/user-attachments/assets/0e901b36-88aa-48da-99bc-017f2338d40f">

bkchr:
> Eventually we may also want to use prdocs here if it does not cause too much churn

Generally we can.



> They can then define `major` for breaking changes

However, this sounds like a hack.

ggwpez:
Input from dApp devs https://gist.github.com/mrshiposha/c734f2258c4562003532441b4e51a931 @mrshiposha

PR #535, opened 2025-01-09
Add remote proxy support to AH Kusama

bkchr:
This adds remote proxy support to AssetHub on Polkadot and Kusama. Currently it is configured only to support using proxies registered on the relay chain to work on AssetHub. In the future this can be expanded.

### DIFF
--- a/integration-tests/emulated/chains/parachains/testing/penpal/Cargo.toml
+++ b/integration-tests/emulated/chains/parachains/testing/penpal/Cargo.toml
@@ -29,5 +29,5 @@ runtime-benchmarks = [
 	"cumulus-primitives-core/runtime-benchmarks",
 	"frame-support/runtime-benchmarks",
 	"parachains-common/runtime-benchmarks",
-	"penpal-runtime/runtime-benchmarks"
+	"penpal-runtime/runtime-benchmarks",
 ]


pallets/remote-proxy/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+/// The remote proxy interface.
+pub trait RemoteProxyInterface<AccountId, ProxyType, BlockNumber> {
+	/// The remote account id.
+	type RemoteAccountId: Parameter + MaxEncodedLen;
+	/// The remote proxy type.
+	type RemoteProxyType: Parameter + MaxEncodedLen;
+	/// The remote block number.
+	type RemoteBlockNumber: Parameter + Saturating + MaxEncodedLen + Default;
+	/// The hash type used by the remote chain.
+	type Hash: Parameter + MaxEncodedLen;
```
xlc, line 42: this pallet is in the runtime repo instead of polkadot-sdk so presumably it is only intended to be used by runtimes here. in that case, do we really need this to be that generic?

xlc (commented):
this is not in polkadot-sdk repo so how do we want to test it on westend?

pallets/remote-proxy/src/lib.rs
pandres95: Not in the PSDK doesn't necessarily mean not reusable, thought.

Take for example [frame-contrib](https://github.com/virto-network/frame-contrib), a collection of pallets that are intended to be initially used in [Kreivo](https://github.com/virto-network/kreivo) but are generic and extensible enough so they can be used anywhere else.

Maybe by passing it in the future to the PSDK, or by being used by other chains, it's desirable to be generic.

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
```diff
@@ -992,6 +1038,7 @@ construct_runtime!(
 		Utility: pallet_utility = 40,
 		Multisig: pallet_multisig = 41,
 		Proxy: pallet_proxy = 42,
+		RemoteProxyRelayChain: pallet_remote_proxy::<Instance1> = 43,
```
xlc, line 1041: why Instance1?

xlc (commented):
this is hardcoded in many ways to only support relaychain as remote and no easy wa

pallets/remote-proxy/src/lib.rs
```diff
@@ -0,0 +1,419 @@
...
+
+	/// Get the latest block to storage root mapping.
+	fn block_to_storage_root(
+		validation_data: &PersistedValidationData,
+	) -> Option<(Self::RemoteBlockNumber, <Self::Hasher as Hasher>::Out)>;
+
+	/// The storage key where to find the [`ProxyDefinition`] for the given proxy account in the
+	/// remote chain.
+	fn proxy_definition_storage_key(proxy: &Self::RemoteAccountId) -> Vec<u8> {
+		let mut key = storage_prefix("Proxy".as_bytes(), "Proxies".as_bytes()).to_vec();
```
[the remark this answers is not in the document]
bkchr: The keys are configurable. This is a trait method with a default implementation that could just be changed.

acatangiu:
I'd be more comfortable if we also did a security audit for this before adding it to live Polkadot AssetHub

bkchr:
> I'd be more comfortable if we also did a security audit for this before adding it to live Polkadot AssetHub

The scope of the changes is quite narrow. Doing some proper review should be fine. The only difference to the real `proxy` pallet is that the data is coming in from the outside.

bkchr:
Okay removed the Polkadot integration for now. It can come later after the audit.

Tbaut:
Any chance the people chains can also be included while we're at it? This would help change the identity of pure proxies.

pallets/remote-proxy/src/lib.rs
```diff
@@ -0,0 +1,468 @@
...
+		/// remote proxy call. This is for example useful when having a multisig operation. The
+		/// multisig call can use [`Self::remote_proxy_with_registered_proof`] to get an approval by
+		/// the members of the multisig. The final execution of the multisig call should be at least
+		/// a batch of `register_remote_proxy_proof` and the multisig call that uses
+		/// `remote_proxy_with_registered_proof`. This way the final approver can use a recent proof
+		/// to prove the existence of the remote proxy. Otherwise it would require the multisig
+		/// members to approve the call in [`Config::MaxStorageRootsToKeep`] amount of time.
+		///
+		/// It is supported to register multiple proofs, but the proofs need to be consumed in the
+		/// reverse order as they were registered. Basically this means last in, last out.
```
xlc, line 288:
```suggestion
		/// reverse order as they were registered. Basically this means last in, first out.
```

xlc (commented):
This means remove proxy is only fully effective after `MaxStorageRootsToKeep` blocks. It could be possible to construct some attack based on this behaivour. We should at very minimal document and communicate this change to ensure people are aware of the consequence.
We could ask the UI to display a warning to user when they are trying to remove a proxy indicating the account still have permission on other chain for X blocks.
Another way is somehow be able to blacklist this account for X blocks.

bkchr:
> We should at very minimal document and communicate this change to ensure people are aware of the consequence.

I mean I have documented it. Or where should I document this?

pallets/remote-proxy/src/lib.rs
bkchr: Actually the sentence before was incorrect :see_no_evil:

bkchr:
> #441 we haven't figured out the whole process yet. I see a changelog entry but it does not have the warning. No one really read doc comments.

What kind of warning should there be? I mean users will not see this changelog and would not understand it either. For them nothing breaks, this is more about telling UI builders to show this warning? And they hopefully read the docs to understand on how to implement it? We can also reduce the "keep alive" timer to 1 minute. Should still be enough time to build the proof and get it included.

xlc:
If the keep alive timer is only 1 minute then I wouldn't worry about it too much. Still mention it changelog will be nice, which I expect UI builders to read and think about it.

muharem (commented):
Nice.
It looks good to me. I’m not very familiar with the sp_trie API you’re using, though.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -966,6 +966,52 @@ impl pallet_xcm_bridge_hub_router::Config<ToPolkadotXcmRouterInstance> for Runti
...
+		ProxyDefinition<AccountId, kusama_runtime_constants::proxy::ProxyType, BlockNumber>,
+		Option<ProxyDefinition<AccountId, ProxyType, BlockNumber>>,
+	> for RelayChainToLocalProxyTypeConverter
+{
+	fn convert(
+		a: ProxyDefinition<AccountId, kusama_runtime_constants::proxy::ProxyType, BlockNumber>,
+	) -> Option<ProxyDefinition<AccountId, ProxyType, BlockNumber>> {
+		let proxy_type = match a.proxy_type {
+			kusama_runtime_constants::proxy::ProxyType::Any => ProxyType::Any,
+			kusama_runtime_constants::proxy::ProxyType::NonTransfer => ProxyType::NonTransfer,
```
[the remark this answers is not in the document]
bkchr: In what way they are different? The point being that you can not do any transfers. For me that is exactly the same on every chain.

pallets/remote-proxy/Cargo.toml
```diff
@@ -0,0 +1,70 @@
+[package]
+name = "pallet-remote-proxy"
```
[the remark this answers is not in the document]
eskimor: Disregard, was already asked.

xlc (approved):

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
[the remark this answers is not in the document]
muharem: I see your perspective. My perspective was that if I gave someone `NonTransfer` when it did not include some of the Asset Hub pallets, I would expect it to be the same. But just heads up. 

@ggwpez we might want to take the same approach then for the migration.

muharem (approved):

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1002,6 +1049,7 @@ construct_runtime!(
 		Utility: pallet_utility = 40,
 		Multisig: pallet_multisig = 41,
 		Proxy: pallet_proxy = 42,
+		RemoteProxyRelayChain: pallet_remote_proxy = 43,
```
bkontur, line 1052: Shouldn't we also set `OnSystemEvent` here: https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs#L684-L686?
```
impl cumulus_pallet_parachain_system::Config for Runtime {
-       type OnSystemEvent = ();
+	type OnSystemEvent = (RemoteProxyRelayChain, );
        ...
```

I also think this feature is a great candidate for adding some end-to-end (e2e) or integration tests. The Chopsticks ecosystem tests PR hasn't been merged yet, but at the very least, this should be achievable with emulated integration tests.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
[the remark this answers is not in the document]
ggwpez: In the proxy migration we currently widen the permissions of `NonTransfer` by adding more pallets to the list.  
I wasn't sure if this could be a security issue, since the permissions of the proxy get silently increased without any user action.  
But maybe the intention of no-transfer stays, so could both be fine.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
bkchr: Ha! Good find! :D :see_no_evil:

bkchr: Embarrassing xD

bkontur:
> Command "bench --pallet pallet_remote_proxy" has started 🚀 [See logs here](https://github.com/polkadot-fellows/runtimes/actions/runs/13080682309/job/36503922730)

needs https://github.com/polkadot-fellows/runtimes/pull/573/files#diff-80668ff8e793b64f36a9a3ec512df5cbca4ad448c157a5d81abda1b15f35f1daR1129
```
+ [pallet_remote_proxy, RemoteProxyRelayChain]
```
plus feature propagation and compilation is fixed here: https://github.com/polkadot-fellows/runtimes/pull/573

pallets/remote-proxy/src/lib.rs
```diff
@@ -0,0 +1,468 @@
...
+		///
+		/// It is supported to register multiple proofs, but the proofs need to be consumed in the
+		/// reverse order as they were registered. Basically this means last in, first out.
+		///
+		/// The [`dispatch_context`] spans the entire lifetime of a transaction and every call in
+		/// the transaction gets access to the same context.
+		///
+		/// # Example
+		///
+		/// ```
```
ggwpez, line 295:
```suggestion
		/// ```ignore
```
It tries to compile this with `cargo test`.

pallets/remote-proxy/src/lib.rs
```diff
@@ -0,0 +1,468 @@
...
+
+	impl<T: Config, I: 'static> OnSystemEvent for Pallet<T, I> {
+		fn on_validation_data(validation_data: &PersistedValidationData) {
+			let Some((block, hash)) = T::RemoteProxy::block_to_storage_root(validation_data) else {
+				return;
+			};
+
+			// Update the block to root mappings.
+			BlockToRoot::<T>::insert(&block, hash);
+			BlockToRoot::<T>::remove(block.saturating_sub(T::MaxStorageRootsToKeep::get()));
```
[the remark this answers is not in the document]
bkchr: Good find 🙈

bkchr:
Fuck why was this merged :see_no_evil:

bkchr:
Apparently some checks were not part of the required checks :see_no_evil:

pallets/remote-proxy/src/lib.rs
```diff
@@ -0,0 +1,496 @@
...
+		<T as pallet_proxy::Config>::ProxyType,
+		BlockNumberFor<T>,
+	>>::RemoteProxyType;
+	type WeightInfoOf<T, I> = <T as Config<I>>::WeightInfo;
+
+	#[pallet::pallet]
+	pub struct Pallet<T, I = ()>(_);
+
+	/// Stores the last [`Config::MaxStorageRootsToKeep`] block to storage root mappings of the
+	/// target chain.
```
ggwpez, line 176: One comment please on the order of entries here.

pallets/remote-proxy/src/lib.rs
```diff
@@ -0,0 +1,496 @@
...
+				let delete_up_to =
+					block.clone().saturating_sub(T::MaxStorageRootsToKeep::get().into());
+
+				while roots.first().is_some_and(|f| f.0 <= delete_up_to) {
+					roots.remove(0);
+				}
+
+				// We always remove all the old items before, thus there should always be space in
+				// the vector.
+				let _res = roots.try_push((block, hash));
```
ggwpez, line 224: Could use `defensive_proof` here to get a production error log.

pallets/remote-proxy/src/lib.rs
```diff
@@ -0,0 +1,496 @@
...
+			);
+
+			Ok(())
+		}
+
+		/// Dispatch the given `call` from an account that the sender is authorised on a remote
+		/// chain.
+		///
+		/// The dispatch origin for this call must be _Signed_. The difference to
+		/// [`Self::remote_proxy`] is that the proof nees to registered before using
```
ggwpez, line 350: typo

pallets/remote-proxy/src/lib.rs
```diff
@@ -0,0 +1,496 @@
...
+					>|
+					 -> bool {
+						x.delegate == who &&
+							force_proxy_type.as_ref().map_or(true, |y| &x.proxy_type == y)
+					};
+
+					proxy_definitions
+						.into_iter()
+						.filter_map(T::RemoteProxy::remote_to_local_proxy_defintion)
+						.find(f)
```
ggwpez, line 446: I guess in the case that there are multiple proxies and no force type, it will just pick the first one?  
Should not produce issues i think, just an observation.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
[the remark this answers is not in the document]
bkchr: The idea of `non-transfer` is that the proxy can do *anything* besides transferring funds. Every pallet that was added on the relay chain side, was also just added to this filter if it did not transfer funds.

PR #538, opened 2025-01-10
force unstake the last remaining corrupt ledger

kianenigma:
One last staking ledger that due to the bug found last year is in an inconsistent state. 

The Support team is in touch with the owner, and force_staking them is the preferred way. 

Note that this is only an action that we are doing because the root cause of the inconsistent stake was buggy code in the polkadot runtime, not any user interaction. 

See: https://forum.polkadot.network/t/recover-corrupted-staking-ledgers-in-polkadot-and-kusama/9796


bkchr:
This could be just its own referendum?

kianenigma:
> This could be just its own referendum?

yeah it could, but since it would require Root, and we have previously done other fixes related to this bug as runtime upgrade, I opted for this. 

We can still do Root, but it will certainly be longer, and a more cumbersome process.

bkchr:
This could also be done via the whitelisted track? I mean this is just fixing up stuff, nothing controversial.

kianenigma:
Yeah since it is a technical bug, fellowship can be asked to whitelist. 

I think there is an easier path if we merge this as-is, since we are close to publishing 1.4 anyways, but if you think otherwise, we can also pursue it via the fellowship.

relay/polkadot/src/lib.rs
```diff
@@ -2013,158 +1988,61 @@ pub mod migrations {
...
+		// 12gmcL9eej9jRBFT26vZLF4b7aAe4P9aEYHGHFzJdmf5arPi
+		pub CorruptStash: AccountId = hex_literal::hex!(
+			"4a90f8d375290b428e580408f18359f66ef367f0f8d1d56c5d3e002ad29c8e00"
+		).into();
+	}
+	pub struct Migrate;
+	impl OnRuntimeUpgrade for Migrate {
+		fn on_runtime_upgrade() -> frame_election_provider_support::Weight {
+			// ensure this only runs once, in the 1.4.0 release
+			if VERSION.spec_version == 1_400_000 {
```
bkchr, line 2019:
```suggestion
			if System::last_runtime_upgrade_spec_version() < 1_400_000 {
```

bkchr (approved):

acatangiu (approved):

acatangiu:
do we do CHANGELOG for this?

seadanda (approved):

bkontur:
In my PR on the actual `main`, polkadot check-migrations CI job failed:
https://github.com/polkadot-fellows/runtimes/actions/runs/12947252387/job/36113430989?pr=547

```
[2025-01-24T10:40:51Z ERROR runtime] panicked at /home/runner/work/runtimes/runtimes/relay/polkadot/src/lib.rs:2039:13:
    assertion failed: pallet_staking::Ledger::<Runtime>::get(CorruptStash::get()).is_some()
thread 'main' panicked at cli/main.rs:326:10:
called `Result::unwrap()` on an `Err` value: Input("failed to execute TryRuntime_on_runtime_upgrade: Execution aborted due to trap: wasm trap: wasm `unreachable` instruction executed\nWASM backtrace:\nerror while executing at wasm backtrace:\n    0: 0xa96e - polkadot_runtime.wasm!rust_begin_unwind\n    1: 0x3ca5 - polkadot_runtime.wasm!core::panicking::panic_fmt::hf7f2b23977fc8496\n    2: 0x9b68 - polkadot_runtime.wasm!core::panicking::panic::h6230965df90522ef\n    3: 0x5cae5b - polkadot_runtime.wasm!<(TupleElement0,TupleElement1,TupleElement2,TupleElement3,TupleElement4,TupleElement5,TupleElement6,TupleElement7,TupleElement8,TupleElement9) as frame_support::traits::hooks::OnRuntimeUpgrade>::try_on_runtime_upgrade::h3468e69caa69cbad\n    4: 0x6c9492 - polkadot_runtime.wasm!TryRuntime_on_runtime_upgrade")
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Relevant code: https://github.com/polkadot-fellows/runtimes/blob/main/relay/polkadot/src/lib.rs#L2039

Why? Was this fixed on chain somehow?

bkchr:
@kianenigma ^^^^

Looks like we can remove the migration?

bkchr:
Hmm, so the `CorruptStash` still exists on-chain.

bkontur:
so why did my run fail? Is this corrupted stash "sometimes ok" and "sometimes not" (Schrödinger's cat)?

Shouldn't we be less strict and fix this `pre_upgrade/post_upgrade` something like?
```
		fn pre_upgrade() -> Result<Vec<u8>, sp_runtime::TryRuntimeError> {
                        let found_corrupted = pallet_staking::Ledger::<Runtime>::get(CorruptStash::get()).is_some(); 
			Ok(found_corrupted.encode())
		}

			total_weight
		fn post_upgrade(state: Vec<u8>) -> Result<(), sp_runtime::TryRuntimeError> {
                        let found_corrupted: bool = Decode::decode(&mut &state:[..)?;]
                        if found_corrupted {
			        assert!(pallet_staking::Ledger::<Runtime>::get(CorruptStash::get()).is_none());
			}
			Ok(())
		}
```

bkchr:
@bkontur yeah sounds good.

PR #539, opened 2025-01-10
Add PoolAssetsExchanger to asset hubs

franciscoaguirre:
This adds the new `AssetExchanger` introduced in https://github.com/paritytech/polkadot-sdk/pull/5130 to asset hubs.
This exchanger is necessary to enable [delivery fees to be paid in assets other than DOT/KSM](https://github.com/paritytech/polkadot-sdk/pull/5131), for example, it will enable cross-chain transfers from one parachain to another, going through Asset Hub, using only USDT, no DOT/KSM involved.


system-parachains/asset-hubs/asset-hub-kusama/src/xcm_config.rs
```diff
@@ -197,6 +197,28 @@ pub type PoolFungiblesTransactor = FungiblesAdapter<
...
+/// Used to convert one asset to another, when there is a pool available between the two.
+/// This type thus allows paying delivery fees with any asset as long as there is a pool between said
+/// asset and the asset required for fee payment.
+pub type PoolAssetsExchanger = SingleAssetExchangeAdapter<
+	AssetConversion,
+	NativeAndAssets,
+	(
+		TrustBackedAssetsAsLocation<TrustBackedAssetsPalletLocation, Balance, xcm::v4::Location>,
+		ForeignAssetsConvertedConcreteId,
+		// `ForeignAssetsConvertedConcreteId` excludes the relay token, so we add it back here.
```
acatangiu, line 210: this comment is confusing, what do you mean it "excludes", maybe more accurate will be?:

```suggestion
		// `ForeignAssetsConvertedConcreteId` doesn't include Relay token, so we handle it explicitly here.
```

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -196,6 +196,28 @@ pub type PoolFungiblesTransactor = FungiblesAdapter<
...
+/// Used to convert one asset to another, when there is a pool available between the two.
+/// This type thus allows paying delivery fees with any asset as long as there is a pool between said
+/// asset and the asset required for fee payment.
+pub type PoolAssetsExchanger = SingleAssetExchangeAdapter<
+	AssetConversion,
+	NativeAndAssets,
+	(
+		TrustBackedAssetsAsLocation<TrustBackedAssetsPalletLocation, Balance, xcm::v4::Location>,
+		ForeignAssetsConvertedConcreteId,
+		// `ForeignAssetsConvertedConcreteId` excludes the relay token, so we add it back here.
```
acatangiu, line 209: same, maybe like this?

```suggestion
		// `ForeignAssetsConvertedConcreteId` doesn't include Relay token, so we handle it explicitly here.
```

acatangiu (approved):

bkontur (approved):

eskimor (approved):

PR #542, opened 2025-01-13
[AHM] Proxy migration

ggwpez:
👉 To be merged into the AHM dev branch.

quote from the `proxy.md` doc:

## Pallet Proxy

The proxy pallet consists of two storage variables.
## Storage: Proxies

The [Proxies](https://github.com/paritytech/polkadot-sdk/blob/7c5224cb01710d0c14c87bf3463cc79e49b3e7b5/substrate/frame/proxy/src/lib.rs#L564-L579) storage map maps a delegator to its delegates. It can be translated one-to-one by mapping the `ProxyType` and `Delay` fields.
### Proxy Type Translation
The different kinds that are possible for a proxy are a [runtime injected type](https://github.com/paritytech/polkadot-sdk/blob/7c5224cb01710d0c14c87bf3463cc79e49b3e7b5/substrate/frame/proxy/src/lib.rs#L119-L125). Since these are different for each runtime, we need a converter that maps the Relay to AH `ProxyType` as close as possible to keep the original intention. The Relay kind is defined [here](https://github.com/polkadot-fellows/runtimes/blob/dde99603d7dbd6b8bf541d57eb30d9c07a4fce32/relay/polkadot/src/lib.rs#L1000-L1010) and the AH version [here](https://github.com/polkadot-fellows/runtimes/blob/fd8d0c23d83a7b512e721b1fde2ba3737a3478d5/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L453-L468). This is done by injecting a `RcToProxyType` converter into the Asset Hub migration pallet. This is not bullet proof since it relies on some copy&paste code instead of pulling in the Polkadot runtime into the AH runtime but it is the simplest solution.

Mapping from Relay to AH looks as follows:
- Any: same
- NonTransfer: same
- Governance: newly added
- Staking: newly added
- Variant 4: ignore as it is a historic remnant
- Variant 5: ignore ditto
- CancelProxy: same
- Auction: dropped
- NominationPools: newly added

All variants that serve no purpose anymore on the Relay Chain are deleted from there. For example `Staking`. The ones that are still usable on the relay like `NonTransfer` are **also deleted** since there is no storage deposit taken anymore. (TODO think about what is best here)
### Translation of the Delay

The [delay of a ProxyDefinition](https://github.com/paritytech/polkadot-sdk/blob/7c5224cb01710d0c14c87bf3463cc79e49b3e7b5/substrate/frame/proxy/src/lib.rs#L77) is measured in blocks. These are currently 6 seconds Relay blocks. To translate them to 12s AH blocks, we can divide the number by two.
## Storage: Announcements

The [Announcements](https://github.com/paritytech/polkadot-sdk/blob/7c5224cb01710d0c14c87bf3463cc79e49b3e7b5/substrate/frame/proxy/src/lib.rs#L581-L592) storage maps proxy account IDs to [Accouncement](https://github.com/paritytech/polkadot-sdk/blob/7c5224cb01710d0c14c87bf3463cc79e49b3e7b5/substrate/frame/proxy/src/lib.rs#L80-L89). Since an announcement contains a call hash, we cannot translate them for the same reason as with the Multisigs; the preimage of the hash would be either undecodable, decode to something else (security issue) or accidentally decode to the same thing.  

We therefore do not migrate the announcements.
## User Impact
- Announcements need to be re-created
- Proxies of type `Auction` are not migrated and need to be re-created on the Relay


bkchr:
> The [delay of a ProxyDefinition](https://github.com/paritytech/polkadot-sdk/blob/7c5224cb01710d0c14c87bf3463cc79e49b3e7b5/substrate/frame/proxy/src/lib.rs#L77) is measured in blocks. These are currently 6 seconds Relay blocks. To translate them to 12s AH blocks, we can divide the number by two.

We should translate the existing proxies to the relay chain and not the other way around.

ggwpez:
> We should translate the existing proxies to the relay chain and not the other way around.

WDYM? We want to remove everything from the Relay chain as much as possible. Including proxies and balances - if possible.

bkchr:
> WDYM? We want to remove everything from the Relay chain as much as possible. Including proxies and balances - if possible.

I expressed myself badly. When we upgrade to the SDK release that includes `BlockNumberProvider`, we should switch to `RelayChainBlockNumber` as provider. Then we migrate all the existing proxies on AH to the relay chain block number. Later when you do the migration, you only need to migrate the data from the relay chain to AH without touching the `delay` field.

ggwpez:
> I expressed myself badly. When we upgrade to the SDK release that includes `BlockNumberProvider`, we should switch to `RelayChainBlockNumber` as provider. Then we migrate all the existing proxies on AH to the relay chain block number. Later when you do the migration, you only need to migrate the data from the relay chain to AH without touching the `delay` field.

Yea sounds good. Will do once we have it in the SDK and upgraded to it. The change should be rather small from what is not being added.

PR #543, opened 2025-01-15
[AHM] Cleanup and add Preimage doc

ggwpez:
👉 To be merged into the Asset Hub Migration working branch.

Changes:
- Factor out XCM sending into common function
- Cleanup
- Add docs for the preimage pallet that I started with


pallets/rc-migrator/src/lib.rs
```diff
@@ -364,10 +369,64 @@ pub mod pallet {
...
+			create_call: impl Fn(Vec<E>) -> types::AhMigratorCall<T>,
+		) -> Result<(), Error<T>> {
+			const MAX_MSG_SIZE: u32 = 50_000; // Soft message size limit. Hard limit is about 64KiB
+
+			while !items.is_empty() {
+				let mut remaining_size: u32 = MAX_MSG_SIZE;
+				let mut batch = Vec::new();
+
+				while !items.is_empty() {
+					// Order does not matter, so we take from the back as optimization
```
muharem, line 394: maybe worth mentioning in the doc

pallets/rc-migrator/src/preimage.md
```diff
@@ -0,0 +1,191 @@
...
+
+Q: One question here would be whether or not we want to translate these calls. I think we can and should. But I am not sure about the best time point to do so. We can translate the preimages calls upon arrival on the Asset Hub, although there is a small change that a preimage that was not intended to be decoded as a call would be translated. After all, the preimage pallet is a universal data storage without any implied semantics about its content. It could therefore be better to translate the preimages once we are migrating the Referenda pallet and then only translate all preimages that occur as hash lookups in a referenda.
+Basically: loop referenda -> load referenda -> load preimage of referenda -> translate preimage -> calculate new preimage hash -> update preimage with new hash -> update referenda with new hash.  
+One further note on the preimage translation: If the length of a preimage is increased by the translation, then we should not reconsider the deposit but keep the original deposit as to not punish users for this. The cases that the translation increases the size of a preimage past the 4MiB hard limit should be negligible.
```
muharem, line 9: the deposit should be lower on the AH. probably wont be a problem. we keep the deposit same and provide a call to update it (which should release some funds)

pallets/rc-migrator/src/preimage.md
```diff
@@ -0,0 +1,191 @@
+The preimage pallet consists of three storage maps, one of which is a legacy item. The interesting one are `RequestStatusFor` and `PreimageFor`.
+
+## Storage: PreimageFor
+
+[Maps](https://github.com/paritytech/polkadot-sdk/blob/00946b10ab18331f959f5cbced7c433b6132b1cb/substrate/frame/preimage/src/lib.rs#L185) a hash and the length of a preimage to its preimage data. Preimages can be migrated rather easily by sending them in chunks from the Relay and appending them on the Asset Hub. The preimages are often used to store large governance calls.
+
+Q: One question here would be whether or not we want to translate these calls. I think we can and should. But I am not sure about the best time point to do so. We can translate the preimages calls upon arrival on the Asset Hub, although there is a small change that a preimage that was not intended to be decoded as a call would be translated. After all, the preimage pallet is a universal data storage without any implied semantics about its content. It could therefore be better to translate the preimages once we are migrating the Referenda pallet and then only translate all preimages that occur as hash lookups in a referenda.
```
muharem, line 7: agree. something we also have discussed with Joe and Donal, we wanna map them.

muharem, line 7: only scheduler which used by referenda or root can request a preimage. in theory there should only preimages with valid calls having request count > 0.

muharem (approved):

pallets/rc-migrator/src/lib.rs
```diff
@@ -364,10 +369,64 @@ pub mod pallet {
...
-			log::info!(target: LOG_TARGET, "[Block {:?}] Stage transition: {:?} -> {:?}", frame_system::Pallet::<T>::block_number(), old, new_stage);
+			RcMigrationStage::<T>::put(&new);
+			log::info!(target: LOG_TARGET, "[Block {:?}] Stage transition: {:?} -> {:?}", frame_system::Pallet::<T>::block_number(), &old, &new);
+			Self::deposit_event(Event::StageTransition { old, new });
+		}
+
+		/// Split up the items into chunks of `MAX_MSG_SIZE` and send them as separate XCM
+		/// transacts.
+		///
+		/// This is done to avoid exceeding the XCM message size limit.
```
muharem, line 382:
```suggestion
		/// This is done to avoid exceeding the XCM message size limit.
		///
		/// Note: The order of `items` is not preserved.
		/// The `items` are batched and processed in reverse order for optimization.
```

muharem, line 382:
```suggestion
```suggestion
		/// This is done to avoid exceeding the XCM message size limit.
		///
		/// Note: The order of `items` is not preserved.
		/// The `items` are batched and processed in reverse order for optimization.
```
```

pallets/rc-migrator/src/lib.rs
ggwpez: Oh yea, I will reverse the vector first to get rid of this.

pallets/rc-migrator/src/lib.rs
ggwpez: reverted

pallets/rc-migrator/src/preimage.md
ggwpez: Oh right, so we could also do the translation as we migrate the preimages.

But then when we do the governance migration afterwards it will not find the hashes anymore. So they kind of need to happen together anyway.

muharem: Right. Might be worth migrating them together

Issue #446, opened 2024-08-23
investigate errors when skipping xcm benchmarks for collectives-polkadot


brenzi:
While fixing XCM benchmarking for encointer in #440, I got an error while running xcm::generic benchmarks:
https://gist.github.com/brenzi/bb6050f9cb00d34f279cf1f4e5b6e4ba

I did 
```
~/bin/polkadot-parachain-v1.14.0-runtime-benchmarks benchmark pallet --chain ./encointer-kusama-chain-spec.json --pallet=pallet-xcm-benchmarks::generic --extrinsic="*"
``` 
and got `Error executing and verifying runtime benchmark: Benchmark not found for this pallet.` 

This error should be ignored because the benchmark is skipped for our runtime, as this functionality isn't provided by encointer (exactly like people chain for example).

I relalized that a check needs to be removed and that the coretime chain still has this check implemented. I guess it should be removed too?

https://github.com/polkadot-fellows/runtimes/blob/790ef91d8ba3215df1b6d76b49547d404f07e83f/system-parachains/collectives/collectives-polkadot/src/lib.rs#L1164

bkontur:
Will be fixed by: https://github.com/polkadot-fellows/runtimes/pull/547

bkontur:
@brenzi can you please close this? It is merged in `main`.

PR #547, opened 2025-01-16
Measured XCM weights and benchmarking for collectives-polkadot

bkontur:
Closes: https://github.com/paritytech/polkadot-sdk/issues/2904
Closes: https://github.com/polkadot-fellows/runtimes/issues/446
Continuation: https://github.com/paritytech/polkadot-sdk/pull/6820

The `collectives-polkadot` uses `FixedWeightBounds`. This PR introduces a properly measured XCM benchmarking setup and updates the weights with fresh data.

## TODO
- [x] check if we use somewhere some hard-coded weights for `Transact` that goes to the `collectives-polkadot`


bkontur:
>     * [ ]  check if we use somewhere some hard-coded weights for `Transact` that goes to the `collectives-polkadot`

@seadanda @muharem any idea?

seadanda:
> > ```
> > * [ ]  check if we use somewhere some hard-coded weights for `Transact` that goes to the `collectives-polkadot`
> > ```
> 
> @seadanda @muharem any idea?

There is nothing talking to collectives in this way AFAIK. We have only used that pattern for the People Chain migration and coretime interface so far.

system-parachains/collectives/collectives-polkadot/src/weights/xcm/pallet_xcm_benchmarks_fungible.rs
```diff
@@ -0,0 +1,190 @@
...
+            .saturating_add(T::DbWeight::get().writes(3))
+    }
+    /// Storage: `Benchmark::Override` (r:0 w:0)
+    /// Proof: `Benchmark::Override` (`max_values`: None, `max_size`: None, mode: `Measured`)
+    pub(crate) fn reserve_asset_deposited() -> Weight {
+        // Proof Size summary in bytes:
+        //  Measured:  `0`
+        //  Estimated: `0`
+        // Minimum execution time: 18_446_744_073_709_551_000 picoseconds.
+        Weight::from_parts(18_446_744_073_709_551_000, 0)
```
muharem, line 103: quite big, but seems not only for this runtime

system-parachains/collectives/collectives-polkadot/src/lib.rs
```diff
@@ -1164,12 +1174,113 @@ impl_runtime_apis! {
...
+
+			type XcmBalances = pallet_xcm_benchmarks::fungible::Pallet::<Runtime>;
+			type XcmGeneric = pallet_xcm_benchmarks::generic::Pallet::<Runtime>;
+
 			let whitelist: Vec<TrackedStorageKey> = AllPalletsWithSystem::whitelisted_storage_keys();
 			let mut batches = Vec::<BenchmarkBatch>::new();
 			let params = (&config, &whitelist);
 			add_benchmarks!(params, batches);
 
-			if batches.is_empty() { return Err("Benchmark not found for this pallet.".into()) }
```
muharem, line 1172: why not needed anymore?

muharem (approved):

muharem:
> > ```
> > * [ ]  check if we use somewhere some hard-coded weights for `Transact` that goes to the `collectives-polkadot`
> > ```
> 
> @seadanda @muharem any idea?

I do not remember any. But I tried to cover all important cases with integration tests. I would check them, if anything should be updated or added and we should be good

system-parachains/collectives/collectives-polkadot/src/weights/xcm/pallet_xcm_benchmarks_fungible.rs
acatangiu: this is basically `Weight::MAX` and used for unsupported instructions - in this particular case `ReserveAssetDeposited` is disabled on collectives because it doesn't support any reserve assets, it only supports DOT through teleports

system-parachains/collectives/collectives-polkadot/src/lib.rs
bkontur: This has been removed from all Polkadot-SDK and Polkadot-Fellows runtimes, so why keep it here?  

https://github.com/polkadot-fellows/runtimes/issues/446

I'm not entirely sure about the purpose of this `IF`, but I can imagine it might have caused issues for scripting/bots. For example, when regenerating a specific `pallet_xyz` for a set of runtimes, if one of those runtimes doesn’t include `pallet_xyz`, the process might abort instead of simply skipping the runtime without that pallet.  

However, I believe that for scripting/bots, we now use `benchmark pallet --list`, so this shouldn't be an issue anymore.

system-parachains/collectives/collectives-polkadot/src/weights/xcm/pallet_xcm_benchmarks_fungible.rs
bkontur: > this is basically `Weight::MAX` and used for unsupported instructions - in this particular case `ReserveAssetDeposited` is disabled on collectives because it doesn't support any reserve assets, it only supports DOT through teleports

:point_up_2:

acatangiu (approved):

system-parachains/collectives/collectives-polkadot/src/weights/xcm/pallet_xcm_benchmarks_fungible.rs
seadanda: Maybe a DQ: why not just use `Weight::Max`?

muharem (approved):

system-parachains/collectives/collectives-polkadot/src/weights/xcm/pallet_xcm_benchmarks_fungible.rs
acatangiu: This is how benchmarks generate it, not sure about the reasoning why it was implemented like this.

PR #550, opened 2025-01-17
[AHM] Improve accounts migration

muharem:
Meant to be merged into `dev-asset-hub-migration` branch.

Some improvements and fixes to the accounts migration.

### DIFF
--- a/pallets/rc-migrator/src/accounts.rs
+++ b/pallets/rc-migrator/src/accounts.rs
@@ -189,15 +189,17 @@ impl<T: Config> PalletMigration for AccountsMigrator<T> {
 			return Err(Error::OutOfWeight);
 		}
 
-		let mut iter = if let Some(last_key) = last_key {
+		let mut iter = if let Some(ref last_key) = last_key {
 			SystemAccount::<T>::iter_from_key(last_key)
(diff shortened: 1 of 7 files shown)


pallets/rc-migrator/src/accounts.rs
```diff
@@ -298,18 +329,36 @@ impl<T: Config> Pallet<T> {
 			pallet_balances::Freezes::<T>::get(&who).into();
 
 		for freeze in &freezes {
-			<T as Config>::Currency::thaw(&freeze.id, &who)
-				// TODO: handle error
-				.unwrap();
+			if let Err(e) = <T as Config>::Currency::thaw(&freeze.id, &who) {
```
ggwpez, line 332:
```suggestion
			if let Err(e) = <T as Config>::Currency::thaw(&freeze.id, &who).defensive() {
```
Then it also fails the test directly.

ggwpez (approved):

pallets/rc-migrator/src/accounts.rs
muharem: there is actually defensive one level up if this function return an error

PR #551, opened 2025-01-17
Prepare for self-hosted runner

bkchr:

### DIFF
--- a/.github/install-deps.sh
+++ b/.github/install-deps.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+
+sudo apt update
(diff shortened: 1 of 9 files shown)


ggwpez (approved):

xlc (approved):
we are also using self-hosted runner and we bake essential deps on the runner so that there is no need to install them repeatedly in workflows

seadanda (approved):

muharem (approved):

bkchr:
> we are also using self-hosted runner and we bake essential deps on the runner so that there is no need to install them repeatedly in workflows

Yeah I also proposed that, but for now let's move with this and have something working.

PR #553, opened 2025-01-20
Activate 10 Mb PoVs

s0me0ne-unkn0wn:
This will activate 10 Mb PoVs on Kusama as soon as the runtime is upgraded.
The point is that we want to activate it anyway, and we better have it activated sooner rather than later to allow more time for testing before proposing that for Polkadot.

CC @eskimor @sandreim


relay/kusama/src/lib.rs
```diff
@@ -1902,6 +1906,50 @@ parameter_types! {
...
+	T: parachains_configuration::Config,
+{
+	#[cfg(feature = "try-runtime")]
+	fn pre_upgrade() -> Result<Vec<u8>, sp_runtime::TryRuntimeError> {
+		// The pre-upgrade state doesn't matter
+		Ok(vec![])
+	}
+
+	fn on_runtime_upgrade() -> Weight {
+		match parachains_configuration::Pallet::<T>::set_max_pov_size(
```
bkchr, line 1923:
```suggestion
		match parachains_configuration::Pallet::<Runtime>::set_max_pov_size(
```

Then the struct doesn't need to be generic.

s0me0ne-unkn0wn: I thought you loved generic structs :sweat_smile: 
Fixed in https://github.com/polkadot-fellows/runtimes/pull/553/commits/6e3353a0a20be5b862a67f88833b27cabb071073

bkchr (approved):

ordian (approved):
I am assuming the concerns with high CPU usage at full utilization from erasure coding are not a blocker?

acatangiu (approved):

skunert (approved):

eskimor (approved):

PR #559, opened 2025-01-23
[AHM] Add Polkadot call filtering

ggwpez:
To be merged into the AHM working branch.

Changes:
- Configure the RC Migrator pallet as call filter for the Polkadot Relay chain
- Test

### DIFF
--- a/pallets/rc-migrator/src/lib.rs
+++ b/pallets/rc-migrator/src/lib.rs
@@ -208,15 +208,12 @@ pub mod pallet {
 		type AhWeightInfo: AhWeightInfo;
 		/// The existential deposit on the Asset Hub.
 		type AhExistentialDeposit: Get<<Self as pallet_balances::Config>::Balance>;
-		/// Whether this call is permanently disabled on the Relay Chain.
-		///
-		/// This includes all pallets that will be moved to the Asset Hub and we dont want anyone to
-		/// keep using them.
-		type RcCallEnabledAfterMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
-		/// Whether this Relay Chain call is disabled temporarily during the migration.
+		/// Contains all calls that are allowed during the migration.
 		///
 		/// The calls in here will be available again after the migration.
 		type RcCallEnabledDuringMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
+		/// Contains all calls that are allowed after the migration finished.
+		type RcCallEnabledAfterMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
 	}
 
 	#[pallet::error]
--- a/pallets/rc-migrator/src/lib.rs
+++ b/pallets/rc-migrator/src/lib.rs
@@ -211,9 +211,9 @@ pub mod pallet {
 		/// Contains all calls that are allowed during the migration.
 		///
 		/// The calls in here will be available again after the migration.
-		type RcCallEnabledDuringMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
+		type RcCallsEnabledDuringMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
 		/// Contains all calls that are allowed after the migration finished.
-		type RcCallEnabledAfterMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
+		type RcCallsEnabledAfterMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
 	}
 
 	#[pallet::error]
--- a/pallets/rc-migrator/src/lib.rs
+++ b/pallets/rc-migrator/src/lib.rs
@@ -211,9 +211,9 @@ pub mod pallet {
 		/// Contains all calls that are allowed during the migration.
 		///
 		/// The calls in here will be available again after the migration.
-		type RcCallsEnabledDuringMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
+		type RcIntraMigrationCalls: Contains<<Self as frame_system::Config>::RuntimeCall>;
 		/// Contains all calls that are allowed after the migration finished.
-		type RcCallsEnabledAfterMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
+		type RcPostMigrationCalls: Contains<<Self as frame_system::Config>::RuntimeCall>;
 	}
 
 	#[pallet::error]
@@ -612,11 +612,11 @@ impl<T: Config> Contains<<T as frame_system::Config>::RuntimeCall> for Pallet<T>
 		const ALLOWED: bool = true;
 		const FORBIDDEN: bool = false;
 
-		if is_finished && !T::RcCallEnabledAfterMigration::contains(call) {
+		if is_finished && !T::RcIntraMigrationCalls::contains(call) {
 			return FORBIDDEN;
 		}
 
-		if is_ongoing && !T::RcCallEnabledDuringMigration::contains(call) {
+		if is_ongoing && !T::RcPostMigrationCalls::contains(call) {
 			return FORBIDDEN;
 		}
 
--- a/pallets/rc-migrator/src/lib.rs
+++ b/pallets/rc-migrator/src/lib.rs
@@ -46,7 +46,7 @@ use frame_support::{
 	traits::{
 		fungible::{Inspect, InspectFreeze, Mutate, MutateFreeze, MutateHold},
 		tokens::{Fortitude, Precision, Preservation},
-		Defensive, LockableCurrency, ReservableCurrency,
+		Contains, Defensive, LockableCurrency, ReservableCurrency,
 	},
 	weights::WeightMeter,
 };
@@ -96,7 +96,19 @@ impl<T: Config> From<OutOfWeightError> for Error<T> {
 	}
 }
 
-#[derive(Encode, Decode, Clone, PartialEq, Eq, Default, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+#[derive(
+	Encode,
+	Decode,
+	Clone,
+	PartialEq,
+	Eq,
+	Default,
+	RuntimeDebug,
+	TypeInfo,
+	MaxEncodedLen,
+	PartialOrd,
+	Ord,
+)]
 pub enum MigrationStage<AccountId> {
 	/// The migration has not yet started but will start in the next block.
 	#[default]
@@ -196,6 +208,15 @@ pub mod pallet {
 		type AhWeightInfo: AhWeightInfo;
 		/// The existential deposit on the Asset Hub.
 		type AhExistentialDeposit: Get<<Self as pallet_balances::Config>::Balance>;
+		/// Whether this call is permanently disabled on the Relay Chain.
+		///
+		/// This includes all pallets that will be moved to the Asset Hub and we dont want anyone to
+		/// keep using them.
+		type RcCallEnabledAfterMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
+		/// Whether this Relay Chain call is disabled temporarily during the migration.
+		///
+		/// The calls in here will be available again after the migration.
+		type RcCallEnabledDuringMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
 	}
 
 	#[pallet::error]
@@ -249,16 +270,6 @@ pub mod pallet {
 
 	#[pallet::pallet]
 	pub struct Pallet<T>(_);
-
-	#[pallet::call]
-	impl<T: Config> Pallet<T> {
-		/// TODO
-		#[pallet::call_index(0)]
-		#[pallet::weight({1})]
-		pub fn do_something(_origin: OriginFor<T>) -> DispatchResultWithPostInfo {
-			Ok(().into())
-		}
-	}
 
 	#[pallet::hooks]
 	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
@@ -591,5 +602,27 @@ pub mod pallet {
 
 			Ok(())
 		}
+	}
+}
+
+impl<T: Config> Contains<<T as frame_system::Config>::RuntimeCall> for Pallet<T> {
+	fn contains(call: &<T as frame_system::Config>::RuntimeCall) -> bool {
+		let stage = RcMigrationStage::<T>::get();
+		let is_finished = stage >= MigrationStage::MigrationDone;
+		let is_ongoing = stage > MigrationStage::Pending && !is_finished;
+
+		// We have to return whether the call is allowed:
+		const ALLOWED: bool = true;
+		const FORBIDDEN: bool = false;
+
+		if is_finished && !T::RcCallEnabledAfterMigration::contains(call) {
+			return FORBIDDEN;
+		}
+
+		if is_ongoing && !T::RcCallEnabledDuringMigration::contains(call) {
+			return FORBIDDEN;
+		}
+
+		ALLOWED
 	}
 }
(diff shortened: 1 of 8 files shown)


pallets/rc-migrator/src/lib.rs
```diff
@@ -196,6 +208,12 @@ pub mod pallet {
 		type AhWeightInfo: AhWeightInfo;
 		/// The existential deposit on the Asset Hub.
 		type AhExistentialDeposit: Get<<Self as pallet_balances::Config>::Balance>;
+		/// Contains all calls that are allowed during the migration.
+		///
+		/// The calls in here will be available again after the migration.
+		type RcCallEnabledDuringMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
+		/// Contains all calls that are allowed after the migration finished.
+		type RcCallEnabledAfterMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
```
seadanda, line 216:
```suggestion
		type RcCallsEnabledDuringMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
		/// Contains all calls that are allowed after the migration finished.
		type RcCallsEnabledAfterMigration: Contains<<Self as frame_system::Config>::RuntimeCall>;
```

Or even something like `RcIntraMigrationCalls` and `RcPostMigrationCalls` for shorter naming.

relay/polkadot/src/lib.rs
```diff
@@ -1576,6 +1576,113 @@ impl pallet_rc_migrator::Config for Runtime {
...
+///       Start        End
+///
+/// This call returns a 2-tuple to indicate whether a call is disabled during these periods.
+pub fn call_disable_status(call: &<Runtime as frame_system::Config>::RuntimeCall) -> (bool, bool) {
+	use RuntimeCall::*;
+	const ON: bool = false;
+	const OFF: bool = true;
+
+	match call {
+		System(..) => (OFF, ON),
```
seadanda, line 1627: Should we consider an escape hatch for System? If we bork the migration for some unknown reason and get stuck and don't get to the `MigrationDone` stage for more than an era (or some reasonable time period) we should allow some subset of root-only calls to rescue ourselves from doomsday gov. CC @bkontur

relay/polkadot/src/lib.rs
```diff
@@ -1576,6 +1576,113 @@ impl pallet_rc_migrator::Config for Runtime {
...
+		ParaInherent(..) => (ON, ON),    // only inherents
+		// ParaScheduler has no calls
+		Paras(..) => (OFF, ON), // TODO only root so could think about keeping it on
+		Initializer(..) => (OFF, ON), // only root so could think about keeping it on
+		// Dmp // has no calls and deprecated
+		Hrmp(..) => (OFF, OFF),
+		// ParaSessionInfo has no calls
+		ParasDisputes(..) => (OFF, ON), // TODO check with security
+		ParasSlashing(..) => (OFF, ON), // TODO check with security
+		OnDemand(..) => (OFF, ON),
```
seadanda, line 1672: If we get https://github.com/paritytech/polkadot-sdk/pull/5990 merged in time we should try to avoid filtering on-demand, as there will no longer be a balances dependency and we should avoid a situation where parachains can't make blocks through the migration.

relay/polkadot/src/lib.rs
```diff
@@ -1576,6 +1576,113 @@ impl pallet_rc_migrator::Config for Runtime {
...
+		// CoretimeAssignmentProvider has no calls
+		Registrar(..) => (OFF, ON),
+		Slots(..) => (OFF, ON), // TODO not sure
```
seadanda, line 1675:
```suggestion
		Slots(..) => (OFF, OFF),
```

We don't need these extrinsics with coretime.

relay/polkadot/src/lib.rs
```diff
@@ -1576,6 +1576,113 @@ impl pallet_rc_migrator::Config for Runtime {
...
+		Auctions(..) => (OFF, OFF),
+		Crowdloan(..) => (OFF, ON), // TODO maybe only payouts
```
seadanda, line 1677: Yea probably just `withdraw`, `refund` and `dissolve`.
Ideally we'd transfer this state to AH and allow people to withdraw or refund there directly, but it would be basically an entirely new pallet and would break all the crowdloan apps maybe unnecessarily. That's likely going to be the next phase of AHM though when the EDs go up. Still like 20 months until the last crowdloans expire so we need something long term

relay/polkadot/src/lib.rs
```diff
@@ -1576,6 +1576,113 @@ impl pallet_rc_migrator::Config for Runtime {
...
+		Coretime(..) => (OFF, ON),
```
seadanda, line 1678: These are all privileged calls anyway, I think they should be allowed except for request_revenue_info_at, but we should be careful about the timing of this.

relay/polkadot/src/lib.rs
```diff
@@ -1576,6 +1576,113 @@ impl pallet_rc_migrator::Config for Runtime {
...
+		StateTrieMigration(..) => (OFF, OFF), // Deprecated
+		XcmPallet(..) => (OFF, ON),
```
seadanda, line 1680: Again, this would cause problems across the coretime interface. We should find a clean way to allow this for root or paras only during this period of time

relay/polkadot/src/lib.rs
```diff
@@ -1576,6 +1576,113 @@ impl pallet_rc_migrator::Config for Runtime {
...
+		Multisig(..) => (OFF, ON),
+		Bounties(..) => (OFF, OFF),
+		ChildBounties(..) => (OFF, OFF),
+		ElectionProviderMultiPhase(..) => (OFF, OFF),
+		VoterList(..) => (OFF, OFF),
+		NominationPools(..) => (OFF, OFF),
+		FastUnstake(..) => (OFF, OFF),
+		// DelegatedStaking has on calls
+		// ParachainsOrigin has no calls
+		Configuration(..) => (OFF, ON),
```
seadanda, line 1660:
```suggestion
		Configuration(..) => (ON, ON),
```
All root calls

relay/polkadot/src/lib.rs
```diff
@@ -1576,6 +1576,113 @@ impl pallet_rc_migrator::Config for Runtime {
...
+		ParasShared(..) => (OFF, OFF), /* Has no calls but a call enum https://github.com/paritytech/polkadot-sdk/blob/ee803b74056fac5101c06ec5998586fa6eaac470/polkadot/runtime/parachains/src/shared.rs#L185-L186 */
+		ParaInclusion(..) => (OFF, OFF), /* Has no calls but a call enum https://github.com/paritytech/polkadot-sdk/blob/74ec1ee226ace087748f38dfeffc869cd5534ac8/polkadot/runtime/parachains/src/inclusion/mod.rs#L352-L353 */
+		ParaInherent(..) => (ON, ON),    // only inherents
+		// ParaScheduler has no calls
+		Paras(..) => (OFF, ON), // TODO only root so could think about keeping it on
+		Initializer(..) => (OFF, ON), // only root so could think about keeping it on
```
seadanda, line 1666:
```suggestion
		Initializer(..) => (ON, ON),
```

relay/polkadot/src/lib.rs
```diff
@@ -1576,6 +1576,113 @@ impl pallet_rc_migrator::Config for Runtime {
...
+		NominationPools(..) => (OFF, OFF),
+		FastUnstake(..) => (OFF, OFF),
+		// DelegatedStaking has on calls
+		// ParachainsOrigin has no calls
+		Configuration(..) => (OFF, ON),
+		ParasShared(..) => (OFF, OFF), /* Has no calls but a call enum https://github.com/paritytech/polkadot-sdk/blob/ee803b74056fac5101c06ec5998586fa6eaac470/polkadot/runtime/parachains/src/shared.rs#L185-L186 */
+		ParaInclusion(..) => (OFF, OFF), /* Has no calls but a call enum https://github.com/paritytech/polkadot-sdk/blob/74ec1ee226ace087748f38dfeffc869cd5534ac8/polkadot/runtime/parachains/src/inclusion/mod.rs#L352-L353 */
+		ParaInherent(..) => (ON, ON),    // only inherents
+		// ParaScheduler has no calls
+		Paras(..) => (OFF, ON), // TODO only root so could think about keeping it on
```
seadanda, line 1665:
```suggestion
		Paras(..) => (ON, ON),
```
I agree, and also this (and the other parachains pallet extrinsics) shouldn't have any bearing on the balances or state that we're migrating, but CC @x3c41a would be good to test that the migrations work even if these extrinsics run mid migration. Maybe more like fuzzing since there are so many calls and stage combinations and a stretch, but would be nice to have if we have time.

relay/polkadot/tests/ahm/mod.rs
```diff
@@ -48,3 +49,101 @@ async fn remote_ext_test_setup() -> Option<RemoteExternalities<Block>> {
 
 	Some(ext)
 }
+
+#[test]
+fn call_filter_works() {
+	let mut t: sp_io::TestExternalities =
+		frame_system::GenesisConfig::<T>::default().build_storage().unwrap().into();
+
+	// MQ calls are never filtered:
```
seadanda, line 58: Did we get to the bottom of the prioritisation of system messages? I lost track of the status of your streaking work

relay/polkadot/tests/ahm/mod.rs
```diff
@@ -48,3 +49,101 @@ async fn remote_ext_test_setup() -> Option<RemoteExternalities<Block>> {
...
+
+			assert!(!is_forbidden(&mq_call));
+			assert!(!is_forbidden(&system_call));
+			assert!(is_forbidden(&indices_call));
+		}
+	});
+}
+
+fn is_forbidden(call: &polkadot_runtime::RuntimeCall) -> bool {
+	let Err(err) = call.clone().dispatch(RuntimeOrigin::signed(AccountId32::from([0; 32]))) else {
```
seadanda, line 143: This is actually a nice test to run on basically every pallet and extrinsic during the migration except for a few exceptions that are explicitly allowed from signed origins - the ones that are root only will fail anyway from this origin.

We could then have a separate check for the root calls that should be filtered, with the baseline being that all root calls are allowed except the ones we know will cause problems.

seadanda (commented):
I like the approach! We should get input from a wider group on the individual filtered calls/pallets, maybe based on the owners on the excel sheet

relay/polkadot/src/lib.rs
muharem: We also can have a similar impl for the EnsureOrigin if needed to let only Fellows to command some calls during migration. I do not think we can execute the root fast enough.

relay/polkadot/src/lib.rs
muharem: The impact is very low, I would just disable.

relay/polkadot/src/lib.rs
```diff
@@ -1576,6 +1576,113 @@ impl pallet_rc_migrator::Config for Runtime {
...
+		// CoretimeAssignmentProvider has no calls
+		Registrar(..) => (OFF, ON),
+		Slots(..) => (OFF, ON), // TODO not sure
+		Auctions(..) => (OFF, OFF),
+		Crowdloan(..) => (OFF, ON), // TODO maybe only payouts
+		Coretime(..) => (OFF, ON),
+		StateTrieMigration(..) => (OFF, OFF), // Deprecated
+		XcmPallet(..) => (OFF, ON),
+		MessageQueue(..) => (ON, ON), // TODO think about this
+		AssetRate(..) => (OFF, OFF),  // TODO @muharem
```
muharem, line 1682: it can be just disabled for simplicity. it usually being altered once in a quarter if price changes and only by Gov. but even if left on, no problem. this only serves for the price rate between the some coin and DOT to assert the right Gov origin.

relay/polkadot/src/lib.rs
```diff
@@ -1576,6 +1576,113 @@ impl pallet_rc_migrator::Config for Runtime {
 	type AhExistentialDeposit = AhExistentialDeposit;
 	type RcWeightInfo = ();
 	type AhWeightInfo = ();
+	type RcCallEnabledDuringMigration = CallsEnabledDuringMigration;
+	type RcCallEnabledAfterMigration = CallsEnabledAfterMigration;
+}
+
+/// Contains all calls that are enabled during the migration.
+pub struct CallsEnabledDuringMigration;
```
muharem, line 1584: should we move all new type concerning only the migration to some migration.rs module? easier to navigate and clean up it later.

relay/polkadot/src/lib.rs
```diff
@@ -1576,6 +1576,113 @@ impl pallet_rc_migrator::Config for Runtime {
...
+/// --------|-----------|--------->
+///       Start        End
+///
+/// This call returns a 2-tuple to indicate whether a call is disabled during these periods.
+pub fn call_disable_status(call: &<Runtime as frame_system::Config>::RuntimeCall) -> (bool, bool) {
+	use RuntimeCall::*;
+	const ON: bool = false;
+	const OFF: bool = true;
+
+	match call {
```
muharem, line 1626: may be some TODO to review the list after we progress more with the pallets migration. we might find out new or drop some constrains.

muharem (approved):

relay/polkadot/src/lib.rs
ggwpez: Will turn it on for now. TX fees should be no issue i think since we move the accounts first anyway.

relay/polkadot/src/lib.rs
ggwpez: I will add a TODO

relay/polkadot/src/lib.rs
ggwpez: >The impact is very low, I would just disable.

You mean also after the migration? I was thinking about enabling the withdraw function after the migration again.

relay/polkadot/tests/ahm/mod.rs
ggwpez: Yea it worked here but we should get it audited & merged https://github.com/paritytech/polkadot-sdk/pull/6059

relay/polkadot/tests/ahm/mod.rs
ggwpez: Yes we could do that. Currently we need to manually construct the call arguments as well since our RuntimeCall enum does not implement an `Arbitrary` trait.

relay/polkadot/src/lib.rs
ggwpez: Change it to this now:  
```rust
Coretime(coretime::Call::<Runtime>::request_revenue_at { .. }) => (OFF, ON),
Coretime(..) => (ON, ON),
```

relay/polkadot/src/lib.rs
ggwpez: Changed to:  
```rust
Crowdloan(crowdloan::Call::<Runtime>::dissolve { .. } | crowdloan::Call::<Runtime>::refund { .. } | crowdloan::Call::<Runtime>::dissolve { .. }) => (OFF, ON),
Crowdloan(..) => (OFF, OFF),
```

relay/polkadot/src/lib.rs
ggwpez: Yea good idea.

relay/polkadot/src/lib.rs
ggwpez: Yes good idea, will do.

relay/polkadot/src/lib.rs
ggwpez: I will create an issue to check it in the end so we dont miss.

relay/polkadot/src/lib.rs
ggwpez: I can enable it since it should not mess with the migration.

ggwpez:
Going to merge. Tests should run after https://github.com/polkadot-fellows/runtimes/pull/561

relay/polkadot/src/lib.rs
muharem: this actually only filters extrinsics to the pallet, and does not lock messages received via UMP and its processing via xcm-executor

PR #563, opened 2025-01-28
[AHM] Fast unstake and Bags list

ggwpez:
To be merged into the AHM dev branch.

There should be no user impact here - both pallets have their storage migrated exactly as is and the used times are always in Epoch.


system-parachains/asset-hubs/asset-hub-polkadot/src/staking/bags_thresholds.rs
ggwpez, line 1: Need to check if we have to re-generate these values.

muharem: No proper review given from me for this file. But I will do it on request. May be we better consult with staking team on this, will take more time from me.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -15,5 +15,34 @@
...
+pub mod bags_thresholds;
 pub mod nom_pools;
+
+use crate::*;
+
+impl pallet_fast_unstake::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type Currency = Balances;
+	type BatchSize = frame_support::traits::ConstU32<16>;
+	type Deposit = frame_support::traits::ConstU128<{ UNITS }>;
```
muharem, line 30: might need to be updated, since should cost less on a parachain

pallets/rc-migrator/src/lib.rs
```diff
@@ -151,17 +158,30 @@ pub enum MigrationStage<AccountId> {
...
+	BagsListMigrationInit,
+	BagsListMigrationOngoing {
+		next_key: Option<BagsListStage<AccountId, BagsListScore>>,
+	},
+	BagsListMigrationDone,
 	MigrationDone,
 }
 
-pub type MigrationStageFor<T> = MigrationStage<<T as frame_system::Config>::AccountId>;
+pub type MigrationStageFor<T> = MigrationStage<
```
muharem, line 179: Do we have some non written convention when we use `For` or `Of` suffix? I wished to keep it consistent (= 
I had a feeling that Gav used `For` for storage items and `Of` for types, but I am not sure.

pallets/rc-migrator/src/lib.rs
```diff
@@ -575,6 +596,39 @@ pub mod pallet {
...
+						Ok(None) => {
+							Self::transition(MigrationStage::FastUnstakeMigrationDone);
+						},
+						Ok(Some(next_key)) => {
+							Self::transition(MigrationStage::FastUnstakeMigrationOngoing {
+								next_key: Some(next_key),
+							});
+						},
+						e => {
+							log::error!(target: LOG_TARGET, "Error while migrating fast unstake: {:?}", e);
```
muharem, line 626: actually no need for error, since defensive is below. just would be nice to pass the `e` for defensive as well

pallets/rc-migrator/src/staking/bags_list.rs
```diff
@@ -0,0 +1,179 @@
...
+				.try_consume(<T as frame_system::Config>::DbWeight::get().reads_writes(1, 1))
+				.is_err()
+			{
+				if messages.is_empty() {
+					return Err(Error::OutOfWeight);
+				} else {
+					break;
+				}
+			}
+			if messages.len() > 10_000 {
```
muharem, line 76: temporary measure or not? can you explain please

pallets/ah-migrator/src/staking/bags_list.rs
```diff
@@ -0,0 +1,72 @@
...
+
+use crate::*;
+use pallet_rc_migrator::staking::bags_list::alias;
+
+impl<T: Config> Pallet<T> {
+	pub fn do_receive_bags_list_messages(
+		messages: Vec<RcBagsListMessage<T>>,
+	) -> Result<(), Error<T>> {
+		let (mut good, mut bad) = (0, 0);
+		log::info!("Integrating {} BagsListMessages", messages.len());
```
muharem, line 28: log target?

muharem (approved):

pallets/rc-migrator/src/lib.rs
ggwpez: I was wondering about the same thing. Core issue is that there is `AccountIdOf` and `BlockNumberFor`, so two very central types in contradiction 😆 🤦‍♂️   

I dont really mind when we use what, we can use `Of` for types if you want.

pallets/rc-migrator/src/staking/bags_list.rs
ggwpez: I was not sure here how exactly we benchmark things. Depending that, this would be a failsafe in case that our weights are wrong to not completely overload the system.  
It is not great or precise, but better than some infinite loop.

pallets/ah-migrator/src/staking/bags_list.rs
ggwpez: Not sure if it adds much. Do you use it often to filter?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
ggwpez: Updated to 1 / 100 UNIT.

PR #569, opened 2025-01-30
[AHM] Scheduler Pallet

muharem:
meant to be merged into `dev-asset-hub-migration`

The migration for the Scheduler pallet from Relay Chain to Asset Hub.

### DIFF
--- a/pallets/ah-migrator/src/call.rs
+++ b/pallets/ah-migrator/src/call.rs
@@ -42,7 +42,7 @@ impl<T: Config> Pallet<T> {
 			return Err(Error::<T>::FailedToConvertCall);
 		};
 
-		log::info!("MAPPED CALL: {:?}", call);
+		log::debug!(target: LOG_TARGET, "mapped call: {:?}", call);
 
 		let ah_bounded_call = T::Preimage::bound(call).map_err(|err| {
 			defensive!("Failed to bound call: {:?}", err);
@@ -52,7 +52,7 @@ impl<T: Config> Pallet<T> {
 		if ah_bounded_call.lookup_needed() {
 			// Noted preimages for referendums that did not pass will need to be manually removed
 			// later.
-			log::info!("New preimage was noted for call");
+			log::info!(target: LOG_TARGET, "New preimage was noted for call");
 		}
 
 		Ok(ah_bounded_call)
@@ -66,7 +66,7 @@ impl<T: Config> Pallet<T> {
 					encoded
 				} else {
 					// not an error since a submitter can delete the preimage for ongoing referendum
-					log::warn!("No preimage found for call hash: {:?}", hash);
+					log::warn!(target: LOG_TARGET, "No preimage found for call hash: {:?}", hash);
 					return Err(Error::<T>::PreimageNotFound);
 				};
 				Ok(encoded.into_owned())
@@ -76,7 +76,7 @@ impl<T: Config> Pallet<T> {
 					encoded
 				} else {
 					// not an error since a submitter can delete the preimage for ongoing referendum
-					log::warn!("No preimage found for call hash: {:?}", (hash, len));
+					log::warn!(target: LOG_TARGET, "No preimage found for call hash: {:?}", (hash, len));
 					return Err(Error::<T>::PreimageNotFound);
 				};
 				Ok(encoded.into_owned())
--- a/pallets/ah-migrator/src/call.rs
+++ b/pallets/ah-migrator/src/call.rs
@@ -1,0 +1,86 @@
+// Copyright (C) Parity Technologies (UK) Ltd.
+// This file is part of Polkadot.
+
+// Polkadot is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+use crate::*;
+use frame_support::traits::Bounded;
+
+pub type BoundedCallOf<T> =
+	Bounded<<T as frame_system::Config>::RuntimeCall, <T as frame_system::Config>::Hashing>;
+
+impl<T: Config> Pallet<T> {
+	pub fn map_rc_ah_call(
+		rc_bounded_call: &BoundedCallOf<T>,
+	) -> Result<BoundedCallOf<T>, Error<T>> {
+		let encoded_call = if let Ok(e) = Self::fetch_preimage(rc_bounded_call) {
+			e
+		} else {
+			return Err(Error::<T>::PreimageNotFound);
+		};
+
+		if let Some(hash) = rc_bounded_call.lookup_hash() {
+			if T::Preimage::is_requested(&hash) {
+				T::Preimage::unrequest(&hash);
+			}
+		}
+
+		let call = if let Ok(call) = T::RcToAhCall::try_convert(&encoded_call) {
+			call
+		} else {
+			return Err(Error::<T>::FailedToConvertCall);
+		};
+
+		log::info!("MAPPED CALL: {:?}", call);
+
+		let ah_bounded_call = T::Preimage::bound(call).map_err(|err| {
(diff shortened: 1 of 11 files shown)


pallets/ah-migrator/src/call.rs
```diff
@@ -0,0 +1,86 @@
...
+
+		let call = if let Ok(call) = T::RcToAhCall::try_convert(&encoded_call) {
+			call
+		} else {
+			return Err(Error::<T>::FailedToConvertCall);
+		};
+
+		log::info!("MAPPED CALL: {:?}", call);
+
+		let ah_bounded_call = T::Preimage::bound(call).map_err(|err| {
```
ggwpez, line 47: This is not noting the preimage, right? So it just returns a lookup hash?  
so when or where will we note it?

pallets/ah-migrator/src/lib.rs
```diff
@@ -267,6 +274,16 @@ pub mod pallet {
 			count_bad: u32,
 		},
 		ReferendaProcessed,
+		SchedulerMessagesReceived {
```
ggwpez, line 277: I added a generic variant with a `PalletName` that we can use as to not copy&paste this always for new pallets.

pallets/ah-migrator/src/scheduler.rs
```diff
@@ -0,0 +1,115 @@
...
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+use crate::*;
+use frame_support::traits::DefensiveTruncateFrom;
+use pallet_rc_migrator::scheduler::{alias::Scheduled, RcSchedulerMessage};
+
+/// Messages set from the RC Migrator concerning the Scheduler pallet.
```
ggwpez, line 21:
```suggestion
/// Messages sent from the RC Migrator concerning the Scheduler pallet.
```

pallets/ah-migrator/src/scheduler.rs
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	fn do_process_scheduler_message(message: RcSchedulerMessageOf<T>) -> Result<(), Error<T>> {
+		log::debug!("Processing scheduler message: {:?}", message);
+
+		match message {
+			RcSchedulerMessage::IncompleteSince(block_number) => {
+				pallet_scheduler::IncompleteSince::<T>::put(block_number);
+			},
+			RcSchedulerMessage::Agenda((block_number, tasks)) => {
```
ggwpez, line 57: I wonder how many of the agendas we need to migrate? We can also re-set the alarms for all agendas after migrating them.  
Referendas should be the only thing using the scheduler AFAIK.

pallets/ah-migrator/src/scheduler.rs
```diff
@@ -0,0 +1,115 @@
...
+		log::info!(target: LOG_TARGET, "Processed {} scheduler messages", count_good);
+
+		Ok(())
+	}
+
+	fn do_process_scheduler_message(message: RcSchedulerMessageOf<T>) -> Result<(), Error<T>> {
+		log::debug!("Processing scheduler message: {:?}", message);
+
+		match message {
+			RcSchedulerMessage::IncompleteSince(block_number) => {
```
ggwpez, line 54: The assumption is that we use Relay block numbers and an updated scheduler paller, right?

pallets/ah-migrator/src/scheduler.rs
```diff
@@ -0,0 +1,115 @@
...
+						let task = Scheduled {
+							maybe_id: task.maybe_id,
+							priority: task.priority,
+							call,
+							maybe_periodic: task.maybe_periodic,
+							origin,
+						};
+						Some(task)
+					} else {
+						None
```
ggwpez, line 92: Actually the `None` entries are noise. We can clean them up while we are at it.

pallets/rc-migrator/src/scheduler.rs
```diff
@@ -0,0 +1,196 @@
...
+impl<T: Config> PalletMigration for SchedulerMigrator<T> {
+	type Key = SchedulerStage<BlockNumberFor<T>>;
+	type Error = Error<T>;
+	fn migrate_many(
+		last_key: Option<Self::Key>,
+		weight_counter: &mut WeightMeter,
+	) -> Result<Option<Self::Key>, Self::Error> {
+		let mut last_key = last_key.unwrap_or(SchedulerStage::IncompleteSince);
+		let mut messages = Vec::new();
+
```
ggwpez, line 56: We also have to go through all pallets `on_initialize` code and possibly disable it during the migration. Otherwise some pallets may break from partial state being migrated. I will add it as TODO

pallets/rc-migrator/src/scheduler.rs
```diff
@@ -0,0 +1,196 @@
...
+			last_key = match last_key {
+				SchedulerStage::IncompleteSince => {
+					if let Some(since) = pallet_scheduler::IncompleteSince::<T>::take() {
+						messages.push(RcSchedulerMessage::IncompleteSince(since));
+					}
+					SchedulerStage::Agenda(None)
+				},
+				SchedulerStage::Agenda(last_key) => {
+					let mut iter = if let Some(last_key) = last_key {
+						alias::Agenda::<T>::iter_from_key(last_key)
```
ggwpez, line 82: Not sure if you have to call `hashed_key_for`. Did you try it it migrates all?

ggwpez (approved):
Looks good, just one Q if we need to migrate it in general 😅

pallets/ah-migrator/src/call.rs
```diff
@@ -0,0 +1,86 @@
...
+			}
+		}
+
+		let call = if let Ok(call) = T::RcToAhCall::try_convert(&encoded_call) {
+			call
+		} else {
+			return Err(Error::<T>::FailedToConvertCall);
+		};
+
+		log::info!("MAPPED CALL: {:?}", call);
```
muharem, line 45: ups, my debug log

pallets/ah-migrator/src/call.rs
muharem: This does `note` if it does not fit into `inline` size limit.

pallets/ah-migrator/src/scheduler.rs
muharem: On AH currently we only need a scheduler with RC block number provider. It will serve the referenda. For now I do not see any other use cases for the scheduler. 
Why RC block provider for referenda? Because for referendums we want the times as decision period to be close to real world period of 28 days for example.

pallets/ah-migrator/src/scheduler.rs
muharem: We can simply disable scheduler for the time of the migration (I will create a PR for it later), and the scheduler should just process those overdue tasks when migration will end (at least now it is how it works, I will keep my eyes on the last change we introduce with block number provider for the scheduler).

pallets/rc-migrator/src/scheduler.rs
muharem: Right. I have an idea how to do it for the scheduler

pallets/rc-migrator/src/scheduler.rs
muharem: I think so, I will double check. I used `iter_from_key` for all other pallets migrations

muharem:
> Looks good, just one Q if we need to migrate it in general 😅

I thought about it, and I think the simplest is to just migrate it. We might have passed referendums' calls that scheduled to be enacted. Also we have referenda pallet's service calls that already scheduled. We pause the scheduler for the time of migration and those call would just be dispatched when migration ends. Otherwise we need some custom logic for these calls. The presented approach seems to be the simplest to me.

pallets/ah-migrator/src/call.rs
ggwpez: I dont find where it notes it, i checked this [bound](https://github.com/paritytech/polkadot-sdk/blob/f4f56f6cf819472fcbab7ef367ec521f26cb85cb/substrate/frame/support/src/traits/preimages.rs#L247-L260) function.

pallets/ah-migrator/src/scheduler.rs
ggwpez: Discussed in DM.

pallets/rc-migrator/src/scheduler.rs
ggwpez: Yea but it is tricky since it sometimes does still catch a subset. We will see when testing,  it could be a likely culprit when entries are missing.

ggwpez:
>The presented approach seems to be the simplest to me.

Yea sounds good.

ggwpez (approved):

PR #574, opened 2025-02-03
Update rust and use stable for benches

bkchr:

We are trying to get the command bot working. However, right now it is timing out when trying to build the runtimes plus the omni bencher after 6 hours. This should not take 6 hours ;) So, one guess is that maybe the nightly is doing something that takes quite long to build. We want to try the stable compiler and see if that fixes the issue.


bkontur:
Can we also change profile to `production` here: https://github.com/polkadot-fellows/runtimes/blob/main/.github/scripts/cmd/cmd.py#L82? Polkadot-sdk also uses `production`: https://github.com/paritytech/polkadot-sdk/blob/master/.github/scripts/cmd/cmd.py#L107

pandres95 (approved):

ggwpez (approved):
There is also a `--timings` option on the build command, but i guess that does not help much when it already times out.  
I used this reverse shell SSH stuff in the past to connect to an actions runner, it is a bit awkward but works: https://github.com/marketplace/actions/debugging-with-ssh#shut-down-server-if-user-doesnt-connect

eskimor (approved):

bkontur (approved):

acatangiu (approved):
new rust version comes with some new helper/sugar functions that clippy wants us to use :)

bkchr:
> There is also a `--timings` option on the build command, but i guess that does not help much when it already times out. I used this reverse shell SSH stuff in the past to connect to an actions runner, it is a bit awkward but works: https://github.com/marketplace/actions/debugging-with-ssh#shut-down-server-if-user-doesnt-connect Not sure how it interacts with self-hosted runners though 🤔

We already tried to run the commands manually on the machine and it worked.

PR #576, opened 2025-02-04
Add xcm recorder to all runtimes

franciscoaguirre:
The `XcmRecorder` config item is necessary for the `DryRunApi` to return the `local_xcm`.
`pallet-xcm` can be used as the implementation.
Added it to all runtimes.

### DIFF
--- a/relay/kusama/src/xcm_config.rs
+++ b/relay/kusama/src/xcm_config.rs
@@ -189,7 +189,7 @@ pub struct XcmConfig;
 impl xcm_executor::Config for XcmConfig {
 	type RuntimeCall = RuntimeCall;
 	type XcmSender = XcmRouter;
-	type XcmRecorder = ();
+	type XcmRecorder = XcmPallet;
 	type AssetTransactor = LocalAssetTransactor;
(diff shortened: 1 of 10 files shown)


acatangiu (approved):

pandres95 (approved):
Lgtm.

Was this related to the dry run API not returning proper values in the past?

relay/kusama/src/xcm_config.rs
```diff
@@ -189,7 +189,7 @@ pub struct XcmConfig;
 impl xcm_executor::Config for XcmConfig {
 	type RuntimeCall = RuntimeCall;
 	type XcmSender = XcmRouter;
-	type XcmRecorder = ();
+	type XcmRecorder = XcmPallet;
```
bkontur, line 192: maybe we could regenerate xcm weights also, this adds one more read :),
but also we can leave it to the next time

bkontur (approved):

franciscoaguirre:
> Lgtm.
> 
> Was this related to the dry run API not returning proper values in the past?

Yes

bkchr (approved):

PR #584, opened 2025-02-09
[AHM] Stage management

muharem:
meant to be merged into ahm dev branch

Stage management for the Asset Hub migration


pallets/ah-migrator/src/lib.rs
```diff
@@ -172,6 +193,10 @@ pub mod pallet {
 	pub type RcAccounts<T: Config> =
 		StorageMap<_, Twox64Concat, T::AccountId, RcAccountFor<T>, OptionQuery>;
 
+	/// The Asset Hub migration state.
+	#[pallet::storage]
+	pub type AhMigrationStage<T: Config> = StorageValue<_, MigrationStage, ValueQuery>;
```
ggwpez, line 198: Not sure if `ValueQuery` makes sense if our default is `Pending` which says that the migration will start in the next block.

pallets/ah-migrator/src/lib.rs
```diff
@@ -542,6 +599,48 @@ pub mod pallet {
...
+			let old = AhMigrationStage::<T>::get();
+			AhMigrationStage::<T>::put(&new);
+			log::info!(
+				target: LOG_TARGET,
+				"[Block {:?}] Stage transition: {:?} -> {:?}",
+				frame_system::Pallet::<T>::block_number(),
+				&old,
+				&new
+			);
+			Self::deposit_event(Event::StageTransition { old, new });
```
ggwpez, line 614: Maybe this should send an UMP each time there is a stage transition? Then the relay can keep track of progress and we only have to send it from one spot in the code.

pallets/ah-migrator/src/types.rs
```diff
@@ -25,3 +25,17 @@ pub fn map_lock_reason(reasons: LockReasons) -> LockWithdrawReasons {
...
+pub enum RcPalletConfig {
+	#[codec(index = 255)]
+	RcmController(RcMigratorCall),
+}
+
+/// Call encoding for the calls needed from the rc-migrator pallet.
+#[derive(Encode, Decode)]
+pub enum RcMigratorCall {
+	#[codec(index = 2)]
+	StartDataMigration,
```
ggwpez, line 40: I think this should maybe be something like `AhStageChanged(stage)` and then define the stage type either in the RC pallet or in a shared crate.

pallets/rc-migrator/src/lib.rs
```diff
@@ -283,6 +291,10 @@ pub mod pallet {
 	{
 		/// The overarching event type.
 		type RuntimeEvent: From<Event<Self>> + IsType<<Self as frame_system::Config>::RuntimeEvent>;
+		/// The origin that can perform permissioned operations like setting the migration stage.
+		///
+		/// This is generally root and Fellows origins.
```
ggwpez, line 296: We could try to tie this in with a point on this plan here: https://docs.google.com/document/d/1urpWdOkvQuhROeXf3FyjkdOcLiC36NpfKVADmig0lQI/edit?tab=t.0#heading=h.nc2w1yrxt8yf  
Not sure if Donal sent it to you yet

pallets/rc-migrator/src/lib.rs
```diff
@@ -365,18 +377,75 @@ pub mod pallet {
...
 	pub struct Pallet<T>(_);
 
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Set the migration stage.
+		///
+		/// This call is intended for emergency use only and is guarded by the
+		/// [`Config::ManagerOrigin`].
+		#[pallet::call_index(0)]
+		pub fn set_stage(origin: OriginFor<T>, stage: MigrationStageOf<T>) -> DispatchResult {
```
ggwpez, line 387:
```suggestion
		pub fn force_set_stage(origin: OriginFor<T>, stage: MigrationStageOf<T>) -> DispatchResult {
```
So we dont call it by accident.

pallets/rc-migrator/src/lib.rs
```diff
@@ -365,18 +377,75 @@ pub mod pallet {
...
+			Self::transition(MigrationStage::Scheduled { block_number });
+			Ok(())
+		}
+
+		/// Start the data migration.
+		///
+		/// This is typically called by the Asset Hub to indicate it's readiness to receive the
+		/// migration data.
+		#[pallet::call_index(2)]
+		pub fn start_data_migration(origin: OriginFor<T>) -> DispatchResult {
```
ggwpez, line 411:
```suggestion
		pub fn ah_ready_for_data(origin: OriginFor<T>) -> DispatchResult {
```
? I think it could help to include `ah` in there to show that it comes from the AH.

pallets/rc-migrator/src/lib.rs
```diff
@@ -121,6 +121,12 @@ pub enum MigrationStage<AccountId, BlockNumber, BagsListScore, AccountIndex, Vot
 	/// The migration has not yet started but will start in the next block.
 	#[default]
 	Pending,
+	/// The migration has been scheduled to start at the given block number.
+	Scheduled {
+		block_number: BlockNumber,
+	},
+	/// The migration is initializing.
+	Initializing,
```
ggwpez, line 129:
```suggestion
	WaitingForAhReadiness,
```

pallets/rc-migrator/src/lib.rs
```diff
@@ -365,18 +377,75 @@ pub mod pallet {
...
+		}
+
+		/// Start the data migration.
+		///
+		/// This is typically called by the Asset Hub to indicate it's readiness to receive the
+		/// migration data.
+		#[pallet::call_index(2)]
+		pub fn start_data_migration(origin: OriginFor<T>) -> DispatchResult {
+			<T as Config>::ManagerOrigin::ensure_origin(origin)?;
+			Self::transition(MigrationStage::AccountsMigrationInit);
```
ggwpez, line 413: Maybe defensive assert to check that we are in the right phase.

pallets/rc-migrator/src/lib.rs
```diff
@@ -365,18 +377,75 @@ pub mod pallet {
...
 				MigrationStage::Pending => {
-					// TODO: not complete
+					// TODO: we should do nothing on pending stage.
+					// On production the AH will send a message and initialize the migration.
+					// Now we transition to `AccountsMigrationInit` to run tests
 					Self::transition(MigrationStage::AccountsMigrationInit);
 				},
+				MigrationStage::Scheduled { block_number } =>
+					if now >= block_number {
+						match Self::send_xcm(types::AhMigratorCall::<T>::StartMigration) {
```
ggwpez, line 434:
```suggestion
						match Self::send_xcm(types::AhMigratorCall::<T>::IsAhReadyToMigrate) {
```
Sorry for the silly naming suggestions, but i think it can help later to see what is happening.

pallets/rc-migrator/src/lib.rs
```diff
@@ -365,18 +377,75 @@ pub mod pallet {
...
+							Err(_) => {
+								defensive!(
+									"Failed to send StartMigration message to AH, \
+									retry with the next block"
+								);
+							},
+						}
+					},
+				MigrationStage::Initializing => {
+					// waiting AH to send a message and to start sending the data
```
ggwpez, line 447: We could add some timeout in case AH does not send back the Ok, but should not do now.

pallets/rc-migrator/src/lib.rs
ggwpez: PS: comment should include Asset Hub as origin.

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
```diff
@@ -1120,6 +1120,10 @@ impl pallet_claims::Config for Runtime {
 
 impl pallet_ah_migrator::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
+	type ManagerOrigin = EitherOfDiverse<
+		EnsureRoot<AccountId>,
```
ggwpez, line 1124: The relay chain will come in as root, right? So we dont need to use this `EnsureXcm`?

pallets/ah-migrator/src/lib.rs
muharem: I would first see the results of dry run. It would be more clear what we need, and do we need optimize on migration time for example.

pallets/ah-migrator/src/types.rs
muharem: May be `receive_stage_update(stage)`? If we have many, it makes total sense. If we have three, may be three different calls is simpler solution. But no strong opinion. Again I would may be wait until we see the results of dry run.

pallets/rc-migrator/src/lib.rs
muharem: may be I better update the doc? It will be more general and if needed we can include more into it.

pallets/rc-migrator/src/lib.rs
muharem: > We could try to tie this in with a point on this plan here: https://docs.google.com/document/d/1urpWdOkvQuhROeXf3FyjkdOcLiC36NpfKVADmig0lQI/edit?tab=t.0#heading=h.nc2w1yrxt8yf Not sure if Donal sent it to you yet

It opens a doc for me at the first line.

pallets/rc-migrator/src/lib.rs
muharem: I think can be more general, we might include more. But also no strong opinion. I just trying to not create too many stages to not waste time.

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
muharem: Right

pallets/ah-migrator/src/lib.rs
muharem: I changed the doc for `Pending` in both pallets (...will start in the future). In both pallets it's `ValueQuery`. I think its better than the alternative with `OptionQuery` and no `Pending` variants at all (otherwise its never reachable stage).

pallets/rc-migrator/src/lib.rs
muharem: updated the doc

pallets/ah-migrator/src/lib.rs
muharem: But what we achieve by that? AH does not really know much about the stage (it does not know if all pallets data is sent). 
I can see that we can have AH sending an XCM message after it processes every XCM message, but it might increase the migration time significantly. 
Or we can let AH know that this message with the data is the last on this pallet, but AH still can get congested just on one pallet with a lot of data.

pallets/rc-migrator/src/lib.rs
```diff
@@ -122,9 +122,18 @@ pub enum PalletEventName {
 #[derive(Encode, Decode, Clone, Default, RuntimeDebug, TypeInfo, MaxEncodedLen, PartialEq, Eq)]
 pub enum MigrationStage<AccountId, BlockNumber, BagsListScore, AccountIndex, VotingClass, AssetKind>
 {
-	/// The migration has not yet started but will start in the next block.
+	/// The migration has not yet started but will start in the future.
 	#[default]
 	Pending,
```
acatangiu, line 127: make sure to change code that was relying on old meaning of `Pending` (at the very least `self.is_ongoing()`)

pallets/ah-migrator/src/lib.rs
ggwpez: Unclear yet on what we need exactly, seems fine.

pallets/rc-migrator/src/lib.rs
ggwpez: Okay we can rename when it needs to be more specific later.

pallets/rc-migrator/src/lib.rs
ggwpez: Yea, it is not really clear what is initalized and when that stage will be done.

pallets/rc-migrator/src/lib.rs
ggwpez: Good for later.

pallets/rc-migrator/src/lib.rs
ggwpez: for later

ggwpez (approved):

pallets/rc-migrator/src/lib.rs
muharem: updated

PR #587, opened 2025-02-10
[AHM] Example Benchmark

ggwpez:
DNM

To run these benchmarks, you will need a modified version of `frame-omni-bencher` that can load
snapshots of the relay and asset hub. You can find it on branch `oty-ahm-omni-bencher` of the
SDK. Install it with
`cargo install --path substrate/utils/frame/omni-bencher --profile production`
```bash
frame-omni-bencher v1 benchmark pallet --runtime=target/release/wbuild/asset-hub-polkadot-runtime/asset_hub_polkadot_runtime.wasm --pallet "pallet-ah-migrator" --extrinsic "" --snap=ah-polkadot.snap --rc-snap=polkadot.snap
```

@muharem seems to work. Only downside is that we need to slightly alter the migrations to not immediately enqueue the XCM but rather return them. Possibly we can also somehow hack the ´SendXcm` implementation for benchmarks so that it requires no changes but i did not try that yet.


pallets/ah-migrator/src/benchmarking.rs
```diff
@@ -0,0 +1,105 @@
...
+
+#[benchmarks(where T: pallet_balances::Config)]
+mod benchmarks {
+	use super::*;
+
+	#[benchmark]
+	fn receive_multisigs() {
+		verify_snapshot::<T>();
+		let (messages, _cursor) = relay_snapshot(|| {
+			unwrap_no_debug(pallet_rc_migrator::multisig::MultisigMigrator::<T>::migrate_out_many(
```
ggwpez, line 44: We possibly need an alternative limit here, like 100 or so. Otherwise if we only pass the weight then we dont know how many it will be before hand.

ggwpez:
Already merged this into another MR

PR #593, opened 2025-02-17
state-trie-version for  AH Kusama and AH Polkadot

nkpar:

### DIFF
--- a/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
+++ b/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
@@ -1758,7 +1758,7 @@ impl pallet_state_trie_migration::Config for Runtime {
 
 	type MaxKeyLen = MigrationMaxKeyLen;
 }
-
+// Statemint State Migration Controller
 ord_parameter_types! {
 	pub const MigController: AccountId = AccountId::from(hex_literal::hex!("8458ed39dc4b6f6c7255f7bc42be50c2967db126357c999d44e12ca7ac80dc52"));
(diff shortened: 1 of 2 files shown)


system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1775,27 +1755,11 @@ impl pallet_state_trie_migration::Config for Runtime {
 	type MaxKeyLen = MigrationMaxKeyLen;
 }
 
-#[cfg(feature = "state-trie-version-1")]
 frame_support::ord_parameter_types! {
 	pub const MigController: AccountId = AccountId::from(hex_literal::hex!("8458ed39dc4b6f6c7255f7bc42be50c2967db126357c999d44e12ca7ac80dc52"));
 	pub const RootMigController: AccountId = AccountId::from(hex_literal::hex!("8458ed39dc4b6f6c7255f7bc42be50c2967db126357c999d44e12ca7ac80dc52"));
 }
 
-#[cfg(feature = "state-trie-version-1")]
```
kianenigma, line 1784: Okay about parameters: 

1. Signed migration is safe, and generally cannot go wrong. I think the account listed here is controlled by parity devops (let's double check -- will ping you about it), and can stay as-is.
2. The `ControlOrigin` IMO should not be account. If you inspect the code of the pallet, you will see that this account can trigger the automatic migration, and if this account makes a mistake the parachain is doomed. We should see what is the precedence here: If we have used this for the relay chains in the past, we can do the same here. If not, we should set `ControlOrigin` to be the fellowship. 
3. `MigrationMaxKeyLen` should be a smaller value. If we have used 512 on the relay chains, we should use a smaller value here.

PR #594, opened 2025-02-18
[AHM] Sovereign Account Translation

ggwpez:
To be merged into the AHM working branch. Quote from `accounts.md`:

# Account Migration

Accounts are migrated with all their balance, locks and reserves at the beginning of the Asset Hub
migration.

## User Impact

Users need to be aware that all of their funds will be moved from the Relay chain to the Asset Hub.
The Account ID will stay the same. This ensures that normal user accounts will be to control their
funds on Asset Hub.

## Sovereign Account Translation

For parachain sovereign accounts, it is not possible to just use the same account ID. The sovereign
account address of a parachain is calculated differently, depending on whether it is the account on
the Relay or a parachain (like Asset Hub).  

There are different kinds of sovereign accounts. In this context, we only focus on these parachain
sovereign accounts:
- On the Relay: derived from `"para" ++ para_id ++ 00..`
- On the Asset Hub and all other sibling parachains: derived from `"sibl" ++ para_id ++ 00..`

Our translation logic inverts the derivation and changes the prefix from `"para"` to `"sibl"` for
all accounts that match the pattern `"para" ++ para_id ++ 00..`. The full list of translated
accounts is in [this CSV file](./sovereign_account_translation.csv).

It is advised that parachains check that they can control their account on Asset Hub. They can also
forego this check if they do not need control thereof - for example when they are not holding any
funds on their relay sovereign account. However, please note that someone could still send funds to
that address before or after the migration.

Example for Bifrost: this is the [relay sovereign account](https://polkadot.subscan.io/account/13YMK2eeopZtUNpeHnJ1Ws2HqMQG6Ts9PGCZYGyFbSYoZfcm) and it gets translated to this [sibling sovereign account](https://assethub-polkadot.subscan.io/account/13cKp89TtYknbyYnqnF6dWN75q5ZosvFSuqzoEVkUAaNR47A).


pallets/rc-migrator/src/accounts.rs
```diff
@@ -542,4 +544,32 @@ impl<T: Config> AccountsMigrator<T> {
...
 		Weight::from_all(1)
 	}
+
+	/// Try to translate a Parachain sovereign account to the Parachain AH sovereign account.
+	///
+	/// Returns:
+	/// - `Ok(None)` if the account is not a Parachain sovereign account
+	/// - `Ok(Some((ah_account, para_id)))` with the translated account and the para id
+	/// - `Err(())` otherwise
+	pub fn try_translate_rc_sovereign_to_ah(acc: T::AccountId) -> Result<Option<(T::AccountId, u16)>, ()> {
```
ggwpez, line 554: This is the important function

pallets/rc-migrator/src/accounts.rs
```diff
@@ -542,4 +544,34 @@ impl<T: Config> AccountsMigrator<T> {
 		// TODO: define actual weight
 		Weight::from_all(1)
 	}
+
+	/// Try to translate a Parachain sovereign account to the Parachain AH sovereign account.
```
muharem, line 548: I would provide a link to the code base where these encodings originate

muharem (approved):

acatangiu (commented):
looks good so far! afaict still needs:

- [ ] equivalent function for translating derived SA (given a derivation path param)
- [ ] hooking this to the migration - do you plan to test every address if it is derived or use some static list of accounts that should go through this translation? I am guessing the latter.

ggwpez:
>do you plan to test every address if it is derived or use some static list of accounts that should go through this translation? I am guessing the latter.

We wanted to generate this static list as first step of the migration - automatically on-chain.  
I think this still makes sense, since there could be new parachains registered between the time when we generate the list and when the migration actually happens.

> equivalent function for translating derived SA (given a derivation path param)

Yes, I did not look into accounts that are derived from SAs. Going to do that in the next MR.

PR #599, opened 2025-02-21
Rename OnDemandAssignmentProvider to OnDemand on Kusama

DrW3RK:
Not sure why we have two different pallet names for this on Kusama and Polkadot. This ensures that on-demand chain docs are the same for both. If we have to preserve different names on Kusama and Polkadot, will work on the docs side instead.

### DIFF
--- a/relay/kusama/src/lib.rs
+++ b/relay/kusama/src/lib.rs
@@ -1839,7 +1839,7 @@ construct_runtime! {
 		ParaSessionInfo: parachains_session_info = 61,
 		ParasDisputes: parachains_disputes = 62,
 		ParasSlashing: parachains_slashing = 63,
-		OnDemandAssignmentProvider: parachains_on_demand = 64,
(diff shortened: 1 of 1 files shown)


relay/kusama/src/lib.rs
```diff
@@ -1839,7 +1839,7 @@ construct_runtime! {
 		ParaSessionInfo: parachains_session_info = 61,
 		ParasDisputes: parachains_disputes = 62,
 		ParasSlashing: parachains_slashing = 63,
-		OnDemandAssignmentProvider: parachains_on_demand = 64,
+		OnDemand: parachains_on_demand = 64,
```
tdimitrov, line 1842: This rename will require a migration of the storage too. 

I was working on this on the past but at that point there were conflicting migrations. I don't remember the details but I can dig up my old PR.

tdimitrov (changes requested):

PR #600, opened 2025-02-21
[AHM] Attempt to parse RC proposals with `XcmPallet::send`

bkontur:
Relates to: https://github.com/paritytech/polkadot-sdk/issues/7577

TDB:

Basically, it rewrites `(0, [Parachain(id)])` to `(1, [Parachain(id)])` and corrects `XcmPallet` -> `PolkadotXcm` mapping.

This could increase successful execution rate of migrated ongoing referenda.


bkontur:
cc: @ggwpez @muharem

muharem:
@bkontur do you have a referenda example with xcm send call?

bkontur:
> @bkontur do you have a referenda example with xcm send call?

Yes, sure, see the test at the end: https://github.com/polkadot-fellows/runtimes/pull/600/files#diff-9d44fd5501c89210fc95ef759c1cfc6e946714f8f75dcc905c9868d23abb40ddR374

system-parachains/asset-hubs/asset-hub-polkadot/src/migration.rs
```diff
@@ -331,6 +340,75 @@ impl RcToAhCall {
...
+					log::error!(
+						"Failed to decode `RcRuntimeCall::XcmPallet(RcXcmPalletCall::send` call to AH PolkadotXcm call - invalid dest: {:?} version!",
+						dest
+					);
+				})?;
+				let self_para_id: u32 = ParachainInfo::parachain_id().into();
+				match dest.unpack() {
+					(0, [Parachain(para_id)]) if para_id != &self_para_id => {
+						// The call is dedicated to the child parachain, so we change it to the sibling parachain (0->1).
+						let dest: Location = Location::new(1, [Parachain(*para_id)]);
```
bkontur, line 354: @muharem @ggwpez @acatangiu @seadanda 

This `pallet_xcm::Call::send` conversion requires that the target destination parachain has already accepted `AssetHub` as a new governance (root) before this call is executed on the migrated AssetHub governance.  

However, there is another potential fix, hack, or workaround to consider:  

Instead of rewriting child-to-sibling `parents: 0 -> 1`, we could wrap the `message` (`RcXcmPalletCall::send { dest, message }`) from the relay chain with another `pallet_xcm::Call::send(dest: Location::parent(), ...)`, like this:
```
// Send `message` back to the relaychain
RuntimeCall::PolkadotXcm(pallet_xcm::Call::<Runtime>::send {
       dest: Box::new(Location::parent()), // back to the relaychain
       Xcm(vec![
              UnpaidExecution {...},
              Transact { call: message.encode(), ...}
       ]),
})
```

So in the end, the call that was originally supposed to be sent from the relay chain to the `child` parachain will first be migrated to `AssetHub`. There, it will be wrapped with `pallet_xcm::Call::send` and sent back to the relay chain, which will then forward it to the `child` parachain. Since the `child` parachain already trusts `Location::parent()` as root, this approach should work.  

Essentially, with this roundtrip, **we don't need** to coordinate with other parachains to upgrade their XCM configuration before AHM starts.

However, this might not be worth the effort. When (start) moving governance to `AssetHub`, we need to ensure that other parachains trust `AssetHub` as governance/root, at least for system parachains.

Another possibility is to migrate only referenda that have **already been executed** on the relay chain and wait until the remaining ones are executed on the relaychain and then migrate them also to AH.

acatangiu:
I will say it again:

IMO this is a complicated problem with complicated fixes but with small impact.

Some referenda which are on-going during migration might fail after the migration because of this. IF there will actually be any that do, they can be simply repeated after the migration. It's not that big of a deal IMO.

I would not put too much resources into it. Let's figure out everything else then leave this as a "nice-to-have" and come back to it if we have time.

PR #612, opened 2025-02-28
[AHM] Pallet testing framework

ggwpez:
Main part looks like this. New pallet checks can be added to the tuple:

```rust
type RcChecks = (
	pallet_rc_migrator::preimage::PreimageChunkMigrator<Polkadot>,
	pallet_rc_migrator::indices::IndicesMigrator<Polkadot>,
	// other pallets go here
);

type AhChecks = (
	pallet_rc_migrator::preimage::PreimageChunkMigrator<AssetHub>,
	pallet_rc_migrator::indices::IndicesMigrator<AssetHub>,
	// other pallets go here
);

#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
async fn pallet_migration_works() {
	let Some((mut rc, mut ah)) = load_externalities().await else { return };

	// Pre-checks on the Relay
	let rc_pre_payload = rc.execute_with(RcChecks::pre_check);

	// Pre-checks on the Asset Hub
	let ah_pre_payload = ah.execute_with(|| {
		let payload = AhChecks::pre_check(rc_pre_payload.clone());
		payload
	});

	// Migrate the Relay Chain
	let dmp_messages = rc_migrate(&mut rc);

	// Post-checks on the Relay
	rc.execute_with(|| RcChecks::post_check(rc_pre_payload.clone()));

	// Migrate the Asset Hub
	ah_migrate(&mut ah, dmp_messages);

	// Post-checks on the Asset Hub
	ah.execute_with(|| {
		AhChecks::post_check(rc_pre_payload, ah_pre_payload);
	});
}
```

Payload dependencies for custom checks are passed:
- rc_pre -> RcPrePayload
- ah_pre(RcPrePayload) -> AhPrePayload
- rc_post(RcPrePayload)
- ah_post(RcPrePayload, AhPrePayload)


pallets/rc-migrator/src/indices.rs
```diff
@@ -67,16 +69,12 @@ impl<T: Config> PalletMigration for IndicesMigrator<T> {
 				break;
 			}
 
-			let mut iter = match inner_key {
```
ggwpez, line 70: @muharem i removed the stored key here and just use next now, since otherwise it seems that not all keys are removed.  
Issue with this is that it is _really_ slow. Maybe we need a two step approach where we first iterate key like we usually do and then a second sweep phase to get the ones that the first phase missed.

integration-tests/ahm/src/mock.rs
```diff
@@ -113,23 +110,21 @@ pub fn enqueue_dmp(msgs: Vec<InboundDownwardMessage>) {
 // both the DMP messages sent from the relay chain to asset hub, which will be used to perform the
 // migration, and the relay chain payload, which will be used to check the correctness of the
 // migration process.
-pub fn rc_migrate<RcMigrator: RcPalletMigrationChecks>(
-	mut relay_chain: RemoteExternalities<PolkadotBlock>,
-) -> (Vec<InboundDownwardMessage>, RcMigrator::RcPayload) {
+pub fn rc_migrate(
```
muharem, line 113: should we drop these abstractions? the `rc_migrate` and `ah_migrate` and have these code just within the `pallet_migration_works` function?

integration-tests/ahm/src/tests.rs
```diff
@@ -43,17 +44,41 @@ use xcm_emulator::ConvertLocation;
...
-		pallet_ah_migrator::preimage::PreimageMigrationCheck<AssetHub>,
-	>(ah, rc_payload, dmp_messages);
+async fn pallet_migration_works() {
+	let Some((mut rc, mut ah)) = load_externalities().await else { return };
+
+	// Pre-checks on the Relay
+	let rc_pre_payload = rc.execute_with(RcChecks::pre_check);
+
+	// Pre-checks on the Asset Hub
+	let ah_pre_payload = ah.execute_with(|| AhChecks::pre_check(rc_pre_payload.clone()));
```
muharem, line 67: how about not running checks if `START_STAGE` is provided? we consider it as a debug no checks run?

pallets/rc-migrator/src/indices.rs
muharem: with the key it just skips some?

integration-tests/ahm/src/mock.rs
ggwpez: I think its fine now, the main testing function looks quite clean. We can switch these functions to accept tokio channels in the if we want to run it in parallel.

pallets/rc-migrator/src/indices.rs
ggwpez: @cheme

ggwpez: I can try to write a minimal reproducible test it its not clear from this ^.

cheme: Good to hear `next_key` work as intended.

Just looking quickly at the code there is this line:
```
			log::error!("key failed to decode at {:?}: {:?}", self.previous_key, e);
```
that happen when we can't 
```
match (self.closure)(raw_key_without_prefix) {
```
but I don't know this code, yet when it fails it just log the error and then continue, maybe related. 

But TBH I don't see what could go wrong here (as long as it only run on one block.

It may likely be related with the way the `from_key` is build (iter use reverse on it to get original key, at the same time the previous_key stored in the iterator is meant to be the full hashed key so not too sure about this reverse call (also I expect Account :: iter_from_key to rebuild the full hashed key from index, then Hasher::reverse is just to call the closure on the key without prefix, not sure what it's used for).

PR #614, opened 2025-03-03
Fix scheduler weight and add test

ggwpez:
- **Add test to collectives pallet**
- **Manually edit scheduler weight**

## Investigation

The Rust compiler changed how the `stringify!` macro formats paths on [Aug 12th '24](https://github.com/rust-lang/rust/issues/128992#issuecomment-2282908325). I assume that this broke the V1 benchmarking [here](https://github.com/paritytech/polkadot-sdk/blob/7ecf3f757a5d6f622309cea7f788e8a547a5dce8/substrate/frame/benchmarking/src/v1.rs#L1011). We had updated the scheduler pallet to V2 syntax (which is [not affected](https://github.com/paritytech/polkadot-sdk/blob/df99fb9431a579589c1c87832222c9551b5c4f7c/substrate/frame/support/procedural/src/benchmark.rs#L565)) on [Nov 20th '24](https://github.com/paritytech/polkadot-sdk/pull/6292). We did not back-port this since there was no apparent reason for it.  
The is why it seemingly fixed it self on SDK master but not in the runtimes repo (since that is using V1 scheduler benchmarking).  
Another indication for this is the fact that it did work on the [whitelist pallet](https://github.com/polkadot-fellows/runtimes/blob/6b85bf6adb427942976648e6d235e0169dfced16/relay/polkadot/src/weights/pallet_whitelist.rs#L85), which is on V2 much longer but used the same [custom pov mode](https://github.com/paritytech/polkadot-sdk/blob/b76e91acc953e682b3ddcfc45ecacaaf26c694a1/substrate/frame/whitelist/src/benchmarking.rs#L68).

Adding some sanity checks to prevent this in the future here https://github.com/paritytech/polkadot-sdk/pull/7785/


system-parachains/collectives/collectives-polkadot/src/weights/pallet_scheduler.rs
```diff
@@ -82,24 +82,25 @@ impl<T: frame_system::Config> pallet_scheduler::WeightInfo for WeightInfo<T> {
 		Weight::from_parts(4_560_000, 0)
 			.saturating_add(Weight::from_parts(0, 0))
 	}
+	// FAIL-CI I copied this from SDK master now for testing
```
ggwpez, line 85:
```suggestion
```

bkchr (approved):

xlc (approved):

ordian (approved):

acatangiu (approved):

bkontur (approved):

PR #615, opened 2025-03-05
[AHM] Vesting test

eugypalu:
Add vesting tests with pre/post migration checks


pallets/ah-migrator/src/vesting.rs
```diff
@@ -86,3 +89,64 @@ pub mod alias {
...
+#[cfg(feature = "std")]
+impl<T: Config> crate::types::AhMigrationCheck for VestingMigrator<T> {
+	type RcPrePayload =
+		Vec<(Vec<u8>, Vec<BalanceOf<T>>, Vec<GenericVestingInfo<BlockNumberFor<T>, BalanceOf<T>>>)>;
+
+	type AhPrePayload =
+		Vec<(Vec<u8>, Vec<BalanceOf<T>>, Vec<GenericVestingInfo<BlockNumberFor<T>, BalanceOf<T>>>)>;
+
+	fn pre_check(_: Self::RcPrePayload) -> Self::AhPrePayload {
+		pallet_vesting::Vesting::<T>::iter()
```
ggwpez, line 102: Actually there should be no vesting schedules on the asset hub before the migration starts.  
Can you please check that it is empty?

pallets/ah-migrator/src/vesting.rs
```diff
@@ -86,3 +89,64 @@ pub mod alias {
...
+
+		let all_post: BTreeMap<
+			Vec<u8>,
+			(Vec<BalanceOf<T>>, Vec<GenericVestingInfo<BlockNumberFor<T>, BalanceOf<T>>>),
+		> = pallet_vesting::Vesting::<T>::iter()
+			.map(|(who, schedules)| {
+				let balances: Vec<BalanceOf<T>> = schedules.iter().map(|s| s.locked()).collect();
+				let vesting_info: Vec<GenericVestingInfo<BlockNumberFor<T>, BalanceOf<T>>> =
+					schedules
+						.iter()
```
ggwpez, line 139: It can be a bit easier to read if you first create a vector and then `push` into that, but it doesnt really matter for test code.

pallets/rc-migrator/src/vesting.rs
```diff
@@ -94,3 +95,65 @@ impl<T: Config> PalletMigration for VestingMigrator<T> {
...
+							locked: s.locked(),
+							starting_block: s.starting_block(),
+							per_block: s.per_block(),
+						})
+						.collect();
+				(who.encode(), (balances, vesting_info))
+			})
+			.collect();
+
+		assert_eq!(pre_map, current_map, "Vesting data mismatch after migration");
```
ggwpez, line 157: Is this just the same test as in the AH post check?! It should probably check that the data is all gone from the relay.  
Although there could be a bug with the data being removed, hence why we need these tests.  
But it does not make sense to test the same thing twice.

pallets/ah-migrator/src/vesting.rs
eugypalu: Thanks, fixed

pallets/rc-migrator/src/vesting.rs
eugypalu: My idea was to check that the vesting currently present in RC is equal to RcPrePayload. It's a sort of double check to verify that the migrated RcPrePayload is complete and that there were no issues beforehand, making it slightly different from the one in AH. I could be completely wrong, but looking at the vesting migration, it didn't seem like the vesting was being removed, which is why I didn't check if it was empty.

muharem: we probably just forgot to remove the items from the storage on RC. @ggwpez should tell for sure. other than that this PR look good to me.

pallets/ah-migrator/src/vesting.rs
```diff
@@ -86,3 +89,58 @@ pub mod alias {
...
+		assert!(
+			vesting_schedules.is_empty(),
+			"Vesting schedules should be empty before migration starts"
+		);
+	}
+
+	fn post_check(rc_pre_payload: Self::RcPrePayload, _: Self::AhPrePayload) {
+		use std::collections::BTreeMap;
+
+		let rc_pre: BTreeMap<
```
ggwpez, line 111: I think you can also directly put the BTreeMap as `RcPrePayload` instead of the vector?

ggwpez (approved):
Going to merge if the ahm-test turns green in CI

PR #618, opened 2025-03-06
[AHM] Improve weight accounting for accounts migration 

muharem:
mean to be merged into AHM dev branch 

Improve weight accounting for accounts migration.

Please read code and comments for more info.


integration-tests/ahm/src/tests.rs
```diff
@@ -204,3 +204,40 @@ async fn print_sovereign_account_translation() {
...
+					.and_modify(|count| *count += 1)
+					.or_insert(1);
+			}
+		}
+	});
+
+	/*
+	RC Polkadot snapshot from 2025-01-24:
+		total_count ~ 1_434_995
+		total_liquid_count ~ 1_373_890
```
muharem, line 240: here we can see that we definitely need to weight fully liquid accounts

PR #625, opened 2025-03-17
Snowbridge - Integration test for 2-hop transfer from Ethereum to AHP to AHK

claravanstaden:
Adds a test to send Weth bridge from Ethereum to Polkadot AH to Kusama AH, in a second transaction. This shows how we could support bridging to Kusama in a 2-step fashion until Snowbridge V2 is released.


claravanstaden:
> why is this mandatory? the test should still have same value even without sending full amount of Weth, no?

@acatangiu the test is still valid, but practically I think users might want to send the full value to Kusama. So the test is fine but ideally https://github.com/paritytech/polkadot-sdk/pull/7243 should be on Polkadot before we release the 2-hop bridge to Kusama feature.

> also maybe you still send full Weth if you prefund that account with DOT ED so that all Weth can be withdrawn?

I prefunded the[ account with DOT](https://github.com/polkadot-fellows/runtimes/pull/625/files#diff-46a0e8ad1ccebb1fd2c05ed3202f1d08efec5f64c55fafc3504ca8ec462592f0R1337), but it didn't seem to make a difference. The test still fails with `FailedToTransactAsset("Account that is desired to remain would die")` if I send all the WETH.

integration-tests/emulated/chains/parachains/assets/asset-hub-kusama/src/genesis.rs
```diff
@@ -94,6 +103,26 @@ pub fn genesis() -> Storage {
 					false,
 					ED,
 				),
+				// Ether
+				(
+					Location::new(2, [GlobalConsensus(EthereumNetwork::get())]),
+					AssetHubPolkadotSovereignAccount::get(),
+					true,
+					ED,
```
acatangiu, line 111:
```suggestion
					MIN_ETHER_BALANCE,
```

integration-tests/emulated/helpers/src/lib.rs
```diff
@@ -404,3 +406,69 @@ macro_rules! test_chain_can_claim_assets {
 		}
 	};
 }
+
+/// note: $asset needs to be prefunded outside this function
+#[macro_export]
+macro_rules! create_pool_with_native_on {
```
acatangiu, line 412: do we not have this defined somewhere in the SDK so we can reuse instead of duplicating code here?

claravanstaden: It is in the BridgeHub Westend integration tests crate: https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/parachains/integration-tests/emulated/tests/bridges/bridge-hub-westend/src/tests/mod.rs#L126

So I thought it would be weird to use it here. I think it should be moved to a common crate in the `polkadot-sdk` though, will make a PR and add a comment here to replace it eventually.

claravanstaden: https://github.com/paritytech/polkadot-sdk/pull/8289

acatangiu (approved):

raymondkfcheung (approved):

franciscoaguirre (approved):

integration-tests/emulated/chains/parachains/assets/asset-hub-kusama/src/genesis.rs
```diff
@@ -37,7 +41,14 @@ frame_support::parameter_types! {
 				xcm::v4::Junction::GeneralIndex(penpal_emulated_chain::TELEPORTABLE_ASSET_ID.into()),
 			]
 		);
+	 pub UniversalLocation: InteriorLocation = [GlobalConsensus(Kusama), Parachain(PARA_ID),
+   ].into();
```
serban300, line 45: Nit: Looks badly formatted. Doesn't it fit on a single line ?

integration-tests/emulated/chains/parachains/assets/asset-hub-kusama/src/genesis.rs
```diff
@@ -92,6 +103,29 @@ pub fn genesis() -> sp_core::storage::Storage {
...
+				),
+				// Weth
+				(
+					xcm::v4::Location::new(
+						2,
+						[
+							xcm::v4::Junction::GlobalConsensus(EthereumNetworkXcmV4::get()),
+							xcm::v4::Junction::AccountKey20 { network: None, key: WETH },
+						],
+					),
```
serban300, line 124: Can we put this and `EthereumNetworkXcmV4` in `common.rs` ? I see that it is duplicated here and in `asset-hub-polkadot`

integration-tests/emulated/tests/bridges/bridge-hub-polkadot/src/tests/mod.rs
```diff
@@ -73,7 +74,7 @@ pub(crate) fn weth_at_asset_hubs() -> xcm::v4::Location {
 			xcm::v4::Junction::GlobalConsensus(xcm::v4::NetworkId::Ethereum {
 				chain_id: snowbridge::CHAIN_ID,
 			}),
-			xcm::v4::Junction::AccountKey20 { network: None, key: snowbridge::WETH },
+			xcm::v4::Junction::AccountKey20 { network: None, key: WETH },
 		],
 	)
```
serban300, line 79: If we move 

```
xcm::v4::Location::new(
						2,
						[
							xcm::v4::Junction::GlobalConsensus(EthereumNetworkXcmV4::get()),
							xcm::v4::Junction::AccountKey20 { network: None, key: WETH },
						],
					),
```

into common, we can remove this duplicate as well.

claravanstaden: Addressed in https://github.com/polkadot-fellows/runtimes/pull/625/commits/748af5085e2289813ea502301c4c7173e33633ae.

integration-tests/emulated/chains/parachains/assets/asset-hub-kusama/src/genesis.rs
claravanstaden: Addressed in https://github.com/polkadot-fellows/runtimes/pull/625/commits/748af5085e2289813ea502301c4c7173e33633ae.

serban300 (approved):

PR #628, opened 2025-03-18
[AHM] Fix accounts migration failures

muharem:
The PR addresses the following issues in account migrations:

Problem 1:
An account has a free balance below the existential deposit (ED) but a reserved balance above ED. This causes the free balance to be burned when releasing the hold, additionally leading to a panic in the test environment due to an existing debug_assert.

Solution:
The hold is partially released first to meet the ED requirement before releasing the remaining hold.

Example: https://polkadot.subscan.io/account/16HwqWaypD6acNeUu3faVGVowG86UvG1ieLrXZSWVxsaNyLS

Problem 2:
An account has no free balance but a held balance above ED. This results in an error when attempting to recover the hold on Asset Hub (AH).

Solution:
A provider reference is added on AH if the account lacks a free balance to meet the ED requirement.

Example: https://polkadot.subscan.io/account/167NvqSJt5wKFDR4RFYkXA1piJfEKVH8MG4fs4nX9MnbNPQa

Problem 3:
An account’s total balance (free + reserved) is below ED, causing most fungible implementation operations to fail.

Solution:
Such accounts are ignored and not migrated.

Example (currently only one account): https://polkadot.subscan.io/account/15QBmCuy3aRxn8xuzfaRvzaLdawsGbPmttbcvBU8q71kEb7a

### DIFF
--- a/pallets/rc-migrator/src/accounts.rs
+++ b/pallets/rc-migrator/src/accounts.rs
@@ -294,7 +294,7 @@ impl<T: Config> AccountsMigrator<T> {
 	/// Migrate a single account out of the Relay chain and return it.
 	///
 	/// The account on the relay chain is modified as part of this operation.
-	fn withdraw_account(
+	pub fn withdraw_account(
 		who: T::AccountId,
 		account_info: AccountInfoFor<T>,
 		rc_weight: &mut WeightMeter,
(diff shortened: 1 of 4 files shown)


pallets/rc-migrator/src/accounts.rs
```diff
@@ -388,19 +374,56 @@ impl<T: Config> AccountsMigrator<T> {
...
+			let IdAmount { id, amount } = hold.clone();
+			let free = <T as Config>::Currency::balance(&who);
+
+			// When the free balance is below the minimum balance and we attempt to release a hold,
+			// the `fungible` implementation would burn the entire free balance while zeroing the
+			// hold. To prevent this, we partially release the hold just enough to raise the free
+			// balance to the minimum balance, while maintaining some balance on hold. This approach
+			// prevents the free balance from being burned.
+			// This scenario causes a panic in the test environment - see:
+			// https://github.com/paritytech/polkadot-sdk/blob/35e6befc5dd61deb154ff0eb7c180a038e626d66/substrate/frame/balances/src/impl_fungible.rs#L285
```
ggwpez, line 391: We probably need to give the whole account migration another throughout review at the end, so am just going to rely on tests for now and merge.

ggwpez (approved):

PR #632, opened 2025-03-19
[AHM] bags_list and conviction_voting tests 

eugypalu:
Add pre/post migration checks to bags_list and conviction_voting

### DIFF
--- a/pallets/ah-migrator/src/conviction_voting.rs
+++ b/pallets/ah-migrator/src/conviction_voting.rs
@@ -18,7 +18,7 @@ use crate::*;
 use frame_support::traits::{ClassCountOf, DefensiveTruncateFrom};
 use pallet_conviction_voting::TallyOf;
 use pallet_rc_migrator::conviction_voting::{
-	alias, RcConvictionVotingMessage, RcConvictionVotingMessageOf,
+	alias, ConvictionVotingMigrator, RcConvictionVotingMessage, RcConvictionVotingMessageOf,
 };
 
 impl<T: Config> Pallet<T> {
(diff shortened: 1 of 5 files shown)


pallets/ah-migrator/src/conviction_voting.rs
```diff
@@ -70,3 +70,39 @@ impl<T: Config> Pallet<T> {
...
+		for (account_id, balance_per_class) in pallet_conviction_voting::ClassLocksFor::<T>::iter()
+		{
+			let balance_per_class: Vec<_> = balance_per_class.into_iter().collect();
+			ah_messages
+				.push(RcConvictionVotingMessage::ClassLocksFor(account_id, balance_per_class));
+		}
+
+		assert_eq!(
+            rc_pre_payload, ah_messages,
+            "Conviction voting data mismatch: Asset Hub data differs from original Relay Chain data"
```
muharem, line 105: looks like something wrong with formatting. check CIs, we use a specific version + rust nightly for fmt

pallets/rc-migrator/src/conviction_voting.rs
```diff
@@ -250,3 +250,38 @@ pub mod alias {
...
+
+impl<T: Config> crate::types::RcMigrationCheck for ConvictionVotingMigrator<T> {
+	type RcPrePayload = Vec<RcConvictionVotingMessageOf<T>>;
+
+	fn pre_check() -> Self::RcPrePayload {
+		let mut messages = Vec::new();
+
+		// Collect VotingFor
+		for (account_id, class, voting) in alias::VotingFor::<T>::iter() {
+			messages.push(RcConvictionVotingMessage::VotingFor(account_id, class, voting));
```
muharem, line 262: for conviction voting we filter zero votings, this test should fail on ah post check

pallets/ah-migrator/src/conviction_voting.rs
```diff
@@ -70,3 +70,39 @@ impl<T: Config> Pallet<T> {
...
+			alias::VotingFor::<T>::iter().next().is_none(),
+			"VotingFor should be empty before migration starts"
+		);
+		assert!(
+			pallet_conviction_voting::ClassLocksFor::<T>::iter().next().is_none(),
+			"ClassLocksFor should be empty before migration starts"
+		);
+	}
+
+	fn post_check(rc_pre_payload: Self::RcPrePayload, _: Self::AhPrePayload) {
```
muharem, line 89: we should be checking everywhere that these payloads are not empty

pallets/ah-migrator/src/conviction_voting.rs
ggwpez: `cargo +nightly-2024-09-11-aarch64-apple-darwin fmt` is what i am using on macos.

pallets/rc-migrator/src/conviction_voting.rs
eugypalu: You're right! Fixed, thanks

ggwpez (approved):
LGTM, will merge soon.

Can you please add the tested properties to the second table here? https://docs.google.com/spreadsheets/d/1Mjb7iyxINSLMI8D7RhMKceYiCj65a3-h7zCc1j_gaa8/edit?gid=0#gid=0

It is a bit of an experiment, but i would hope that it helps us to track where we are missing test coverage

PR #641, opened 2025-03-21
[AHM] Rebase

ggwpez:
Changes:
- Rebase and delete some stuff that currently wont compile, can do more throughout cleanup afterwards


.github/workflows/test.yml
```diff
@@ -132,88 +122,113 @@ jobs:
...
         run: |
-          df -h
-          sudo apt -y autoremove --purge
-          sudo apt -y autoclean
-          sudo rm -rf /usr/share/dotnet
-          sudo rm -rf /opt/ghc
-          sudo rm -rf "/usr/local/share/boost"
-          sudo rm -rf "$AGENT_TOOLSDIRECTORY"
-          df -h
+          curl -sL 'https://sole-amaranth-pinniped.myfilebase.com/ipfs/QmTdrtye7Vt2xDybZHR3Xjf8kbxXkSfk54AUz2DuKKGmrG?download=true&filename=try-runtime-snap-polkadot-25075273.snap.lz4' -o polkadot.snap.lz4
```
muharem, line 184: may be some general name so that we can just replace the file by newer version?

muharem (approved):

.github/workflows/test.yml
ggwpez: This is far from ideal - we need some S3 bucket or something that we can update with a single command and that also replaces it here.  
I added this just as a first version, we can improve it then.

PR #642, opened 2025-03-21
[AHM] Run AHM test in CI

ggwpez:


muharem (approved):

.github/workflows/test.yml
```diff
@@ -183,7 +228,7 @@ jobs:
       - name: Build
         run: cargo test -p chain-spec-generator --release --locked -q --features=runtime-benchmarks
         env:
-          RUSTFLAGS: "-C debug-assertions -D warnings"
+          RUSTFLAGS: "-C debug-assertions -A warnings" # FAIL-CI AHM
```
bkontur, line 231: @ggwpez off-topic, fellows repo does not supprot `FAIL-CI` check as polkadot-sdk, I think it could be useful, several times I was thinking about porting this features to the fellows or is there a reason why it was not added to the fellows, wdyt?

PR #645, opened 2025-03-26
[CI] Align ubuntu version for runners and use fixed version

bkontur:
Addressing comment: https://github.com/polkadot-fellows/runtimes/pull/637/files#r2012149573

Actually, some workflows use `runs-on: ubuntu-22.04`, while others use `runs-on: ubuntu-latest`. Additionally, there are several other places where either `ubuntu-22.04` or `ubuntu-latest` is used.  

## **Open questions**  

- [ ] Decide whether to use `ubuntu-22.04` or a higher version like `ubuntu-24.04` ([Ubuntu Releases](https://releases.ubuntu.com/))  
- [ ] Consider an easier configuration method:  
  - Using `runs-on: ${{ vars.UBUNTU_VERSION }}` (but this requires repo admin access to modify variables, which is probably not ideal)  
  - Using `$GITHUB_ENV` and an `.env` file (though, as I understand, this only works at the step level, not the job level)  
  - Using a matrix strategy with several versions (probably not needed)
  - Sticking to manual updates (maybe that's the best option. Anyway, how often do we actually need to change it?)

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   set-auto-merge:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     environment: master
     timeout-minutes: 10
     # Important! This forces the job to run only on comments on Pull Requests that starts with '/merge'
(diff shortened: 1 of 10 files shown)


.github/workflows/auto-merge.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   set-auto-merge:
-    runs-on: ubuntu-latest
```
bkchr, line 10: Please use latest everywhere. This is for example the label used by our self hosted runners.

PR #655, opened 2025-03-31
Improve account balances integration tests

re-gius:
- Adding extensive testing coverage for the balances kept on the relay chain
- Adding tests on checking account balances before and after migration

**To Do**
Complete checks on total issuance before and after the migration

### DIFF
--- a/pallets/ah-migrator/src/account.rs
+++ b/pallets/ah-migrator/src/account.rs
@@ -188,7 +188,7 @@ impl<T: Config> Pallet<T> {
 #[cfg(feature = "std")]
 impl<T: Config> crate::types::AhMigrationCheck for AccountsMigrator<T> {
 	// rc_total_issuance_before
-	type RcPrePayload = BalanceOf<T>;
+	type RcPrePayload = (BalanceOf<T>, BalanceOf<T>);
 	type AhPrePayload = ();
 
 	/// Run some checks on asset hub before the migration and store intermediate payload.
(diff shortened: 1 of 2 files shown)


re-gius:
Some identities do not match https://github.com/paritytech/polkadot-sdk/issues/7622 , so further investigations are needed

re-gius:
Removed checks that are not supported yet

pallets/ah-migrator/src/account.rs
```diff
@@ -188,16 +188,44 @@ impl<T: Config> Pallet<T> {
...
 	fn pre_check(_: Self::RcPrePayload) -> Self::AhPrePayload {
 		let check_account = T::CheckingAccount::get();
-		let _checking_balance = <T as Config>::Currency::total_balance(&check_account);
-		//assert_eq!(checking_balance, 100000000); // TODO Adrian
+		let checking_balance = <T as Config>::Currency::total_balance(&check_account);
+		assert_eq!(
+			checking_balance,
+			<T as Config>::Currency::minimum_balance(),
+			"Checking account on Asset Hub should have only existential deposit before migration."
+		);
```
acatangiu, line 204: This is only the case on Polkadot

re-gius: So in general we cannot say anything about that account before migration?

ggwpez (approved):

PR #657, opened 2025-04-01
[AHM] add sudo behind feature flag

pepoviola:

Add `sudo` behind `zombie-bite-sudo` feature to use with zombie-bite for migration tests.

cc: @ggwpez

### DIFF
--- a/relay/polkadot/src/genesis_config_presets.rs
+++ b/relay/polkadot/src/genesis_config_presets.rs
@@ -189,6 +189,10 @@ fn polkadot_testnet_genesis(
 			"forceEra": Forcing::NotForcing,
 			"slashRewardFraction": Perbill::from_percent(10),
 		},
+		#[cfg(feature="zombie-bite-sudo")]
+		"sudo": {
+			"key": Some(_root_key)
+		},
 		"babe": {
 			"epochConfig": Some(BABE_GENESIS_EPOCH_CONFIG),
 		},
--- a/relay/polkadot/src/lib.rs
+++ b/relay/polkadot/src/lib.rs
@@ -20,6 +20,9 @@
 // `construct_runtime!` does a lot of recursion and requires us to increase the limit to 256.
 #![recursion_limit = "512"]
 
+// AHM enable to easily add sudo key in genesis
+#![feature(stmt_expr_attributes)]
+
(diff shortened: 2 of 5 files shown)


relay/polkadot/src/lib.rs
```diff
@@ -1666,13 +1669,23 @@ construct_runtime! {
 		Mmr: pallet_mmr = 201,
 		BeefyMmrLeaf: pallet_beefy_mmr = 202,
 
+		#[cfg(feature="zombie-bite-sudo")]
```
ggwpez, line 1672:
```suggestion
		#[cfg(feature="zombie-bite-sudo")] // FAIL-CI
```
so that we wont forget to remove it.

ggwpez (approved):

relay/polkadot/src/genesis_config_presets.rs
```diff
@@ -135,6 +135,73 @@ fn default_parachains_host_configuration() -> HostConfiguration<polkadot_primiti
 	}
 }
 
+#[cfg(feature = "zombie-bite-sudo")] // FAIL-CI
+#[allow(clippy::type_complexity)]
+fn polkadot_testnet_genesis(
```
ggwpez, line 140: We dont really use genesis anyway? Like we just overwrite it in ZB, or?

relay/polkadot/src/lib.rs
```diff
@@ -1673,6 +1673,137 @@ construct_runtime! {
 	}
 }
 
+#[cfg(feature = "zombie-bite-sudo")] // FAIL-CI
```
ggwpez, line 1676: So we enable this feature by default now?

pepoviola: no, is not enabled by default we should add to the just _recipe_ in ahm-dryrun.

relay/polkadot/src/genesis_config_presets.rs
pepoviola: That correct, we can remove this.

x3c41a (approved):

ggwpez:
Could be problematic when we add / remove pallets now, but okay we can probably switch to construct-runtime-v2 soon.

PR #659, opened 2025-04-02
Enable Async Backing for Kusama Asset Hub

muharem:
Enable Async Backing for Kusama Asset Hub.

Main configuration changes:
- `UNINCLUDED_SEGMENT_CAPACITY` from 1 to 3
- `MILLISECS_PER_BLOCK` from 12_000 to 6000
- `MAXIMUM_BLOCK_WEIGHT` from 0.5 sec to 2 sec
- `NORMAL_DISPATCH_RATIO` from 75% to 85%, but still 2x (0.125 -> 0.3 sec) increase for operational transactions

NOTE: `SLOT_DURATION` stays at 12 sec.

The Kusama People Chain, which is already running with async backing, has also migrated to use the constant parameters shared with the Kusama Asset Hub.

### Motivation
This PR enables async backing specifically for the Asset Hub to support the upcoming Asset Hub migration. Only Asset Hub is included, as this work is part of the ongoing migration effort.

### Impact:
- block time is 6 seconds (previously 12 seconds);
- `pallet_collator_selection`: collators that authored blocks before the migration to async backing, may be removed for being stale more quickly since `KickThreshold` used timings based on 12 sec block time;
- `pallet_proxy`: announcements for the proxies with specified `delay`, may become active earlier than expected. Currently, there are no proxies with a specified `delay` exists;
- `pallet_nfts`: collections with defined `mint.start_block` and `mint.end_block` may start and end their minting period earlier. At present, there are no NFT collections with `start_block` or `end_block` set.

Once SDK 2503 is integrated into the runtimes, the last two pallets can be configured to use the more time-accurate Relay Chain block provider.


system-parachains/constants/src/kusama.rs
```diff
@@ -23,6 +23,16 @@ pub mod consensus {
...
 	pub const RELAY_CHAIN_SLOT_DURATION_MILLIS: u32 = 6000;
+
+	/// Parameters enabling async backing functionality.
+	///
+	/// Once all system chains have migrated to the new async backing mechanism, the parameters
+	/// in this namespace will replace those currently defined in `super::*`.
+	pub mod fast {
+		/// Maximum number of blocks simultaneously accepted by the Runtime, not yet included into
+		/// the relay chain.
+		pub const UNINCLUDED_SEGMENT_CAPACITY: u32 = 3;
```
skunert, line 34: Why is this single constant in `kusama.rs`, instead of bundled with the other constants?

system-parachains/constants/src/lib.rs
```diff
@@ -57,3 +57,42 @@ pub const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
...
+	pub const HOURS: BlockNumber = MINUTES * 60;
+	pub const DAYS: BlockNumber = HOURS * 24;
+
+	/// We assume that ~5% of the block weight is consumed by `on_initialize` handlers. This is
+	/// used to limit the maximal weight of a single extrinsic.
+	pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(5);
+
+	/// We allow `Normal` extrinsics to fill up the block up to 85%, the rest can be used by
+	/// Operational  extrinsics.
+	pub const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(85);
```
skunert, line 91: Saw it in the description, but why now 85 instead of 75?

system-parachains/constants/src/kusama.rs
```diff
@@ -23,6 +23,16 @@ pub mod consensus {
 	pub const BLOCK_PROCESSING_VELOCITY: u32 = 1;
 	/// Relay chain slot duration, in milliseconds.
 	pub const RELAY_CHAIN_SLOT_DURATION_MILLIS: u32 = 6000;
+
+	/// Parameters enabling async backing functionality.
+	///
+	/// Once all system chains have migrated to the new async backing mechanism, the parameters
+	/// in this namespace will replace those currently defined in `super::*`.
+	pub mod fast {
```
skunert, line 31: We could just call it `async_backing` or something, at least would be more clear to me.

skunert (commented):
Overall changes look good, I did a quick verification locally that the chain is indeed producing at 6s interval with these parameters.

system-parachains/constants/src/kusama.rs
muharem: I may not understood the comment, but this constant here because it overloads the `super:: UNINCLUDED_SEGMENT_CAPACITY` and stays under `consensus` namespace, later we can drop this module as mentioned above.

system-parachains/constants/src/lib.rs
muharem: before it was 75%, ~ 0.125 of a sec. with the same ratio, its gonna be 0.5 sec for operational, which seems to be quite high, since its allocated in every block, even if there is no operational work. With 85% we have 0.3sec which is 2 times more than previous capacity anyway.

system-parachains/constants/src/kusama.rs
muharem: renamed

system-parachains/constants/src/kusama.rs
skunert: The question was why this is not in `system-parachains/constants/src/lib.rs`, because it is just a normal parachain config parameter and not really kusama specific.

skunert (approved):
Changes to enable async backing look good. 

The points listed unter "Impact" seem acceptable to me, but I am not super familiar with the pallet internals.

system-parachains/constants/src/kusama.rs
muharem: I see, just bc the constant it's overloading is here. I can update it when we enable async backing everywhere.

system-parachains/constants/src/lib.rs
```diff
@@ -57,3 +57,42 @@ pub const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
...
+
+	/// This determines the average expected block time that we are targeting.
+	///
+	/// Blocks will be produced at a minimum duration defined by `SLOT_DURATION`. `SLOT_DURATION` is
+	/// picked up by `pallet_timestamp` which is in turn picked up by `pallet_aura` to implement `fn
+	/// slot_duration()`.
+	///
+	/// Change this to adjust the block time.
+	pub const MILLISECS_PER_BLOCK: u64 = 6_000;
+	pub const SLOT_DURATION: u64 = MILLISECS_PER_BLOCK;
```
eskimor, line 78: We should keep this at 12s. Reduces stress once we enable elastic scaling. (Less frequent handovers)

muharem: You mean only `SLOT_DURATION`? We keep 6s for `MILLISECS_PER_BLOCK`? Should it also be readjusted for People Chain? Now its set to 6s.

bkchr: > We should keep this at 12s. Reduces stress once we enable elastic scaling. (Less frequent handovers)

I'm for going down to 6s. If one guy being down, we only loose 6s and not 12s.

eskimor: We can do so later, if we think it is worthwhile. Going up again is less straight forward.

Risk wise it also seems more likely that we run into a situation where a longer slot duration is beneficial than the other way round.

eskimor: Do we have any data on how frequent collators go down in practice?

bkchr (approved):
Generally fine with this.

system-parachains/constants/src/lib.rs
bkchr: > Going up again is less straight forward.

Going up is a straight forward as going down. (yeah right now requires a small migration, but if we really want the flexibility we can fix it in `polkadot-sdk`.

Other chains like Solana change the leader aka block author every 1.6 seconds :)

eskimor: We also had a discussion about censorship resistance yesterday with Web3. This is a problem right now and should be fixed especially for crucial chains like asset hub. While longer slot times don't directly fix the issue, shorter ones certainly don't help. I would keep it as is, until we have a solution that actually works. 12s downtime was fine until now and to my knowledge nobody complained, because it rarely seems to happen in practice. 

> Other chains like Solana change the leader aka block author every 1.6 seconds :)

Do they have any censorship resistance? In theory it might work for them, but their architecture is fundamentally different. E.g. they don't have collators and for validators we do have byzantine assumptions and high economic guarantees.

eskimor: > For testnets we have SLOT_DURATION set to 6s.

Yes this was necessary until recently. @skunert improved the collator to be able to produce multiple blocks per slot so now we don't have to reduce slot time to have more frequent blocks anymore.

eskimor: Also if potential longer downtime is the only issue: Any ideas we have right now (including Web3) to restore censorship resistance involves accepting longer worst case downtimes.

Apart from censorship resistance longer slot times also help with maximizing throughput of elastic scaling.

TL;DR: I don't see any real advantages with going for 6s, while I do foresee at least potential issues with doing so.

s0me0ne-unkn0wn (approved):
With the session duration being `6 * HOURS` if the runtime is upgraded in the middle of a session, we're just getting one shorter-than-normal session, and nothing explodes, right?

system-parachains/constants/src/lib.rs
eskimor: Ok. I am fine with 6s too. Mostly because for censorship, I agree we very likely will need another solution for censorship anyway and for the perf concerns, Mythical is already at 6s ... so if 12s would turn out beneficial for that reason, we would need to find a way to fix that there likely earlier anyway.

eskimor (approved):

alindima (approved):

eskimor:
@skunert just brought up another good reason for 12s slot times, which I had forgotten about: Censorship is indeed becoming easier with lower slot times, because by the time people could have recovered, multiple slots will have passed. So a single dishonest collator will be able to censor multiple good ones.

TL;DR: Let's keep at 12s please.

eskimor (changes requested):
Please change slot time to 12s.

kianenigma (approved):

muharem:
set the `SLOT_DURATION` back to 12 sec, waiting @skunert to confirm the validate the setup.

.github/workflows/runtimes-matrix.json
```diff
@@ -31,7 +31,7 @@
     "path": "system-parachains/asset-hubs/asset-hub-kusama",
     "uri": "wss://kusama-asset-hub-rpc.polkadot.io:443",
     "is_relay": false,
-    "blocktime": 6000,
+    "blocktime": 12000,
```
skunert, line 34: Wondering about this change, not familiar with this matrix. Should it be the block time or slot duration here? Because blocktime will be 6s.

skunert (approved):
Changes LGTM, ran a local zombienet for verification, looks good :+1: 
image

Slight annoyance is that polkadot.js is still showing 12s as target duration, probably because its using the slot_duration to estimate this.

eskimor (approved):

muharem:
@bkchr @kianenigma @s0me0ne-unkn0wn @alindima re-requesting your reviews since I have set the `SLOT_DURATION` to `12s` by @eskimor. We also have some verification from @skunert with zombienet, please check the comment above.

system-parachains/constants/src/lib.rs
```diff
@@ -57,3 +57,42 @@ pub const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
...
+/// Parameters supporting async backing functionality.
+///
+/// Once all system chains have migrated to the new async backing mechanism, the parameters
+/// in this namespace will replace those currently defined in `super::*`.
+pub mod async_backing {
+	use frame_support::weights::{constants::WEIGHT_REF_TIME_PER_SECOND, Weight};
+	pub use parachains_common::BlockNumber;
+	use sp_runtime::Perbill;
+
+	/// This determines the average expected block time that we are targeting.
```
alindima, line 70: the comment is not really accurate anymore. I don't see `MILLISECS_PER_BLOCK` being used anywhere for determining the block time

eskimor (approved):

s0me0ne-unkn0wn (approved):

system-parachains/constants/src/lib.rs
```diff
@@ -59,3 +59,36 @@ pub const MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
...
+	pub const SLOT_DURATION: u64 = 12_000;
+
+	// Time is measured by number of blocks.
+	pub const MINUTES: BlockNumber = 60_000 / (MILLISECS_PER_BLOCK as BlockNumber);
+	pub const HOURS: BlockNumber = MINUTES * 60;
+	pub const DAYS: BlockNumber = HOURS * 24;
+
+	/// We assume that ~5% of the block weight is consumed by `on_initialize` handlers. This is
+	/// used to limit the maximal weight of a single extrinsic.
+	pub const AVERAGE_ON_INITIALIZE_RATIO: Perbill = Perbill::from_percent(5);
```
ggwpez, line 83: Is this not normally higher? I know that we have more ref time now with async backing, but not more PoV size, or?  
It should still be 5MiB?

ggwpez (approved):

CHANGELOG.md
```diff
@@ -36,9 +42,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Disable MBM migrations for all runtimes for check-migrations CI ([polkadot-fellows/runtimes/pull/590](https://github.com/polkadot-fellows/runtimes/pull/590))
 - chain-spec-generator supports conditional building (`--no-default-features --features <runtime>` or `--no-default-features --features all-runtimes` or
   `--no-default-features --features all-polkadot` or `--no-default-features --features all-kusama`)([polkadot-fellows/runtimes/pull/637](https://github.com/polkadot-fellows/runtimes/pull/637))
-- Increase spend payout period for treasuries from 30 to 90 days to provide sufficient time to address issues with insufficient balance of a specific asset in the treasury pot ([polkadot-fellows/runtimes/pull/647](https://github.com/polkadot-fellows/runtimes/pull/647))
-- Kusama: disable/filter `Nis` and `NisCounterpartBalances` pallets calls ([polkadot-fellows/runtimes/pull/656](https://github.com/polkadot-fellows/runtimes/pull/656))
```
bkchr, line 40: Why did you moved them up?

bkchr: I mean is this correct?

bkchr (approved):

system-parachains/constants/src/lib.rs
muharem: Right, it use to be 5%. Still will be higher in absolute numbers.

CHANGELOG.md
muharem: it is wrong in master now. the 1.4.3 does not include them, you can check it here https://github.com/polkadot-fellows/runtimes/releases/tag/v1.4.3

system-parachains/people/people-kusama/src/lib.rs
```diff
@@ -88,42 +98,6 @@ use xcm_runtime_apis::{
...
 
-/// This determines the average expected block time that we are targeting.
-///
-/// Blocks will be produced at a minimum duration defined by `SLOT_DURATION`. `SLOT_DURATION` is
-/// picked up by `pallet_timestamp`, which is in turn picked up by `pallet_aura` to implement `fn
-/// slot_duration()`.
-///
-/// Change this to adjust the block time.
-pub const MILLISECS_PER_BLOCK: u64 = 6_000;
-pub const SLOT_DURATION: u64 = MILLISECS_PER_BLOCK;
```
skunert, line 99: People Kusama can not just switch, this would need a migration. Please leave at 6s.

skunert (changes requested):

system-parachains/people/people-kusama/src/lib.rs
muharem: reverted slot duration to 6s for Kusama People Chain - https://github.com/polkadot-fellows/runtimes/pull/659/commits/d1056625018da9fa56c2133b15fedb35e9dac1ef

skunert (approved):

CHANGELOG.md
```diff
@@ -27,6 +27,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   - Balances: fix: do not emit Issued { amount: 0 } event ([paritytech/polkadot-sdk/pull/5946](https://github.com/paritytech/polkadot-sdk/pull/5946))
   - Snowbridge: Support bridging native ETH ([paritytech/polkadot-sdk/pull/7090](https://github.com/paritytech/polkadot-sdk/pull/7090))
   - Runtime-APIs: Fix DryRunApi client-facing XCM versions ([paritytech/polkadot-sdk/pull/7689](https://github.com/paritytech/polkadot-sdk/pull/7689))
+- Kusama: disable/filter `Nis` and `NisCounterpartBalances` pallets calls ([polkadot-fellows/runtimes/pull/656](https://github.com/polkadot-fellows/runtimes/pull/656))
+- Increase spend payout period for treasuries from 30 to 90 days to provide sufficient time to address issues with insufficient balance of a specific asset in the treasury pot ([polkadot-fellows/runtimes/pull/647](https://github.com/polkadot-fellows/runtimes/pull/647))
+- Kusama: disable/filter `Nis` and `NisCounterpartBalances` pallets calls ([polkadot-fellows/runtimes/pull/656](https://github.com/polkadot-fellows/runtimes/pull/656))
```
bkchr, line 32:
```suggestion
```

CHANGELOG.md
```diff
@@ -27,6 +27,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- Asset Hub: remove XCM sufficient asset fee trader ([polkadot-fellows/runtimes#502](https://github.com/polkadot-fellows/runtimes/pull/502))
+- Increase spend payout period for treasuries from 30 to 90 days to provide sufficient time to address issues with insufficient balance of a specific asset in the treasury pot ([polkadot-fellows/runtimes/pull/647](https://github.com/polkadot-fellows/runtimes/pull/647))
```
bkchr, line 34:
```suggestion
```

PR #660, opened 2025-04-02
[AHM] asset_rate test

eugypalu:
Add pre/post migration checks to asset_rate


ggwpez (approved):
Thanks, just nit comments.

PR #663, opened 2025-04-04
[AHM] Benchmarks

muharem:
Asset Hub migrator benchmarks


integration-tests/ahm/Cargo.toml
```diff
@@ -9,44 +9,48 @@ publish = false
 
 [dev-dependencies]
 codec = { workspace = true, default-features = true }
-asset-hub-polkadot-runtime = { workspace = true }
+asset-hub-polkadot-runtime = { workspace = true, default-features = true, features = ["runtime-benchmarks"]}
```
ggwpez, line 12: You cannot always add it, it will not be correct since we also need to run tests without it.

ggwpez: fixed

muharem:
/cmd bench --runtime asset-hub-polkadot --pallet pallet_ah_migrator

PR #667, opened 2025-04-08
[AHM] Bump polkadot relay and AH versions

seadanda:
This will allow upgrading the chains locally for tests. Spec version will change before release.


x3c41a (approved):

acatangiu (approved):
fine since this is on `dev-asset-hub-migration` branch

PR #668, opened 2025-04-08
Update value of MigrationMaxKeyLen and and MigController to control migration

nkpar:
Update value of MigrationMaxKeyLen to proceed with state trie migration via bot on both KAH and PAH

### DIFF
--- a/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
+++ b/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
@@ -1777,7 +1777,7 @@ parameter_types! {
 	// The deposit configuration for the singed migration. Specially if you want to allow any signed account to do the migration (see `SignedFilter`, these deposits should be high)
 	pub const MigrationSignedDepositPerItem: Balance = CENTS;
 	pub const MigrationSignedDepositBase: Balance = 2_000 * CENTS;
-	pub const MigrationMaxKeyLen: u32 = 32;
+	pub const MigrationMaxKeyLen: u32 = 512;
 }
 
 impl pallet_state_trie_migration::Config for Runtime {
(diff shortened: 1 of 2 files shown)


kianenigma (approved):
This is a good example of why `pallet-parameters` is a good idea 📝

ggwpez (approved):

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1786,10 +1786,11 @@ impl pallet_state_trie_migration::Config for Runtime {
 	type RuntimeHoldReason = RuntimeHoldReason;
 	type SignedDepositPerItem = MigrationSignedDepositPerItem;
 	type SignedDepositBase = MigrationSignedDepositBase;
-	// An origin that can control the whole pallet: should be Root or the Fellowship.
+	// An origin that can control the whole pallet: Should be a Fellowship member or the controller
+	// of the migration.
 	type ControlOrigin = EitherOfDiverse<
-		EnsureRoot<AccountId>,
 		EnsureXcm<IsVoiceOfBody<FellowshipLocation, FellowsBodyId>>,
+		EnsureSignedBy<MigControllerRoot, AccountId>,
```
bkchr, line 1793: Why is this account set as control origin?

ggwpez:
Btw this should have a CHANGELOG entry, please only check the box that it does not require it for no-ops like comment fix or typo.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
ggwpez: The Kusama fellowship ref delay of 5 days is holding us up a bit with the migration.  
Had to do two so far...

PR #674, opened 2025-04-14
[AHM] Complete account balances tests

re-gius:
**Migration Changes**
- If a partially migrated account has not enough free balance (_free_) to cover for the Relay Chain existential deposit + Asset Hub existential deposit (1 DOT + 0,01 DOT for Polkadot), we unreserve some of the reserved balance kept on the Relay Chain to cover for the existential deposits. (On Polkadot, if _free_ < 1,01 DOT, then we unreserve _free_ - 1,01 DOTs from the reserved balance kept on the relay chain).

**New tests**

- [x] Extensive Relay Chain pre-migration on post-migration checks on all the accounts.
- [x] Balances cross-checks on RC and AH post-migration for checking account balance and total issuance
- [x] AH checks on correctly migrated freezes/locks and holds/reserves

For more details, you can take a look at the comments or at [this document](https://hackmd.io/DieSAxrLSuGgFzpgzNd24Q?both).


pallets/ah-migrator/src/account.rs
```diff
@@ -257,15 +260,18 @@ impl<T: Config> crate::types::AhMigrationCheck for AccountsMigrator<T> {
...
 
-		// TODO: Giuseppe @re-gius
-		//   run post migration sanity checks like:
-		//    - rc_migrated_out == ah_migrated_in - failed accounts
+		let (rc_checking_before, _) = rc_pre_payload;
+		let (ah_checking_before, _) = ah_pre_payload;
+
+		let ah_checking_after = <T as Config>::Currency::total_balance(&T::CheckingAccount::get());
+
+		assert_eq!(ah_checking_after, ah_checking_before + rc_checking_before, "Checking account balance on Asset Hub after migration should be the sum of the checking account balance before migration and the checking account balance on the relay chain before migration");
```
acatangiu, line 275: nope, this is not correct. [look here](https://github.com/polkadot-fellows/runtimes/blob/f09750296a8ad40d64e295ff6568acac36c978f7/pallets/ah-migrator/src/account.rs#L200) for the explanation.

pallets/ah-migrator/src/account.rs
```diff
@@ -257,15 +260,18 @@ impl<T: Config> crate::types::AhMigrationCheck for AccountsMigrator<T> {
...
 		assert!(
 			RcAccounts::<T>::iter().next().is_none(),
 			"Failed accounts should not remain in storage after migration"
 		);
 
-		// TODO: Giuseppe @re-gius
-		//   run post migration sanity checks like:
-		//    - rc_migrated_out == ah_migrated_in - failed accounts
+		let (rc_checking_before, _) = rc_pre_payload;
+		let (ah_checking_before, _) = ah_pre_payload;
```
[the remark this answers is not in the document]
re-gius: Yes, it's still a draft - I'm just having a hard time using the relay chain post-migration balances in AH tests, I'll probably need to make a small change to the trait or use a different trait

pallets/rc-migrator/src/accounts.rs
```diff
@@ -804,17 +804,18 @@ impl<T: Config> AccountsMigrator<T> {
...
+		let checking_balance = <T as Config>::Currency::total_balance(&T::CheckingAccount::get());
+		(checking_balance, total_issuance)
 	}
 
-	fn post_check(rc_total_issuance_before: Self::RcPrePayload) {
+	fn post_check(rc_pre_payload: Self::RcPrePayload) {
+		let (_rc_checking_before, rc_total_issuance_before) = rc_pre_payload;
 		// Check that all accounts have been processed correctly
-		let mut kept = 0;
 		for (who, acc_state) in RcAccounts::<T>::iter() {
```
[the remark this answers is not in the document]
muharem: RcAccounts of id -> none basically means -> Migrate

re-gius:
I'm back on this PR, probably adding a separate test for more complex checks.

pallets/ah-migrator/src/account.rs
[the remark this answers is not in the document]
re-gius: The problem with this is that some reserves should be thrown away (e.g., multisig ones) while some others should stay (proxy ones).

pallets/rc-migrator/src/accounts.rs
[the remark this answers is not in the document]
re-gius: Changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/b10d4bfd9ac20d4e3da4285160d82ddf3df4f5d3 + still investigating if it can be improved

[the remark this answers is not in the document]
re-gius: Final version in https://github.com/polkadot-fellows/runtimes/pull/674/commits/22b99a285ab0bdd83ce4f487fb9ee0325f1f6529 , I don't think we can assert anything else on non-migratable accounts

pallets/ah-migrator/src/account.rs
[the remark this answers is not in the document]
re-gius: For holds, freezes and locks, I would need to check case by case to see which ones are migrated to asset hub and **not** deleted right after. This may also depend on pallets like staking that haven't been migrated yet, I'm not sure.

I'm looking into that.

integration-tests/ahm/src/balances_test.rs
```diff
@@ -0,0 +1,141 @@
...
+		// This is fine, we can just ignore/accept this small amount.
+		#[cfg(not(feature = "ahm-westend"))]
+		defensive_assert!(
+			ah_checking_balance_before == pallet_balances::Pallet::<AhRuntime>::minimum_balance()
+		);
+
+		(ah_total_issuance_before, ah_checking_balance_before)
+	}
+
+	fn post_check(rc_pre_payload: Self::RcPrePayload, ah_pre_payload: Self::AhPrePayload) {
```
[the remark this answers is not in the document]
re-gius: This doesn't work now, I'm getting

```
assertion `left == right` failed: Checking account migration is incorrect. Total issuance not correctly tracked.
  left: 15613640545174476238
 right: 15613640545428934061
```

They're like 0.025 DOT apart and I'm trying to understand why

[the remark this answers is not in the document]
acatangiu: Ideally they should match 100% since it's just arithmetic, but IMO we can also live with an impl and a check that keeps us within a few DOTs apart.

The most important bit is to have the check that verifies that it 99.999% works 👍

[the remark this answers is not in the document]
re-gius: I still didn't find the lost DOTs, but added a check on https://github.com/polkadot-fellows/runtimes/pull/674/commits/11df9b14988f26b6a30d459db98515972649b6cf about the difference being < 0.1 DOT. I think that can be good enough

acatangiu (approved):
Looks good!

acatangiu (approved):
LGTM! @ggwpez or @muharem to ack too

re-gius:
@ggwpez @muharem this should be ready to be merged

re-gius:
I'm still working on some final checks on holds/reserves and locks/freezes, both on RC and AH. [in progress...]

re-gius:
Tests should be complete now

integration-tests/ahm/src/balances_test.rs
```diff
@@ -0,0 +1,155 @@
...
+use frame_support::{defensive_assert, traits::Currency};
+use pallet_ah_migrator::types::AhMigrationCheck;
+use pallet_rc_migrator::types::RcMigrationCheck;
+use std::sync::Mutex;
+
+// Total balances and checking balances are migrated correctly.
+pub struct BalancesCrossChecker;
+
+// Used to store the balance kept on the relay chain after migration.
+static RC_KEPT_AFTER: Mutex<Option<u128>> = Mutex::new(None);
```
ggwpez, line 38: Mutex? Where do we access this from more than one thread?  
Or the issue is that the post_check of Rc is not returning any payload?

pallets/ah-migrator/src/account.rs
```diff
@@ -84,6 +86,20 @@ impl<T: Config> Pallet<T> {
 		};
 		debug_assert!(minted == total_balance);
 
+		AhMigratedBalance::<T>::mutate(|balance| {
+			*balance = match (*balance).checked_add(total_balance) {
```
ggwpez: I think there should also be a `defensive_saturating_add` which basically does this but also fails in tests.

pallets/ah-migrator/src/account.rs
```diff
@@ -263,15 +284,113 @@ impl<T: Config> crate::types::AhMigrationCheck for AccountsMigrator<T> {
...
+			// locks on AH after migration are the same as on RC before migration.
+			assert_eq!(
+				summary.locks,
+				locks_enc,
+				"Locks mismatch for account {:?} between RC pre-migration and AH post-migration",
+				who.to_ss58check()
+			);
+
+			// There should be no freezes on AH before the migration so we just need to check that
+			// the freezes on AH after migration are the same as on RC before migration.
```
ggwpez: I guess this is by chance? There could be some, or?

pallets/rc-migrator/src/accounts.rs
```diff
@@ -371,7 +377,28 @@ impl<T: Config> AccountsMigrator<T> {
...
 			};
 
-			if let Err(e) = <T as Config>::Currency::release(&id, &who, amount, Precision::Exact) {
+			// If the hold amount is greater than the reserved balance (inconsistent state), we just
+			// release the entire reserved balance
+			let reserved_balance = <T as Config>::Currency::reserved_balance(&who);
+			let mut release_precision = Precision::Exact;
+			if reserved_balance < amount {
+				log::error!(target: LOG_TARGET,
+					"Hold amount is greater than the reserved balance. \
```
ggwpez: Does this happen with the westend state? Maybe put a `defensive` instead so that the test fails.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -775,29 +806,287 @@ impl<T: Config> AccountsMigrator<T> {
 	}
 }
 
+#[derive(Default, Clone, PartialEq, Eq, Debug)]
```
ggwpez:
```suggestion
// Balance summary of an account of the Relay chain. Only used for testing.
#[derive(Default, Clone, PartialEq, Eq, Debug)]
```
?

pallets/rc-migrator/src/accounts.rs
```diff
@@ -775,29 +806,287 @@ impl<T: Config> AccountsMigrator<T> {
...
+			b"vesting " => 1,
+			b"pyconvot" => 2,
+			_ => {
+				log::error!(
+					target: LOG_TARGET,
+					"Unknown lock id: {:?}",
+					lock_id
+				);
+				3
+			},
```
ggwpez:
```suggestion
```
Since this is a test, it should fail loudly when something is clearly wrong.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -775,29 +806,287 @@ impl<T: Config> AccountsMigrator<T> {
...
 #[cfg(feature = "std")]
-impl<T: Config> crate::types::RcMigrationCheck for AccountsMigrator<T> {
-	// rc_total_issuance_before
-	type RcPrePayload = BalanceOf<T>;
+impl<T> AccountsMigrationChecker<T> {
+	// Translate the lock id to a u8 for both RC and AH using the different encodings for easier
+	// comparison. Lock Ids type is  `LockIdentifier = [u8; 8]` on Polkadot.
+	pub fn lock_id_to_u8(lock_id: [u8; 8], _chain_type: ChainType) -> u8 {
+		match lock_id.as_slice() {
+			b"staking " => 0,
```
ggwpez: You could also define an enum instead of just using u8, but both is fine IMHO.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -775,29 +806,287 @@ impl<T: Config> AccountsMigrator<T> {
...
+			},
+		}
+	}
+
+	// Translate the freeze id to a u8 for both RC and AH using the different encodings for easier
+	// comparison. Freeze Ids type is enum `RuntimeFreezeReason` on Polkadot. Here we use the
+	// encoded enum.
+	pub fn freeze_id_encoding_to_u8(freeze_id: Vec<u8>, chain_type: ChainType) -> u8 {
+		let nomination_pools_encoding = match chain_type {
+			ChainType::RC => [39, 0],
```
ggwpez:
```suggestion
			ChainType::RC => [39, 0], // 39 = nom pools pallet index
```
39 is the pallet index of the pools pallet? One comment please since we will need to change this for other runtimes possibly.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -775,29 +806,287 @@ impl<T: Config> AccountsMigrator<T> {
...
+			ChainType::AH => [80, 0],
+		};
+		if freeze_id.as_slice() == nomination_pools_encoding {
+			0
+		} else {
+			log::error!(target: LOG_TARGET, "Unknown freeze id: {:?}", freeze_id);
```
ggwpez:
```suggestion
			panic!("Unknown freeze id: {:?}", freeze_id);
```
I think this is some serious error when it is unknown.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -775,29 +806,287 @@ impl<T: Config> AccountsMigrator<T> {
...
+			// TODO: change to the correct encoding when Staking holds are correctly re-created on
+			// Asset Hub in pallet-staking-async
+			ChainType::AH => [5, 0],
+		};
+		if hold_id.as_slice() == preimage_encoding {
+			0
+		} else if hold_id.as_slice() == staking_encoding {
+			// TODO: change to the correct encoding when Staking holds are correctly re-created on
+			// Asset Hub in pallet-staking-async
+			0
```
ggwpez: I feel now like using an enum would be reasonable.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -775,29 +806,287 @@ impl<T: Config> AccountsMigrator<T> {
...
+		} else {
+			log::error!(target: LOG_TARGET, "Unknown hold id: {:?}", hold_id);
```
ggwpez: Panic is okay in tests.

integration-tests/ahm/src/balances_test.rs
re-gius: The second one:  post_check of Rc is not returning any payload. This is the only case we need a RC post_check payload for now, so I preferred this shortcut rather than changing the checking trait.

pallets/ah-migrator/src/account.rs
re-gius: We don't have pallet `nomination-pools` on Asset Hub for Kusama, Paseo, and Polkadot. We do have it on Westend though. Do you think we should check for existing freezes?

pallets/rc-migrator/src/accounts.rs
re-gius: Changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4bbb3ea51b858323e9c77d7443e49b848684bb26

pallets/rc-migrator/src/accounts.rs
re-gius: Totally agree, changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4bbb3ea51b858323e9c77d7443e49b848684bb26

pallets/rc-migrator/src/accounts.rs
re-gius: Yes, changed to `panic` in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4bbb3ea51b858323e9c77d7443e49b848684bb26

pallets/rc-migrator/src/accounts.rs
re-gius: Added in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4bbb3ea51b858323e9c77d7443e49b848684bb26

pallets/rc-migrator/src/accounts.rs
re-gius: Totally agree, changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4bbb3ea51b858323e9c77d7443e49b848684bb26

pallets/rc-migrator/src/accounts.rs
re-gius: Changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4bbb3ea51b858323e9c77d7443e49b848684bb26

pallets/rc-migrator/src/accounts.rs
re-gius: Changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4bbb3ea51b858323e9c77d7443e49b848684bb26

pallets/rc-migrator/src/accounts.rs
re-gius: Changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4bbb3ea51b858323e9c77d7443e49b848684bb26 . It doesn't seem to happen in Westend, will double-check

pallets/ah-migrator/src/account.rs
re-gius: Changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4bbb3ea51b858323e9c77d7443e49b848684bb26

pallets/rc-migrator/src/accounts.rs
```diff
@@ -310,6 +314,8 @@ impl<T: Config> AccountsMigrator<T> {
 
 		if !Self::can_migrate_account(&who, &account_info) {
 			log::info!(target: LOG_TARGET, "Account '{}' is not migrated", who.to_ss58check());
+			// Will be used for post-migration checks
+			RcAccounts::<T>::insert(&who, AccountState::Preserve);
```
muharem: I wouldn’t do this. We have only a few clear reasons for Preserve state (e.g. when we know that a specific account must stay on RC) and we handle that during the pre-data migration job. The `can_migrate_account` function have different semantic and checks if migration is even possible (balance >= ED) for current account.
Also, we have a dedicated migration stage where this map is mutated — it doesn’t feel right to mutate it here while we’re using it.
And finally, I don’t think tests should influence the implementation to the point where we add extra data purely for them.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -345,7 +351,7 @@ impl<T: Config> AccountsMigrator<T> {
 			// prevents the free balance from being burned.
 			// This scenario causes a panic in the test environment - see:
 			// https://github.com/paritytech/polkadot-sdk/blob/35e6befc5dd61deb154ff0eb7c180a038e626d66/substrate/frame/balances/src/impl_fungible.rs#L285
-			let amount = if free < ed && amount.saturating_sub(ed - free) > 0 {
+			let mut amount = if free < ed && amount.saturating_sub(ed - free) > 0 {
```
muharem: you can keep it immutable and redefine later

pallets/rc-migrator/src/accounts.rs
```diff
@@ -371,7 +377,23 @@ impl<T: Config> AccountsMigrator<T> {
 				amount
 			};
 
-			if let Err(e) = <T as Config>::Currency::release(&id, &who, amount, Precision::Exact) {
+			// If the hold amount is greater than the reserved balance (inconsistent state), we just
+			// release the entire reserved balance
+			let reserved_balance = <T as Config>::Currency::reserved_balance(&who);
+			let mut release_precision = Precision::Exact;
+			if reserved_balance < amount {
```
muharem: do we have such account/state?

pallets/rc-migrator/src/accounts.rs
```diff
@@ -371,7 +377,23 @@ impl<T: Config> AccountsMigrator<T> {
...
+				defensive!(
+					"Hold amount for account {:?} is greater than the reserved balance.",
+					who.to_ss58check()
+				);
+				log::debug!(target: LOG_TARGET,
```
muharem: why you need debug if you have error above?

pallets/rc-migrator/src/accounts.rs
```diff
@@ -640,7 +662,10 @@ impl<T: Config> AccountsMigrator<T> {
 			if deposit == 0 {
 				return;
 			}
-			reserves.entry(id).and_modify(|e| *e += deposit).or_insert(deposit);
+			reserves
+				.entry(id)
+				.and_modify(|e: &mut u128| *e = e.saturating_add(deposit))
```
muharem: sum of all deposits cannot be bigger than total issuance, hence cannot overflow. I am not against saturating function even in such cases, but I am not in favor of extra changes that needs to be reviewed.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -775,29 +801,301 @@ impl<T: Config> AccountsMigrator<T> {
 	}
 }
 
+// Only used for testing.
```
muharem: if only for testing, why not have it under test or std feature? all these can be under sub module which is under a feature.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -775,29 +801,301 @@ impl<T: Config> AccountsMigrator<T> {
...
+	Staking,
+	Vesting,
+	ConvictionVoting,
+}
+
+// Balance summary of an account of the Relay chain. Only used for testing.
+#[derive(Default, Clone, PartialEq, Eq, Debug)]
+pub struct BalanceSummary {
+	// Balance that can be still reserved
+	pub migrated_free: u128,
```
muharem: comment? not a doc? technically these all public items that can be accessed from external crates. better to place them under a feature.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -775,29 +801,301 @@ impl<T: Config> AccountsMigrator<T> {
...
+		}
+	}
+
+	// Translate the freeze id to enum for both RC and AH using the different encodings for easier
+	// comparison. Freeze Ids type is enum `RuntimeFreezeReason` on Polkadot. Here we use the
+	// encoded enum.
+	pub fn freeze_id_encoding(freeze_id: Vec<u8>, chain_type: ChainType) -> FreezeReason {
+		let nomination_pools_encoding = match chain_type {
+			ChainType::RC => [39, 0], // 39 = nom pools pallet index on Polkadot
+			ChainType::AH => [80, 0], // 80 = nom pools pallet index on Asset Hub
```
muharem: indexes might be different for Polkadot and Kusama, when we use these pallets for Kusama someone might spend some time to figure out why these tests failing

pallets/rc-migrator/src/accounts.rs
```diff
@@ -775,29 +801,301 @@ impl<T: Config> AccountsMigrator<T> {
...
+			panic!("Unknown freeze id: {:?}", freeze_id);
+		}
+	}
+
+	// Translate the hold id to enum for both RC and AH using the different encodings for easier
+	// comparison. Hold Ids type is enum `RuntimeHoldReason` on Polkadot. Here we use the encoded
+	// enum.
+	pub fn hold_id_encoding(hold_id: Vec<u8>, chain_type: ChainType) -> HoldReason {
+		let preimage_encoding: [u8; 2] = match chain_type {
+			ChainType::RC => [10, 0], // 10 = preimage pallet index on Polkadot
```
muharem: here same. you can pass is with some type argument for this test type. or you can just have two different types with these functions and pass them as type argument where you need them.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -775,29 +801,301 @@ impl<T: Config> AccountsMigrator<T> {
...
+			panic!("Unknown hold id: {:?}", hold_id);
+		}
+	}
+
+	// Get the AH expected hold amount for a RC migrated hold.
+	// This is used to check that the hold amount is correct after migration.
+	pub fn ah_hold_amount_from_rc(hold_id: Vec<u8>, hold_amount: u128) -> u128 {
+		match hold_id.as_slice() {
+			// Preimage deposits are divided by 100 when migrated to Asset Hub.
+			[10, 0] => hold_amount.saturating_div(100),
```
muharem: @ggwpez if we have poke dispatchables for all these pallets, may be we do not update deposits during migration?

pallets/rc-migrator/src/accounts.rs
muharem: but its good the you are not using the mappers from config. ty

pallets/rc-migrator/src/accounts.rs
```diff
@@ -775,29 +801,301 @@ impl<T: Config> AccountsMigrator<T> {
...
-	// rc_total_issuance_before
-	type RcPrePayload = BalanceOf<T>;
+pub struct RcKeptBalance<T: Config> {
+	pub rc_reserved_kept: BTreeMap<T::AccountId, u128>,
+	pub rc_free_kept: BTreeMap<T::AccountId, u128>,
+}
+
+#[cfg(feature = "std")]
+impl<T: Config> RcKeptBalance<T> {
+	pub fn new() -> Self {
```
muharem: `build` would be more intuitive

integration-tests/ahm/src/balances_test.rs
```diff
@@ -0,0 +1,155 @@
...
+// Used to store the balance migrated out of the relay chain.
+static RC_MIGRATED_BALANCE: Mutex<Option<u128>> = Mutex::new(None);
+
+// Min tolerance for some balance checks in the tests, currently 0.1 DOT.
+const MIN_DOT_ERROR: u128 = 1000000000;
```
muharem, line 41: this would be different for Polkadot and Kusama

integration-tests/ahm/src/balances_test.rs
```diff
@@ -0,0 +1,155 @@
...
+		let checking_balance = pallet_balances::Pallet::<RcRuntime>::total_balance(&check_account);
+		assert_eq!(
+			checking_balance, 0,
+			"Checking account should have no balance on the relay chain after migration"
+		);
+
+		let rc_kept_after = pallet_balances::Pallet::<RcRuntime>::total_issuance();
+		*RC_KEPT_AFTER.lock().unwrap() = Some(rc_kept_after);
+
+		let rc_migrated_balance =
```
muharem: you can also assert this with `{rc_pre_payload.rc_total_issuance_before - rc_kept_after }`

integration-tests/ahm/src/balances_test.rs
```diff
@@ -0,0 +1,155 @@
...
+	fn pre_check(_: Self::RcPrePayload) -> Self::AhPrePayload {
+		let ah_total_issuance_before = pallet_balances::Pallet::<AhRuntime>::total_issuance();
+		let ah_checking_balance_before = pallet_balances::Pallet::<AhRuntime>::total_balance(
+			&<AhRuntime as pallet_ah_migrator::Config>::CheckingAccount::get(),
+		);
+
+		// AH checking account has incorrect 0.01 DOT balance because of the DED airdrop which
+		// added DOT ED to all existing AH accounts.
+		// This is fine, we can just ignore/accept this small amount.
+		#[cfg(not(feature = "ahm-westend"))]
```
muharem: not needed anymore

integration-tests/ahm/src/balances_test.rs
```diff
@@ -0,0 +1,155 @@
...
+			.lock()
+			.unwrap()
+			.expect("rc_kept_after should be set by RcMigrationCheck::post_check");
+		// `checking_balance` = ah_check_before + rc_check_before
+		let checking_balance = rc_checking_balance_before + ah_checking_balance_before;
+
+		// (0) rc_check_before = `checking_balance` - ah_check_before
+		defensive_assert!(
+			checking_balance - ah_checking_balance_before == rc_checking_balance_before
+		);
```
muharem: I think this does not make much sense, you checking rust + and - operations (=

integration-tests/ahm/src/balances_test.rs
```diff
@@ -0,0 +1,155 @@
...
+				ah_total_issuance_before -
+				ah_checking_balance_before,
+			"Checking balance on asset hub after migration is incorrect"
+		);
+
+		let rc_migrated_balance = RC_MIGRATED_BALANCE
+			.lock()
+			.unwrap()
+			.expect("rc_migrated_balance should be set by RcMigrationCheck::post_check");
+		let ah_migrated_balance = pallet_ah_migrator::AhMigratedBalance::<AhRuntime>::get();
```
muharem: it should be good enough if you use `{ ah_total_issuance_before - ah_total_issuance_after }`

muharem: here same comment, I don’t think tests should influence the implementation to the point where we add extra data purely for them. I would remove this new storage item.

pallets/rc-migrator/src/accounts.rs
re-gius: Ok, removed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4039a4de3c05ad2ec10ffc0a99715dad82d7bb87

pallets/rc-migrator/src/accounts.rs
re-gius: It is changed in the body of an `if` statement, so it shouldn't work because the redefine would be local to that.

pallets/rc-migrator/src/accounts.rs
re-gius: Not now, it happened on a previous snapshot

pallets/rc-migrator/src/accounts.rs
re-gius: Right, removed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4039a4de3c05ad2ec10ffc0a99715dad82d7bb87

pallets/rc-migrator/src/accounts.rs
re-gius: Right, reverted in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4039a4de3c05ad2ec10ffc0a99715dad82d7bb87

pallets/rc-migrator/src/accounts.rs
re-gius: Moved everything under a testing module in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4039a4de3c05ad2ec10ffc0a99715dad82d7bb87

pallets/rc-migrator/src/accounts.rs
re-gius: Is it okay now with the testing module?

pallets/rc-migrator/src/accounts.rs
re-gius: Changed to `unimplemented` for every other chain than Polkadot in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4039a4de3c05ad2ec10ffc0a99715dad82d7bb87

pallets/rc-migrator/src/accounts.rs
re-gius: Left `unimplemented` for Kusama in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4039a4de3c05ad2ec10ffc0a99715dad82d7bb87

pallets/rc-migrator/src/accounts.rs
re-gius: Changed to `build` in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4039a4de3c05ad2ec10ffc0a99715dad82d7bb87

integration-tests/ahm/src/balances_test.rs
re-gius: It's just a tolerance threshold, ideally the error should be 0 on Kusama, we'll see if there's a bigger mismatch there

integration-tests/ahm/src/balances_test.rs
re-gius: Removed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4039a4de3c05ad2ec10ffc0a99715dad82d7bb87

integration-tests/ahm/src/balances_test.rs
re-gius: Removed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4039a4de3c05ad2ec10ffc0a99715dad82d7bb87

integration-tests/ahm/src/balances_test.rs
re-gius: Right, removed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4039a4de3c05ad2ec10ffc0a99715dad82d7bb87
I also noticed that this check on migrated balance is redundant after the check on the total balance

integration-tests/ahm/src/balances_test.rs
re-gius: Yes, I removed it in https://github.com/polkadot-fellows/runtimes/pull/674/commits/4039a4de3c05ad2ec10ffc0a99715dad82d7bb87 because the corresponding check on migrated balance becomes redundant after the check on the total balance

pallets/rc-migrator/src/accounts.rs
muharem: cannot we add `#[cfg(feature = "std")]` ?

pallets/ah-migrator/src/account.rs
```diff
@@ -213,11 +219,15 @@ impl<T: Config> Pallet<T> {
 }
 
 #[cfg(feature = "std")]
-impl<T: Config> crate::types::AhMigrationCheck for AccountsMigrator<T> {
-	// rc_total_issuance_before
-	type RcPrePayload = BalanceOf<T>;
-	// ah_checking_account_before
-	type AhPrePayload = BalanceOf<T>;
+impl<T: Config> crate::types::AhMigrationCheck for AccountsMigrationChecker<T> {
```
muharem: may be `#[cfg(feature = "std")] pub mod tests { ... }`.
the new imports in this file also would move under the mod.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -120,10 +124,17 @@ pub enum AccountState<Balance> {
...
 		/// In practice reserved by old `Currency` api and has no associated reason.
-		reserved: Balance,
+		reserved_amount: Balance,
+		/// The reserved balance that is actually preserved on RC.
+		///
+		/// If the account has not enough free balance for the relay chain existential deposit and
+		/// asset hub existential deposit, some of the `reserved_amount` balance needs to be
+		/// unreserved to cover for the existential deposits. As a consequence, this is always
+		/// smaller than or equal to the `reserved_amount` balance.
+		actual_reserved: Balance,
```
muharem: may than, `expected_reserve` and `actual_reserve`? `_amount` suffix does not adds much meaning

pallets/rc-migrator/src/accounts.rs
```diff
@@ -26,6 +26,8 @@ use frame_system::Account as SystemAccount;
 use pallet_balances::{AccountData, BalanceLock};
 use sp_core::ByteArray;
 use sp_runtime::{traits::Zero, BoundedVec};
+#[cfg(feature = "std")]
+use std::collections::BTreeMap;
```
muharem: can be under tests mod

pallets/rc-migrator/src/accounts.rs
```diff
@@ -695,8 +746,9 @@ impl<T: Config> AccountsMigrator<T> {
 			let rc_ed = <T as Config>::Currency::minimum_balance();
 			let ah_ed = T::AhExistentialDeposit::get();
 
-			// we prioritize the named holds over the unnamed reserve. if any named holds we will
-			// send them to the AH and keep up to the `rc_reserved` on the RC.
+			// we prioritize the named holds over the unnamed reserve. If the account to preserve
+			// has any named holds, we will send them to the AH and keep up to the unnamed reserves
+			// `rc_reserved` on the RC.
```
muharem, line 758: I think most of the changes wont be needed for accounts migration if we do this way. Please double check

``` rust
		for (id, expected_rc_reserved) in reserves {
			weight += T::DbWeight::get().reads_writes(6, 1);
			let free = <T as Config>::Currency::balance(&id);
			let total_reserved = <T as Config>::Currency::reserved_balance(&id);
			let total_hold = pallet_balances::Holds::<T>::get(&id)
				.into_iter()
				// we do not expect more holds
				.take(5)
				.map(|h| h.amount)
				.sum::<T::Balance>();

			let rc_ed = <T as Config>::Currency::minimum_balance();
			let ah_ed = T::AhExistentialDeposit::get();

			defensive_assert!(total_reserved >= total_hold, "total_reserved >= total_hold");

			let missing_free = (rc_ed + ah_ed).saturating_sub(free);
			// we prioritize the named holds over the unnamed reserve. If the account to preserve
			// has any named holds, we will send them to the AH and keep up to the unnamed reserves
			// `rc_reserved` on the RC.
			let actual_rc_reserved = expected_rc_reserved
				.min(total_reserved.saturating_sub(total_hold).saturating_sub(missing_free));

			if actual_rc_reserved == 0 {
				log::info!(
					target: LOG_TARGET,
					"Account doesn't have enough reserved balance to keep on RC. account: {:?}.",
					id.to_ss58check(),
				);
				continue;
			}

			if missing_free == 0 {
				RcAccounts::<T>::insert(
					&id,
					// one consumer reference of reserved balance.
					AccountState::Part {
						free: rc_ed,
						reserved: actual_rc_reserved,
						consumers: 1,
					},
				);
			} else {
				let failed = <T as Config>::Currency::unreserve(&id, missing_free);
				defensive_assert!(failed == 0, "failed to unreserve");

				RcAccounts::<T>::insert(
					&id,
					// one consumer reference of reserved balance.
					AccountState::Part {
						free: rc_ed,
						reserved: actual_rc_reserved,
						consumers: 1,
					},
				);
			}
		}
```

re-gius: Changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/0a23e92adbd93fcdfe5545e3c97d639aa7fb8725

pallets/rc-migrator/src/accounts.rs
re-gius: Changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/0a23e92adbd93fcdfe5545e3c97d639aa7fb8725 but also moved it closer to tests (as in other pallets)

pallets/rc-migrator/src/accounts.rs
re-gius: Removed both and went back to `reserved` in https://github.com/polkadot-fellows/runtimes/pull/674/commits/0a23e92adbd93fcdfe5545e3c97d639aa7fb8725

pallets/ah-migrator/src/account.rs
re-gius: Moved imports there in https://github.com/polkadot-fellows/runtimes/pull/674/commits/0a23e92adbd93fcdfe5545e3c97d639aa7fb8725 , but we still need to re-use them outside of the tests mod

pallets/rc-migrator/src/accounts.rs
muharem: I am not sure this will work. Bc we gonna have some hold left, and most probably we wont be able to withdraw the account fully. Also you changing the amount of a hold here, but you did not change the amount we send to RC. 

(1) We ether make this correct, by sending the right hold amount to RC (which will complicate this code) and we write the tests for this case to actually understand what would happen, if this even gonna work.

(2) Or if we do not have this case in the recent snapshots, we revert this changes and use defensive instead of log::error below. It will error if it failed to release the exact `amount`. 

If this not a problem right now (no such case in the latest snapshot) I would go for (2) at least for the scope of this PR.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -455,16 +470,16 @@ impl<T: Config> AccountsMigrator<T> {
...
-			.checked_sub(account_state.get_rc_free())
-			.defensive_unwrap_or_default();
 		let teleport_reserved = account_data
 			.reserved
 			.checked_sub(account_state.get_rc_reserved())
 			.defensive_unwrap_or_default();
+		let teleport_free = account_data
+			.free
+			.checked_sub(account_state.get_rc_free())
+			.defensive_unwrap_or_default();
```
muharem, line 490: not necessary diff

pallets/rc-migrator/src/accounts.rs
```diff
@@ -709,29 +730,38 @@ impl<T: Config> AccountsMigrator<T> {
...
+				defensive_assert!(failed == 0, "failed to unreserve");
+
+				let new_free = <T as Config>::Currency::balance(&id);
+				if new_free < rc_ed + ah_ed {
+					log::error!(
+						target: LOG_TARGET,
+						"We could not unreserve enough balance on the RC for RC and AH existential deposits for partially migrated account {:?}",
+						id.to_ss58check()
+					)
+				}
```
muharem, line 798: you already did this with `defensive_assert!(failed == 0, "failed to unreserve")` above. or you testing additionally if unreserve is working correctly.

pallets/rc-migrator/src/accounts.rs
```diff
@@ -792,30 +822,335 @@ impl<T: Config> AccountsMigrator<T> {
...
+				rc_reserved_kept.insert(who.clone(), reserved_kept);
+				rc_free_kept.insert(who.clone(), rc_ed);
+			}
+			Self { rc_reserved_kept, rc_free_kept }
+		}
+	}
+}
+
+pub struct AccountsMigrationChecker<T>(sp_std::marker::PhantomData<T>);
```
muharem: why not having this type under `mod tests`? together with the types build for this type and have all together under std

pallets/rc-migrator/src/accounts.rs
```diff
@@ -841,38 +1176,58 @@ impl<T: Config> crate::types::RcMigrationCheck for AccountsMigrator<T> {
...
-
-					assert_eq!(
-						<T as Config>::Currency::balance(&who), *free,
-						"Incorrect free balance on the Relay Chain after the migration for account: {:?}, {:?}",
-						who.to_ss58check(), acc_state
-					);
-
+				// This corresponds to AccountState::Migrate: the account should be fully migrated
+				// to Asset Hub.
+				_ => {
```
muharem: I would not use `_`. If someone adds a new rc account state we probably do not wanna silently it fall into this branch

pallets/rc-migrator/src/accounts.rs
```diff
@@ -841,38 +1176,58 @@ impl<T: Config> crate::types::RcMigrationCheck for AccountsMigrator<T> {
...
+						);
+
+						// Preserved accounts should not have enough free balance to be partly
+						// migrated to Asset Hub.
+						let free_balance = <T as Config>::Currency::reducible_balance(
+							&who,
+							Preservation::Expendable,
+							Fortitude::Polite,
+						);
+						assert!(free_balance < <T as Config>::Currency::minimum_balance().saturating_add(T::AhExistentialDeposit::get()), "Preserved account {:?} should have not enough free balance on the relay chain after migration to be migrated to Asset Hub", who.to_ss58check());
```
muharem: please formatting for this, its very hard to read in IDE

pallets/ah-migrator/src/account.rs
```diff
@@ -262,15 +275,111 @@ impl<T: Config> crate::types::AhMigrationCheck for AccountsMigrator<T> {
...
+		let (account_summaries, _) = rc_pre_payload;
+		for (who, summary) in account_summaries {
+			// assert_eq!(summary.migrated_free, <T as Config>::Currency::balance(&who));
+			// assert_eq!(summary.migrated_reserved, <T as
+			// Config>::Currency::reserved_balance(&who));
+			let ah_free_post = <T as Config>::Currency::balance(&who);
+			let ah_reserved_post = <T as Config>::Currency::reserved_balance(&who);
+			let mut ah_free_before = 0;
+			let mut ah_reserved_before = 0;
+			if let Some((ah_reserved_pre, ah_free_pre)) = ah_pre_payload.get(&who) {
```
muharem: let (ah_free_before, ah_reserved_before) if {} else {}, and no need for mutable. always easier to follow, since I know its not gonna be mutated

pallets/ah-migrator/src/account.rs
```diff
@@ -262,15 +275,111 @@ impl<T: Config> crate::types::AhMigrationCheck for AccountsMigrator<T> {
...
 		);
 
-		// TODO: Giuseppe @re-gius
-		//   run post migration sanity checks like:
-		//    - rc_migrated_out == ah_migrated_in - failed accounts
+		let (account_summaries, _) = rc_pre_payload;
+		for (who, summary) in account_summaries {
+			// assert_eq!(summary.migrated_free, <T as Config>::Currency::balance(&who));
+			// assert_eq!(summary.migrated_reserved, <T as
+			// Config>::Currency::reserved_balance(&who));
```
muharem: what is this about?

pallets/ah-migrator/src/account.rs
```diff
@@ -262,15 +275,111 @@ impl<T: Config> crate::types::AhMigrationCheck for AccountsMigrator<T> {
...
+			let ah_migrated_balance = ah_free_post
+				.saturating_sub(ah_free_before)
+				.saturating_add(ah_reserved_post.saturating_sub(ah_reserved_before));
+			let ah_ed = <T as Config>::Currency::minimum_balance();
+
+			// In case the balance migrated to AH is less than the existential deposit, minting the
+			// balance may fail. Therefore, we just check that the difference between the balance
+			// migrated from the RC to AH and the balance delta on AH before and after migration is
+			// less than AH existential deposit.
+			assert!(rc_migrated_balance.saturating_sub(ah_migrated_balance) <= ah_ed,
```
muharem: do we really have this now? we cannot asset for equality?

pallets/ah-migrator/src/account.rs
```diff
@@ -262,15 +275,111 @@ impl<T: Config> crate::types::AhMigrationCheck for AccountsMigrator<T> {
...
+				"Total balance mismatch for account {:?} between RC pre-migration and AH post-migration",
+				who.to_ss58check()
+			);
+
+			// There are several `unreserve` operations on AH after migration (e.g., unreserve
+			// deposits for multisigs because they are not migrated to AH, adjust deposits for
+			// preimages, ...). Therefore, we just check that the change in reserved balance on AH
+			// after migration is less than the migrated reserved balance from RC.
+			assert!(ah_reserved_post.saturating_sub(ah_reserved_before) <= summary.migrated_reserved, "Change in reserved balance on AH after migration for account {:?} is greater than the migrated reserved balance from RC", who.to_ss58check());
```
muharem: also I would say not good enough. quite important assertion, but this approach leaves a lot of room for mistake. 
may be we collect also into pre check data all expected reserve releases

muharem: please also formatting

pallets/ah-migrator/src/account.rs
```diff
@@ -262,15 +275,111 @@ impl<T: Config> crate::types::AhMigrationCheck for AccountsMigrator<T> {
...
+
+			// There should be no frozen balance on AH before the migration so we just need to check
+			// that the frozen balance on AH after migration is the same as on RC before migration.
+			assert_eq!(summary.frozen, frozen, "Frozen balance mismatch for account {:?} between RC pre-migration and AH post-migration", who.to_ss58check());
```
muharem: also please formatting

pallets/ah-migrator/src/account.rs
```diff
@@ -262,15 +275,111 @@ impl<T: Config> crate::types::AhMigrationCheck for AccountsMigrator<T> {
...
+					}
+				}
+				assert_eq!(
+					rc_hold_amount_for_id,
+					hold_amount,
+					"Hold amount mismatch for account {:?} and hold id {:?} between RC pre-migration and AH post-migration",
+					who.to_ss58check(),
+					hold_id
+				);
+			}
```
muharem: I do not understand this. why not just equality? do you expect multiple holds with the same id?

pallets/ah-migrator/src/account.rs
```diff
@@ -262,15 +275,111 @@ impl<T: Config> crate::types::AhMigrationCheck for AccountsMigrator<T> {
...
+			if let Some((ah_reserved_pre, ah_free_pre)) = ah_pre_payload.get(&who) {
+				ah_free_before = *ah_free_pre;
+				ah_reserved_before = *ah_reserved_pre;
+			}
+
+			let frozen = frame_system::Account::<T>::get(&who).data.frozen;
+			let mut holds_enc = Vec::new();
+			for hold in pallet_balances::Holds::<T>::get(&who) {
+				holds_enc
+					.push((Self::hold_id_encoding(hold.id.encode(), ChainType::AH), hold.amount));
```
muharem: you could just pass the hold ids with pre check data as `Vec<u8>` (result of encode) and have a mapper type passed to the check type as type argument. that mapper type would map it to expected RC encoded id (`Vec<u8>`). I think that would look simpler and would result much less code. 

But it’s a matter of taste - your approach works. I just don’t see how we’re going to adapt it for Kusama.

pallets/ah-migrator/src/account.rs
muharem: this is minor

pallets/rc-migrator/src/accounts.rs
```diff
@@ -792,30 +822,335 @@ impl<T: Config> AccountsMigrator<T> {
...
+			account_summaries.insert(who.clone(), balance_summary);
+		}
+		(account_summaries, total_issuance)
 	}
 
-	fn post_check(rc_total_issuance_before: Self::RcPrePayload) {
+	fn post_check(rc_pre_payload: Self::RcPrePayload) {
+		let (_, rc_total_issuance_before) = rc_pre_payload;
+
+		let mut acc_state_maybe: Option<AccountStateFor<T>>;
```
muharem: minor: `acc_` not really a common var prefix for accounts in our code base

integration-tests/ahm/src/balances_test.rs
```diff
@@ -0,0 +1,133 @@
...
+		let ah_checking_balance_after = pallet_balances::Pallet::<AhRuntime>::total_balance(
+			&<AhRuntime as pallet_ah_migrator::Config>::CheckingAccount::get(),
+		);
+		// Use the rc_kept_after value computed in RcMigrationCheck::post_check
+		let rc_kept_after = RC_KEPT_AFTER
+			.lock()
+			.unwrap()
+			.expect("rc_kept_after should be set by RcMigrationCheck::post_check");
+		// `checking_balance` = ah_check_before + rc_check_before
+		let checking_balance = rc_checking_balance_before + ah_checking_balance_before;
```
muharem: not clear what is the meaning for this `checking_balance` var. by name I could imagine is what we expect the `checking_balance` to be. but its probably not.

integration-tests/ahm/src/balances_test.rs
```diff
@@ -0,0 +1,133 @@
...
+
+		// ah_check_after = `checking_balance` + rc_balance_kept - ah_total_before - ah_check_before
+		// explanation [here](https://github.com/polkadot-fellows/runtimes/blob/dev-asset-hub-migration/pallets/rc-migrator/src/accounts.md#tracking-total-issuance-post-migration)
+		assert_eq!(
+			ah_checking_balance_after,
+			checking_balance + rc_kept_after -
+				ah_total_issuance_before -
+				ah_checking_balance_before,
+			"Checking balance on asset hub after migration is incorrect"
+		);
```
muharem: here if we unwrap the `checking_balance` to `{ rc_checking_balance_before + ah_checking_balance_before }`, we get the exact assertion as above.

integration-tests/ahm/src/balances_test.rs
```diff
@@ -0,0 +1,133 @@
...
+
+		let ah_total_issuance_after = pallet_balances::Pallet::<AhRuntime>::total_issuance();
+		// There is a small difference between the total issuance before and after migration but the
+		// reason is unknown. Currently ignoring it since the difference is less than 0.1 DOT.
+		assert!(
+			ah_total_issuance_after.abs_diff(rc_total_issuance_before) < MIN_DOT_ERROR,
+			"Total issuance is not correctly tracked: before migration {} after migration {}.",
+			rc_total_issuance_before,
+			ah_total_issuance_after
```
muharem, line 127: its not equal right now?

re-gius: No, they are currently 0,0101 DOTs apart and I couldn't figure out why

integration-tests/ahm/src/balances_test.rs
re-gius: Yes, so would you suggest removing one of them?

integration-tests/ahm/src/balances_test.rs
re-gius: Yes, that's because there was a 0.01 DOT airdrop to AH checking balance account which messes up with all calculations (that's `ah_checking_balance_before`)

pallets/rc-migrator/src/accounts.rs
re-gius: Now I get it, changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/b1a9a0fe622c5c653aebe33268f2539f5668d32c

pallets/rc-migrator/src/accounts.rs
re-gius: `missing_free` may be less than the actual required free balance `rc_ed + ah_ed` because there's a `saturating_sub` before, so here I'm just checking that the account had enough rc reserved balance to free

pallets/rc-migrator/src/accounts.rs
re-gius: Why? We may keep `get_rc_free()` balance on the RC

pallets/ah-migrator/src/account.rs
re-gius: Formatted in https://github.com/polkadot-fellows/runtimes/pull/674/commits/b4cb049460c6f8a57a631d64f49e5216a471c132

pallets/ah-migrator/src/account.rs
re-gius: Formatted in https://github.com/polkadot-fellows/runtimes/pull/674/commits/b4cb049460c6f8a57a631d64f49e5216a471c132 -- I agree it's not good enough, but exact checks would be quite complex: we are changing reserves on AH after the migration, so we need to run through all pallets that change these reserves on AH and implement the logic to compute the reserved balance for such accounts before and after migration (e.g., the logic to compute deposits for preimages, which is not straightforward). Moreover, we'd need to handle inconsistencies with reserved balances. We may choose to do it just for Polkadot though if we deem it necessary

pallets/ah-migrator/src/account.rs
re-gius: Removed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/b4cb049460c6f8a57a631d64f49e5216a471c132 - it was just a leftover

pallets/rc-migrator/src/accounts.rs
re-gius: Formatted in https://github.com/polkadot-fellows/runtimes/pull/674/commits/b4cb049460c6f8a57a631d64f49e5216a471c132

pallets/rc-migrator/src/accounts.rs
re-gius: Changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/b4cb049460c6f8a57a631d64f49e5216a471c132

pallets/ah-migrator/src/account.rs
re-gius: We still need it, our code fails for this account 
5GTtcseuBoAVLbxQ32XRnqkBmxxDaHqdpPs8ktUnH1zE4Cg3
https://polkadot.subscan.io/account/15QBmCuy3aRxn8xuzfaRvzaLdawsGbPmttbcvBU8q71kEb7a

integration-tests/ahm/src/balances_test.rs
re-gius: Actually it's 0,0101087698 DOTs, of which:
- 0,0001087698 DOTs come from this account free balance, which is burnt on AH [5HBpFvUckfYEevbMnGXgGidcCRBygFww1FyksaJXYxjagPCK](https://polkadot.subscan.io/account/1687QFjgcSoi6TbsjuagQsTm43BdNZW55kiF2sHt73m6ronH)
- 0,01 DOT is a single existential deposit created out of nowhere?

pallets/ah-migrator/src/account.rs
re-gius: My bad, that account is kept on the relay chain because its total balance is less than RC existential deposit, so it shouldn't be in tests. I removed it in https://github.com/polkadot-fellows/runtimes/pull/674/commits/dd41023d90be3ab9796956429cdbae8b54a2258c , but the test with equality still fails because of account [5HBpFvUckfYEevbMnGXgGidcCRBygFww1FyksaJXYxjagPCK](https://polkadot.subscan.io/account/1687QFjgcSoi6TbsjuagQsTm43BdNZW55kiF2sHt73m6ronH) for which the free balance disappears on AH

re-gius: What happens with that account is basically:
- on RC before migration: 2,2085 total balance of which 2,2084 held (nom-pools) and 0,0001 free
- we migrate the whole balance on AH, so we have 2,2085 total balance on AH
- when we apply the hold on AH, the free balance is burned and we only end up with 2,2084 held balance (and 0 free balance)

pallets/ah-migrator/src/account.rs
re-gius: Yes, there are multiple holds with the same id sometimes

pallets/ah-migrator/src/account.rs
re-gius: If we have an easy way to understand which chain we're in, we can just have conditional mappings based on the chain

pallets/rc-migrator/src/accounts.rs
re-gius: Changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/16964690c5be01edbbd1ae08ee70cff9de1b79fd

pallets/ah-migrator/src/account.rs
re-gius: Changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/a16f02827b0064fdc91e2c252c9187649fde5572

pallets/rc-migrator/src/accounts.rs
muharem: I mean its just the same as the statement above that was removed. but this is minor

pallets/rc-migrator/src/accounts.rs
```diff
@@ -431,16 +434,10 @@ impl<T: Config> AccountsMigrator<T> {
...
-					amount: {:?}, \
-					with error: {:?}",
-					id,
-					who.to_ss58check(),
-					amount,
-					e
+			if let Err(_) = <T as Config>::Currency::release(&id, &who, amount, Precision::Exact) {
+				defensive!(
+					"There is not enough reserved balance to release the hold for account: {:?}",
+					who.to_ss58check()
```
muharem: lets include hold id, amount

re-gius: Changed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/5a8cbabe3f4d1224cf431d12b239f54f4c64b401

pallets/ah-migrator/src/account.rs
re-gius: My bad, that's not true - reverting this, only checking that holds are correctly migrated

pallets/rc-migrator/src/accounts.rs
re-gius: Removed duplicate check in https://github.com/polkadot-fellows/runtimes/pull/674/commits/b039f4d12e8d330e34dbe7237b8ffac497ec99d7

pallets/ah-migrator/src/account.rs
re-gius: Will open an issue for this

pallets/ah-migrator/src/account.rs
re-gius: Will do it with feature flags

pallets/rc-migrator/src/accounts.rs
```diff
@@ -483,21 +480,29 @@ impl<T: Config> AccountsMigrator<T> {
...
+			.free
+			.checked_sub(account_state.get_rc_free())
+			.defensive_unwrap_or_default();
 
+		// This is a debug assertion because it is expected to be true for many accounts.
+		// The RC migration of nomination pools to delegated-staking holds in the past caused
+		// many accounts to have less than the RC existential deposit free balance. When these
+		// accounts are migrated to AH, their free balance will be dusted if less than the AH
+		// existential deposit.
+		debug_assert!(teleport_free >= ah_ed, "teleport_free >= ah_ed");
```
muharem: why `debug_assert` but not `defensive_assert`? also if we consider this acceptable, this should not be `defensive_assert`. we can write an warn log. 

also why it does not panic? may be disabled somewhere and logs errors. please check also other new `defensive_assert` you are adding. we should not add new unhandled issues. we should address them or decide its acceptable and add a warning.

pallets/ah-migrator/src/account.rs
```diff
@@ -213,64 +223,180 @@ impl<T: Config> Pallet<T> {
...
+				);
+
+				let rc_holds = summary
+					.holds
+					.iter()
+					.map(|(id, amount)| (Self::rc_hold_id_encoding_to_ah(id.clone()), *amount))
+					.collect::<Vec<(Vec<u8>, u128)>>();
+				// Check that all holds from RC are present on AH post-migration.
+				assert!(
+					ah_holds.iter().all(|(id, amount)| rc_holds.contains(&(id.clone(), *amount))),
```
muharem: it should be at least `rc_holds.iter`, otherwise if some hold is not migrated this test gonna miss it. 
or you can just sort vectors and assert equal

pallets/ah-migrator/src/account.rs
```diff
@@ -213,64 +223,180 @@ impl<T: Config> Pallet<T> {
...
+					.collect::<Vec<(Vec<u8>, u128)>>();
+				// There should be no freezes on AH before the migration so we just need to check
+				// that the freezes on AH after migration are the same as on RC before
+				// migration.
+				assert_eq!(
+					rc_freezes,
+					ah_freezes,
+					"Freezes mismatch for account {:?} between RC pre-migration and AH post-migration",
+					who.to_ss58check()
+				);
```
muharem: are you sure your vectors sorter? this works now, but might not later.

re-gius: Sorted everything in https://github.com/polkadot-fellows/runtimes/pull/674/commits/2e0e83d0c7210e6b2ec8a6c8b70db6fc7b05c6d2

pallets/ah-migrator/src/account.rs
re-gius: Fixed in https://github.com/polkadot-fellows/runtimes/pull/674/commits/2e0e83d0c7210e6b2ec8a6c8b70db6fc7b05c6d2
Sorted holds by ids + removed diff with ah holds pre migration

pallets/rc-migrator/src/accounts.rs
re-gius: Changed to `log::warn` since this is expected and we cannot really do anything about it
https://github.com/polkadot-fellows/runtimes/pull/674/commits/2e0e83d0c7210e6b2ec8a6c8b70db6fc7b05c6d2

muharem (approved):

integration-tests/ahm/src/balances_test.rs
ggwpez: Its a good start. We can fix those things and then decrease the min error.

pallets/ah-migrator/src/account.rs
```diff
@@ -224,64 +234,197 @@ impl<T: Config> Pallet<T> {
...
+		/// Run some checks after the migration and use the intermediate payload.
+		///
+		/// The expected input should contain the data just transferred out of the relay chain, to
+		/// allow the check that data has been correctly migrated to asset hub. It should also
+		/// contain the data previously stored in asset hub, allowing for more complex logical
+		/// checks on the migration outcome.
+		fn post_check(rc_pre_payload: Self::RcPrePayload, ah_pre_payload: Self::AhPrePayload) {
+			// Check that no failed accounts remain in storage
+			assert!(
+				RcAccounts::<T>::iter().next().is_none(),
```
ggwpez, line 304: FYI if you want the console to print the erroneous account it would be better to use `assert_eq!(RcAccounts::<T>::iter().next(), None)`

ggwpez (approved):

PR #677, opened 2025-04-17
[AHM] Nom_pools test

eugypalu:
Add pre/post migration checks to nom_pools

### DIFF
--- a/pallets/ah-migrator/src/types.rs
+++ b/pallets/ah-migrator/src/types.rs
@@ -65,7 +65,7 @@ pub trait AhMigrationCheck {
 	fn post_check(rc_pre_payload: Self::RcPrePayload, ah_pre_payload: Self::AhPrePayload);
 }
 
-#[impl_trait_for_tuples::impl_for_tuples(16)]
+#[impl_trait_for_tuples::impl_for_tuples(24)]
 impl AhMigrationCheck for Tuple {
 	for_tuples! { type RcPrePayload = (#( Tuple::RcPrePayload ),* ); }
 	for_tuples! { type AhPrePayload = (#( Tuple::AhPrePayload ),* ); }
--- a/pallets/rc-migrator/src/types.rs
+++ b/pallets/rc-migrator/src/types.rs
@@ -171,7 +171,7 @@ pub trait RcMigrationCheck {
 	fn post_check(rc_pre_payload: Self::RcPrePayload);
 }
 
-#[impl_trait_for_tuples::impl_for_tuples(16)]
+#[impl_trait_for_tuples::impl_for_tuples(24)]
 impl RcMigrationCheck for Tuple {
 	for_tuples! { type RcPrePayload = (#( Tuple::RcPrePayload ),* ); }
 


pallets/rc-migrator/src/staking/nom_pools.rs
```diff
@@ -328,3 +334,284 @@ impl<T: pallet_nomination_pools::Config> NomPoolsMigrator<T> {
 		LastPoolId::<T>::put(values.last_pool_id);
 	}
 }
+
+#[derive(Encode, Decode, MaxEncodedLen, TypeInfo, DebugNoBound, PartialEq, Clone)]
+pub struct GenericCommission<AccountId: Debug, BlockNumber: Debug> {
```
muharem, line 339: these types and imports probably causing unused import warnings. you can just create sub modules under cfg feature std here, and move everything under it.

pallets/ah-migrator/src/staking/nom_pools.rs
```diff
@@ -141,3 +144,230 @@ impl<T: Config> Pallet<T> {
...
 }
+
+fn normalize_generic_bonded_pool<T: Config>(
+	mut pool: GenericBondedPoolInner<BalanceOf<T>, T::AccountId, BlockNumberFor<T>>,
+) -> GenericBondedPoolInner<BalanceOf<T>, T::AccountId, BlockNumberFor<T>> {
+	use crate::Pallet;
+
+	if let Some(ref mut throttle_from) = pool.commission.throttle_from {
+		*throttle_from = Pallet::<T>::rc_to_ah_timepoint(*throttle_from);
+		// .saturating_add(One::one());
```
muharem, line 155: should be removed?

pallets/ah-migrator/src/staking/nom_pools.rs
```diff
@@ -141,3 +144,230 @@ impl<T: Config> Pallet<T> {
...
 	}
 }
+
+fn normalize_generic_bonded_pool<T: Config>(
+	mut pool: GenericBondedPoolInner<BalanceOf<T>, T::AccountId, BlockNumberFor<T>>,
+) -> GenericBondedPoolInner<BalanceOf<T>, T::AccountId, BlockNumberFor<T>> {
+	use crate::Pallet;
+
+	if let Some(ref mut throttle_from) = pool.commission.throttle_from {
+		*throttle_from = Pallet::<T>::rc_to_ah_timepoint(*throttle_from);
```
muharem, line 154: not the best way to test it. ideally you would write your own map function to test the result of `rc_to_ah_timepoint`.
but I think we wont need the `rc_to_ah_timepoint`, since we should use RC block times when we get block provider types with 2503 sdk. so you can keep this as it is, just take in consideration for other tests.

pallets/ah-migrator/src/staking/nom_pools.rs
```diff
@@ -141,3 +144,230 @@ impl<T: Config> Pallet<T> {
...
+		// .saturating_add(One::one());
+	}
+	if let Some(ref mut change_rate) = pool.commission.change_rate.as_mut() {
+		change_rate.min_delay =
+			T::RcToAhDelay::convert(change_rate.min_delay).saturating_add(One::one());
```
muharem, line 159: same here, ideally you need your own mapping, better even not looking into impl of `RcToAhDelay`, but from knowing what is expected. otherwise your test is:
``` rust
fn test_sum() {
  let expected_sum = sum(a + b);
  let actual_sum = sum(a + b);
  assert_eq(expected_sum, actual_sum);
}
```

pallets/rc-migrator/src/staking/nom_pools.rs
```diff
@@ -328,3 +334,284 @@ impl<T: pallet_nomination_pools::Config> NomPoolsMigrator<T> {
...
+		}
+
+		// Collect bonded pools
+		for (pool_id, pool) in pallet_nomination_pools::BondedPools::<T>::iter() {
+			let generic_pool = GenericBondedPoolInner {
+				commission: GenericCommission {
+					current: pool.commission.current,
+					max: pool.commission.max,
+					change_rate: pool.commission.change_rate,
+					throttle_from: None, // None to avoid discrepancies during the AH postcheck
```
muharem, line 486: I would calculate here what is expected on post check and not follow what the migration does (extract with RC pallet, map with AH pallet). Like with AAA tests, pre_check here for me would be first A, second A would be the migration itself, and last A would be in AH post check.

pallets/rc-migrator/src/staking/nom_pools.rs
eugypalu: Fixed, thanks

pallets/ah-migrator/src/staking/nom_pools.rs
eugypalu: Thanks, I missed it. Removed

pallets/rc-migrator/src/staking/nom_pools.rs
eugypalu: You're right. I moved the calculation of `change_rate` to the RC pre-check and removed `normalize_generic_bonded_pool` function from AH

muharem:
CIs jobs failing

pallets/rc-migrator/src/staking/nom_pools.rs
```diff
@@ -19,8 +19,17 @@
...
+use pallet_nomination_pools::{
+	BondedPoolInner, ClaimPermission, CommissionChangeRate, CommissionClaimPermission, PoolId,
+	PoolMember, PoolRoles, PoolState,
+};
+use sp_runtime::{
+	traits::{One, Zero},
+	Perbill, Saturating,
+};
+use sp_staking::EraIndex;
+use sp_std::fmt::Debug;
```
muharem, line 32: these imports if only used for std items should trigger warnings. 
you could do something like

``` rust 
#[cfg(feature = "std")]
pub mod tests {
  use ...; // all your imports 

  // all your new types 
  pub struct GenericCommission
  
  // the type implementing test trait 
  pub struct NomPoolsMigratorTest
}
```

eugypalu:
> CIs jobs failing

Fixed, now it works

PR #679, opened 2025-04-17
[Encointer] use XCM V5 to remotely transfer funds from encointer treasury accounts

clangenb:
WIP


clangenb:
@franciscoaguirre, @acatangiu taging you so that you won't forget ;)

acatangiu:
Not forgotten, but still busy fighting fires, didn't yet look at this 😢

clangenb:
Bump :)

system-parachains/encointer/src/treasuries_xcm_payout.rs
```diff
@@ -0,0 +1,179 @@
...
+	type Error = xcm::latest::Error;
+
+	fn transfer(
+		from: &Self::Payer,
+		to: &Self::Beneficiary,
+		asset_kind: Self::AssetKind,
+		amount: Self::Balance,
+	) -> Result<Self::Id, Self::Error> {
+		let locatable = AssetKindToLocatableAsset::try_convert(asset_kind)
+			.map_err(|_| xcm::latest::Error::InvalidLocation)?;
```
franciscoaguirre: We are moving towards not ignoring this error and logging it

system-parachains/encointer/src/treasuries_xcm_payout.rs
```diff
@@ -0,0 +1,179 @@
...
+		let beneficiary = TransactorRefToLocation::try_convert(&to)
+			.map_err(|_| xcm::latest::Error::InvalidLocation)?;
+
+		let query_id = Querier::new_query(
+			asset_location.clone(),
+			Timeout::get(),
+			from_location.interior.clone(),
+		);
+
+		let fee_asset = fee_asset(4 * ONE_KSM / 10);
```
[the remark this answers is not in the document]
franciscoaguirre: Correct, but adding the additional condition that the `WeightToFee` implementation doesn't change. For using `query_xcm_weight` you have to first build an initial XCM, get the weight and use that to put the correct fee in `WithdrawAsset` and `PayFees`

system-parachains/encointer/src/treasuries_xcm_payout.rs
```diff
@@ -0,0 +1,179 @@
...
+				RefundSurplus,
+				DepositAsset {
+					assets: AssetFilter::Wild(WildAsset::All),
+					beneficiary: from_at_target,
+				},
+			])),
+			TransferAsset { beneficiary, assets: (asset_id, amount).into() },
+		]);
+
+		let (ticket, _) = Router::validate(&mut Some(destination), &mut Some(message))?;
```
franciscoaguirre: If this is always executed by the treasury then it's fine to ignore the delivery fee (the second element in the tuple) but I would still add a comment and name it `_delivery_fee` instead.

system-parachains/encointer/src/tests/unit_tests.rs
```diff
@@ -0,0 +1,164 @@
...
+			Weight::zero(),
+		);
+
+		assert_eq!(mock::Assets::balance(1, &recipient), amount);
+
+		let expected_lower_bound = INITIAL_BALANCE - amount - fee_amount;
+
+		// Fixme: Why is the lower bound == actual, even when changing the fee amount
+		println!("Lower Bound	{:?}", expected_lower_bound);
+		println!("Actual 		{:?}", mock::Assets::balance(1, &sender_account_on_target));
```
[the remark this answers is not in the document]
franciscoaguirre: Could you share the actual values from before and after?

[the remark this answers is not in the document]
clangenb: Thanks for looking into this! I have the full trace log output here, https://gist.github.com/clangenb/a43f0fc549b0c53fd2596f7eeaf6c735

And these are the requested before after values:

```console
---- tests::unit_tests::transfer_over_xcm_works stdout ----
Initial Balance: 100000000000000
TransferAmount: 10000000000000
PayFeesAmount: 400000000000
Balance After Transfer
Expected (Initial - TransferAmount - PayFeesAmount):89600000000000
Actual 89600000000000
thread 'tests::unit_tests::transfer_over_xcm_works' panicked at system-parachains/encointer/src/tests/unit_tests.rs:166:9:
assertion failed: mock::Assets::balance(1, &sender_account_on_target) > expected_lower_bound
```

And if I increase the `PayFeesAmount`, the actual will decrease by the corresponding amount.

system-parachains/encointer/src/treasuries_xcm_payout.rs
clangenb: Cool, thanks for the hint. But what is your reasoning behind the intuition: "it is fine to ignore the delivery fee if this is executed by the treasury"?

system-parachains/encointer/src/tests/unit_tests.rs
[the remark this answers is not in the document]
clangenb: I found the issue. The `DummyWeightTrader` of the mock has not refunded anything. It works now.

integration-tests/emulated/tests/encointer/encointer-kusama/src/tests/remote_treasury_payout.rs
```diff
@@ -0,0 +1,256 @@
...
+		.into()),
+	};
+
+	assert_eq!(treasury_location_on_ah(), loc);
+
+	let account_on_asset_hub =
+		asset_hub_kusama_runtime::xcm_config::LocationToAccountId::convert_location(&loc).unwrap();
+
+	assert_eq!(account_on_asset_hub, treasury_account_on_ah());
+}
```
clangenb: I have no idea why this test fails with `FailedToTransactAssets`. We seem to fund the correct account in the setup, and the almost identical test works in the unit-test with the mock runtime. Do you have any idea what I am missing here? @franciscoaguirre

system-parachains/encointer/src/treasuries_xcm_payout.rs
```diff
@@ -0,0 +1,285 @@
...
+use xcm::{latest::Error, opaque::lts::Weight, prelude::*};
+use xcm_builder::LocatableAssetId;
+use xcm_executor::traits::{ConvertLocation, QueryHandler, QueryResponseStatus};
+
+pub use pallet_encointer_treasuries::Transfer;
+
+// This is the value that has been queried from the Asset Hub Kusama runtime.
+// There is an integration test in `integration-tests/emulated/tests/encointer/encointer-kusama/
+// That verifies that this fee is correct and will catch fee changes in Asset-Hub Kusama
+pub const REMOTE_XCM_TRANSFER_REMOTE_EXECUTION_FEE: u128 = 12749033321;
```
clangenb: Do you agree that this is a safe approach? We hardcode this value, and there is a test in the integration tests that would detect if the fee computation changes in the asset hub.

integration-tests/emulated/tests/encointer/encointer-kusama/src/tests/remote_treasury_payout.rs
clangenb: This was an error on my end. Fixed it.

integration-tests/emulated/tests/encointer/encointer-kusama/src/tests/remote_treasury_payout.rs
```diff
@@ -0,0 +1,242 @@
...
+	<EncointerKusama as TestExt>::execute_with(|| {
+		encointer_kusama_runtime::TransferOverXcm::from_on_remote(
+			&treasury_account,
+			asset_kind.clone(),
+		)
+		.unwrap()
+	})
+}
+
+fn treasury_account_on_ah() -> AccountId {
```
franciscoaguirre:
```suggestion
fn encointer_treasury_sov_account_on_ah() -> AccountId {
```

Would be less confusing

system-parachains/encointer/src/lib.rs
```diff
@@ -585,12 +597,65 @@ impl pallet_encointer_democracy::Config for Runtime {
...
+	pub const AnyNetwork: Option<NetworkId> = None;
 }
+
+pub type TransferOverXcm = crate::treasuries_xcm_payout::TransferOverXcm<
+	crate::xcm_config::XcmRouter,
+	crate::PolkadotXcm,
+	ConstU32<{ 6 * HOURS }>,
+	AccountId,
+	// Needs to match AssetKind of the encointer_treasuries::Config
+	VersionedLocatableAsset,
```
franciscoaguirre: Instead of adding this comment, you could create a common type alias that's used in both places

system-parachains/encointer/src/tests/unit_tests.rs
```diff
@@ -0,0 +1,190 @@
...
+
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Tests for making sure `PayOverXcm::pay` generates the correct message and sends it to the
```
franciscoaguirre:
```suggestion
//! Tests for making sure `TransferOverXcm::transfer` generates the correct message and sends it to the
```

system-parachains/encointer/src/treasuries_xcm_payout.rs
```diff
@@ -0,0 +1,292 @@
...
+//
+// Encointer is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with Encointer.  If not, see <http://www.gnu.org/licenses/>.
+
+//! `PayOverXcm` struct for paying through XCM and getting the status back.
```
franciscoaguirre:
```suggestion
//! `TransferOverXcm` struct for paying through XCM and getting the status back.
```

franciscoaguirre (approved):

system-parachains/encointer/src/lib.rs
clangenb: It does use the same type alias, but yeah. The comment is misleading - I will fix the comment. Thanks a lot!

integration-tests/emulated/tests/encointer/encointer-kusama/src/tests/remote_treasury_payout.rs
clangenb: Thanks, you are absolutely right!

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Use `pallet-assets` instead of `pallet-balances` for XCM benchmarks on asset hubs ([polkadot-fellows/runtimes/pull/758](https://github.com/polkadot-fellows/runtimes/pull/758))
   - This means XCM benchmarks will have a higher weight.
 - All XCM benchmarks use sibling parachain as destination instead of Relay chain to properly adapt weights in context of incoming migration from Relay to Asset Hub ([polkadot-fellows/runtimes/pull/709](https://github.com/polkadot-fellows/runtimes/pull/709))
+- [Encointer] use XCM V5 to remotely spend funds from encointer treasury accounts on AHK [polkadot-fellows/runtimes/pull/679](https://github.com/polkadot-fellows/runtimes/pull/679)
```
acatangiu: this should be under `Added` not `Fixed`

system-parachains/encointer/src/xcm_config.rs
```diff
@@ -73,8 +73,8 @@ pub type LocationToAccountId = (
 	SiblingParachainConvertsVia<Sibling, AccountId>,
 	// Straight up local `AccountId32` origins just alias directly to `AccountId`.
 	AccountId32Aliases<RelayNetwork, AccountId>,
-	// Here/local root location to `AccountId`.
-	HashedDescription<AccountId, DescribeTerminus>,
+	// Foreign locations alias into accounts according to a hash of their standard description.
+	HashedDescription<AccountId, DescribeFamily<DescribeAllTerminal>>,
```
acatangiu: WARNING: [DescribeTerminus](https://github.com/paritytech/polkadot-sdk/blob/1e4af2353ea9dcb9ad0afc1ce63b03df68108ecf/polkadot/xcm/xcm-builder/src/location_conversion.rs#L47) is NOT included in [DescribeFamily<DescribeAllTerminal>>](https://github.com/paritytech/polkadot-sdk/blob/1e4af2353ea9dcb9ad0afc1ce63b03df68108ecf/polkadot/xcm/xcm-builder/src/location_conversion.rs#L122).

(it is included in `DescribeAllTerminal` but notice the `DescribeFamily` outer wrapping that changes the location and derivation seed)

ONLY remove `HashedDescription<AccountId, DescribeTerminus>,` if and ONLY IF you are sure that local Encointer Root location does NOT need a derived account. My guess is that it does, in which case, don't remove it, simply add `HashedDescription<AccountId, DescribeFamily<DescribeAllTerminal>>` after it.

system-parachains/encointer/src/treasuries_xcm_payout.rs
```diff
@@ -0,0 +1,292 @@
...
+use xcm::{latest::Error, opaque::lts::Weight, prelude::*};
+use xcm_builder::LocatableAssetId;
+use xcm_executor::traits::{QueryHandler, QueryResponseStatus};
+
+pub use pallet_encointer_treasuries::Transfer;
+
+// This is the value that has been queried from the Asset Hub Kusama runtime.
+// There is an integration test in `integration-tests/emulated/tests/encointer/encointer-kusama/
+// That verifies that this fee is correct and will catch fee changes in Asset-Hub Kusama
+pub const REMOTE_XCM_TRANSFER_REMOTE_EXECUTION_FEE: u128 = 12905733320;
```
acatangiu, line 33: this is brittle - it can easily break when Asset Hub Kusama and Encointer runtimes are not upgraded at the same time (e.g. if there is a patch release for AHK fixing some bug that's also changing some weights/fees).

if you want to use this in the initial feature launch, that's fine, but I strongly suggest to change it to something more resilient.

ideas/alternatives:
- (flexible) allow configurable fee in the Encointer treasury proposal so that it is controllable off-chain
- (easiest and most robust) allow UnpaidExecution on AHK from Encointer Treasury
- (cool) add a "permanent migration" to AHK that publishes new fees to Encointer whenever they change (could be enhanced to a general on-runtime-upgrade pub-sub pallet)

while writing this I realized only the unpaid option is truly robust, as the fee could change even for same runtime version (no runtime upgrade/change), when/if there is high network congestion.. maybe add some buffer :)

acatangiu (approved):
generally looks good

system-parachains/encointer/src/xcm_config.rs
clangenb: Ohh, thank so very much for the clarification. I absolutely did not intend this. I will reintroduce it. 👍

system-parachains/encointer/src/treasuries_xcm_payout.rs
clangenb: Yes, thanks a lot for the emphasis, and also your creative inspiration. We are aware of the brittleness. We have implemented an offchain-watchdog service for Integritee, and it would probably make sense to have something similar for encointer too. The watchdog will would have two responsibilities:

1. Execute dry-run to see if it is operable, and give a heartbeat if yes.
2. Publish updated fees if it is not operable.

Unpaid execution is unfortunately not an option I assume because there can be unlimited treasury accounts in theory (1 for each community).

I really like the "permanent migration" option, I will think about it some more.

We are aware of the brittleness, and will probably improve the architecture regarding this, but we would definitely love to launch it as is now.

Thanks a lot again for your detailed thoughts!

integration-tests/emulated/tests/encointer/encointer-kusama/src/tests/mod.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2023 Encointer Association
+// This file is part of Encointer
```
ggwpez: Hm, i think we should stick with the fellowship Polkadot license.  

PS: okay i see we also have the encointer runtimes with their own license.

system-parachains/encointer/Cargo.toml
```diff
@@ -107,12 +107,15 @@ polkadot-core-primitives = { workspace = true }
...
-[dev-dependencies]
-parachains-runtimes-test-utils = { workspace = true }
-
 [build-dependencies]
 substrate-wasm-builder = { optional = true, workspace = true }
 
+[dev-dependencies]
+pallet-assets = { workspace = true, features = ["std"] }
+parachains-runtimes-test-utils = { workspace = true }
+sp-io = { workspace = true, features = ["std"] }
```
ggwpez: `default-features = true` should be enough?

ggwpez (approved):
Encointer is not a trusted teleporter, or?

acatangiu:
> Encointer is not a trusted teleporter, or?

it is a trusted teleporter, it's para id is in the sys chains interval, and that's what's used for differentiating trusted teleporters

integration-tests/emulated/tests/encointer/encointer-kusama/src/tests/mod.rs
acatangiu: @clangenb please fix these remaining comments

system-parachains/encointer/Cargo.toml
clangenb: This is true, I don't know where I copied that explicit pattern from. I shall harmonize this in the future if this is the pattern that others use too.

integration-tests/emulated/tests/encointer/encointer-kusama/src/tests/mod.rs
clangenb: I personally don't have a strong opinion on the license. The runtime had it's encointer license due to historical reasons, and I suggest keeping it like that. However, I will change it to the encointer license everywhere else.

system-parachains/encointer/Cargo.toml
clangenb: Fixed it after all

integration-tests/emulated/tests/encointer/encointer-kusama/src/tests/mod.rs
clangenb: I have changed the license in the integration tests, but keep the runtime to our own license for the time being.

cc @brenzi

system-parachains/encointer/src/lib.rs
```diff
@@ -585,12 +598,72 @@ impl pallet_encointer_democracy::Config for Runtime {
...
+	type Paymaster = impls::benchmarks::TransferWithEnsure<
+		TransferOverXcm,
+		impls::benchmarks::OpenHrmpChannel<ConstU32<1000>>,
+	>;
+	#[cfg(feature = "runtime-benchmarks")]
+	type BenchmarkHelper =
+		benchmarks_helper::TreasuryArguments<sp_core::ConstU8<1>, ConstU32<1000>>;
+}
+
+#[cfg(feature = "runtime-benchmarks")]
```
bkontur, line 635: @clangenb just nit, this benchmarking code could also go to the `impl.rs` -> `pub mod benchmarks`

system-parachains/encointer/src/tests/unit_tests.rs
```diff
@@ -0,0 +1,190 @@
...
+/// Account #3 on the local chain, parachain 42, controls an amount of funds on parachain 2.
+/// [`PayOverXcm::pay`] creates the correct message for account #3 to pay another account, account
+/// #5, on parachain 1000, remotely, in its native token.
+#[test]
+fn transfer_over_xcm_works() {
+	sp_tracing::init_for_tests();
+
+	let sender = AccountId::new([1u8; 32]);
+	let sender_location_on_target =
+		Location::new(1, X2([Parachain(42), AccountId32 { network: None, id: [1; 32] }].into()));
```
bkontur, line 71: @clangenb just an nit, using too low values for parachain could be tricky in some scenarios (bellow 2000 are considered system parachains).

system-parachains/encointer/src/tests/unit_tests.rs
```diff
@@ -0,0 +1,190 @@
...
+}
+
+parameter_types! {
+	pub SenderAccount: AccountId = AccountId::new([3u8; 32]);
+	pub InteriorAccount: InteriorLocation = AccountId32 { id: SenderAccount::get().into(), network: None }.into();
+	pub InteriorBody: InteriorLocation = Plurality { id: BodyId::Treasury, part: BodyPart::Voice }.into();
+	pub Timeout: BlockNumber = 5; // 5 blocks
+}
+
+/// Scenario:
```
bkontur, line 61: @clangenb I didn't walk through these test, but just curios, if we could use emulated tests (EncointerKusama, AssetHubKusama, AssetHubPolkadot, PenpalKusama, PenpalPolkadot) to simulate this? That way we won't need all this mocking stuff.

system-parachains/encointer/src/treasuries_xcm_payout.rs
```diff
@@ -0,0 +1,291 @@
...
+		// Transform origin into Location::new(1, X2([Parachain(SourceParaId), from.interior }])
+		DescendOrigin(from_location.interior.clone()),
+		// For simplicity, we assume now that the treasury has KSM and pays fees with KSM.
+		WithdrawAsset(vec![remote_fee.clone()].into()),
+		PayFees { asset: remote_fee },
+		SetAppendix(Xcm(vec![
+			ReportError(QueryResponseInfo {
+				destination: destination.clone(),
+				query_id,
+				max_weight: Weight::zero(),
```
bkontur, line 277: @clangenb is `Weight::zero()` really working? If I am checking it correctly, it should end up in the: https://github.com/paritytech/polkadot-sdk/blob/master/polkadot/xcm/pallet-xcm/src/lib.rs#L4052-L4064
```
if weight.any_gt(max_weight) {
								let e = Event::NotifyOverweight {
```
Ok, maybe not that critical if the `ReportError` does not work, but still :)

system-parachains/encointer/src/weights/pallet_encointer_treasuries.rs
```diff
@@ -62,4 +62,15 @@ impl<T: frame_system::Config> pallet_encointer_treasuries::WeightInfo for Weight
 			.saturating_add(T::DbWeight::get().reads(6))
 			.saturating_add(T::DbWeight::get().writes(5))
 	}
+
+	fn swap_asset() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `733`
+		//  Estimated: `6196`
+		// Minimum execution time: 70_344_000 picoseconds.
+		Weight::from_parts(72_445_000, 0)
```
bkontur, line 71: @clangenb Did we run fresh benchmarks for this? I think we should better run maybe as follow up - with mentioned small nits

system-parachains/encointer/src/treasuries_xcm_payout.rs
```diff
@@ -0,0 +1,291 @@
...
+		asset_kind: <Self as Transfer>::AssetKind,
+	) -> Result<LocatableAssetId, Error> {
+		AssetKindToLocatableAsset::try_convert(asset_kind).map_err(|e| {
+			log::error!("Could not convert asset kind to locatable asset: {e:?}");
+			Error::InvalidLocation
+		})
+	}
+}
+
+pub fn remote_transfer_xcm(
```
bkontur, line 254: @clangenb I would either add some docs for `pub` or just remove `pub`, because it is not used anywhere else

system-parachains/encointer/src/treasuries_xcm_payout.rs
```diff
@@ -0,0 +1,291 @@
...
+			RefundSurplus,
+			DepositAsset { assets: AssetFilter::Wild(WildAsset::All), beneficiary: from_at_target },
+		])),
+		TransferAsset { beneficiary, assets: (asset_id, amount).into() },
+	]);
+
+	Ok(xcm)
+}
+
+pub fn append_from_to_target(from: Location, target: Location) -> Result<Location, Error> {
```
bkontur, line 288: @clangenb also, remove `pub` please

system-parachains/encointer/src/tests/unit_tests.rs
clangenb: I intend to [upstream](https://github.com/paritytech/polkadot-sdk/pull/9173) the `Transfer` trait and a `TransferOverXCM` generic implementation. In that process, I will move this test upstream. There is already [mocking](https://github.com/paritytech/polkadot-sdk/tree/1a512570552119a49a8ecb2abfb7021954c4422d/polkadot/xcm/xcm-builder/src/tests/pay) in place there, which I intend to re-use.

I agree with you that we can then delete the mocking and this unit test from here. 👍

system-parachains/encointer/src/treasuries_xcm_payout.rs
clangenb: Hmm, interesting. I suspect that the mock that I copied from upstream is lacking some logic here. I will have a closer look here while upstreaming that part. 👍

system-parachains/encointer/src/weights/pallet_encointer_treasuries.rs
clangenb: Good catch, you are right. This call has not been properly benchmarked! Do I as a mere rank I member have the rights to run the benchmarks?

bkontur: > ... Do I as a mere rank I member have the rights to run the benchmarks?

Yes, you can :)

system-parachains/encointer/src/lib.rs
clangenb: e9fc1cd40d68df747f44268a6d8b5f6d42c9b620

system-parachains/encointer/src/tests/unit_tests.rs
clangenb: Good point. I copied this from upstreams `Pay` unit test. However, the `Pay` trait assumes free execution on the target parachain, so hence it was a reasonable assumption to put 42 as the parachain. I will make this a higher number when I move these tests upstream. 👍

system-parachains/encointer/src/treasuries_xcm_payout.rs
clangenb: e9fc1cd40d68df747f44268a6d8b5f6d42c9b620

system-parachains/encointer/src/treasuries_xcm_payout.rs
clangenb: e9fc1cd40d68df747f44268a6d8b5f6d42c9b620

Issue #644, opened 2025-03-26
[AHM] [tracking] Integration testing for pallet migration


re-gius:
This is a **tracking issue**.

Follow the pallet testing framework from #612 to implement integration tests for pallet migrations to asset hub. The guidelines (from [here](https://github.com/polkadot-fellows/runtimes/issues/555#issuecomment-2602465484)) are:

> * Ensure that all the data was moved over correctly
> * Ensure that the pallets still work as expected (ie. Proxy types are correctly mapped, Multisig is re-creatable)
> * Check that the accounts have the correct deposits
> * Data was removed from the Relay (if applicable)

You can start working on pallets for which the migration is complete, and you can find them in #555 .

When you write tests, you should follow the comment guidelines from [here](https://github.com/paritytech/ahm-dryrun/pull/14), so that the testing coverage can be automatically updated on this [spreadsheet](https://docs.google.com/spreadsheets/d/1Mjb7iyxINSLMI8D7RhMKceYiCj65a3-h7zCc1j_gaa8/edit?gid=0#gid=0) (second tab).

👉 If you want to work on this: Please check that the pallet migration has been completed and that there is not already a merge request for the pallet that you want to work on (here or in the comments below). Please pick only one or a small set of pallets - not all at once. This keeps review times low. Add a comment below to announce the pallets you're working on.

## Pallet list

- [x] `claims` : #622 
- [ ] `crowdloans` : #622 
- [ ] `pallet-asset-rate` : #660 
- [x] `pallet-bags-list` : #632 
- [x] `pallet-balances` (accounts with balances) : #628 + #655 + #674 
- [ ] `pallet-bounties` : #669 
- [x] `pallet-conviction-voting` : #632 
- [ ] `pallet-fast-unstake` : #658 
- [x] `pallet-indices` : #612 
- [x] `pallet-multisig` #666 
- [ ] `pallet-nomination-pools` #677 
- [ ] `pallet-preimage` : #619 + #643 + #838 
- [x] `pallet-proxy` : #621 + #707 
- [x] `pallet-referenda` #672 + #713 
- [x] `pallet-scheduler` #680 
- [ ] `pallet-staking` #727 
- [x] `pallet-treasury` #664 + #671 
- [x] `pallet-vesting` : #615

re-gius:
Since https://github.com/polkadot-fellows/runtimes/pull/642 these tests also run in CI, so make sure your tests pass before merging them into the asset hub development branch.

[...]

PR #680, opened 2025-04-18
[AHM] Scheduler storage item checks

PolkadotDom:
This PR includes checks for the Scheduler pallet storage items #644

Note:
When the Agenda storage map is being placed on AH, the outcome is conditional on pallet-preimage state. Not only this, but it alters the preimage state as well. So for the reconstruction of this logic in the checks I had to remove this dependency. I did so by grabbing the dependent data (Each agenda's schedule's call) and sending it over.

Everything else is more or less as you'd expect.


integration-tests/ahm/src/tests.rs
```diff
@@ -83,6 +83,7 @@ pub type RcPolkadotChecks = (
 	pallet_rc_migrator::referenda::ReferendaMigrator<Polkadot>,
 	pallet_rc_migrator::claims::ClaimsMigrator<Polkadot>,
 	pallet_rc_migrator::crowdloan::CrowdloanMigrator<Polkadot>,
+	pallet_rc_migrator::scheduler::SchedulerMigrator<Polkadot>,
```
ggwpez, line 86: It does not work on Westend? But westend also has the scheduler pallet, right?

pallets/ah-migrator/src/scheduler.rs
```diff
@@ -136,3 +137,159 @@ impl<T: Config> Pallet<T> {
...
+			.into_iter()
+			.filter_map(|(block_number, rc_tasks, rc_task_call_encodings)| {
+				let mut ah_tasks = Vec::new();
+				let tasks = rc_tasks.into_iter();
+				let encoded_calls = rc_task_call_encodings.into_iter();
+
+				// Iterate over task and it's corresponding encoded call.
+				for (rc_task_opt, encoded_call_opt) in tasks.zip(encoded_calls) {
+					// Check both task and encoded call exist.
+					if let (Some(rc_task), Some(encoded_call)) = (rc_task_opt, encoded_call_opt) {
```
ggwpez, line 207: You can use let..else to avoid nesting here and keep code more readable.

pallets/ah-migrator/src/scheduler.rs
```diff
@@ -136,3 +137,159 @@ impl<T: Config> Pallet<T> {
...
+						// Attempt origin conversion.
+						let ah_origin =
+							match T::RcToAhPalletsOrigin::try_convert(rc_task.origin.clone()) {
+								Ok(origin) => origin,
+								Err(_) => {
+									// Origin conversion failed, skip task.
```
ggwpez, line 213: How often do we expect this to happen? Maybe it should print some error in the tests?

pallets/ah-migrator/src/scheduler.rs
```diff
@@ -136,3 +137,159 @@ impl<T: Config> Pallet<T> {
...
+									continue;
+								},
+							};
+
+						// Attempt call conversion.
+						let ah_call = match Pallet::<T>::map_rc_ah_call_no_preimage(encoded_call) {
+							Ok(call) => call,
+							// Conversion failed, skip task.
+							Err(_) => return None,
```
ggwpez, line 222: Hm, maybe also print something please.

pallets/ah-migrator/src/scheduler.rs
```diff
@@ -136,3 +137,159 @@ impl<T: Config> Pallet<T> {
...
+
+		// Collect agendas on AH.
+		let mut ah_agenda: Vec<_> = pallet_rc_migrator::scheduler::alias::Agenda::<T>::iter()
+			.map(|(bn, tasks)| (bn, tasks.into_inner()))
+			.collect();
+
+		// Assert storage 'Scheduler::Agenda::ah_post::length'
+		assert_eq!(
+			ah_agenda.len(),
+			expected_ah_agenda.len(),
```
ggwpez, line 253: There is some oddity with the scheduler that it can also have an all empty agenda, so if it fails here then we know that it could be triggered by that.

pallets/rc-migrator/src/scheduler.rs
```diff
@@ -284,3 +284,107 @@ pub mod alias {
...
+		let agenda_and_call_encodings: Vec<_> = alias::Agenda::<T>::iter()
+			.map(|(bn, tasks)| {
+				(bn, tasks.clone().into_inner(), Self::get_task_call_encodings(tasks))
+			})
+			.collect();
+		let retries: Vec<_> = pallet_scheduler::Retries::<T>::iter().collect();
+		let lookup: Vec<_> = alias::Lookup::<T>::iter().collect();
+
+		// (IncompleteSince, Agendas and their schedule's call encodings, Retries, Lookup)
+		(incomplete_since, agenda_and_call_encodings, retries, lookup).encode()
```
ggwpez, line 307: Hm I see, the `RcPalletsOrigin` makes it difficult to also use the struct here...

PolkadotDom: Yeah :/

pallets/ah-migrator/src/scheduler.rs
PolkadotDom: I suppose it's not expected, will add logs and use defensive!

pallets/rc-migrator/src/scheduler.rs
PolkadotDom: Yes I believe that's the issue I was encountering

integration-tests/ahm/src/tests.rs
PolkadotDom: Ah. Good catch, there was a flaw in my thinking. Taking a look, but there's a pre-req PR up now https://github.com/paritytech/polkadot-sdk/pull/8401

muharem (approved):

PR #681, opened 2025-04-22
Enable RFC103 node side feature on Kusama

sandreim:
This will enable candidate receipt v2 format at the runtime upgrade time on Kusama. It is a dependency for Elastic Scaling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   - Bounties Pallet: add approve_bounty_with_curator call ([paritytech/polkadot-sdk/pull/5961](https://github.com/paritytech/polkadot-sdk/pull/5961))
   - Collective: Dynamic deposit based on number of proposals ([paritytech/polkadot-sdk/pull/3151](https://github.com/paritytech/polkadot-sdk/pull/3151))
   - New runtime api that returns the associated pool accounts with a nomination pool ([paritytech/polkadot-sdk/pull/6357](https://github.com/paritytech/polkadot-sdk/pull/6357))
-  - Enable RFC103 on Kusama
+- Enable RFC103 on Kusama ([polkadot-fellows/runtimes/pull/681](https://github.com/polkadot-fellows/runtimes/pull/681/))
 
 ### Changed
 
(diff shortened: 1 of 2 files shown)


ordian (approved):
1. This has been successfully tested on Westend with 3 cores, right?
2. We want to bundle this into a runtime upgrade as opposed to a separate referendum to expedite the process, right?

sandreim:
> 1. This has been successfully tested on Westend with 3 cores, right? 

Yes, we had this running for quite some time. 

> 2. We want to bundle this into a runtime upgrade as opposed to a separate referendum to expedite the process, right?

Yes, so we don't wait an extra referenda.

acatangiu (approved):

bkchr (changes requested):
As with all the proposals before, just do an on chain referendum for this... The host configuration stuff is exposed exactly for this reason.

bkchr:
Also did all the Kusama validators already upgraded to a compatible node version?

sandreim:
> As with all the proposals before, just do an on chain referendum for this... The host configuration stuff is exposed exactly for this reason.

I am aware of the reason and that this is not the normal way to enable, but this time I want speed things up a bit. :)

> Also did all the Kusama validators already upgraded to a compatible node version?

From the data I have,  83% are on a compatible version.

sandreim:
I just pinged validators to push for more upgrades. If we enable the bit now there should be no problem still, because no parachains are using v2 receipts yet. I think there is plenty of time for people to upgrade. We'll be much more strict and enable via referenda on Polkadot.

bkchr:
> because no parachains are using v2 receipts yet.

So, why do we then need to speed up things?

sandreim:
> > because no parachains are using v2 receipts yet.
> 
> So, why do we then need to speed up things?

Two reasons:
- deadlines, we want to ship Elastic Scaling as promised by the end of May.
- I want to get a shot at deploying a YAP parachain with RFC103 as soon as possible before enabling on Polkadot.

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   - Bounties Pallet: add approve_bounty_with_curator call ([paritytech/polkadot-sdk/pull/5961](https://github.com/paritytech/polkadot-sdk/pull/5961))
   - Collective: Dynamic deposit based on number of proposals ([paritytech/polkadot-sdk/pull/3151](https://github.com/paritytech/polkadot-sdk/pull/3151))
   - New runtime api that returns the associated pool accounts with a nomination pool ([paritytech/polkadot-sdk/pull/6357](https://github.com/paritytech/polkadot-sdk/pull/6357))
+- Enable RFC103 on Kusama ([polkadot-fellows/runtimes/pull/681](https://github.com/polkadot-fellows/runtimes/pull/681/))
```
bkchr, line 19:
```suggestion
  - Enable RFC103 on Kusama ([polkadot-fellows/runtimes/pull/681](https://github.com/polkadot-fellows/runtimes/pull/681/))
```

bkchr (approved):

sandreim:
The zombienet test keeps failing. I ran it locally and the parachain doesn't have a core assigned so it makes no progress. 

cc @pepoviola

skunert (approved):

PR #684, opened 2025-04-23
Zombienet worflow

pepoviola:

cc @bkontur, isolated workflow to easily test with diff image versions.

Thx!


.github/workflows/zombienet-smoke.yml
```diff
@@ -0,0 +1,83 @@
...
+        polkadot-image:
+          default: "docker.io/parity/polkadot:latest"
+          type: string
+
+        cumulus-image:
+          default: "docker.io/parity/polkadot-parachain:latest"
+          type: string
+
+env:
+  FRAME_OMNI_BENCHER_RELEASE_VERSION: polkadot-stable2412-4
```
bkontur, line 28: not needed here
```suggestion
```

PR #692, opened 2025-04-26
[AHM] Relay Chain migrator benchmarks

muharem:
Relay Chain migrator benchmarks


muharem:
/cmd bench --runtime polkadot-runtime --pallet pallet_rc_migrator

pallets/rc-migrator/src/crowdloan.rs
```diff
@@ -213,6 +214,16 @@ impl<T: Config> PalletMigration for CrowdloanMigrator<T>
 					let mut contributions_iter = pallet_crowdloan::Pallet::<T>::contribution_iterator(fund.fund_index);
 
 					while let Some((contributor, (amount, memo))) = contributions_iter.next() {
+						if weight_counter.try_consume(T::DbWeight::get().reads_writes(1, 1)).is_err() {
```
muharem, line 217: I would rather use `consume` here, but it might panic bc of `debug_assert`.

pallets/rc-migrator/src/multisig.rs
```diff
@@ -98,15 +98,21 @@ type BalanceOf<T> = <<T as pallet_multisig::Config>::Currency as Currency<
 	<T as frame_system::Config>::AccountId,
 >>::Balance;
 
-pub struct MultisigMigrator<T, W, L> {
-	_marker: sp_std::marker::PhantomData<(T, W, L)>,
```
muharem, line 102: just simplifying this since not use in benchmarks anymore

pallets/rc-migrator/src/crowdloan.rs
ggwpez: Hm, yes looks like we do not have a saturating consume.

PR #718, opened 2025-05-07
Max PoV size 10mb for Kusama Asset Hub and People chains

muharem:
Max PoV size 10mb for Kusama Asset Hub and People chains.

The Kusama and Polkadot Relay Chains have recently increased their max PoV size to 10 MB. Parachain runtime configurations will be updated accordingly through the upcoming SDK release and its integration, which includes this PR: https://github.com/paritytech/polkadot-sdk/pull/5884.

However, this update will not be available before the planned migration. In this PR, I propose updating the block size configuration for Asset Hub ahead of that SDK release to ensure the extra capacity is available for the migration.

This PR updates the proof size specifically for runtimes with async backing enabled (Asset Hub and People chains) to support the upcoming Asset Hub migration. Currently, only Asset Hub is updated - along with the People chain, as it shares constants from the async backing module - since this change is part of the ongoing migration effort.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Changed
+
+- Max PoV size 10mb for Kusama Asset Hub and People chains ([polkadot-fellows/runtimes/pull/718](https://github.com/polkadot-fellows/runtimes/pull/718))
+
 ### Fixed
 
 - Enabled XCM instructions `ExchangeAsset` and `AliasOrigin` on the system parachains ([polkadot-fellows/runtimes/pull/700](https://github.com/polkadot-fellows/runtimes/pull/700))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- Max PoV size 10mb for Kusama Asset Hub and People chains ([polkadot-fellows/runtimes/pull/718](https://github.com/polkadot-fellows/runtimes/pull/718))
+- Max PoV size 10mb for System Parachains ([polkadot-fellows/runtimes/pull/718](https://github.com/polkadot-fellows/runtimes/pull/718))
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- Max PoV size 10mb for System Parachains ([polkadot-fellows/runtimes/pull/718](https://github.com/polkadot-fellows/runtimes/pull/718))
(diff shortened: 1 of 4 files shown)


system-parachains/constants/src/lib.rs
```diff
@@ -86,9 +86,13 @@ pub mod async_backing {
 	/// Operational  extrinsics.
 	pub const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(85);
 
+	/// Maximum PoV size.
+	// TODO: This value is copied from the recent update of the `polkadot_primitives::MAX_POV_SIZE`
+	// constant. When PR https://github.com/paritytech/polkadot-sdk/pull/5884 is included in
+	// the SDK release, this constant should be replaced with a direct reference.
+	pub const MAX_POV_SIZE: u64 = 10 * 1024 * 1024;
```
muharem, line 93: since this under the `async_backing` namespace, this only concerns the Kusama Asset Hub and People chains.

system-parachains/constants/src/lib.rs
```diff
@@ -86,9 +86,13 @@ pub mod async_backing {
 	/// Operational  extrinsics.
 	pub const NORMAL_DISPATCH_RATIO: Perbill = Perbill::from_percent(85);
 
+	/// Maximum PoV size.
+	// TODO: This value is copied from the recent update of the `polkadot_primitives::MAX_POV_SIZE`
```
s0me0ne-unkn0wn, line 90: Not sure it's recorded in the style guide, but generally a TODO item in comments requires a corresponding Github issue

s0me0ne-unkn0wn (approved):
Given that the `NORMAL_DISPATCH_RATIO` is 85% for this runtime, so we shouldn't overshoot even in the worst case, LGTM.

system-parachains/constants/src/lib.rs
muharem: done

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Changed
+
+- Max PoV size 10mb for Kusama Asset Hub and People chains ([polkadot-fellows/runtimes/pull/718](https://github.com/polkadot-fellows/runtimes/pull/718))
```
joepetrowski, line 11: Why only these chains?

muharem: update for all

muharem:
@s0me0ne-unkn0wn I have updated it for all System Parachains. please check this again

system-parachains/constants/src/lib.rs
muharem: not valid anymore

s0me0ne-unkn0wn (approved):

CHANGELOG.md
muharem: reverted it for now - some bridge tests are failing. I tried fixing it, but they pull constants from the SDK for assertions, not sure why. I asked Branislav; if there’s a simple fix, I’ll handle it. I’m trying to keep my focus on the AHM migration, so I’ve limited the changes to what’s directly related. It might look like just changing a constant, but it often brings additional complexity.

muharem:
I will open a new PR for this

PR #720, opened 2025-05-07
[AHM] Increase test coverage

eugypalu:
Added some tests and modified assert messages to make assertions identifiable by the coverage script

### DIFF
--- a/pallets/ah-migrator/src/bounties.rs
+++ b/pallets/ah-migrator/src/bounties.rs
@@ -132,6 +132,7 @@ impl<T: Config> crate::types::AhMigrationCheck for BountiesMigrator<T> {
 		);
 
 		// Assert storage 'Bounties::Bounties::ah_post::correct'
+		// Assert storage 'Bounties::Bounties::ah_post::consistent'
 		assert_eq!(
 			pallet_bounties::Bounties::<T>::iter().collect::<Vec<_>>(),
 			rc_bounties,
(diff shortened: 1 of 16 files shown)


pallets/ah-migrator/src/bounties.rs
```diff
@@ -154,7 +156,15 @@ impl<T: Config> crate::types::AhMigrationCheck for BountiesMigrator<T> {
 			"Bounty descriptions map value on Asset Hub should match RC value"
 		);
 
+		// Assert storage 'Bounties::BountyApprovals::ah_post::length'
```
ggwpez, line 159: We could probably change the script to always mark the length requirement as checked when the `correct` requirement is fulfilled.

pallets/ah-migrator/src/claims.rs
```diff
@@ -96,27 +96,27 @@ impl<T: Config> crate::types::AhMigrationCheck for ClaimsMigrator<T> {
...
 		assert!(
 			!pallet_claims::Total::<T>::exists(),
-			"Claims total should be empty before migration starts"
+			"Assert storage 'Claims::Total::ah_pre::empty'"
 		);
 		// "Assert storage 'Claims::Claims::ah_pre::empty'"
 		assert!(
 			alias::Claims::<T>::iter().next().is_none(),
-			"Claims should be empty before migration starts"
+			"Assert storage 'Claims::Claims::ah_pre::empty'"
```
ggwpez, line 104: You can remove the commend if you add it here, the script only checks if the string appears anywhere.

PR #717, opened 2025-05-07
[AHM] Emulated test for migration with async backing

muharem:
Emulated test for migration with async backing.

Read code and comments for more info.

PR #727, opened 2025-05-10
[AHM] Staking migration storage item checks

PolkadotDom:
Relay chain and AssetHub migration checks for the pallet-staking (to pallet-staking-async) storage items. Part of #644

Note: I used the existing rc-migrator staking/message.rs piping for sending the data over, this made the types easier to convert from pallet-staking to pallet-staking-async.

### DIFF
--- a/pallets/rc-migrator/src/staking/staking.rs
+++ b/pallets/rc-migrator/src/staking/staking.rs
@@ -601,17 +601,6 @@ impl<T: Config> crate::types::RcMigrationCheck for StakingMigrator<T> {
             .collect();
         messages.extend(virtual_stakers_messages);
 
-        // StakingStage::ErasStartSessionIndex
-        let eras_ssi_items: Vec<_> = pallet_staking::ErasStartSessionIndex::<T>::iter().collect();
-        let eras_ssi_messages: Vec<_> = eras_ssi_items
-            .into_iter()
-            .map(|(era, session)| RcStakingMessage::ErasStartSessionIndex {
-                era,     
-                session, 
-            })
-            .collect();
-        messages.extend(eras_ssi_messages);
-
         // StakingStage::ErasStakersOverview
         let eras_so_items: Vec<_> = pallet_staking::ErasStakersOverview::<T>::iter().collect();
         let eras_so_messages: Vec<_> = eras_so_items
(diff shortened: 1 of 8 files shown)


pallets/rc-migrator/src/staking/staking.rs
```diff
@@ -504,3 +496,443 @@ fn resume<Map: frame_support::IterableStorageMap<K, V>, K: FullEncode, V: FullCo
...
+			pallet_staking::ErasStakers::<T>::iter().next().is_none(),
+			"ErasStakers map should be empty on RC after migration (deprecated item)"
+		);
+		// Assert storage 'Staking::ErasStakersClipped::rc_post::empty'
+		assert!(
+			pallet_staking::ErasStakersClipped::<T>::iter().next().is_none(),
+			"ErasStakersClipped map should be empty on RC after migration (deprecated item)"
+		);
+
+		// Fails - should we remove state during migration?
```
PolkadotDom, line 931: Since [this PR](https://github.com/polkadot-fellows/runtimes/pull/717/files#diff-9998e9e841f381c56e5b61c1cefd860dd005a28fa97bb688fb759e48dd44b755L235-L250), ErasStartSessionIndex is no longer migrated, and this check fails. Should we be removing the storage item during the migration even if we're not piping it?

ggwpez: Yea that is not migrated anymore, but good call for removing it from the Relay. I guess it would still be there.  
We can note it for the cleanup phase

ggwpez:
Thanks a lot for the MR! We will plan to do the Westend migration tomorrow, but it we will need this test for Kusama and Polkadot and it is good to see that it runs in the CI

PolkadotDom:
> Thanks a lot for the MR! We will plan to do the Westend migration tomorrow, but it we will need this test for Kusama and Polkadot and it is good to see that it runs in the CI

Very nice. Excited to see how it goes 🙌

pallets/rc-migrator/src/staking/staking.rs
PolkadotDom: Cleanup phase? Is that an extra step after the migration or you mean StakingStage::Finished?

PR #728, opened 2025-05-13
Ensure no era planned when migration starts

Ank4n:
Note: With a snapshot where current_era > active_era, some tests will fail.


seadanda (commented):
> With a snapshot where current_era > active_era, some tests will fail.

Would it be a good idea to gate this so it doesn't run except when it's compiled with `on-chain-release-build` or similar?

pallets/rc-migrator/src/lib.rs
```diff
@@ -635,6 +639,20 @@ pub mod pallet {
 				},
 				MigrationStage::Scheduled { block_number } =>
 					if now >= block_number {
+
+						weight_counter.consume(T::DbWeight::get().reads(2));
+						#[cfg(feature = "ahm-staking-migration")]
```
kianenigma, line 1003: Is the feature still needed?

ggwpez: I am removing it in #821 afterwards

pallets/rc-migrator/src/lib.rs
```diff
@@ -745,11 +745,17 @@ pub mod pallet {
...
+		/// `start_moment` must be a point within the current era before the validator election
+		/// process has started. Typically, this corresponds to the final session of the era, just
+		/// prior to the election kickoff.
+=======
 		/// ### Parameters:
 		/// - `start`: The block number at which the migration will start.
 		/// - `cool_off_end`: The block number at which the cool-off period will end.
 		///
 		/// Read [`MigrationStage::Scheduled`] documentation for more details.
+>>>>>>> origin/dev-asset-hub-migration
```
muharem: upsi

PR #709, opened 2025-05-05
[Bench] Benchmarks for all runtimes to use AssetHub location instead of Parent

karolk91:
Fixes https://github.com/polkadot-fellows/runtimes/issues/546
Relates to: https://github.com/paritytech/polkadot-sdk/pull/7321, https://github.com/polkadot-fellows/runtimes/pull/673

This PR switches destination/location of various XCM related benchmarks to use AssetHub instead of Relay Chain (parent)

Chains to cover:
- [x] Kusama Coretime
- [x] Kusama People
- [x] Kusama Encointer
- [x] Kusama BridgeHub
- [x] Kusama AssetHub
- [x] Polkadot Coretime
- [x] Polkadot People
- [x] Polkadot Collectives
- [x] Polkadot BridgeHub
- [x] Polkadot AssetHub

The change takes into account incoming migration of Relay to Asset Hub so the benchmarks use Asset Hub location in places where Relay location would be used otherwise

TODO:

- [x] Cover all chains
- [x] Re-generate weights

PR #729, opened 2025-05-15
Add balance to StakingPot accounts for benchmarks for all system parachains

bkontur:
An [issue](https://github.com/polkadot-fellows/runtimes/pull/709#issuecomment-2853988376) was uncovered with the new `pay_fees` XCM benchmark. The problem arises when the `Trader` deposits an amount into the `StakingPot`. In some cases, if the amount is below the existential deposit (ED), the deposit fails and is ignored. This causes the benchmark to miss recording an expected storage write.

This PR sets an initial balance for the `StakingPot` accounts in genesis—**only for benchmarking purposes**.


raymondkfcheung (approved):

system-parachains/asset-hubs/asset-hub-kusama/src/genesis_config_presets.rs
```diff
@@ -71,7 +71,16 @@ pub fn asset_hub_kusama_local_testnet_genesis(para_id: ParaId) -> serde_json::Va
...
 fn asset_hub_kusama_development_genesis(para_id: ParaId) -> serde_json::Value {
-	asset_hub_kusama_local_testnet_genesis(para_id)
+	asset_hub_kusama_genesis(
+		invulnerables(),
+		{
+			let mut accounts = testnet_accounts();
+			// Make sure `StakingPot` is funded for benchmarking purposes.
+			accounts.push(StakingPot::get());
+			accounts
+		},
```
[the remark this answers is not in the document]
bkontur: @serban300 something like this would be ok: https://github.com/polkadot-fellows/runtimes/pull/729/commits/5d2fa2a9607b9c6c7e0c017c1074c0f93343ed46 ?

x3c41a (approved):

serban300 (approved):

karolk91 (approved):

tdimitrov (approved):

.github/workflows/test.yml
```diff
@@ -187,7 +187,7 @@ jobs:
           save-if: ${{ github.ref == 'refs/heads/main' }}
 
       - name: Check
-        run: cargo check -p chain-spec-generator --release --locked -q --features=runtime-benchmarks,all-runtimes
```
ggwpez, line 190: Had to do this to make CI go green and unblock the Release.

ggwpez (approved):

.github/workflows/test.yml
bkontur: @ggwpez no problem, and I think we can remove the runtime `runtime-benchmarks` feature from chain-spec-generator, because we don't use it anymore with frame-omni-bencher
https://github.com/polkadot-fellows/runtimes/blob/main/chain-spec-generator/Cargo.toml#L35-L49

PR #730, opened 2025-05-16
Snowbridge: Remove `snowbridge-pallet-system::NativeToForeignId` 

yrong:


yrong:
Just added some tests in https://github.com/polkadot-fellows/runtimes/pull/730/commits/a182c176c5594c4247ba6f2574b8dbc22914b4b2 to demonstrate the following:

- PNA TokenId generated in V4 is exactly the same as in V5.
-  Location encoded in V4 can be correctly decoded in V5.

Does this mean the migration might not even be nessessary?

yrong:
### Try runtime test

`try-runtime --runtime ./target/release/wbuild/bridge-hub-polkadot-runtime/bridge_hub_polkadot_runtime.compact.compressed.wasm on-runtime-upgrade --blocktime 6000 --checks=pre-and-post --disable-spec-version-check live --uri wss://polkadot-bridge-hub-rpc.polkadot.io`

```
[2025-05-16T13:20:31Z INFO  bridge_hub_polkadot_runtime::bridge_to_ethereum_config::migration] PNA migration from v4 Location { parents: 1, interior: X1([GlobalConsensus(Polkadot)]) } to v5 Location { parents: 1, interior: X1([GlobalConsensus(Polkadot)]) }
[2025-05-16T13:20:31Z INFO  bridge_hub_polkadot_runtime::bridge_to_ethereum_config::migration] PNA migration from v4 Location { parents: 1, interior: X3([GlobalConsensus(Polkadot), Parachain(2012), GeneralKey { length: 4, data: [80, 65, 82, 65, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] }]) } to v5 Location { parents: 1, interior: X3([GlobalConsensus(Polkadot), Parachain(2012), GeneralKey { length: 4, data: [80, 65, 82, 65, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0] }]) }
...
[2025-05-16T13:20:31Z INFO  bridge_hub_polkadot_runtime::bridge_to_ethereum_config::migration] PNA migration finished!
```

system-parachains/bridge-hubs/bridge-hub-polkadot/src/bridge_to_ethereum_config.rs
```diff
@@ -226,6 +226,139 @@ pub mod benchmark_helpers {
...
+	use sp_runtime::TryRuntimeError;
+	use xcm::prelude::*;
+
+	pub struct RegisteredPNA {
+		pub location: Location,
+		pub token_id: TokenId,
+	}
+
+	/// Migrate ForeignToNativeId from XCMv4 to XCMv5.
+	pub struct MigratePNAsToXcm5<T: snowbridge_pallet_system::Config>(core::marker::PhantomData<T>);
```
vgeddes, line 247:
```suggestion
	pub struct MigrateLocationsToXcmV5<T: snowbridge_pallet_system::Config>(core::marker::PhantomData<T>);
```

acatangiu:
> @acatangiu Which release do you think we should include this migration? As I mentioned earlier in this [comment](https://github.com/polkadot-fellows/runtimes/pull/730#issuecomment-2886418494), the V4-encoded location can be decoded in V5, which makes this PR technically unnessessary.
> 
> However, I’d still prefer to merge it just for consistency and to avoid confusion.

hmm, I don't understand the question really, this is a runtime migration defined at the runtime level, so not part of any crates released in Polkadot-SDK - nothing to backport

acatangiu (approved):

franciscoaguirre:
The migration is not needed since V5 locations are exactly the same as V4 locations except for testnet network ids. Since all PNAs are for Polkadot the encodings will be the same

system-parachains/bridge-hubs/bridge-hub-polkadot/src/bridge_to_ethereum_config.rs
```diff
@@ -227,6 +227,141 @@ pub mod benchmark_helpers {
...
+
+	impl<T: snowbridge_pallet_system::Config> frame_support::traits::OnRuntimeUpgrade
+		for MigrateLocationsToXcmV5<T>
+	{
+		fn on_runtime_upgrade() -> Weight {
+			let mut weight = T::DbWeight::get().reads_writes(1, 1);
+
+			let translate = |pre: xcm::v4::Location| -> Option<xcm::v5::Location> {
+				weight.saturating_accrue(T::DbWeight::get().reads_writes(1, 2));
+				let location = xcm::v5::Location::try_from(pre.clone()).expect("valid location");
```
bkontur: @yrong I think that using `.expect(` here is not the best thing:

https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/support/src/traits/hooks.rs#L473-L507
```
/// Similar to [`Hooks::on_initialize`], any code in this block is mandatory and MUST execute.
/// It is strongly recommended to dry-run the execution of these hooks using
/// [try-runtime-cli](https://github.com/paritytech/try-runtime-cli) to ensure they will not
/// produce and overweight block which can brick your chain. Use with care!
...
fn on_runtime_upgrade() -> Weight {
```

bkontur:
> The migration is not needed since V5 locations are exactly the same as V4 locations except for testnet network ids. Since all PNAs are for Polkadot the encodings will be the same

I was checking the migration, but then I came to the same conclusion as @franciscoaguirre :point_up_2:  and with more info

Why is this migration even needed?  And why just for `ForeignToNativeId`? What about `NativeToForeignId`?
PBH is 1.5.1, iirc this is based `stable2412` which has XCMv5 and I am checking this crate, which is already using `xcm::v5::Location`?
https://github.com/paritytech/polkadot-sdk/blob/stable2412/bridges/snowbridge/pallets/system/src/lib.rs#L271-L278
```
#[pallet::storage]
	pub type ForeignToNativeId<T: Config> =
		StorageMap<_, Blake2_128Concat, TokenId, xcm::v5::Location, OptionQuery>;

	/// Lookup table for native location relative to ethereum to foreign token ID
	#[pallet::storage]
	pub type NativeToForeignId<T: Config> =
		StorageMap<_, Blake2_128Concat, xcm::v5::Location, TokenId, OptionQuery>;
```

Ok, upcoming 1.6.X uses stable2503, with again xcm latest, which is XCMv5:
https://github.com/paritytech/polkadot-sdk/blob/stable2503/bridges/snowbridge/pallets/system/src/lib.rs#L232-L233
but there is just `ForeignToNativeId` and you are not using `VersionedLocation` for storage, so really there is no need to do any migration here.

---

But I would just add `RemoveStorage` migration for removing `NativeToForeignId` storage, which is removed with stable2503 and nothing else.

bkontur (changes requested):
@yrong please see: https://github.com/polkadot-fellows/runtimes/pull/730#issuecomment-3020602068

yrong:
@bkontur @franciscoaguirre Thanks for the review.

> Why is this migration even needed?

Yeah, it's techniquelly not required, as I mentioned in https://github.com/polkadot-fellows/runtimes/pull/730#issuecomment-2886418494 - it's actually a no-op operation.

>  I would just add RemoveStorage migration for removing NativeToForeignId storage, which is removed with stable2503 and nothing else.

Nice catch. I've made changes accordingly and tested with `try-runtime-cli`.  The logs show the RemoveStorage migration running as expected.

```
[2025-07-01T01:17:56Z INFO  bridge_hub_polkadot_runtime::bridge_to_ethereum_config::migration] Number of PNAs: 24
[2025-07-01T01:17:56Z INFO  bridge_hub_polkadot_runtime::bridge_to_ethereum_config::migration] RemoveNativeToForeignId finished!
```

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
```diff
@@ -154,7 +154,7 @@ pub type Migrations = (
 		Runtime,
 		bridge_to_kusama_config::XcmOverBridgeHubKusamaInstance,
 	>,
-	snowbridge_pallet_system::migration::FeePerGasMigrationV0ToV1<Runtime>,
+	bridge_to_ethereum_config::migration::RemoveNativeToForeignId<Runtime>,
```
bkontur: @yrong this PR should really contain just adding one migration and nothing else :), for example:
```suggestion
	frame_support::migrations::RemoveStorage<
		"EthereumSystem", // <- define as &str param_type
		"NativeToForeignId", // <- define as &str param_type
		RocksDbWeight,
        >,
```


for example check in polkadot-sdk:
```
frame_support::migrations::RemoveStorage<
		BridgeWestendMessagesPalletName,
		OutboundLanesCongestedSignalsKey,
		RocksDbWeight,
>
```

bkontur (changes requested):

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
yrong: @bkontur The [tests](https://github.com/yrong/runtimes/blob/701ac13e4aa971f42864bd9cf74dbd955f7dc3cb/system-parachains/bridge-hubs/bridge-hub-polkadot/tests/snowbridge.rs#L430) added in this PR are meant to ensure that the PNA transfer don't break in future XCM releases (e.g. XCM V6). The reason is that the tokenID - the key of `ForeignToNativeId` storage to identify the PNA - was generated from XCM V4  and has already been finalized and stored on Ethereum side.

So we can't allow it to change. The test will fail if there is any compatalibity issue in a future release, so I prefer to keep it as a safeguard.

Meanwhile, I've simplified the migration and made some cleaned things up as you suggested in https://github.com/polkadot-fellows/runtimes/pull/730/commits/1ffb3a09544dfbd2c110f7c671eb6186299b3854.

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
```diff
@@ -145,6 +145,8 @@ pub type UncheckedExtrinsic =
 parameter_types! {
 	pub const BridgeKusamaMessagesPalletName: &'static str = "BridgeKusamaMessages";
 	pub const OutboundLanesCongestedSignalsKey: &'static str = "OutboundLanesCongestedSignals";
```
bkontur:
```suggestion
```

system-parachains/bridge-hubs/bridge-hub-polkadot/tests/snowbridge.rs
```diff
@@ -416,3 +425,197 @@ fn construct_and_apply_extrinsic(
 	let r = Executive::apply_extrinsic(xt);
 	r.unwrap()
 }
+
+#[test]
+fn check_token_id_on_chain_derived_from_xcm_v4_is_same_as_value_derived_from_xcm_v5() {
```
bkontur: @yrong can you add here please some description and/or what to do when this test starts to fail?

```suggestion
fn check_compatibility_for_token_id_stored_on_ethereum() {
```

system-parachains/bridge-hubs/bridge-hub-polkadot/tests/snowbridge.rs
```diff
@@ -416,3 +425,197 @@ fn construct_and_apply_extrinsic(
...
+				assert_eq!(location, tc.reanchored);
+
+				let token_id = TokenIdOf::convert_location(&location).unwrap();
+				assert_eq!(token_id, tc.foreign);
+			})
+	}
+}
+
+#[test]
+fn check_location_encode_in_xcm_v4_can_be_decoded_by_xcm_v5() {
```
bkontur, line 536: @yrong I think we can remove this test

system-parachains/bridge-hubs/bridge-hub-polkadot/tests/snowbridge.rs
```diff
@@ -416,3 +425,197 @@ fn construct_and_apply_extrinsic(
...
+				],
+			),
+		},
+	];
+	for tc in test_cases.iter() {
+		let location: Location = Decode::decode(&mut tc.v4.encode().as_slice())
+			.expect("Stored data should decode to V5 format correctly");
+		assert_eq!(location, tc.latest);
+	}
+}
```
bkontur: @yrong I think we can remove this test

```suggestion
```

system-parachains/bridge-hubs/bridge-hub-polkadot/tests/snowbridge.rs
yrong: https://github.com/polkadot-fellows/runtimes/pull/730/commits/47961f019f0b31ed732d864961b2481911cc2c89

system-parachains/bridge-hubs/bridge-hub-polkadot/tests/snowbridge.rs
yrong: IMHO, this test is also useful, so I'd prefer to keep it.  I've added comments in 
https://github.com/polkadot-fellows/runtimes/pull/730/commits/055f9ee7a184ba0156899ddce75342b6ba60a38a for more details.

bkontur: > IMHO, this test is also useful, so I'd prefer to keep it.

Well, this test does not test anything special for Snowbridge or specific onchain stuff like the test before, it just tests "some" compatibility for XCMv4 vs XCMv5 - which should better go directly to the `polkadot-sdk/xcm/src/v5` module and not here.

We basically agreed [here](https://github.com/polkadot-fellows/runtimes/pull/753) that XCMv4 vs XCMv5 are compatible, so really this test is not needed here.

yrong: Make sense. Removed in https://github.com/polkadot-fellows/runtimes/pull/730/commits/e051c640b6726de4b798a818452eca07ff639b2e

acatangiu (approved):

acatangiu (approved):

PR #712, opened 2025-05-05
Release 1.5.1

franciscoaguirre:
Addresses https://github.com/polkadot-fellows/runtimes/issues/697.

All runtimes will be bumped now since we have features and fixes for all of them.

PR #735, opened 2025-05-21
bump

franciscoaguirre:
This branch was checked out from 1.4.3, I want to check the bench bot's results to see if the problem in https://github.com/polkadot-fellows/runtimes/pull/712 is with 1.5.x or was already happening before


franciscoaguirre:
Closing since I already saw the issue exists in 1.5.0 as well so it wasn't introduced in 1.5.1

PR #751, opened 2025-06-03
[AHM] Dependency fixes for integration tests and the Polkadot runtime

muharem:
Dependency fixes for integration tests and the Polkadot runtime

As seen in this PR https://github.com/polkadot-fellows/runtimes/pull/742, integration tests are failing because some dependencies are not enabling the std feature. That PR also shows `chain-spec-generator` failing to build its tests, but since there have been updates to its `Cargo.toml` on master, I suggest we wait for those changes to be merged.

The `ahm-test-westend` CI test disabled.

Additionally, this removes the `pallet-ah-migrator` dependency from the polkadot-runtime crate.

### DIFF
--- a/integration-tests/emulated/chains/relays/kusama/Cargo.toml
+++ b/integration-tests/emulated/chains/relays/kusama/Cargo.toml
@@ -25,4 +25,4 @@ emulated-integration-tests-common = { workspace = true }
 
 # Runtimes
 kusama-runtime-constants = { workspace = true, default-features = true }
-kusama-runtime = { workspace = true, default-features = true }
+kusama-runtime = { workspace = true }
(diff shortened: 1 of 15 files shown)


integration-tests/emulated/chains/relays/kusama/Cargo.toml
```diff
@@ -25,4 +25,4 @@ emulated-integration-tests-common = { workspace = true }
 
 # Runtimes
 kusama-runtime-constants = { workspace = true, default-features = true }
-kusama-runtime = { workspace = true }
+kusama-runtime = { workspace = true, default-features = true }
```
ggwpez:
```suggestion
kusama-runtime = { workspace = true }
```
We never disabled them in the workspace, zepter sad.  
But maybe we should 🤷

muharem: ah, for kusama not, for polkadot actually it is disabled

ggwpez: Actually that is what has been causing issues 🙃 removed if from Polkadot runtime now.

PR #755, opened 2025-06-05
Add `pallet_revive` to Kusama AssetHub

athei:
- [x] Run benchmarks
- [x] Clarify if we need to do something to properly set the on chain storage version of the pallet: No

Fixes https://github.com/paritytech/contract-issues/issues/58


system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1102,9 +1137,44 @@ pub type TxExtension = (
...
+	type Error = ();
+
+	fn try_from(value: RuntimeCall) -> Result<Self, Self::Error> {
+		match value {
+			RuntimeCall::Revive(call) => Ok(call),
+			_ => Err(()),
+		}
+	}
+}
+
```
pgherveou, line 1175: maybe add a note that this can be removed  after this  PR
https://github.com/paritytech/polkadot-sdk/pull/8234/files#diff-a81a7d2eaeeb3b6f1256b7b96b1f43a2447943f498f053ecf1e511bdbfa04cd1

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1029,6 +1030,38 @@ impl pallet_remote_proxy::Config for Runtime {
...
+	type RuntimeCall = RuntimeCall;
+	type CallFilter = Nothing;
+	type DepositPerItem = DepositPerItem;
+	type DepositPerByte = DepositPerByte;
+	type WeightPrice = pallet_transaction_payment::Pallet<Self>;
+	type WeightInfo = weights::pallet_revive::WeightInfo<Self>;
+	type ChainExtension = ();
+	type AddressMapper = pallet_revive::AccountId32Mapper<Self>;
+	type RuntimeMemory = ConstU32<{ 128 * 1024 * 1024 }>;
+	type PVFMemory = ConstU32<{ 512 * 1024 * 1024 }>;
```
ggwpez, line 1052: Do you have a link to the SDK where we set this limit on the node side?

athei: AFAIK both of those are set in state. Can somebody verify this? @bkchr @eskimor  ?

bkchr: Yes they are configured via the `HostConfigration`:
```
executorParams: [
    {
      MaxMemoryPages: 8,192
    }
    {
      PvfExecTimeout: [
        Backing
        2,500
      ]
    }
    {
      PvfExecTimeout: [
        Approval
        15,000
      ]
    }
  ]
```

athei: Thanks! What about the memory pages for parachain nodes? Is this still set on the CLI or also in state somewhere?

athei: Just verified both values: Kusama has `8192` memory pages configured as Basti posed above. The runtime memory pages are [hard coded](https://github.com/paritytech/polkadot-sdk/blob/598feddb893f5ad3923a62e41a2f179b6e10c30c/substrate/client/executor/common/src/wasm_runtime.rs#L30) to `2048`.

```
8192 * 64K = 512M (PVFMemory)
2048 * 64K = 128M (RuntimeMemory)
```

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
athei: Done

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
ggwpez: Oh okay, those are WASM memory pages. I assumed 4 KiB pages 👍

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1102,9 +1137,45 @@ pub type TxExtension = (
...
+			RuntimeCall::Revive(call) => Ok(call),
+			_ => Err(()),
+		}
+	}
+}
+
 /// Unchecked extrinsic type as expected by this runtime.
 pub type UncheckedExtrinsic =
-	generic::UncheckedExtrinsic<Address, RuntimeCall, Signature, TxExtension>;
+	pallet_revive::evm::runtime::UncheckedExtrinsic<Address, Signature, EthExtraImpl>;
```
ggwpez, line 1178: Is this a breaking change for wallets or other signers? Then we need to ensure to broadcast that info as part of the release.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1889,6 +1961,173 @@ impl_runtime_apis! {
...
+			Revive::evm_block_gas_limit()
+		}
+
+		fn gas_price() -> U256 {
+			Revive::evm_gas_price()
+		}
+
+		fn nonce(address: H160) -> Nonce {
+			let account = <Runtime as pallet_revive::Config>::AddressMapper::to_account_id(&address);
+			System::account_nonce(account)
```
ggwpez, line 1980: Side note for AHM: When a para sovereign account uses this and we translate it from its child to its sibling representation then they will probably lose access to this derived Eth Address.

@seadanda another reason to hand out free any-proxies in these cases.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
pgherveou: should not be a breaking change, this type should be a zero-size wrapper for generic::UncheckedExtrinsic,

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1889,6 +1961,173 @@ impl_runtime_apis! {
...
+		}
+
+		fn eth_transact(tx: pallet_revive::evm::GenericTransaction) -> Result<pallet_revive::EthTransactInfo<Balance>, pallet_revive::EthTransactError>
+		{
+			let blockweights: BlockWeights = <Runtime as frame_system::Config>::BlockWeights::get();
+			let tx_fee = |pallet_call, mut dispatch_info: DispatchInfo| {
+				let call = RuntimeCall::Revive(pallet_call);
+				dispatch_info.extension_weight = EthExtraImpl::get_eth_extension(0, 0u32.into()).weight(&call);
```
ggwpez, line 1988: Why is the nonce and tip not part of the `GenericTransaction` and used here? Not sure if it can make a difference for weight computation though, maybe there is a special path for tip > 0.

athei: AFAIK wallets just don't pass it when doing a dry run. So we can't really do anything about it. @pgherveou should know more. Tip doesn't even exist on Ethereum.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
kianenigma: eventually, some kind of PET test set in this repo to ensure the PAPI/PJS APIs don't break with such changes will be great. In this case it seems obvious though that it will be fine.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1029,6 +1030,38 @@ impl pallet_remote_proxy::Config for Runtime {
 	type WeightInfo = weights::pallet_remote_proxy::WeightInfo<Runtime>;
 }
 
+parameter_types! {
+	pub const DepositPerItem: Balance = system_para_deposit(1, 0);
+	pub const DepositPerByte: Balance = system_para_deposit(0, 1);
```
kianenigma, line 1035: Hope this implies a standard eg. ERC20 contract will be dirt cheap :)

kianenigma (approved):

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
athei: If this helps: On Westend pjs kept working as-is after this change. It doesn't change the data layout of the extrinsic. Just adds some code that scans for `eth_transact` and treats it differently from the rest of the dispatchables.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1029,6 +1030,38 @@ impl pallet_remote_proxy::Config for Runtime {
 	type WeightInfo = weights::pallet_remote_proxy::WeightInfo<Runtime>;
 }
 
+parameter_types! {
+	pub const DepositPerItem: Balance = system_para_deposit(1, 0);
```
acatangiu, line 1034: @athei is this the intended value? it ultimately translates to `DepositPerItem = 1 key * 20 * CENTS + 0 bytes * MILLICENTS` so 20 CENTS per item

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
acatangiu: this translates to 1 MILLICENTS per byte

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
acatangiu: @gui1117 or @georgepisaltu can you take a look pls?

acatangiu (approved):

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
pgherveou: > Why is the nonce and tip not part of the GenericTransaction

GenericTransaction is an EVM struct, it only defines fields that are available in EVM transaction

> dispatch_info.extension_weight = EthExtraImpl::get_eth_extension(0, 0u32.into()).weight(&call);

we need to set the extension_weight here or the estmated dispatch_weight are not correct, this depends on the tx extensions which get_eth_extension returns, tip and nonce don't really matter here, as they don't impact the weight.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
athei: I just set it to the canonical value for this chain. I assume that `system_para_deposit` is doing the right thing. Is that approach flawed? Not sure what I should be doing here other than trusting that somebody did the economic math when defining those values.

The pallet itself can handle any value. You can increase or decrease it later.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
ggwpez: cc @josepot not sure if relevant for you but just FYI

josepot: > cc @josepot not sure if relevant for you but just FYI

thanks a lot for the ping! I have to run now to catch a plane for Protocol Berg, but I will have a close look at this ASAP 🙏 

> If this helps: On Westend pjs kept working as-is after this change. It doesn't change the data layout of the extrinsic. Just adds some code that scans for `eth_transact` and treats it differently from the rest of the dispatchables.

Sadly PJS doesn't use the metadata for creating extrinsics 😞 Everything is just hardcoded, which is very problematic, as I'm sure that you already know... Also, that means that the fact that it works with PJS is not a very good indicator.

athei: > Everything is just hardcoded, which is very problematic, as I'm sure that you already know... Also, that means that the fact that it works with PJS is not a very good indicator.

Non sequitur:  If a tool that hard codes the extrinsic format keeps working is a very good sign. It means that the on-the-wire encoding did not change. As a matter of fact, it is even a better indicator than when a more flexible tool like Papi kept working.

Yes, maybe Papi flips out when the Metadata doesn't look exactly like it expects to be. But what would be the point of parsing the Metadata in the first place then?

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
kianenigma: `system_para_deposit` is the standard way to do it, but we have also miscalculated it before and then created more issues by trying to change it. If your pallet can handle the deposit variables changing on the fly (some kind of a `poke_deposit`), that is great.

How many bytes is a standard ERC20 contract in PVM? How much should I pay to create a $KIAN token?

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
ggwpez: Either way it is not a blocker, only something that we would need to communicate more explicitly.

ggwpez: Either way it is not a blocker, only something that we would need to communicate more explicitly.

josepot: > Non sequitur: If a tool that hard codes the extrinsic format keeps working is a very good sign. It means that the on-the-wire encoding did not change. As a matter of fact, it is even a better indicator than when a more flexible tool like Papi kept working.

You are completely missing the point. You see, using the metadata for dynamically creating the codecs rather than having things hardcoded is a good thing. It's basically the reason why the PAPI kept on working on Westend AH and rococo [when it was broken with PJS](https://github.com/polkadot-js/apps/issues/11602).

> Yes, maybe Papi flips out when the Metadata doesn't look exactly like it expects to be. But what would be the point of parsing the Metadata in the first place then?

No, you are mistaken. As I have explained [many times](https://forum.polkadot.network/t/papi-and-runtime-upgrades-making-dapps-truly-resilient/13178#p-29894-understanding-runtime-upgrades-breaking-vs-compatible-changes-1): just b/c in 2 different runtimes things are encoded just the same, that doesn't mean that there haven't been breaking changes:

> Example: Imagine a runtime developer renames a struct field. The SCALE data stays the same, but the metadata changes… and that can break DApps relying on the old naming, because decoding logic now interprets it differently.

Just b/c things are encoded the same, that does not imply that there haven't been a breaking change.

All that being said, at a first glance, I really don't think that this particular change is going to break anything with PAPI.

athei: Not saying at all its not the right approach to parse the Metadata. But now you suddenly depend on the stability of two things instead of just one.

In order to make sure Metadata does not change in an incompatible way we need some kind of CI to enforce that. Otherwise stuff will slip through since it is so nuanced.

Maybe the tool you linked can be turned into a CI check on this repo to flag incompatible changes according to the PAPI implementation.

georgepisaltu: The current approach in `pallet-revive` with a wrapper over `UncheckedExtrinsic` has become obsolete since the introduction of `TransactionExtension`. I looked at the implementation of the wrapper to bypass the checks done specifically for `eth_transact` calls and it all looks very hacky.

This behavior of having an essentially free transaction and doing some checks on it before eventually dispatching it is exactly what `TransactionExtension` is supposed to do. `pallet-revive` should provide a `TransactionExtension` instead of this wrapper over `UncheckedExtrinsic`. People who use the pallet in their runtime would have to just include the extension in their runtime's pipeline. The extension would also be able to properly track weight usage for all the checks it does, including possible refunds in `post_dispatch`.

My opinion is that you should make the effort to transition this logic to `TransactionExtension` and ditch the `pallet_revive::evm::runtime::UncheckedExtrinsic` wrapper in order to properly integrate the feature with other verifications done to transactions and plug into the weight accounting system. Then we can go back to the battle-tested `sp_runtime::generic::UncheckedExtrinsic`, which thankfully is used by (almost) everyone everywhere even though the extrinsic type is technically a generic.

I looked at the implementation used in this PR specifically and the metadata and unchecked -> checked verifications look ok to me. Even if you decide to go ahead with this PR and take on this tech debt, we will still have to switch to `TransactionExtension` sooner or later because the wrapper model is not extensible.

gui1117: > can you take a look pls?

But from what I see it doesn't change the extrinsic metadata v15 and v16. (extension, extra_ty, address_ty, signature_ty, call_ty are all the same).
But extrinsic metadata v14 is changed (`ty` is changed into a wrapper of the former type), but I don't think tools actually use the metadata of the extrinsic type available only in v14.

To be 100% sure though, we should compare the extrinsic metadata before and after.

athei: @georgepisaltu We are aware of that. And we also talked about that.

1) We started this work before the new transaction extensions were merged.
2) The new transaction extension required transitioning to the new transaction format (correct?) that was not well supported at the time.
3) We are planning to move to `TransactionExtension` in the future. But we are just under immense time pressure here and can't do a rewrite right now.

> My opinion is that you should make the effort to transition this logic to TransactionExtension and ditch the pallet_revive::evm::runtime::UncheckedExtrinsic wrapper in order to properly integrate the feature with other verifications done to transactions and plug into the weight accounting system. Then we can go back to the battle-tested sp_runtime::generic::UncheckedExtrinsic, which thankfully is used by (almost) everyone everywhere even though the extrinsic type is technically a generic.

Agreed.

acatangiu: @athei please plan explicit work to address this technical debt in the next cycle (or realistically next next cycle)

athei: It is on the roadmap: https://github.com/paritytech/contract-issues/issues/69

But guys really i feel you are a bit hard on me here. Calling it technical debt as if we are cutting corners here.

The reality that transactions extension didn't exist at the time. So we had no way of doing it without "technical deb". So please cut me some slack here.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
athei: >  If your pallet can handle the deposit variables changing on the fly (some kind of a poke_deposit), that is great.

It works by making deposit refunds independent of the current deposit price. Refunds are pro rata: If you remove 1% of storage you get 1% storage deposit refund of all the deposits currently held. This of course crates some opportunity for arbitrage when the price changes but this is fine.

> How many bytes is a standard ERC20 contract in PVM? How much should I pay to create a $KIAN token?

Ok lets do some calculation here:

We assume the code for the contract is already on chain and your contract just references it. This means you pay 30% of the full deposit for the code. This is to prevent griefing and is configured using `CodeHashLockupDepositPercent`. We might decide to lower this later which is not a problem. Let's assume this contract:
```solidity
// SPDX-License-Identifier: MIT
// Compatible with OpenZeppelin Contracts ^5.0.0
pragma solidity ^0.8.27;

import {ERC20} from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
import {Ownable} from "@openzeppelin/contracts/access/Ownable.sol";

contract KIAN is ERC20, Ownable {
    constructor(address initialOwner) ERC20("KIAN", "KI") Ownable(initialOwner) {}
}
```

This compiles to 40KB PolkaVM code.

Each contract creates a new `ContractInfo` storage item of `124` bytes.

The contract has those data fields:
```solidity
uint256 private _totalSupply;
string private _name;
string private _symbol;
address private _owner;
```

Those will be 4 storage items of 64 bytes each. since the strings will be stored inline. Its 64 instead of 32 because we need to charge for the key too (both value and key are both 32 bytes).

Lets' summarize:
``
system_para_deposit(5, 30% * 30KB + 124 + 4 * 64) + ED = 635399964 = 0.00063KSM
``

Calculation here: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=5636e6c0205515cfe9f5a4272037d74c

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
athei: Okay I did some due diligence. 

1) I compared the metadata json files of the runtime before and after this PR. The `UncheckedExtrisnic` looks exactly the same with its contents. There doesn't seem to be a second wrapper type visible in the metadata.
2) I used https://diff.papi.how/ to comapre the two runtimes. No incompatible changes were listed there. But the extrinsic isn't listed at all. So maybe it is just not checked by the tool.

I attach the Metadata so you can use the tool by youself:
[metadatas.zip](https://github.com/user-attachments/files/20713804/metadatas.zip)

gui1117: Yes my comment was wrong about v14, the field is not changing, the type info is implemented in a transparent way. 100% transparent

PR #758, opened 2025-06-05
Use a non-native asset for XCM benchmarks

franciscoaguirre:
XCM benchmarks were always using the native asset which is not the worst case for various instructions and calls.
In this PR, I changed it to a custom asset in the assets pallet which is the worst case since it has to go through a double map instead of a single map to get the amount of an account.

### DIFF
--- a/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
+++ b/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
@@ -1355,7 +1355,26 @@ mod benches {
 		}
 
 		fn get_asset() -> Asset {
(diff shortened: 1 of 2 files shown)


bkontur:
We could merge this Also with this: https://github.com/polkadot-fellows/runtimes/pull/709 (preparation for AHM) because delivering to Parent is also not the worst case

franciscoaguirre:
> We could merge this Also with this: #709 (preparation for AHM) because delivering to Parent is also not the worst case

Good idea. I'll change the base of this PR to point to that branch

franciscoaguirre:
Closing this in favor of https://github.com/karolk91/runtimes/pull/2

PR #765, opened 2025-06-10
[AHM] Slightly improved indices test

x3c41a:
removed merging with always empty array


pallets/ah-migrator/src/indices.rs
```diff
@@ -66,7 +66,6 @@ impl<T: Config> crate::types::AhMigrationCheck for IndicesMigrator<T> {
 		use std::collections::BTreeMap;
 		let all_pre: BTreeMap<_, _> = rc_pre_payload
 			.into_iter()
-			.chain(ah_pre_payload.into_iter())
```
dr333ws: line 60:  `assert!(indices.is_empty(), "Assert storage 'Indices::Accounts::ah_pre::empty'");`
means that we always expect empty `ah_pre_payload` thus this line is not needed

re-gius: Then you can just change `type AhPrePayload = ()` and avoid using it altogether. Am I missing something?

dr333ws: TTBOMK, if we remove the whole `type AhPrePayload` we won't have anything to return from `pre_check` and will break the interface - https://github.com/polkadot-fellows/runtimes/blob/6048e1c18f36a9e00ea396d39b456f5e92ba1552/integration-tests/ahm/src/tests.rs#L157C15-L157C24

acatangiu: only looked at the link you referenced so maybe dumb question, but why does it "break the interface"? the interface stays the same, it still returns something, it just returns `()` instead of the unused value/type.

muharem (approved):

pallets/ah-migrator/src/indices.rs
dr333ws: That's actually a good question. I misunderstood what Giuseppe meant by `and avoid using it altogether` (c). 
I though that it also meant removing return argument from 
`fn pre_check(_: Self::RcPrePayload) -> Self::AhPrePayload {`. 
We discussed it over the Element chat and came to the conclusion -- what needs to be done.

All good, just a little misunderstanding, no interface/contract breakage here.

ggwpez:
Actually it could happen that this will hold entries on Paseo or Kusama. We can re-open of needed, but going to close for now. We do deploy the pallets in advance, so someone could actually use it and make the test fail.

PR #700, opened 2025-04-30
Enable ExchangeAsset in asset hubs

franciscoaguirre:
`ExchangeAsset` was missing its benchmark for being enabled in asset hub.
It already has an implementation so it can be used for swapping two assets using the liquidity pools on asset hub.
Aiming for [release 1.5.1](https://github.com/polkadot-fellows/runtimes/issues/697).

PR #606, opened 2025-02-25
Upgrade polkadot sdk to `stable2412-4`

pandres95:
This Pull Request provides a minimal set of changes to support the upgrade to `stable2412` (to the latest available patch).

## Notable changes

Here are the most notable changes:

- `sp-std` removed for every runtime (except Encointer). Now we use `core` or `alloc` instead.
- Mocked weights for `frame_system_extensions`, `pallet_transaction_payment`, and `xcm` on most runtimes.
- Mocked weights for `pallet_identity` on People runtimes.
- Update to XCMv5 started.

## Notable TODOs

- [x] ~Review whether we need to make more changes regarding XCMv5~ (moved to a follow-up PR). @franciscoaguirre 
- [x] In Asset Hubs, the benchmarking options for `pallet-asset-conversion-tx-payment` is not exposed (or at least, is raising errors when trying to add it).
- [x] For `polkadot-runtime`, do we still need `coretime_migration.rs`? @seadanda 
- [x] Assist with the initial implementation of `TransactionExtension` @georgepisaltu @gui1117 

## Open Question

- Shouldn't we start using `frame`, `#[frame_construct_runtime]` and `#[derive_impl]` since they've already been proven to work?

## Help needed

- [x] @clangenb to update and re-enable Encointer.

## Action List

- [x] ~Remove any reference to either `NetworkId::Westend` and `NetworkId::Rococo` and replace with `NetworkId::ByGenesis`~ (moved to a follow-up PR).
- [x] Upgrade to `stable2412-3`
- [x]  Setup filter for pallet-staking to exclude pool members. See comment [here](https://github.com/polkadot-fellows/runtimes/pull/608#issuecomment-2716068043).

## Follow-up PRs

- [ ] Refresh weights (to do before this release). ;)
- [ ] Move to XCMv5.
- [ ] Remove `sp-std`.

PR #771, opened 2025-06-17
Set up alias_origin weights correctly

bkontur:
This is leftover from bump stable2412 where we added Aliasers but not changed weights:
https://github.com/polkadot-fellows/runtimes/pull/606
for AssetHub weights were fixed here https://github.com/polkadot-fellows/runtimes/pull/700 but not for other chains

### DIFF
--- a/system-parachains/encointer/src/weights/xcm/mod.rs
+++ b/system-parachains/encointer/src/weights/xcm/mod.rs
@@ -200,7 +200,7 @@ impl<Call> XcmWeightInfo<Call> for EncointerKusamaXcmWeight<Call> {
 		XcmGeneric::<Runtime>::clear_transact_status()
 	}
 	fn universal_origin(_: &Junction) -> Weight {
-		XcmGeneric::<Runtime>::universal_origin()
+		Weight::MAX
 	}
 	fn export_message(_: &NetworkId, _: &Junctions, _: &Xcm<()>) -> Weight {
 		Weight::MAX


system-parachains/encointer/src/weights/xcm/mod.rs
```diff
@@ -200,7 +200,7 @@ impl<Call> XcmWeightInfo<Call> for EncointerKusamaXcmWeight<Call> {
 		XcmGeneric::<Runtime>::clear_transact_status()
 	}
 	fn universal_origin(_: &Junction) -> Weight {
-		XcmGeneric::<Runtime>::universal_origin()
+		Weight::MAX
```
franciscoaguirre, line 203: We want to disable this on encointer?

franciscoaguirre (approved):

franciscoaguirre:
Also, should we add bridge hubs here?

system-parachains/encointer/src/weights/xcm/mod.rs
bkontur: It was enabled here by accident with stable2412 bump I guess,
encointer does not support that:
```
type UniversalAliases = Nothing;

...

fn universal_alias() -> Result<(Location, Junction), BenchmarkError> {
			Err(BenchmarkError::Skip)
		}
```

franciscoaguirre:
> > Also, should we add bridge hubs here?
> 
> We could, but do we need `AliasOrigin` for BridgeHubs? I think it was intentionally not added to BHs, do we need it here? If so, let's do follow-up

I don't think it's needed since they're not user facing. Let's keep it like this

PR #772, opened 2025-06-17
[AHM] DMP prioritization on AH

muharem:
DMP queue prioritization over HRMP on AH during migration.

Implements DMP queue prioritization during Asset Hub migration by setting the message queue head to the Relay Chain origin on block finalization. This ensures Relay Chain messages (or DMP) are processed first in the next block instead of following the default round-robin pattern.

To maintain balanced processing of HRMP messages, the system allows configuration of a repeating priority pattern:
- Number of consecutive blocks where Relay Chain messages get priority
- Number of blocks for normal round-robin processing

The priority pattern can be configured either through pallet config or runtime call override, with three options:
- Default pallet configuration
- Custom pattern (e.g., 18 blocks priority, 2 blocks round-robin)
- Disabled priority

Refer to the types documentation for more details.

### DIFF
--- a/pallets/ah-migrator/src/lib.rs
+++ b/pallets/ah-migrator/src/lib.rs
@@ -1056,6 +1056,11 @@ pub mod pallet {
 				"Failed to send XCM message to update XCM data message processed count"
 			);
 		}
+
+		fn integrity_test() {
+			let (dmp_priority_blocks, _) = T::DmpQueuePriorityPattern::get();
+			assert!(!dmp_priority_blocks.is_zero(), "the `dmp_priority_blocks` should be non-zero");
+		}
 	}
 
 	impl<T: Config> Pallet<T> {
--- a/pallets/ah-migrator/src/lib.rs
+++ b/pallets/ah-migrator/src/lib.rs
@@ -70,9 +70,11 @@ use frame_support::{
 		Contains, Defensive, DefensiveTruncateFrom, LockableCurrency, OriginTrait, QueryPreimage,
 		ReservableCurrency, StorePreimage, VariantCount, WithdrawReasons as LockWithdrawReasons,
 	},
+	weights::WeightMeter,
 };
 use frame_system::pallet_prelude::*;
 use pallet_balances::{AccountData, Reasons as LockReasons};
+use pallet_rc_migrator::types::MigrationStatus;
 
 #[cfg(not(feature = "ahm-westend"))]
 use pallet_rc_migrator::bounties::RcBountiesMessageOf;
@@ -83,6 +85,7 @@ use pallet_rc_migrator::crowdloan::RcCrowdloanMessageOf;
 #[cfg(not(feature = "ahm-westend"))]
 use pallet_rc_migrator::treasury::RcTreasuryMessage;
 
+use cumulus_primitives_core::AggregateMessageOrigin;
 use pallet_rc_migrator::{
 	accounts::Account as RcAccount,
 	conviction_voting::RcConvictionVotingMessageOf,
@@ -101,7 +104,7 @@ use scheduler::RcSchedulerMessageOf;
 use sp_application_crypto::Ss58Codec;
 use sp_core::H256;
 use sp_runtime::{
-	traits::{BlockNumberProvider, Convert, TryConvert, Zero},
+	traits::{BlockNumberProvider, Convert, One, TryConvert, Zero},
 	AccountId32, FixedU128,
 };
 use sp_std::prelude::*;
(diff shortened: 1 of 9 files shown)


pallets/ah-migrator/src/lib.rs
```diff
@@ -199,6 +202,47 @@ pub struct BalancesBefore<Balance: Default> {
...
+}
+
+/// The priority of the DMP queue during migration.
+///
+/// Controls how the DMP (Downward Message Passing) queue is processed relative to other queues
+/// during the migration process. This helps ensure timely processing of migration messages.
+/// The default priority pattern is defined in the pallet configuration, but can be overridden
+/// by a storage value of this type.
+#[derive(Encode, Decode, Default, Clone, PartialEq, Eq, RuntimeDebug, TypeInfo, MaxEncodedLen)]
+#[cfg_attr(feature = "stable2503", derive(DecodeWithMemTracking))]
```
ggwpez, line 233: We could merge master 2503 in now. Only downside is that we can probably not decode storage snapshots from RPC anymore.

pallets/ah-migrator/src/lib.rs
```diff
@@ -349,6 +410,16 @@ pub mod pallet {
...
 	pub type AhBalancesBefore<T: Config> = StorageValue<_, BalancesBefore<T::Balance>, ValueQuery>;
 
+	/// The priority of the DMP queue during migration.
+	///
+	/// Controls how the DMP (Downward Message Passing) queue is processed relative to other queues
+	/// during the migration process. This helps ensure timely processing of migration messages.
+	/// The default priority pattern is defined in the pallet configuration, but can be overridden
+	/// by a storage value of this type.
+	#[pallet::storage]
+	pub type DmpQueuePriorityConfig<T: Config> =
```
ggwpez, line 430: I think just using `pallet_parameters` would be enough. It also supports default values and origins can be restricted.

pallets/ah-migrator/src/lib.rs
```diff
@@ -404,6 +475,22 @@ pub mod pallet {
...
+			prioritized: bool,
+			/// Current block number within the pattern cycle (1 to period).
+			cycle_block: BlockNumberFor<T>,
+			/// Total number of blocks in the pattern cycle
+			cycle_period: BlockNumberFor<T>,
+		},
+		/// The DMP queue priority config was set.
+		DmpQueuePriorityConfigSet {
+			/// The new priority pattern.
+			priority: DmpQueuePriority<BlockNumberFor<T>>,
```
ggwpez: The new value can already be fetched from storage, maybe include the old one? Or both?

pallets/ah-migrator/src/lib.rs
```diff
@@ -885,6 +972,19 @@ pub mod pallet {
 			Self::migration_start_hook().map_err(Into::into)
 		}
 
+		/// Set the DMP queue priority configuration.
```
ggwpez, line 991:
```suggestion
		/// Set the DMP queue priority configuration.
		///
		/// Can only be called by the `ManagerOrigin`.
```

pallets/ah-migrator/src/lib.rs
```diff
@@ -885,6 +972,19 @@ pub mod pallet {
...
+		#[pallet::call_index(102)]
+		#[pallet::weight(T::AhWeightInfo::set_dmp_queue_priority())]
+		pub fn set_dmp_queue_priority(
+			origin: OriginFor<T>,
+			priority: DmpQueuePriority<BlockNumberFor<T>>,
+		) -> DispatchResult {
+			<T as Config>::ManagerOrigin::ensure_origin(origin)?;
+			DmpQueuePriorityConfig::<T>::put(priority.clone());
```
ggwpez: Maybe early return if its identical, then we also dont need to emit an event.

bkontur:
cc: @serban300

pallets/ah-migrator/src/lib.rs
```diff
@@ -1047,9 +1157,44 @@ pub mod pallet {
 				None
 			}
 		}
+
+		/// Force the DMP queue priority for the next block.
+		pub fn force_dmp_queue_priority(now: BlockNumberFor<T>) {
+			let (dmp_priority_blocks, round_robin_blocks) = match DmpQueuePriorityConfig::<T>::get()
```
ggwpez, line 1194: Maybe a small struct for this? The we get it as metadata in PJS instead of having to rely on remembering the tuple order.

pallets/ah-migrator/src/lib.rs
```diff
@@ -1047,9 +1157,44 @@ pub mod pallet {
...
+			{
+				DmpQueuePriority::Config => T::DmpQueuePriorityPattern::get(),
+				DmpQueuePriority::OverrideConfig(dmp_priority_blocks, round_robin_blocks) =>
+					(dmp_priority_blocks, round_robin_blocks),
+				DmpQueuePriority::Disabled => return,
+			};
+
+			let period = dmp_priority_blocks + round_robin_blocks;
+			let current_block = now % period;
```
ggwpez, line 1206: Sanity check period > 0 please 😋

pallets/ah-migrator/src/lib.rs
ggwpez: Also maybe check both values > 0.

pallets/ah-migrator/src/lib.rs
```diff
@@ -1047,9 +1157,44 @@ pub mod pallet {
...
+						false
+					},
+				}
+			} else {
+				false
+			};
+
+			Self::deposit_event(Event::DmpQueuePrioritySet {
+				prioritized: is_set,
+				cycle_block: current_block + BlockNumberFor::<T>::one(),
```
ggwpez, line 1225: The + 1 is for UI or why? I think 0 < n < period is normal as well?

pallets/ah-migrator/src/lib.rs
```diff
@@ -1047,9 +1157,44 @@ pub mod pallet {
...
+
+			let is_set = if current_block < dmp_priority_blocks {
+				// it is safe to force set the queue without checking if the DMP queue is empty, as
+				// the implementation handles these checks internally.
+				let dmp = AggregateMessageOrigin::Parent;
+				match T::MessageQueue::force_set_head(&mut WeightMeter::new(), &dmp) {
+					Ok(is_set) => is_set,
+					Err(_) => {
+						defensive!("Failed to force set DMP queue priority");
+						false
```
ggwpez, line 1216: I think it fails if the queue is empty, so we would probably see this quite often? Or do you expect one DMP per block?

pallets/ah-migrator/src/lib.rs
```diff
@@ -1047,9 +1157,44 @@ pub mod pallet {
...
+				DmpQueuePriority::Config => T::DmpQueuePriorityPattern::get(),
+				DmpQueuePriority::OverrideConfig(dmp_priority_blocks, round_robin_blocks) =>
+					(dmp_priority_blocks, round_robin_blocks),
+				DmpQueuePriority::Disabled => return,
+			};
+
+			let period = dmp_priority_blocks + round_robin_blocks;
+			let current_block = now % period;
+
+			let is_set = if current_block < dmp_priority_blocks {
```
serban300, line 1208: Nit: I think it would be easier to follow something like:

```
let mut is_set = false;
if current_block < dmp_priority_blocks {
    match T::MessageQueue::force_set_head(&mut WeightMeter::new(), &dmp) {
					Ok(new_is_set) => is_set = new_is_set,
    ....
}
```

pallets/ah-migrator/src/lib.rs
muharem: it requires migration or why? if migration we can include them in our tests.

pallets/ah-migrator/src/lib.rs
muharem: I thought about it too. but parameters wont work with out special/multi sig account that we will store in this pallet. also its quite low effort to introduce these mutable parameters on a pallet level.

pallets/ah-migrator/src/lib.rs
muharem: matter of taste I believe, I really dislike mutable variables (=

pallets/ah-migrator/src/lib.rs
muharem: in this case it returns false. it fails when there is no enough weight to set the head

pallets/ah-migrator/src/lib.rs
muharem: Yes, just for UI. and also it made more sense bc actually the priority set for the next block, not for the one we are right now.

pallets/ah-migrator/src/lib.rs
muharem: I believe we only care that dmp priority is not zero. Updated.

pallets/ah-migrator/src/lib.rs
ggwpez: Yea true. We could run the storage migrations in the tests

PR #773, opened 2025-06-17
Revert "Add `pallet_revive` to Kusama AssetHub (#755)"

athei:
We need to postpone the launch on Kusama in order to fix those two recently discovered issues that cannot be fixed retroactively:

https://github.com/paritytech/contract-issues/issues/75
https://github.com/paritytech/contract-issues/issues/109


joepetrowski (approved):

ggwpez (approved):
# 😭

PR #774, opened 2025-06-18
[AHM] Disable indirect XCMs from RC to AH and vice versa during migration

muharem:
Disable indirect XCMs from RC to AH and vice versa during migration

While we want to prioritize migration-related XCM messages between the Relay Chain (RC) and the Asset Hub (AH), we must ensure that no one can abuse the routing logic to indirectly send messages between them via XCM forwarding.

Specifically, this PR prevents scenarios where an external parachain (e.g., chain X) crafts an XCM message to the AH that, when executed, results in a forwarded message to the RC — or vice versa. During migration, this kind of message forwarding must be blocked to ensure predictable and secure behavior.

This PR:
- Disables message forwarding in the existing XCM routers between RC and AH during migration.
- Introduces new migration-specific routers that allow controlled message sending.
- Uses these new routers exclusively in the migrator pallets to send messages required for migration.

### DIFF
--- a/integration-tests/ahm/src/xcm_route.rs
+++ b/integration-tests/ahm/src/xcm_route.rs
@@ -128,6 +128,8 @@ fn test_send_to_rc_from_ah() {
 	});
 }
 
+/*
+FIXME v4 -> v5 broke it
 #[test]
 fn test_send_to_ah_from_rc() {
 	let mut t: sp_io::TestExternalities = frame_system::GenesisConfig::<RcRuntime>::default()
@@ -226,4 +228,4 @@ fn test_send_to_ah_from_rc() {
 
 		assert!(result.is_ok(), "fails with error: {:?}", result.err());
 	});
-}
+}*/


relay/polkadot/src/ah_migration/phase1.rs
```diff
@@ -137,7 +137,7 @@ pub fn call_allowed_status(call: &<Runtime as frame_system::Config>::RuntimeCall
 		Coretime(coretime::Call::<Runtime>::request_revenue_at { .. }) => (OFF, ON),
 		Coretime(..) => (ON, ON),
 		StateTrieMigration(..) => (OFF, OFF), // Deprecated
-		XcmPallet(..) => (OFF, ON), /* TODO allow para origins and root to call this during the migration, see https://github.com/polkadot-fellows/runtimes/pull/559#discussion_r1928789463 */
+		XcmPallet(..) => (OFF, ON),
```
muharem: disabled XCM pallet here should not affect the execution of messages from Coretime. messages from Coretime are first fed into the MQ, which then passes them to the XCM executor. this filtering only disables extrinsics imported to this chain that use XCM pallet dispatchables.

cc: @seadanda @ggwpez

relay/polkadot/src/lib.rs
```diff
@@ -1548,7 +1548,7 @@ impl pallet_rc_migrator::Config for Runtime {
 	>;
 	type Currency = Balances;
 	type CheckingAccount = xcm_config::CheckAccount;
-	type SendXcm = xcm_config::XcmRouter;
+	type SendXcm = xcm_config::XcmRouterWithoutException;
```
ggwpez, line 1605: Why do we need a special one here on the relay? I thought we want to use the one that disables during migration.

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -404,14 +404,21 @@ impl xcm_executor::Config for XcmConfig {
...
 	cumulus_primitives_utility::ParentAsUmp<ParachainSystem, PolkadotXcm, PriceForParentDelivery>,
 	// ..and XCMP to communicate with the sibling chains.
 	XcmpQueue,
 );
 
+/// For routing XCM messages which do not cross local consensus boundary.
+type LocalXcmRouter = pallet_ah_migrator::RouteInnerWithException<
+	LocalXcmRouterWithoutException,
+	Equals<ParentLocation>,
+	crate::AhMigrator,
```
ggwpez, line 519: Not sure i understand it here: I thought we want to block messages from all origins that are not `ParentLocation`.  
But looking at the code of `RouteInnerWithException` it seems like it will block `Equals<ParentLocation>` when the migration is ongoing.

relay/polkadot/src/xcm_config.rs
```diff
@@ -130,13 +130,22 @@ parameter_types! {
...
 
+/// The XCM router. When we want to send an XCM message, we use this type. It amalgamates all of our
+/// individual routers.
+///
+/// This router does not route to the Asset Hub if the migration is ongoing.
+pub type XcmRouter = pallet_rc_migrator::types::RouteInnerWithException<
+	XcmRouterWithoutException,
+	Equals<AssetHubLocation>,
+	crate::RcMigrator,
+>;
```
acatangiu: Doesn't this also disable sending migration messages to AH when migration is ongoing?

acatangiu (changes requested):
Before doing detailed review, I have a high-level question:

What are the scenarios where this is needed? Do you have practical examples?

I understand we don't want other chains to be able to get RC or AH to send XCM messages to each other (route through one to get to the other), but I believe we are already covered and do not need this extra mechanism (more complexity, more failure points).

Afaict, the only ways to have chain A initiate a (new) XCM message to chain B are:
- [x] Through extrinsics that end up doing `xcm::send()` - these can all be covered using call filters.
- [ ] Through XCM asset transfers from chain C to B where A acts as the reserve
     - [x] RC is only reserve for DOT, and DOT teleports are already disabled (will fail) during migration,
     - [ ] AH is reserve for many assets, and could act as router to RC (even if the assets are effectively lost/trapped on RC because it doesn't accept them) - we can also cover this case by using existing reserve transfers filters to disallow reserve transfers to RC.

relay/polkadot/src/xcm_config.rs
muharem: I actually described this in the PR description. There is new routers that used by migrator pallets, these routers do not have filtering.

relay/polkadot/src/lib.rs
muharem: bc its the only one that can send to AH and we need it for the migrator pallet to send data. the rest cannot.

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
muharem: we do not lock anyone from sending to RC or AH and they will use that 2 blocks from our pattern (lets say (18, 2)). but we do not let their xcm programs to send messages from RC or AH to AH or RC and use that 18 blocks from our pattern.

acatangiu:
> you can also reroute via QueryPallet, ReportError, ReportHolding and ReportTransactStatus instructions.

today I learned that [QueryResponseInfo](https://github.com/paritytech/polkadot-sdk/blob/cc5065dab6ab4c40d2901981481afdad8e34d4c2/polkadot/xcm/src/v5/mod.rs#L320C12-L320C29) takes a `destination: Location` parameter that specifies where the response should go... 🤦 unexpected

pallets/rc-migrator/src/types.rs
```diff
@@ -212,6 +213,45 @@ impl<Status: MigrationStatus, Default: Get<Weight>> Get<Weight> for ZeroWeightOr
 		Status::is_ongoing().then(Weight::zero).unwrap_or_else(Default::get)
 	}
 }
+
+/// A wrapper around `Inner` that routes messages through `Inner` unless `Exception` is true and
+/// `MigrationState` is ongoing.
+pub struct RouteInnerWithException<Inner, Exception, MigrationState>(
+	PhantomData<(Inner, Exception, MigrationState)>,
+);
```
acatangiu: some improvement can be made to `Exception` naming to explain that it is a condition applied to the `destination`, but just nice to have, the type defined here will be short-lived anyway (removed after AHM)

acatangiu (approved):
Looks good! 👍

pallets/rc-migrator/src/types.rs
```diff
@@ -212,6 +213,45 @@ impl<Status: MigrationStatus, Default: Get<Weight>> Get<Weight> for ZeroWeightOr
...
+{
+	type Ticket = Inner::Ticket;
+	fn validate(
+		destination: &mut Option<Location>,
+		message: &mut Option<Xcm<()>>,
+	) -> SendResult<Self::Ticket> {
+		if MigrationState::is_ongoing() &&
+			Exception::contains(destination.as_ref().ok_or(SendError::MissingArgument)?)
+		{
+			Err(SendError::Unroutable)
```
bkontur: very nit - I would maybe use different code:
```
	/// Destination is routable, but there is some issue with the transport mechanism.
	#[codec(index = 14)]
	Transport(#[codec(skip)] &'static str),
```
```suggestion
			Err(SendError::Transport("Migration ongoing - routing is temporary blocked!"))
```

pallets/rc-migrator/src/types.rs
```diff
@@ -212,6 +213,45 @@ impl<Status: MigrationStatus, Default: Get<Weight>> Get<Weight> for ZeroWeightOr
...
+		} else {
+			Inner::validate(destination, message)
+		}
```
bkontur: I would switch the conditions - `MigrationState::is_ongoing()` is touching the storage, right?
so in most cases we don't need to hit the storage, right?

```suggestion
		if Exception::contains(destination.as_ref().ok_or(SendError::MissingArgument)?) && MigrationState::is_ongoing()
		{
			Err(SendError::Transport("Migration ongoing - routing is temporary blocked!"))
		} else {
			Inner::validate(destination, message)
		}
```

pallets/rc-migrator/src/types.rs
bkontur: > some improvement can be made to `Exception` naming to explain that it is a condition applied to the `destination`

Agree, I would vote for renaming `Exception` to `BlockFor` -  `BlockFor::contains(destination`

muharem: @bkontur `RouteInnerWithException<.., BlockFor, ..>` ?

pallets/rc-migrator/src/types.rs
muharem: migration stage gonna be read at every block anyway when we deploy this runtime version

pallets/rc-migrator/src/types.rs
```diff
@@ -206,13 +208,56 @@ pub trait MigrationStatus {
...
+	fn get_messages() -> Vec<(VersionedLocation, Vec<VersionedXcm<()>>)> {
+		Inner::get_messages()
+	}
+}
+
+pub struct ExceptResponseFor<Querier>(PhantomData<Querier>);
+impl<Querier: Contains<Location>> Contains<Xcm<()>> for ExceptResponseFor<Querier> {
+	fn contains(l: &Xcm<()>) -> bool {
+		match l.first() {
+			Some(QueryResponse { querier: Some(querier), .. }) => !Querier::contains(querier),
```
muharem, line 279: we have to whitelist query responses for our flow control system

muharem: `querier` here from the perspective of the `destination`

pallets/rc-migrator/src/types.rs
muharem: updated the doc

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
```diff
@@ -1159,7 +1159,7 @@ impl pallet_ah_migrator::Config for Runtime {
 	type RcPalletsOrigin = ah_migration::RcPalletsOrigin;
 	type RcToAhPalletsOrigin = ah_migration::RcToAhPalletsOrigin;
 	type Preimage = Preimage;
-	type SendXcm = xcm_config::XcmRouter;
+	type SendXcm = xcm_config::LocalXcmRouterWithoutException;
```
bkontur: I am just thinking loud, does this AH migrator send messages just to the RC? Or possibly to other parachains?
Can we just use `ParentAsUmp` directly here? (I am ok, also with `LocalXcmRouterWithoutException` tuple - not that big deal)
```suggestion
	type SendXcm = WithUniqueTopic<
		cumulus_primitives_utility::ParentAsUmp<ParachainSystem, PolkadotXcm, PriceForParentDelivery>,
	>;
```

But the other thing - the `XcmRouter` was previously wrapped with `WithUniqueTopic`, but whatever we are adding here isn't. That means the `SetTopic` feature for tracking XCM messages won't apply probably.

Maybe we don’t need to track XCM messages at all. I need to go through the entire flow starting from the initial migration message (`first-tracking-id`). Most likely, the up-and-down migration messages won’t be tracked under the same `first-tracking-id` due to the query/response behavior (also need to check).

Anyway, just wanted to raise the point: without `WithUniqueTopic`, we most probably lose the ability to track messages using the `SetTopic` feature from the AssetHub migrator to the RC (maybe not that big issue also).

cc: @raymondkfcheung maybe, it could be worth to double-check all the XCM routing set it up for this migration, when everyhting will be ready (all similar PRs are merged to this dev-ahm branch).

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
bkontur: @muharem can you please add some docs for:
```
// TODO: some docs
FromContains<Equals<ParentLocation>, pallet_ah_migrator::ExceptResponseFor<Equals<Here>>>,
```
Could you please add a test covering this? Is `Equals<Here>` the correct value? Does `Here` applies from the `ParentLocation` point of view?

Where do we even send `QueryResponse`? Is it because the `xcm-executor` can send it internally? Can I see somewhere the flow which triggers `QueryResponse`?

I'm trying to read the `to_querier` function, the `QueryResponse` instruction, and the surrounding code, but it's not exactly easy bedtime reading :)

pallets/rc-migrator/src/types.rs
bkontur: ok, no problem, we can keep as it is.

Or other suggestions :D
```
/// A wrapper around `Inner` that routes messages through `Inner` unless `Exception` is true and
/// `MigrationState` is ongoing.
pub struct RouteInnerUnlessException<Inner, Exception, MigrationState>(

or shorther

pub struct RouteInnerUnless<Inner, Exception, MigrationState>(

or even shorter

pub struct RouteUnless<Inner, Exception, MigrationState>(
```

muharem: I asked a large language model 
Screenshot 2025-07-03 at 10 26 47

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
muharem: I usually try to introduce as little entropy as possible. The `LocalXcmRouterWithoutException` type is exactly the same as the one currently used (prior to this PR); only the name has been changed. This means I do not introduce any behaviour change to the migrator pallets with this PR.

pallets/rc-migrator/src/types.rs
muharem: also, my thinking when writing this was: if the migration isn’t ongoing, I don’t even want to bother checking those conditions. That means outside of the migration, we’re not relying on the correctness of those checks - less risk overall.

pallets/rc-migrator/src/types.rs
bkontur: haha, suggestions above were also based on AI :D :D :D,
let's keep it as it is :)

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
muharem: > Could you please add a test covering this? Is Equals<Here> the correct value? 

`migration_works_time` test covers this, without this our flow-control system wont simply work

> Does Here applies from the ParentLocation point of view?

yes

> Where do we even send QueryResponse? Is it because the xcm-executor can send it internally? Can I see somewhere the flow which triggers QueryResponse?

our flow control system for migration

I will add more docs.

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
bkontur: > This means I do not introduce any behaviour change to the migrator pallets with this PR.

Well, using `LocalXcmRouterWithoutException` instead of `XcmRouter` avoids using `SetTopic` with `WithUniqueTopic` - which may not be relevant behaviour change for migration. As I said before, just raising the point, we can go with or without `WithUniqueTopic` probably.

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
muharem: I have added more docs

bkontur (approved):

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
muharem: I see now. I was trying to avoid touching the `ToKusamaXcmRouter` and `SovereignPaidRemoteExporter<...>` routers, this sounds good to you? I am adding now the `SetTopic` for the migrator router.

muharem: @bkontur I added topic for the migrator router - https://github.com/polkadot-fellows/runtimes/pull/774/commits/e4c42efa2f26d3b0dfd956e0a28c171c493d6244

PR #780, opened 2025-06-23
[AHM]  Flow Control System

muharem:
Flow Control System

The new system uses the XCM buildin reporting mechanism to notify the Relay Chain about processed messages. For every data message sent to Asset Hub, we register a query and use the XCM `ReportTransactStatus` instruction within the `Appendix` to initiate reporting after the XCM message is processed and its status is known. On the Relay Chain side, we keep the entire XCM message until confirmation is received. If confirmation is never received, we can manually resend it with a call. Individual messages can also be manually acknowledged. The PR also introduces a storage value for unprocessed message buffer size to override the config value on the fly if needed, with a dedicated dispatchable call. The old counter-based backpressure system is removed.

### DIFF
--- a/pallets/ah-migrator/src/lib.rs
+++ b/pallets/ah-migrator/src/lib.rs
@@ -569,6 +569,8 @@ pub mod pallet {
 		) -> DispatchResult {
 			ensure_root(origin)?;
 
+			ensure!(values.len() == 1, Error::<T>::InvalidParameters);
+
 			let (referendum_count, deciding_count, track_queue) =
 				values.pop().ok_or(Error::<T>::InvalidParameters)?;
 
--- a/pallets/ah-migrator/src/lib.rs
+++ b/pallets/ah-migrator/src/lib.rs
@@ -338,11 +338,6 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type AhMigrationStage<T: Config> = StorageValue<_, MigrationStage, ValueQuery>;
 
-	/// The total number of XCM data messages processed from the Relay Chain and the number of XCM
-	/// messages that encountered an error during processing.
-	#[pallet::storage]
-	pub type DmpDataMessageCounts<T: Config> = StorageValue<_, (u32, u32), ValueQuery>;
-
 	/// Helper storage item to store the total balance / total issuance of native token at the start
 	/// of the migration. Since teleports are disabled during migration, the total issuance will not
 	/// change for other reason than the migration itself.
@@ -374,7 +369,8 @@ pub mod pallet {
 		FailedToCalculateCheckingAccount,
 		/// Vector did not fit into its compile-time bound.
 		FailedToBoundVector,
-		Unreachable,
+		/// Invalid parameters.
+		InvalidParameters,
 	}
 
 	#[pallet::event]
@@ -440,11 +436,7 @@ pub mod pallet {
 		) -> DispatchResult {
 			ensure_root(origin)?;
 
-			let res = Self::do_receive_accounts(accounts);
-
-			Self::increment_msg_received_count(res.is_err());
-
-			res.map_err(Into::into)
+			Self::do_receive_accounts(accounts).map_err(Into::into)
 		}
 
 		/// Receive multisigs from the Relay Chain.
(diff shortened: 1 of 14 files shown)


pallets/ah-migrator/src/lib.rs
```diff
@@ -590,32 +550,32 @@ pub mod pallet {
...
+				Vec<(TrackIdOf<T, ()>, u32)>,
+				// track_queue (referendum_id, votes)
+				Vec<(TrackIdOf<T, ()>, Vec<(u32, u128)>)>,
+			)>,
 		) -> DispatchResult {
 			ensure_root(origin)?;
 
-			let res =
-				Self::do_receive_referenda_values(referendum_count, deciding_count, track_queue);
+			ensure!(values.len() == 1, Error::<T>::InvalidParameters);
```
ggwpez: Why do you use a vec then?

pallets/rc-migrator/src/lib.rs
```diff
@@ -527,13 +565,15 @@ pub mod pallet {
 	pub type RcMigratedBalance<T: Config> =
 		StorageValue<_, MigratedBalances<T::Balance>, ValueQuery>;
 
-	/// The total number of XCM data messages sent to the Asset Hub and the number of XCM messages
-	/// the Asset Hub has confirmed as processed.
-	///
-	/// The difference between these two numbers are the messages that are "in-flight". We aim to
-	/// keep this number low to not accidentally overload the asset hub.
+	/// The pending XCM messages.
```
ggwpez, line 615: These are indexed by query_id? Please put some more comments :)

pallets/rc-migrator/src/lib.rs
```diff
@@ -597,12 +637,86 @@ pub mod pallet {
...
+			};
+
+			if matches!(response, MaybeErrorCode::Success) {
+				log::info!(
+					target: LOG_TARGET,
+					"Received success response for query id: {}",
+					query_id
+				);
+				PendingXcmMessages::<T>::remove(query_id);
+			}
```
ggwpez, line 766: Maybe print an error otherwise?

ggwpez (approved):

pallets/ah-migrator/src/lib.rs
muharem: bc of the `send_chunked_xcm_and_track` signature. I used `send_xcm` before for it. but since only the `send_chunked_xcm_and_track` has tracking I have to use it and I did not want to change the working code a lot. the `receive_referenda_values` we designed first only to receive a single message, not a batch.

pallets/rc-migrator/src/lib.rs
muharem: add the doc

pallets/rc-migrator/src/lib.rs
muharem: added the log. but note that we have this response in the even. I believe that where we really will be following it.

pallets/rc-migrator/src/lib.rs
```diff
@@ -670,19 +718,100 @@ pub mod pallet {
...
+				log::error!(target: LOG_TARGET, "Error while sending XCM message: {:?}", err);
+				Self::deposit_event(Event::<T>::XcmResendAttempt {
+					query_id,
+					send_error: Some(err),
+				});
+			} else {
+				Self::deposit_event(Event::<T>::XcmResendAttempt { query_id, send_error: None });
+			}
+
+			Ok(Pays::No.into())
```
bkontur, line 791: Why was `Ok(Pays::No.into())` added only here and not to the other calls?
All of those calls are guarded by `ManagerOrigin::ensure_origin`, and I expect them to be called with `Unpaid + Transact` from other chains, right?

But if we plan to allow `ManagerOrigin` for specific accounts so they can call these extrinsics directly (without XCM from other chain), then we should add `Ok(Pays::No.into())` to all of them. Otherwise, when we migrate balances, those accounts won't be able to call.

muharem: good point. I will do it. but also those accounts still would need some balance. I actually would filter that account and not migrate it on RC

PR #776, opened 2025-06-20
[AHM] OpenGov for Kusama AH

bkontur:
Closes: https://github.com/paritytech/polkadot-sdk/issues/7578
Relates: https://github.com/paritytech/polkadot-sdk/issues/7541
Relates: https://github.com/paritytech/polkadot-sdk/issues/7566

TBD: as a preparation for AHM for Kusama (intentionally not based on `dev-asset-hub-migration` for now - I will left TODO when needed some migration state or whatever)

This PR introduces all relevant changes and settings for RC and system parachains related to the migration of OpenGov to **Kusama** AssetHub. The counterpart for **Polkadot** AssetHub can be found here: https://github.com/polkadot-fellows/runtimes/pull/626.

## Key features

- Kusama Collectives/Fellows stays on the RC (AHM does not migrate them)
  - Means that AH needs to accept whitelisting call from the RC, 
    ```
    /// Origin commanded by Polkadot Fellows (3rd Dan fellows or greater).
    Fellows,
    ```
- Kusama People needs to accept `GeneralAdmin` voice plurality from AH and RC
- All system parachains needs to accept `StakingAdmin` from AH and RC because of `CollatorSelectionUpdateOrigin`
  - Except AssetHub needs to accept `StakingAdmin` from local OpenGov AH and RC

## Plan
- [ ] let's collect all the changes for OpenGov on AH (final, post migration)
  - We could even merge this to AHK with some disabled call filtering for OpenGov 
- [ ] then add tweaks to RC/AH for migration

## TODO
- [ ] collect all TODOs on the way
- [x] adjust all system paras to trust OpenGov on AH
- [x] add all the relevant tests from https://github.com/polkadot-fellows/runtimes/pull/626 @karolk91 
- [ ] add all the relevant tests for treasury spending
- [ ] check Proxy setting
- [ ] separate security audit?
- [x] make sure that Kusama Fellows on the RC can whitelist for AH OpenGov
- [ ] check AuctionAdmin, LeaseAdmin origin propagation for RC

## Questions
- [ ] `type Paymaster =` correct impl for local assets/accounts (VersionedAssetId / VersionedLocation is enought instead of locatable ones?)
  - [ ] Align the LocalPay for AHP
- [ ] `TreasuryInteriorLocation` how is the account with PALLET_ID=18 migrated on AH?
[...]

PR #626, opened 2025-03-17
[AHM] Governance nits&tweaks (very draft)

bkontur:
Relates: https://github.com/paritytech/polkadot-sdk/issues/7541
Relates: https://github.com/paritytech/polkadot-sdk/issues/7566

TDB:

## TODO

- [ ] check/setup governance origin/root/plurality propagation from migrated AssetHub governance to relaychain (before/during/after AHM)
- [ ] check/setup governance origin/root/plurality propagation from migrated AssetHub governance to system parachains (before/during/after AHM)
- [ ] add integration tests that ensure governance can `authorize_upgrade` relay/system chains

PR #783, opened 2025-06-24
System chains governace authorizes upgrade tests

karolk91:
This PR introduces test case scenarios to verify if parachain is able to process `authorize_upgrade` call as if it was send via governance chain (a Relay Chain for purpose of this PR)

These test cases were introduced to [polkadot-sdk](https://github.com/search?q=repo%3Aparitytech%2Fpolkadot-sdk%20governance_authorize_upgrade_works&type=code) and suggested via [comment](https://github.com/paritytech/polkadot-sdk/issues/7566#issuecomment-2879702028) that it would be nice to port, so this is just integration of existing testnets cases for production networks

This PR also introduces integration tests covering all system chains that simulates real network upgrade using whitelisting (via Collectives chain in case of Polkadot)

### Test Coverage

| Chain                   | Unit Tests                       | Integration Tests              |
|-------------------------|--------------------------------------|-------------------------------------|
| Polkadot Asset Hub      | ✅                                   | ✅                                  |
| Polkadot Coretime       | ✅                                   | ✅                                  |
| Polkadot Collectives    | ✅                                   | ✅                                  |
| Polkadot Bridge Hub     | ✅                                   | ✅                                  |
| Polkadot People         | ✅                                   | ✅                                  |
| Kusama Asset Hub        | ✅                                   | ✅                                  |
| Kusama Coretime         | ✅                                   | ✅                                  |
| Kusama Bridge Hub       | ✅                                   | ✅                                  |
| Kusama People           | ✅                                   | ✅                                  |
| Kusama Encointer        | ✅                                  |_Lack of mocked network definition_ [here](https://github.com/polkadot-fellows/runtimes/blob/2fb47566718a974c261b68fffaae500be5581820/integration-tests/emulated/networks/kusama-system/src/lib.rs#L40) |
| Kusama Relay            | _Not applicable_        | ✅                                  |
| Polkadot Relay          | _Not applicable_        | ✅                                  |
| Kusama Glutton          |  _Not supported: `BasicParachainRuntime` not implemented_ | _Lack of mocked network definition_ [here](https://github.com/polkadot-fellows/runtimes/blob/2fb47566718a974c261b68fffaae500be5581820/integration-tests/emulated/networks/kusama-system/src/lib.rs#L40)|

### Follow-up:

For AHM, we will add also the tests for OpenGov on AHs:
https://github.com/polkadot-fellows/runtimes/pull/776
https://github.com/polkadot-fellows/runtimes/pull/626


system-parachains/encointer/src/xcm_config.rs
```diff
@@ -53,6 +53,7 @@ parameter_types! {
 	pub UniversalLocation: InteriorLocation =
 	[GlobalConsensus(RelayNetwork::get()), Parachain(ParachainInfo::parachain_id().into())].into();
 	pub StakingPot: AccountId = CollatorSelection::account_id();
+	pub const GovernanceLocation: Location = Location::parent();
```
bkontur: @karolk91 just thinking, can we just define this in the test file directly, when it is not used here? I think later we will have this constant here: https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/constants/src/kusama.rs#L153 right?

bkontur (approved):
@karolk91 Good job! This PR captures the current state as-is with tests and will be useful when moving OpenGov to the AssetHubs - https://github.com/polkadot-fellows/runtimes/pull/626 and https://github.com/polkadot-fellows/runtimes/pull/776

system-parachains/encointer/src/xcm_config.rs
acatangiu: ➕

acatangiu (approved):

system-parachains/encointer/src/xcm_config.rs
karolk91: > @karolk91 just thinking, can we just define this in the test file directly, when it is not used here? I think later we will have this constant here: https://github.com/polkadot-fellows/runtimes/blob/main/system-parachains/constants/src/kusama.rs#L153 right?

I should probably just introduce this common constant as part of this PR

karolk91: I have introduced `GovernanceLocation` to common package in https://github.com/polkadot-fellows/runtimes/pull/783/commits/b5a8792cca4e575ba8ae037d0c4221e995ec923e. I have dropped `const` modifier because in future I expect it to point to `AssetHubLocation` and we would need to drop `const` anyway

@bkontur, I have followed the convention that `GovernanceLocation` is re-exported via xcm_config for every chain. In case of Encointer its only used for testing, do you think its worth breaking the convention for this single occurence and import straight in tests instead of via xcm_config ?

integration-tests/emulated/helpers/src/lib.rs
```diff
@@ -552,3 +559,87 @@ macro_rules! test_chain_can_claim_assets {
 		}
 	};
 }
+
+// TODO: remove when stable2503 / stable2506 released
```
bkontur:
```suggestion
// TODO: remove when stable2503-7 / stable2506 released
```

integration-tests/emulated/helpers/src/lib.rs
```diff
@@ -552,3 +559,87 @@ macro_rules! test_chain_can_claim_assets {
...
+		let whitelist_call: T::RuntimeCall =
+			pallet_whitelist::Call::<T::Runtime>::dispatch_whitelisted_call_with_preimage {
+				call: Box::new(call.into()),
+			}
+			.into();
+		whitelist_call.dispatch(origin)
+	})
+}
+
+// TODO: remove when stable2503 / stable2506 released
```
bkontur:
```suggestion
// TODO: remove when stable2503-7 / stable2506 released
```

integration-tests/emulated/helpers/src/lib.rs
```diff
@@ -552,3 +559,87 @@ macro_rules! test_chain_can_claim_assets {
...
+				fallback_max_weight,
+				call: transact_call.encode().into(),
+			}
+		]))),
+	}
+	.into();
+	call
+}
+
+// TODO: remove when stable2503 / stable2506 released
```
bkontur:
```suggestion
// TODO: remove when stable2503-7 / stable2506 released
```

integration-tests/emulated/helpers/src/lib.rs
```diff
@@ -552,3 +559,87 @@ macro_rules! test_chain_can_claim_assets {
...
+pub fn call_hash_of<T>(call: &T::RuntimeCall) -> H256
+where
+	T: Chain,
+	T::Runtime: frame_system::Config<Hash = H256>,
+	T::RuntimeCall: Encode,
+{
+	<T::Runtime as frame_system::Config>::Hashing::hash_of(&call)
+}
+
+// TODO: remove when stable2503 / stable2506 released
```
bkontur:
```suggestion
// TODO: remove when stable2503-7 / stable2506 released
```

bkontur (commented):
@karolk91 please, add to the desc, that we are adding also governance integration emulated tests, which simulates the real usage how we upgrade live chains and

## Follow-up section: 

For AHM, we will add also the tests for OpenGov on AHs:
https://github.com/polkadot-fellows/runtimes/pull/776
https://github.com/polkadot-fellows/runtimes/pull/626

integration-tests/emulated/tests/governance/kusama/src/open_gov_on_relay.rs
```diff
@@ -0,0 +1,210 @@
...
+		),
+		DispatchError::Module(sp_runtime::ModuleError {
+			index: 44,
+			error: [3, 0, 0, 0],
+			message: Some("CallIsNotWhitelisted")
+		})
+	);
+
+	// whitelist
+	Kusama::execute_with(|| {
```
bkontur, line 69: @karolk91 Karol, we forgot to add and use here `macro_rules! assert_whitelisted {` :), please

karolk91: done

bkontur (approved):

bkontur (approved):

acatangiu (approved):

Cargo.lock
```diff
@@ -4092,9 +4092,9 @@ dependencies = [
 
 [[package]]
 name = "emulated-integration-tests-common"
-version = "20.1.0"
+version = "21.0.0"
```
pandres95, line 4095: Open question: Wouldn't it make sense to upgrade the entire set of dependencies to the latest patch version, now that we're making changes on `Cargo.lock`?

Cargo.toml
```diff
@@ -65,7 +65,7 @@ frame-benchmarking = { version = "40.2.0", default-features = false }
 frame-election-provider-support = { version = "40.1.1", default-features = false }
 frame-executive = { version = "40.0.1", default-features = false }
 frame-support = { version = "40.1.0", default-features = false }
-frame-system = { version = "40.1.0", default-features = false }
```
pandres95, line 68: Think it'd definitely be better if we update other related deps in the PSDK, to prevent having multiple versions of the crates pinned to multiple dependants. This tends to be messy sometimes,

Cargo.lock
bkontur: > Open question: Wouldn't it make sense to upgrade the entire set of dependencies to the latest patch version, now that we're making changes on `Cargo.lock`?

Well, in the past we did both - PR with patch upgrade and also PR with just few patch deps upgrade - I think both ways are ok. We can do follow-up to bring the full stable2503-7

PR #784, opened 2025-06-25
[AHM] Disable #[pallet::hooks] for migrating pallets during migration

muharem:
Disable #[pallet::hooks] for migrating pallets during migration

Since there is no built-in feature to disable specific hooks at the runtime level, we configured certain parameters to cause those hooks to return early. To properly review this PR, you’ll need to look into the implementation of the affected hooks.

An alternative would be to write a custom container instead of using the AllPalletsWithSystem type for executables, but I believe that approach would be significantly more complex and the presented solution is more reasonable. 

These disables hooks for `treasury`, `scheduler` and `fast_unstake` pallets on RC and AH.

### DIFF
--- a/pallets/rc-migrator/src/types.rs
+++ b/pallets/rc-migrator/src/types.rs
@@ -215,6 +215,33 @@ impl<Status: MigrationStatus, Left: TypedGet, Right: Get<Left::Type>> Get<Left::
 	}
 }
 
+/// A weight that is `Weight::MAX` if the migration is ongoing, otherwise it is the [`Inner`]
+/// weight of the [`pallet_fast_unstake::weights::WeightInfo`] trait.
+pub struct MaxOnIdleOrInner<Status, Inner>(PhantomData<(Status, Inner)>);
+impl<Status: MigrationStatus, Inner: pallet_fast_unstake::weights::WeightInfo>
+	pallet_fast_unstake::weights::WeightInfo for MaxOnIdleOrInner<Status, Inner>
+{
+	fn on_idle_unstake(b: u32) -> Weight {
+		Status::is_ongoing()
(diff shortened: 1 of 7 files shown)


pallets/rc-migrator/src/types.rs
```diff
@@ -205,13 +205,43 @@ pub trait MigrationStatus {
...
+			.then(|| Weight::MAX)
+			.unwrap_or_else(|| Inner::on_idle_unstake(b))
+	}
+	fn on_idle_check(v: u32, b: u32) -> Weight {
+		Status::is_ongoing()
+			.then(|| Weight::MAX)
+			.unwrap_or_else(|| Inner::on_idle_check(v, b))
+	}
+	fn register_fast_unstake() -> Weight {
+		Inner::register_fast_unstake()
```
ggwpez, line 235: Not MAX here?

ggwpez: okay i get it when looking at the usage, but doc could mention it

relay/polkadot/src/lib.rs
```diff
@@ -756,7 +761,8 @@ impl pallet_treasury::Config for Runtime {
 	type Currency = Balances;
 	type RejectOrigin = EitherOfDiverse<EnsureRoot<AccountId>, Treasurer>;
 	type RuntimeEvent = RuntimeEvent;
-	type SpendPeriod = SpendPeriod;
+	type SpendPeriod =
+		pallet_rc_migrator::types::LeftOrRight<RcMigrator, DisableSpends, SpendPeriod>;
```
ggwpez, line 765: This would enable spending again after the migration, but I guess it does not matter since we already migrated everything.

PR #786, opened 2025-06-26
[AHM] Introducing the Cool-Off stage

muharem:
Introducing the Cool-Off stage

Introduces the `CoolOff` stage with a `cool_off_end` block number. The migration should be scheduled with two time points: `start` and `cool_off_end`. When the chain reaches the `start` block, migration begins on the Relay Chain by filtering the listed dispatchables for extrinsics, prioritizing messages from the Asset Hub (AH), and sending a message to AH indicating the start of the migration. Upon receiving this message, AH also begins filtering the listed dispatchables and prioritizes messages from the Relay Chain. The Relay Chain does not send any data to AH until the `cool_off_end` block has passed. This cool-off period allows time to ensure everything is functioning correctly and gives an opportunity to halt the migration if issues arise.


pallets/rc-migrator/src/lib.rs
```diff
@@ -593,7 +628,11 @@ pub mod pallet {
 		#[pallet::weight(T::RcWeightInfo::start_data_migration())]
 		pub fn start_data_migration(origin: OriginFor<T>) -> DispatchResult {
 			<T as Config>::ManagerOrigin::ensure_origin(origin)?;
-			Self::transition(MigrationStage::Starting);
+			let cool_off_end = match RcMigrationStage::<T>::get() {
+				MigrationStage::WaitingForAh { cool_off_end } => cool_off_end,
+				_ => return Err(Error::<T>::UnreachableStage.into()),
```
ggwpez: defensive

pallets/rc-migrator/src/lib.rs
```diff
@@ -646,13 +685,26 @@ pub mod pallet {
 							},
 						}
 					},
-				MigrationStage::WaitingForAh => {
+				MigrationStage::WaitingForAh { .. } => {
 					// waiting AH to send a message and to start sending the data. This stage may be
 					// skipped if AH is fast.
```
ggwpez: In the past there was a case that this stage can be skipped, but I think it cannot be skipped any more. So maybe adjust the comment.

pallets/rc-migrator/src/lib.rs
```diff
@@ -1482,7 +1534,12 @@ pub mod pallet {
 
 			if new == MigrationStage::Starting {
 				defensive_assert!(
-					matches!(old, MigrationStage::WaitingForAh | MigrationStage::Scheduled { .. }),
+					matches!(
+						old,
+						MigrationStage::WaitingForAh { .. } |
+							MigrationStage::Scheduled { .. } |
+							MigrationStage::CoolOff { .. }
+					),
```
ggwpez, line 1595: Please also check the `is_ongoing` of the Stage enum, but i think it does not need modifications.

ggwpez (approved):

pallets/rc-migrator/src/lib.rs
muharem: yes, it looks correct

``` rust
	pub fn is_ongoing(&self) -> bool {
		!matches!(
			self,
			MigrationStage::Pending |
				MigrationStage::Scheduled { .. } |
				MigrationStage::MigrationDone
		)
	}
```

pallets/rc-migrator/src/lib.rs
muharem: updated

pallets/rc-migrator/src/lib.rs
muharem: updated

integration-tests/ahm/src/tests.rs
```diff
@@ -532,13 +505,17 @@ async fn scheduled_migration_works() {
...
+
 	// Schedule the migration on RC.
 	let dmp_messages = rc.execute_with(|| {
 		log::info!("Scheduling the migration on RC");
 		next_block_rc();
 
 		let now = frame_system::Pallet::<Polkadot>::block_number();
-		let scheduled_at = now + 2;
+		start = now + 2;
+		cool_off_end = start + 3;
```
dr333ws, line 518: Are there any suggested values for `start` and `cool_off_end`? We can also discover those values during dry-running but just in case they're already known...

PR #787, opened 2025-06-26
[AHM] AH messages prioritization on RC

muharem:
Asset Hub messages prioritization over other parachains messages on RC during migration.

The solution is identical to the one introduced for DMP on AH by the PR https://github.com/polkadot-fellows/runtimes/pull/772

Implements Asset Hub messages prioritization during Asset Hub migration by setting the message queue head on RC to the Asset Hub origin on block finalization. This ensures Asset Hub messages are processed first in the next block instead of following the default round-robin pattern.

To maintain balanced processing of UMP messages, the system allows configuration of a repeating priority pattern:
- Number of consecutive blocks where Asset Hub messages get priority
- Number of blocks for normal round-robin processing

The priority pattern can be configured either through pallet config or runtime call override, with three options:
- Default pallet configuration
- Custom pattern (e.g., 18 blocks priority, 2 blocks round-robin)
- Disabled priority

Refer to the types documentation for more details.

### DIFF
--- a/pallets/rc-migrator/src/weights.rs
+++ b/pallets/rc-migrator/src/weights.rs
@@ -58,6 +58,8 @@ pub trait WeightInfo {
 	fn start_data_migration() -> Weight;
 	fn send_chunked_xcm_and_track() -> Weight;
 	fn update_ah_msg_processed_count() -> Weight;
+	fn set_ah_ump_queue_priority() -> Weight;
+	fn force_ah_ump_queue_priority() -> Weight;
 }
 
 /// Weights for `pallet_rc_migrator` using the Substrate node and recommended hardware.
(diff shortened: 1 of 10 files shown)


pallets/rc-migrator/src/weights.rs
```diff
@@ -147,6 +149,12 @@ impl<T: frame_system::Config> WeightInfo for SubstrateWeight<T> {
 			.saturating_add(T::DbWeight::get().reads(1_u64))
 			.saturating_add(T::DbWeight::get().writes(1_u64))
 	}
+	fn set_ah_ump_queue_priority() -> Weight {
+		Weight::from_parts(1, 1)
```
ggwpez, line 153: Need to re-run weights before we do the migration

PR #788, opened 2025-06-30
[AHM] Cache test snapshots and include more tests into CI job

muharem:
Cache test snapshots and include more tests into CI job

Caching test snapshots allows us to run tests significantly faster by avoiding repeated reads of multi-gigabyte files from the file system.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -187,11 +187,16 @@ jobs:
           lz4 -d ah-polkadot.snap.lz4 ah-polkadot.snap
 
       - name: Test Asset Hub Migration (Polkadot)
-        run: cargo test --locked -q -p polkadot-integration-tests-ahm -r pallet_migration_works -- --nocapture
+        run: cargo test --locked -q -p polkadot-integration-tests-ahm -r -- --nocapture
         env:
           RUST_BACKTRACE: 1
           SNAP_RC: "../../polkadot.snap"
           SNAP_AH: "../../ah-polkadot.snap"
+          RUST_LOG: "warn"
+      - name: Test Benchmarks Asset Hub Migration (Polkadot)
+        run: cargo test --locked -q -p polkadot-integration-tests-ahm -r --features runtime-benchmarks -- bench_rc bench_ops bench_ah --nocapture
+        env:
+          RUST_BACKTRACE: 1
           RUST_LOG: "warn"
 
   # TODO: remove if not needed. or reuse for Paseo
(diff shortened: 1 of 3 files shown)


integration-tests/ahm/src/mock.rs
```diff
@@ -33,39 +31,80 @@ use polkadot_primitives::UpwardMessage;
...
+	}
+
+	log::info!("Loading {} snapshot", chain.to_string());
+
+	// Load snapshot.
+	let snap = std::env::var(chain.to_string()).ok()?;
 	let abs = std::path::absolute(snap.clone());
 
-	let ext = Builder::<Block>::default()
+	let ext = Builder::<PolkadotBlock>::default()
```
muharem: it requires it to access the hasher function. we know it is the same for RC and AH

integration-tests/ahm/src/mock.rs
```diff
@@ -33,39 +31,80 @@ use polkadot_primitives::UpwardMessage;
...
-	env: &str,
-) -> Option<RemoteExternalities<Block>> {
+pub async fn remote_ext_test_setup(chain: Chain) -> Option<TestExternalities> {
 	sp_tracing::try_init_simple();
-	let snap = std::env::var(env).ok()?;
+	log::info!("Checking {} snapshot cache", chain.to_string());
+
+	// Check cache first
+
+	// Hold the lock for the entire function to avoid loading the snapshot twice.
```
ggwpez: Sound like use-case for `OnceCell`? I will give it a shot

muharem: yes, it should work with tokio::OnceCell. but wont make it simpler, will look same.

ggwpez: A bit simpler since it has `get_or_init` 😄 i changed it now

muharem: ok, it does look nicer tho

.github/workflows/test.yml
```diff
@@ -187,12 +187,17 @@ jobs:
           lz4 -d ah-polkadot.snap.lz4 ah-polkadot.snap
 
       - name: Test Asset Hub Migration (Polkadot)
-        run: cargo test --locked -q -p polkadot-integration-tests-ahm -r pallet_migration_works -- --nocapture
+        run: cargo test --locked -q -p polkadot-integration-tests-ahm -r -- --nocapture
```
ggwpez:
```suggestion
        run: cargo test --locked -q -p polkadot-integration-tests-ahm -r -- --nocapture --test-threads 1
```
See if CI works since it currently is cancelled.

PR #789, opened 2025-06-30
[AHM] Cleanup Westend feature

ggwpez:
Cleanup westend feature gates. Changes:
- Delete `ahm-westend` feature
- Remove usage of `ahm-polkadot` feature since it is default
- Use `ahm-kusama` in places where Kusama AHM will be special. Can be implemented later


integration-tests/ahm/src/tests.rs
```diff
@@ -192,6 +182,16 @@ async fn pallet_migration_works() {
...
+
+	// To check if anything changed
+	rc.execute_with(|| {
+		let root = sp_io::storage::root(sp_runtime::StateVersion::V1);
+		println!("Post RC root hash: {:?}", hex::encode(root));
+	});
+	ah.execute_with(|| {
+		let root = sp_io::storage::root(sp_runtime::StateVersion::V1);
+		println!("Post AH root hash: {:?}", hex::encode(root));
+	});
```
ggwpez, line 199: We could also assert these hashes to ensure that it is deterministic across CI and devs.

muharem: should we add same to `migration_works_time`?

ggwpez: Yea I will factor it out into a check.

integration-tests/ahm/Cargo.toml
```diff
@@ -96,21 +95,3 @@ runtime-benchmarks = [
 	"sp-runtime/runtime-benchmarks",
 	"xcm-builder/runtime-benchmarks",
 ]
-
-ahm-westend = [
-	"pallet-ah-migrator/ahm-westend",
-	"pallet-ah-ops/ahm-westend",
-	"pallet-rc-migrator/ahm-westend",
-]
-ahm-polkadot = [
```
muharem, line 105: don't we need to forward ahm-polkadot feature to all these pallets?

ggwpez: I removed it everywhere since we only focus on Polkadot for now. The only difference will be for Kusama, hence the use of `ahm-kusama`  where needed.

PR #802, opened 2025-07-04
🚧 [AHM] Implement RC->AH account translation migration across all involved pallets

sigurpol:
## 🎯 Goal

The goal is to ensure that all `AccountIDs` are properly translated when moved from RC to AH, particularly for sovereign accounts that need translation from `para` prefix to `sibl` prefix.

During the migration from Relay Chain to Asset Hub, accounts need to be translated in two scenarios:

1. **Regular user accounts**: Keep the same AccountID
2. **Parachain sovereign accounts**: Translate from RC format (`"para" ++ para_id ++ 00..`) to AH format (`"sibl" ++ para_id ++ 00..`)

Account translation logic has now been implemented in `pallets/ah-migrator/src/lib.rs` as part of the `AccountTranslator` trait. 

This logic needs to be applied by all pallets that store AccountIDs in their storage maps before inserting data on Asset Hub.

##  Migration Data Flow

- **RC-Migrator** sends accounts with their **original AccountIds** (including sovereign accounts with `para` prefix)
- **AH-Migrator** receives the accounts and translates them (e.g. converting `para` prefix into `sibl`)

## Pallets requiring account translation
	
- ✅ Balances (`System::Account` storage)
- ✅ Vesting
-  👷‍♂️**TODO:**   add a complete list of pallets to migrate

##  🙋 Open points

- Do we need to benchmark the account translation?

### DIFF
--- a/integration-tests/ahm/src/tests.rs
+++ b/integration-tests/ahm/src/tests.rs
@@ -248,7 +248,8 @@ fn sovereign_account_translation() {
 		let rc_acc = AccountId32::from_str(rc_acc).unwrap();
 		let ah_acc = AccountId32::from_str(ah_acc).unwrap();
 
-		let (translated, _para_id) = pallet_rc_migrator::accounts::AccountsMigrator::<Polkadot>::try_translate_rc_sovereign_to_ah(rc_acc).unwrap().unwrap();
+		let (translated, para_id) = pallet_ah_migrator::translate_rc_sovereign_to_ah(&rc_acc);
+		assert!(para_id.is_some());
 		assert_eq!(translated, ah_acc);
 	}
 
@@ -264,8 +265,9 @@ fn sovereign_account_translation() {
 	for rc_acc in bad_cases {
 		let rc_acc = AccountId32::from_str(rc_acc).unwrap();
 
-		let translated = pallet_rc_migrator::accounts::AccountsMigrator::<Polkadot>::try_translate_rc_sovereign_to_ah(rc_acc).unwrap();
-		assert!(translated.is_none());
+		let (translated, para_id) = pallet_ah_migrator::translate_rc_sovereign_to_ah(&rc_acc);
+		assert!(para_id.is_none());
+		assert_eq!(translated, rc_acc); // Should return original account
 	}
 }
 
--- a/integration-tests/ahm/src/tests.rs
+++ b/integration-tests/ahm/src/tests.rs
@@ -172,6 +172,7 @@ mod test_helpers {
 
 	/// Verify account translation event was emitted
 	pub fn verify_account_translation_event(
+		pallet: pallet_ah_migrator::PalletEventName,
 		test_case: &AccountTestCase,
 		original_account: &AccountId32,
 	) {
@@ -196,6 +197,7 @@ mod test_helpers {
 					event_record.event,
 					asset_hub_polkadot_runtime::RuntimeEvent::AhMigrator(
 						pallet_ah_migrator::Event::AccountTranslated {
+							pallet,
 							original: original_account.clone(),
 							translated: test_case.expected_translated.clone().unwrap(),
 							para_id: test_case.expected_para_id.unwrap(),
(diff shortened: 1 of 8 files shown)


pallets/ah-migrator/src/lib.rs
```diff
@@ -1078,6 +1176,28 @@ pub mod pallet {
...
 }
 
+impl<T: Config> AccountTranslator<T::AccountId> for Pallet<T> {
+	fn translate_account(account: &T::AccountId) -> T::AccountId {
+		// T::AccountId is constrained to be AccountId32 in the Config trait
+		let (translated_account, _para_id) = translate_rc_sovereign_to_ah(account);
+		translated_account
+	}
+
+	fn emit_translation_event(
```
ggwpez, line 1186: Do we need this? I would argue that the `translate_account` function can emit any events needed.

ggwpez:
>During the migration from Relay Chain to Asset Hub, accounts need to be translated in two scenarios:

Please keep it *abstract*. I mentioned in the description that you can mock the translation. This is since there will be more scenarios added. You can probably use the `Convert` trait with `AccountId`.  
It really just needs to provide `AccountID -> AccountID`.

pallets/ah-migrator/src/account.rs
```diff
@@ -68,7 +68,20 @@ impl<T: Config> Pallet<T> {
...
+		if let Some(para_id) = para_id {
+			Self::deposit_event(Event::<T>::AccountTranslated {
+				pallet: PalletEventName::Balances,
+				original: account.who.clone(),
+				translated: translated_account.clone(),
+				para_id,
+			});
+		}
+
+		let who = translated_account;
```
ggwpez, line 84:
```suggestion
let who = crate::translate_account_rc_to_ah(account.who);
```
Sorry for not specifying intended usage; i was thinking a bit simpler. The para_id is an implementation detail of the translator.   
The event can be emitted by the translator.

pallets/ah-migrator/src/lib.rs
sigurpol: Yes we don't need it

pallets/ah-migrator/src/lib.rs
```diff
@@ -1078,6 +1176,28 @@ pub mod pallet {
 	}
 }
 
+impl<T: Config> AccountTranslator<T::AccountId> for Pallet<T> {
+	fn translate_account(account: &T::AccountId) -> T::AccountId {
```
re-gius, line 1180: Where is this used?

pallets/ah-migrator/src/account.rs
```diff
@@ -68,7 +68,20 @@ impl<T: Config> Pallet<T> {
 			frame_system::Pallet::<T>::inc_providers(&account.who);
 		}
 
-		let who = account.who;
+		// Extract and translate the account
+		let (translated_account, para_id) = crate::translate_rc_sovereign_to_ah(&account.who);
```
re-gius, line 72: This should go at the beginning of the method, before checking for existential deposit

integration-tests/ahm/src/tests.rs
```diff
@@ -678,10 +854,140 @@ async fn some_account_migration_works() {
...
+					expected_translated.unwrap().to_ss58check(),
+					expected_para_id.unwrap()
+				);
+			}
+		});
+	}
+}
+
+#[tokio::test]
+async fn vesting_account_migration_works() {
```
re-gius, line 897: I would move vesting-specific tests either to a `vesting.rs` file in this same folder or to `RcMigrationCheck`/`AhMigrationCheck` tests in the migration files for pallets `rc-migrator`/`ah-migrator`

integration-tests/ahm/src/tests.rs
```diff
@@ -63,6 +63,194 @@ use polkadot_primitives::UpwardMessage;
...
+			0,
+		)
+		.unwrap_or_else(|err| {
+			log::error!("Account withdrawal failed: {:?}", err);
+			None
+		})
+	}
+
+	/// Create a vesting schedule on the relay chain for an account
+	pub fn create_vesting_schedule_on_rc(
```
re-gius, line 148: Also moving this together with vesting-specific tests

pallets/ah-migrator/src/vesting.rs
```diff
@@ -53,7 +53,23 @@ impl<T: Config> Pallet<T> {
...
+		if let Some(para_id) = para_id {
+			Self::deposit_event(Event::<T>::AccountTranslated {
+				pallet: PalletEventName::Vesting,
+				original: original_account,
+				translated: translated_account.clone(),
+				para_id,
+			});
+		}
+
+		let target_account = translated_account;
```
re-gius, line 70: Maybe it's better to move all these lines to a single function, so that it can be called in a compact way where we need it? Unless we don't want to emit translation events every time (see main comment)

re-gius (commented):
I'm okay with emitting a translation event when we migrate an account to AH, but I don't think we should emit it _every_ time a storage item is changed. Let's see what others think about that.

pallets/ah-migrator/src/lib.rs
sigurpol: The reason I add it initially was because I wanted to decouple the logic of the translation from the one of events assuming that each pallet might have different needs vs events. But 1. it was probably a wrong assumption or a debatable one 2.  doesn't matter now that we switch to the mock approach

pallets/ah-migrator/src/vesting.rs
sigurpol: as per Oliver's comment, emit translation logic will be handled within the translation fn directly but will be out of scope of this initial PR

pallets/ah-migrator/src/account.rs
sigurpol: yes you are definitely right, good catch

integration-tests/ahm/src/tests.rs
sigurpol: I am actually thinking that the integration-ahm/tests e.g. migration_works_time() where we call in cascade all the post_check() per pallet are already testing what each pallet is doing in theory and if the snapshots containg e.g. vesting stuff etc. Again, since the scope of the PR will be reduced to just a mock implementation, that's probably a discussion for later

PR #806, opened 2025-07-05
Add pallet BlackHole to Kusama AssetHub following WFC #539

olanod:
The Wish For Change in [ref#539](https://kusama.subsquare.io/referenda/539) approved the addition of the "BlackHole pallet" to the Kusama AssetHub runtime. 

[`fc-pallet-black-hole`](https://github.com/virto-network/frame-contrib/tree/main/pallets/black-hole) is a simple pallet that periodically burns the balance of the pallet's account(the *event horizon*), a permissioned extrinsic is also included to dispatch signed calls from said account, it's configured to the general admin and will be specially useful for calling pallet revive's map_account to allow contracts and evm wallets send funds to the well-known address of the pallet.


pandres95 (changes requested):
Good implementation! There are some things missing, though:

- Add the pallet to the list of benchmarks.
- Calculate Weights.
- Add some Changelog.

bkchr (changes requested):
Don't want to play the bad boy here, but why isn't that just a function of the  `balances` pallet?

olanod:
> Don't want to play the bad boy here, but why isn't that just a function of the `balances` pallet?

An agile Kusama experiment? not bloating the PolkadotSDK with features that might not be used in Polkadot or the general public? Leave room for future more interesting experiments like "hawking radiation" to address the burn redirect of [WFC#437](https://kusama.subsquare.io/referenda/437)? I personally find it interesting to have burn related functionality concentrated in one place with a fun concept that is easy to understand :)

bkchr:
> An agile Kusama experiment? not bloating the PolkadotSDK with features that might not be used in Polkadot or the general public?

We are talking about one dispatchable and we already had discussed this some years ago, because people thought you could use the all zeros address for this. So, yes I think we should add it. 

Ahh lol, we actually already [have it](https://github.com/paritytech/polkadot-sdk/blob/44ae6a8bebd23a8ffac02d71c6e74ee889c3ab00/substrate/frame/balances/src/lib.rs#L839). Not sure why I did not have seen this yesterday.

> I personally find it interesting to have burn related functionality concentrated in one place with a fun concept that is easy to understand :)

For that we should really use contracts and also not have funds send to something else where people think that they burned it.

olanod:
> Ahh lol, we actually already [have it](https://github.com/paritytech/polkadot-sdk/blob/44ae6a8bebd23a8ffac02d71c6e74ee889c3ab00/substrate/frame/balances/src/lib.rs#L839). Not sure why I did not have seen this yesterday.

Also didn't see this extrinsic before 😅 But still it's not what we want, the main selling point in the proposal was the constant auto burning from the well known address and dispatching from said address without requiring root. E.g. I have a use case where I want to the dark hole address to create an nft collection where each item is priced in KSM, buyers pay the owner of the collection(the event horizon) and earnings are later burned. With root+scheduler+the burn extrinsic this stuff can be done but way more painful. To integrate well with contracts the best would be to map the 0 address to the pallet with a [special mapper](https://github.com/virto-network/frame-contrib/pull/54/files), it just can't be done yet as the relevant trait is sealed but that can be changed.

pandres95:
/cmd bench --runtime asset-hub-kusama --pallet fc_pallet_black_hole

olanod:
> @olanod can you please also copy the pallet to this repo. This ensures we are not blocked by this when upgrading the SDK version.

I see this small pallet as the PoC for having external pallets from token-holder-approved contributors in the runtime, more [pallets are coming](https://kusama.subsquare.io/referenda/538) from frame-contrib(communities+referenda-tracks) so the fun exercise here should be about finding good ways to cooperate and move things forward knowing there might be this kind of blockers but luckily it should be a smooth process with proper communication in place :)

PR #807, opened 2025-07-06
[AHM / ah-migrator]: Implement RC->AH  account translation `mocked` migration across all involved pallets

### LINKED ISSUE #811: [AHM]  Add account translation to treasury spend beneficiaries during RC->AH migration
## 🎯 Core Issue

The `beneficiary` field in treasury spends can contain XCM `Location` structures with `AccountId32` junctions that reference specific account IDs. These AccountIds need to be translated from RC format to AH format

#807 has introduced a mocked implemenation  in  `system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/mod.rs` of `translate_location_account_ids()` which just returns the same location. 

The proper final implementation needs  to  apply the correct account translation to AccountId32 junctions instead, see details later on.

It's also important to remember to enforce the requirement for idempotency as stated in #808. 

See [this thread](https://github.com/polkadot-fellows/runtimes/pull/807#discussion_r2189640529).

### Current flow

1. **RC Side**: Sends `RcTreasuryMessage::Spends` with beneficiary as `VersionedLocation`. The `translate_location_account_ids` introduced by #807 is just a no-op for the time being
2. **AH Side**: Processes in `do_process_treasury_message()` using `T::RcToAhTreasurySpend::convert()`
3. **Conversion**: `RcToAhTreasurySpend::convert()` transforms:
   - `(VersionedLocatableAsset, VersionedLocation)` 
   - → `(VersionedLocatableAsset, VersionedLocatableAccount)`

### Option 1: Translate in Treasury.rs (Pallet Level)

**Location**: `pallets/ah-migrator/src/treasury.rs`

**Approach**: Add account translation in `do_process_treasury_message()` before calling `T::RcToAhTreasurySpend::convert()`.

```rust
RcTreasuryMessage::Spends { id: spend_index, status: spend } => {
    // Extract beneficiary and apply account translation
    let translated_beneficiary = Self::translate_beneficiary_location(spend.beneficiary);
    
    // Create updated spend with translated beneficiary
    let updated_spend = treasury_alias::SpendStatus {
        asset_kind: spend.asset_kind,
        amount: spend.amount,
        beneficiary: translated_beneficiary,
        valid_from: spend.valid_from,
        expire_at: spend.expire_at,
        status: spend.status,
    };
    
    // Apply high-level conversion
    let (asset_kind, beneficiary) = T::RcToAhTreasurySpend::convert((
        updated_spend.asset_kind, 
        updated_spend.beneficiary
    ))?;
    
    // Store result...
}
```

### Option 2: Translate in mod.rs (Runtime Level)

**Location**: `system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/mod.rs`

**Approach**: Modify  mocked `translate_location_account_ids()`  to apply the proper translation

Both options are fine.

sigurpol:
## 🎯 Goal
We want to ensure that all AccountIDs are translated when moved from RC to AH via a new ad-hoc `translate_account_rc_to_ah()` function in the `ah-migrator` pallet.

This PR introduces
- the logic to properly call the function above in all the pallets to migrate
- an implementation for `translate_account_rc_to_ah()` in `ah-migrator` pallet

## How to translate

In this PR, `translate_account_rc_to_ah()` is just a mock function that returns the input account.

[Here](https://github.com/polkadot-fellows/runtimes/pull/807/commits/830da1384d662260cf654f082cef310ccbdfd6ba) I introduced as PoC a version handling specifically  sovereign accounts by converting "para" prefixed accounts to "sibl" prefixed accounts, preserving the para_id and maintaining the same account format:
```
- **Sovereign accounts**: `para{para_id}{zeros}` → `sibl{para_id}{zeros}`
```
I have then reverted since we will have dedicated and tested functions we will be use later on from the translation (ideally a static map containing existing RC accountId -> AH conversion, if it doesn't bloat the WASM binary too much).

## Event handling

`translate_account_rc_to_ah()` will also be responsible to emit a translation event if and when needed (see TODO section).

##  Account migration logic

While migrating pallet X from RC to AH,  there are usually two places where we need to translate accounts:

1. in the pallet-specific `do_receive_*`  method  where we handle the whole translation logic
2. in the pallet-specific `post_check()` method,  to ensure consistent account ID comparison between RC and AH

For all pallets the pattern is the same when it comes to accountIDs:

1. **RC Side**: Stores original account IDs in pre-check data
2. **Migration Process**: Translates `original_account` → `translated_account`
3. **AH Side**: Stores translated account IDs in post-migration state
4. **Post-Check Comparison**:  compare  pre-check RC original_account with AH translated_account. This is why pre-check RC original_account needs to be translated to the new AH format.

## Integrated Pallets

- ✅ **`vesting`** 
- ✅ **`multisig`** 
- ✅ **`proxy`** 
- ✅ **`conviction_voting`**
- ✅ **`staking/nom_pools`** 
- ✅ **`staking/bags_list`** 
- ✅ **`staking/fast_unstake`** 
- ✅ **`staking/delegated_staking`** 
- ✅ **`referenda`** 
- ✅ **`bounties`** 
- ✅ **`treasury`** : NOTE: Extended the pre-check payload to store full proposal data, including account fields, instead of just proposal IDs. Close #811 to properly migrate the `beneficiary` field in `treasury` spends
- ⛔︎ **`scheduler`** :   nothing to do. No direct account storage: unlike other pallets that store account fields directly,
  scheduler stores origins and calls.  Origins are properly converted via T::RcToAhPalletsOrigin::try_convert(). Calls are properly converted via Self::map_rc_ah_call() which should handle account as well. The post-check reconstructs expected state using the same conversion logic
- ✅ **`preimage`** . NOTE: preimage doesn't require post-check account translation. The account translation is applied during processing and the post-checks focus on hash integrity rather than account-based comparisons.
- ✅ **`claims`** 
- ✅ **`crowdloan`** 
- ⛔︎ **`asset_rate`** : no AccountId fields. It only deals with asset types and numeric rates
- ✅ **`indices`** 

## TODO

- Put in place the proper translation
- Add event handling once requirements are clarified


pallets/ah-migrator/src/indices.rs
```diff
@@ -64,11 +65,28 @@ impl<T: Config> crate::types::AhMigrationCheck for IndicesMigrator<T> {
...
 	fn post_check(rc_pre_payload: Self::RcPrePayload, ah_pre_payload: Self::AhPrePayload) {
 		use std::collections::BTreeMap;
-		let all_pre: BTreeMap<_, _> = rc_pre_payload
+		
+		let translated_rc_pre: BTreeMap<_, _> = rc_pre_payload
+			.into_iter()
+			.map(|RcIndicesIndex { index, who, deposit, frozen }| {
+				let who_encoded = who.encode();
+				let translated_encoded = Pallet::<T>::translate_encoded_account_rc_to_ah(who_encoded);
+				let translated_who = T::AccountId::decode(&mut &translated_encoded[..])
```
ggwpez: The en- and decoding is just as a test?

sigurpol: more of a brain fart TBH, pointless + removed. Same in bags_list

pallets/ah-migrator/src/bounties.rs
```diff
@@ -125,18 +182,20 @@ impl<T: Config> crate::types::AhMigrationCheck for BountiesMigrator<T> {
 		);
 
 		// Assert storage 'Bounties::Bounties::ah_post::length'
+		// Check RC alias storage length since that's where we store the migrated bounties
 		assert_eq!(
-			pallet_bounties::Bounties::<T>::iter_keys().count() as u32,
+			pallet_rc_migrator::bounties::alias::Bounties::<T>::iter_keys().count() as u32,
```
ggwpez: We create these aliases when the original pallet does something annoying, like making the storage type private or using private members inside the stored values.  
They must be 100% equivalent though, so testing the original storage type here makes sense.

They are both pointers to the same value.

pallets/ah-migrator/src/bounties.rs
```diff
@@ -125,18 +182,20 @@ impl<T: Config> crate::types::AhMigrationCheck for BountiesMigrator<T> {
...
 			rc_bounties.len() as u32,
-			"Bounties map length on Asset Hub should match the RC value"
+			"Bounties map length in RC alias storage on Asset Hub should match the RC value"
 		);
 
-		// Assert storage 'Bounties::Bounties::ah_post::correct'
-		// Assert storage 'Bounties::Bounties::ah_post::consistent'
+		// Verify that bounties were migrated successfully by checking the keys match
+		let ah_bounty_keys: Vec<_> =
+			pallet_rc_migrator::bounties::alias::Bounties::<T>::iter_keys().collect();
```
ggwpez:
```suggestion
			pallet_bounties::Bounties::<T>::iter_keys().collect();
```

pallets/ah-migrator/src/bounties.rs
sigurpol: Done

pallets/ah-migrator/src/bounties.rs
sigurpol: Applied

pallets/ah-migrator/src/treasury.rs
```diff
@@ -54,7 +54,13 @@ impl<T: Config> Pallet<T> {
...
 			RcTreasuryMessage::Proposals((proposal_index, proposal)) => {
-				pallet_treasury::Proposals::<T>::insert(proposal_index, proposal);
+				let translated_proposal = pallet_treasury::Proposal {
+					proposer: Self::translate_account_rc_to_ah(proposal.proposer),
+					value: proposal.value,
+					beneficiary: Self::translate_account_rc_to_ah(proposal.beneficiary),
+					bond: proposal.bond,
+				};
+				pallet_treasury::Proposals::<T>::insert(proposal_index, translated_proposal);
 			},
```
re-gius, line 64: You should translate the `RcTreasuryMessage::Spends` too, right?

pallets/ah-migrator/src/lib.rs
```diff
@@ -932,6 +932,35 @@ pub mod pallet {
...
+		/// - Sovereign accounts (para -> sibl)
+		/// - Sovereign derived accounts
+		/// - General derived accounts
+		/// - ...and more
+		///
+		/// TODO: Will also be responsible to emit a translation event.
+		/// TODO: The current signature suggests that the function is intended to be infallible and
+		/// always return a valid account. This should be revisited when we replace the mock
+		/// implementation with the real one.
+		pub fn translate_account_rc_to_ah(account: T::AccountId) -> T::AccountId {
```
ggwpez, line 957: In the future we can give RC and AH different AccountId types and just require that `RC::AccountId: IntoAhTranslated<Ah::AccountId>` but it would be too much for this MR.  
Then we can be sure to not having forgotten it anywhere.

pallets/ah-migrator/src/lib.rs
```diff
@@ -932,6 +932,35 @@ pub mod pallet {
...
+			// Mock implementation - return unchanged for now
+			account
+		}
+
+		/// Helper function for migration post-check validation.
+		///
+		/// Should be used to apply account translation to RC pre-check data for consistent
+		/// comparison with AH post-state.
+		pub fn translate_encoded_account_rc_to_ah(who_encoded: Vec<u8>) -> Vec<u8> {
```
ggwpez, line 966: Do we still need / use it?

pallets/ah-migrator/src/multisig.rs
```diff
@@ -61,29 +61,47 @@ impl<T: Config> Pallet<T> {
 	}
 
 	pub fn do_receive_multisig(multisig: RcMultisigOf<T>) -> Result<(), Error<T>> {
-		log::debug!(target: LOG_TARGET, "Integrating multisig {}, {:?}", multisig.creator.to_ss58check(), multisig.deposit);
+		// Translate the creator account from RC to AH format
+		let translated_creator = Self::translate_account_rc_to_ah(multisig.creator.clone());
+
+		// Translate the details account (derived multisig account) if present
+		let translated_details = multisig
+			.details
```
ggwpez, line 71: Note that in some cases we never expect the translation to do anything, but it is still fine to keep. We can remove it later in cases were we know it to not trigger.

pallets/ah-migrator/src/multisig.rs
```diff
@@ -61,29 +61,47 @@ impl<T: Config> Pallet<T> {
...
+		);
 
 		let missing = <T as pallet_multisig::Config>::Currency::unreserve(
-			&multisig.creator,
+			&translated_creator,
 			multisig.deposit,
 		);
 
 		if missing != Default::default() {
 			if KNOWN_BAD_MULTISIGS.contains(&multisig.creator) {
```
ggwpez, line 87: You can probably use the new one as well, i think the translation is no-op here.

pallets/ah-migrator/src/proxy.rs
```diff
@@ -184,38 +214,54 @@ where
...
+				// All RC delegations that could be translated are still here
+				for rc_pre_d in &rc_pre.get(original_rc_delegator).cloned().unwrap_or_default() {
+					let translated: T::RcProxyType = rc_pre_d.0.clone().into();
+					let Ok(translated_kind) = T::RcToProxyType::try_convert(translated.clone())
+					else {
+						// Best effort sanity checking that only Auction and ParaRegistration dont
+						// work
+						let k = translated.encode().get(0).cloned();
+						assert!(
+							k == Some(7) || k == Some(9), /* TODO @ggwpez make all work and add
```
ggwpez:
```suggestion
							k == Some(7) || k == Some(9), /* FAIL-CI @ggwpez make all work and add
```
so that i dont forget it 😆

pallets/ah-migrator/src/bounties.rs
```diff
@@ -126,17 +183,17 @@ impl<T: Config> crate::types::AhMigrationCheck for BountiesMigrator<T> {
...
-			pallet_bounties::Bounties::<T>::iter_keys().count() as u32,
-			rc_bounties.len() as u32,
-			"Bounties map length on Asset Hub should match the RC value"
+			pallet_bounties::Bounties::<T>::iter_keys().count(),
+			rc_bounties.len(),
+			"Bounties map length in RC alias storage on Asset Hub should match the RC value"
 		);
 
-		// Assert storage 'Bounties::Bounties::ah_post::correct'
-		// Assert storage 'Bounties::Bounties::ah_post::consistent'
```
re-gius, line 135: These "ugly" comments are used by an automated script to check for testing coverage, do you have a good reason to remove them?

pallets/ah-migrator/src/staking/bags_list.rs
```diff
@@ -50,13 +50,31 @@ impl<T: Config> Pallet<T> {
...
-				alias::ListNodes::<T>::insert(&id, &node);
-				log::debug!(target: LOG_TARGET, "Integrating BagsListNode: {:?}", &id);
+				let translated_id = Self::translate_account_rc_to_ah(id);
+				debug_assert!(!alias::ListNodes::<T>::contains_key(&translated_id));
+
+				// Translate all AccountId fields in the node structure
+				let translated_node = alias::Node {
+					id: Self::translate_account_rc_to_ah(node.id),
+					prev: node.prev.map(|account| Self::translate_account_rc_to_ah(account)),
+					next: node.next.map(|account| Self::translate_account_rc_to_ah(account)),
```
ggwpez: Hm I think this will break the pallet. These bags list are sorted. We will probably need to re-create the bags.  
Can you please revert this file and create an issue for it? thanks

ggwpez: Maybe it could be a good next task for you, since its also staking related :)

pallets/ah-migrator/src/treasury.rs
sigurpol: Excellent point! 
This same `do_process_treasury_message()` fn    applies in the `RcTreasuryMessage::Spends` branch (see line 82)
  `T::RcToAhTreasurySpend::convert((asset_kind, beneficiary))` and I was assuming `T::RcToAhTreasurySpend::convert()` handles any AccountId translation within the beneficiary structure that may be needed.
  But it's not the case case as you can see https://github.com/polkadot-fellows/runtimes/blob/dev-asset-hub-migration/system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/mod.rs#L445.
 So the beneficiary conversion does NOT apply account translation to AccountIds within the `VersionedLocation` structure.
  
  So my question for you @re-gius and @ggwpez : where does the translation belong more? In treasury.rs (which seems where it belongs more logically and keep the account translation logic centralized in the ah-migrator pallet but which also introduces the need to deal with XCM handling, multiple versions potentially etc) or translate in mod.rs (simple but it would need to access to `translate_account_rc_to_ah()` fn?

pallets/ah-migrator/src/lib.rs
sigurpol: Fair, I can add at least a TODO in the scope of this PR and we can do it later on

pallets/ah-migrator/src/lib.rs
sigurpol: we are using it in `vesting.rs`'s `post_check()`

pallets/ah-migrator/src/staking/nom_pools.rs
```diff
@@ -89,16 +100,76 @@ impl<T: Config> Pallet<T> {
...
+				pallet_nomination_pools::ClaimPermissions::<T>::insert(translated_account, perms.1);
 			},
 		}
 	}
 
+	/// Translate accounts in bonded pool roles from RC to AH format
+	fn translate_bonded_pool_accounts(mut pool: BondedPoolInner<T>) -> BondedPoolInner<T> {
+		// Translate accounts in pool roles using functional approach
+		pool.roles.depositor = Self::translate_account_rc_to_ah(pool.roles.depositor);
+		pool.roles.root = pool.roles.root.map(|acc| Self::translate_account_rc_to_ah(acc));
```
ggwpez:
```suggestion
		pool.roles.root = pool.roles.root.map(Self::translate_account_rc_to_ah);
```
Even more functional 😋

pallets/ah-migrator/src/staking/nom_pools.rs
```diff
@@ -89,16 +100,76 @@ impl<T: Config> Pallet<T> {
...
+		}
+		if let Some((rate, ref mut account)) = pool.commission.current {
+			pool.commission.current =
+				Some((rate, Self::translate_account_rc_to_ah(account.clone())));
+		}
+
+		pool
+	}
+
+	/// Translate accounts in a generic bonded pool structure
```
ggwpez: One comment please how this differs from the fn above.

pallets/ah-migrator/src/staking/bags_list.rs
sigurpol: sure, let's see if I can squeeze the fix directly in this PR or I will create an issue. Do we have a specific project for this migration?

sigurpol:
> Maybe we also need to translate here? @muharem
> 
> https://github.com/polkadot-fellows/runtimes/blob/bfb1c0aef62279f9952f1c5ea19bf2958dbdaeed/system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/mod.rs#L429

Yes, it's my ancestral doubt that I am referring to [here](https://github.com/polkadot-fellows/runtimes/pull/807#discussion_r2190007631)

pallets/ah-migrator/src/treasury.rs
ggwpez: I dont have strong opinion on where to put it. Just need to make sure it gets called at least once. It should be idempotent, so no big issue when called more than once by accident.

ggwpez:
🙈 did not read the discussion before reviewing, yes.

pallets/ah-migrator/src/staking/bags_list.rs
ggwpez: I think there is no project. Just put AHM into the title please.

pallets/ah-migrator/src/referenda.rs
```diff
@@ -81,6 +81,67 @@ pub type AhReferendumStatusOf<T, I> = ReferendumStatus<
...
+			ReferendumInfo::Ongoing(mut status) => {
+				status.submission_deposit.who =
+					Self::translate_account_rc_to_ah(status.submission_deposit.who.clone());
+				if let Some(ref mut decision_deposit) = status.decision_deposit {
+					decision_deposit.who =
+						Self::translate_account_rc_to_ah(decision_deposit.who.clone());
+				}
+				ReferendumInfo::Ongoing(status)
+			},
+			ReferendumInfo::Approved(block, submission_deposit, decision_deposit) => {
```
re-gius, line 97: It seems that both `submission_deposit` and `decision_deposit` are `Options`, so you should check that is not `None`
https://github.com/paritytech/polkadot-sdk/blob/436b4935b52562f79a83b6ecadeac7dcbc1c2367/substrate/frame/referenda/src/types.rs#L342

pallets/ah-migrator/src/staking/bags_list.rs
re-gius: `Prev` and `Next` should be `Option<AccountId>`

pallets/ah-migrator/src/staking/bags_list.rs
```diff
@@ -50,13 +50,31 @@ impl<T: Config> Pallet<T> {
...
+				alias::ListNodes::<T>::insert(&translated_id, &translated_node);
+				log::debug!(target: LOG_TARGET, "Integrating BagsListNode: {:?}", &translated_id);
 			},
 			RcBagsListMessage::Bag { score, bag } => {
 				debug_assert!(!alias::ListBags::<T>::contains_key(&score));
-				alias::ListBags::<T>::insert(&score, &bag);
+
+				// Translate all AccountId fields in the bag structure
+				let translated_bag = alias::Bag {
+					head: bag.head.map(|account| Self::translate_account_rc_to_ah(account)),
```
re-gius: `head` and `tail` should be `Option<AccountId>`

pallets/ah-migrator/src/staking/nom_pools.rs
```diff
@@ -89,16 +100,76 @@ impl<T: Config> Pallet<T> {
...
+					perms.0.to_polkadot_ss58(), translated_account.to_polkadot_ss58());
+				pallet_nomination_pools::ClaimPermissions::<T>::insert(translated_account, perms.1);
 			},
 		}
 	}
 
+	/// Translate accounts in bonded pool roles from RC to AH format
+	fn translate_bonded_pool_accounts(mut pool: BondedPoolInner<T>) -> BondedPoolInner<T> {
+		// Translate accounts in pool roles using functional approach
+		pool.roles.depositor = Self::translate_account_rc_to_ah(pool.roles.depositor);
```
re-gius: `root`, `nominator`, and `bouncer` are `Option<AccountId>`

pallets/ah-migrator/src/bounties.rs
sigurpol: No 😄 Readded.

pallets/ah-migrator/src/lib.rs
sigurpol: Added a TODO / comment

pallets/ah-migrator/src/multisig.rs
sigurpol: Added a note

pallets/ah-migrator/src/proxy.rs
sigurpol: Done

pallets/ah-migrator/src/referenda.rs
sigurpol: you are right but .map() handles it for us for free:
  - If the Option is Some(deposit), it applies the closure to translate the account
  - If the Option is None, it returns None skipping the clousre

pallets/ah-migrator/src/staking/bags_list.rs
sigurpol: Correct and `.map()` handles everything for us (similar to previous comment)

pallets/ah-migrator/src/staking/bags_list.rs
sigurpol: same comment around map()

pallets/ah-migrator/src/staking/nom_pools.rs
sigurpol: Done

sigurpol: Done

pallets/ah-migrator/src/staking/nom_pools.rs
sigurpol: same as before, map() handles it for us for free

pallets/ah-migrator/src/staking/nom_pools.rs
sigurpol: The reason for having both was type compatibility:
  - The migration process uses the actual pallet types (BondedPoolInner<T>)
  - The test/verification code uses generic test types (tests::GenericBondedPoolInner) that can be
  serialized/deserialized for cross-chain verification and it's compiled only for testing.

But it's an overkill, added a macro instead.

pallets/ah-migrator/src/staking/bags_list.rs
sigurpol: Done  [here](https://github.com/polkadot-fellows/runtimes/pull/807/commits/810a5cae2d0c08be8279a47627d112c85c5ed587). Thanks for catching the issue 🙇

pallets/ah-migrator/src/treasury.rs
sigurpol: For the time being I went for a poor-man solution aka  a mocked fn (see [here](https://github.com/polkadot-fellows/runtimes/pull/807/commits/a8834c966dd16483779d3d12190f8f9b3a99593d)) and I have created #811  to properly tackle the issue  cc @re-gius , @ggwpez , @muharem

pallets/ah-migrator/src/staking/bags_list.rs
sigurpol: mmmh. I can do better :) 
```
thread 'tests::migration_works_time' panicked at /home/runner/work/runtimes/runtimes/pallets/ah-migrator/src/staking/bags_list.rs:194:3:
assertion `left == right` failed: Bags list data mismatch: Asset Hub data differs from original Relay Chain data
```
see https://github.com/polkadot-fellows/runtimes/actions/runs/16122934107/job/45493024877
I will investigate

sigurpol: Ok I am reverting the change where I was recreating the bags lists. See [here](https://github.com/polkadot-fellows/runtimes/pull/807/commits/338ff3f931f07171216f0061416c03ba18c04ce7). Reason being that bags are sorted but FIFO insertion is used within a bag. So we want to preserve the order per score and we want to preserve the prev / next relationship within a bag, and that's lost if we recreate from scratch

sigurpol: CI seems to like it https://github.com/polkadot-fellows/runtimes/actions/runs/16128043663/job/45509570666?pr=807 😄

pallets/ah-migrator/src/treasury.rs
sigurpol: I've tried to tackle #811 directly here,  doing the conversion directly in pallet-ah-migrator. Why? Mostly to keep account translation logic self-contained in this pallet so it's the only place where we call `translate_account_rc_to_ah()`. This means though that I need to handle XCM versions and translate location and junctions, see xcm_translations.rs

sigurpol:
@ggwpez , @re-gius , @muharem  I've addressed the two remaining critical issues:
- proper conversion of bags lists preserving prev/next relationship within a bag (which would be lost if we recreate them from scratch) and preserving score order between bags
- proper handling of #811  done within `pallet-ah-migrator`. The logic lives in  `xcm_translation.rs`. Not an XCM expert to say the least, but AFAIK the approach should work vs multiple XCM versions.

Can you please review and let me know what we are missing to be able to merge 🙏 ?

pallets/ah-migrator/src/xcm_translation.rs
```diff
@@ -0,0 +1,118 @@
...
+	fn translate_location_latest(location: Location) -> Result<Location, Error<T>> {
+		let translated_junctions = Self::translate_junctions_latest(location.interior)?;
+		Ok(Location { parents: location.parents, interior: translated_junctions })
+	}
+
+	/// Translate junctions in the latest XCM format
+	fn translate_junctions_latest(junctions: Junctions) -> Result<Junctions, Error<T>> {
+		junctions
+			.iter()
+			.map(|junction| Self::translate_junction_latest(junction.clone()))
```
muharem: I wonder if its valid to translate all `AccountId32` `Junction`s within any possible `Junctions`, or we should white list the `Junctions` we wanna attempt to translate. for example whitelist: X1(Junction::AccountId32), etc.

muharem: now I am not sure if its even needed.

I've seem one translation case for account id, is `para{para_rd}` -> `sibl{para_rd}`. The assumption here is -> someone created a spend with beneficiary on Asset Hub being the `X1(AccountId32(derive(`para{para_rd}`)))`? Not sure its possible. I think the parachain wont be able to access this account. If a parachain wanna receive a treasury spend on its account, it should specify its account as `(1, X1(Parachain({para_id})))`, which gonna be the derivation of `sibl{para_rd}`.

^ this all needs to be double checked, but this is my view to it. 

do we know the exact case we solving with this?

muharem: I mean do we have an example, Location X should be translated into Location Y

pallets/ah-migrator/src/treasury.rs
muharem: I am not sure if the locations needs to be translated. Here is my comment - https://github.com/polkadot-fellows/runtimes/pull/807/files#r2192762149

muharem (changes requested):
Lets do not merge this until we do not have the list of the derivation pattern that we need to map. I think currently the mapping for location is wrong, and not needed if we consider only parachains sov accounts. Additionally we might not need the translation for cases like bounties curators. I do not like the idea of blindly applying the translation everywhere.

pallets/ah-migrator/src/xcm_translation.rs
sigurpol: Great point! Here I tried to be generic because  I don't know the answer of the million dollar question: what are the legitimate patterns for treasury spend beneficiaries that would need account translation?
I  don' t know myself all possible cases and so I tried to go for a `broad` approach. 

If I understand correctly, we are saying the following.

### Standard pattern (No translation needed) - maybe the only real  use case?!?
```rust
  // Treasury spend targeting parachain treasury directly
  beneficiary: Location::new(1, Parachain(2030)) // Bifrost parachain
``` 
  This creates a spend where:
  - Asset Hub derives the beneficiary as sibl2030 automatically
  - No AccountId32 junction involved
  - No translation needed
  - All good
  
### Sovereign account pattern (Translation needed - but is it a real use case?!?)

What if  someone creates a treasury spend on Relay Chain targeting a Bifrost sovereign account?
``` 
  TreasurySpend {
      beneficiary: Location::new(0, AccountId32 {
          network: None,
          id: [/* Direct bytes of para2030 sovereign account */]
          // 13YMK2eeopZtUNpeHnJ1Ws2HqMQG6Ts9PGCZYGyFbSYoZfcm
      }),
      amount: 1000_DOT,
  }
  ```  
  So the spend creator wants to target a Bifrost specific sovereign account that holds DOT on RC. 
  This references the sovereign account directly by its AccountId32 bytes.  
  And needs to be translated like this
  ```
  TreasurySpend {
      beneficiary: Location::new(0, AccountId32 {
          network: None,
          id: [/* NOW translated to sibl2030 sovereign account */]
          // 13cKp89TtYknbyYnqnF6dWN75q5ZosvFSuqmcnN4AaNR47A
      }),
      amount: 1000_DOT,
  }
  ``` 
We actually have one example in the tests like this  -> https://github.com/polkadot-fellows/runtimes/blob/main/integration-tests/emulated/tests/assets/asset-hub-kusama/src/tests/treasury.rs?plain=1#L110

```rust
  // Test case: Treasury spend with direct AccountId32 beneficiary
  let alice_location: Location = [
      Junction::AccountId32 {
          network: None,
          id: Kusama::account_id_of(ALICE).into()
      }
  ].into();

  let treasury_spend_call = RuntimeCall::Treasury(pallet_treasury::Call::spend {
      beneficiary: bx!(VersionedLocation::from(alice_location)), // Direct AccountId32
      // ...
  });
  ```
  so it seems that at least we should be able to manage two cases if the test makes sense:
  1. Parachain(2030) = "Send to Bifrost parachain" 
  2. AccountId32(para2030_bytes) = "Send to Bifrost's sovereign account" (specific account)
  
  What do you think @muharem , @ggwpez ?
  
 (UPDATE: I've added a `TODO` in the code to summarize this thread)

**Comment by sigurpol** [comment]
> Lets do not merge this until we do not have the list of the derivation pattern that we need to map. I think currently the mapping for location is wrong, and not needed if we consider only parachains sov accounts. Additionally we might not need the translation for cases like bounties curators. I do not like the idea of blindly applying the translation everywhere.

I am fine with not merging now but to be fair, in this PR the translation is no-op, it just attaching the infrastructure to convert accounts and should be safe in that perspective (especially now that I have reverted the change about re-creationg of bags lists).

**Comment by sigurpol** [comment]
> > Lets do not merge this until we do not have the list of the derivation pattern that we need to map. I think currently the mapping for location is wrong, and not needed if we consider only parachains sov accounts. Additionally we might not need the translation for cases like bounties curators. I do not like the idea of blindly applying the translation everywhere.
> 
> I am fine with not merging now but to be fair, in this PR the translation is no-op, it just attaching the infrastructure to convert accounts and should be safe in that perspective (especially now that I have reverted the change about re-creationg of bags lists).

@ggwpez , @re-gius, update:  as discussed via DM with @muharem , we park temporarily this PR until we clarify a couple of key points:
-  evaluate the approach of this PR (translate `everywhere`) vs a more targeted one (translate only where needed)
   - this requires to answer the question: what are the pallets where must translate vs the one a translation is not needed?
   - we need to answer the question in any case even if we go for the approach of this PR when we can actually implement the translation function
-  understand the exact requirements / use cases for treasury spends (i.e. do we need to translate location?  no / yes / yes but let's whitelist the 1-2 cases that matter instead of the broad approach of this PR that translates any possible junction)


We will resurrect the PR when we have a better understanding of the topic

File: pallets/ah-migrator/src/staking/nom_pools.rs [hunk 18]
```diff
@@ -86,15 +123,50 @@ impl<T: Config> Pallet<T> {
...
+					perms.0.to_polkadot_ss58(), translated_account.to_polkadot_ss58());
+				pallet_nomination_pools::ClaimPermissions::<T>::insert(translated_account, perms.1);
 			},
 		}
 	}
+
+	/// Translate a bonded RC pool to an AH one.
+	pub fn rc_to_ah_bonded_pool(mut pool: BondedPoolInner<T>) -> BondedPoolInner<T> {
+		if let Some(ref mut throttle_from) = pool.commission.throttle_from {
+			// Plus one here to be safe for the pool member just in case that the pool operator
```
sigurpol: should we keep this plus one ? Does it make sense? I can definitely remove it (here and the equivalent plus one in the post check) if we feel it is not needed.
Wdyt @ggwpez , @muharem ?

sigurpol:
An attempt to actually replace the mocked translation with a real one has been introduced with the commit [https://github.com/polkadot-fellows/runtimes/pull/807/commits/830da1384d662260cf654f082cef310ccbdfd6ba](https://www.google.com/url?q=https://github.com/polkadot-fellows/runtimes/pull/807/commits/830da1384d662260cf654f082cef310ccbdfd6ba&sa=D&source=docs&ust=1752072531871108&usg=AOvVaw3ER-uaHB-bUOnPM-Irj0Lv)

Doesn't mean we will go with the approach of #807  but I wanted to show a possible implementation to compare with alternative approaches @ggwpez  @muharem  @re-gius

ggwpez:
We have special function sin the AH-ops pallet that do the translation, no need to replicate it here :)

sigurpol:
I've reverted the para -> sibl conversion so this PR is back to the initial state where the translation is just a mocked function.

Please review  @ggwpez , @muharem, @re-gius  if you have time.

sigurpol:
> We have special function sin the AH-ops pallet that do the translation, no need to replicate it here :) And lets just keep it mocked for now, i want to do separation of concerns so we can merge this here quicker.

Agreed. Back to a mocked state now.

pallets/ah-migrator/src/staking/bags_list.rs
ggwpez: There should also be an integrity check for the bags-list pallet that we currently dont call. Probably worth to call that, this one  
https://github.com/paritytech/polkadot-sdk/blob/6824a6782354e274cf308be314f142551c5457ab/substrate/frame/bags-list/src/lib.rs#L405 (try_state).  
You probably have to call it directly, as the function is feature gated `<Self as SortedListProvider<T::AccountId>>::try_state()`.

But if it does not work then we need to just revert and do later. As I want to get this merged soon.

pallets/ah-migrator/src/staking/nom_pools.rs
```diff
@@ -16,15 +16,43 @@
...
 };
+use sp_runtime::{traits::One, Saturating};
+
+/// Macro to generate account translation logic for bonded pool structures.
+///
+/// This macro works with two different pool types that share the same field structure:
+/// 1. `BondedPoolInner<T>` - The actual runtime pallet type used during migration
+/// 2. `tests::GenericBondedPoolInner<Balance, AccountId, BlockNumber>` - Generic test type used for
+///    verification
+macro_rules! translate_bonded_pool_accounts {
```
ggwpez: Generally not such a big fan or macros, since they are easy to write but difficult to read (and audit).  
Could be okay for now, but for Kusama it should be cleaned up IMHO.

pallets/ah-migrator/src/staking/nom_pools.rs
ggwpez: No both are not needed anymore, was just me being paranoid :)

ggwpez: Didnt we also remove this `rc_to_ah_timepoint` function recently?

pallets/ah-migrator/src/staking/nom_pools.rs
```diff
@@ -177,6 +249,51 @@ impl<T: Config> crate::types::AhMigrationCheck for NomPoolsMigrator<T> {
...
+							Pallet::<T>::translate_account_rc_to_ah(account_id.clone());
+						PoolMembers { member: (translated_account, member.clone()) }
+					},
+					BondedPools { pool: (pool_id, pool) } => {
+						let mut translated_pool = pool.clone();
+						translate_bonded_pool_accounts!(
+							translated_pool,
+							Pallet::<T>::translate_account_rc_to_ah
+						);
+						// Apply timepoint transformation to match what rc_to_ah_bonded_pool does
```
ggwpez: I think its not needed anymore to translate timpoints, Muharem removed it recently.

ggwpez (approved):

pallets/ah-migrator/src/staking/nom_pools.rs
sigurpol: sure, can go back to a non-macro variant in this PR already, not a big deal

pallets/ah-migrator/src/scheduler.rs
```diff
@@ -28,6 +28,11 @@ pub type RcScheduledOf<T> =
 	Scheduled<call::BoundedCallOf<T>, BlockNumberFor<T>, <T as Config>::RcPalletsOrigin>;
 
 impl<T: Config> Pallet<T> {
+	// NOTE: no direct account migration is provided for scheduler. The scheduler doesn't store
```
muharem, line 31: 👍

pallets/ah-migrator/src/referenda.rs
```diff
@@ -92,7 +153,8 @@ impl<T: Config> Pallet<T> {
 		let (mut count_good, mut count_bad) = (0, 0);
 
 		for (id, referendum) in referendums {
-			match Self::do_receive_referendum(id, referendum) {
+			let translated_referendum = Self::translate_referendum_info(referendum);
```
muharem: minor: may be `translate_referendum_accounts(rc_info)`, the actual translation happens later.

pallets/ah-migrator/src/referenda.rs
```diff
@@ -378,7 +440,9 @@ impl<T: Config> crate::types::AhMigrationCheck for ReferendaMigrator<T> {
 		fn convert_rc_to_ah_referendum<T: Config>(
 			rc_info: RcReferendumInfoOf<T, ()>,
 		) -> AhReferendumInfoOf<T, ()> {
-			match rc_info {
+			let translated_rc_info = crate::Pallet::<T>::translate_referendum_info(rc_info);
```
muharem: not really safe, the test uses the same map function as the code it testing. but may be a good review it enough.

pallets/ah-migrator/src/conviction_voting.rs
```diff
@@ -95,33 +124,63 @@ impl<T: Config> crate::types::AhMigrationCheck for ConvictionVotingMigrator<T> {
...
+		// Build expected data by applying account translation to RC pre-payload data
+		let expected_ah_messages: Vec<_> = rc_pre_payload
+			.iter()
+			.map(|message| match message {
+				RcConvictionVotingMessage::VotingFor(account_id, class, voting) => {
+					// Translate the voter account
+					let translated_account =
+						Pallet::<T>::translate_account_rc_to_ah(account_id.clone());
+					// Translate delegate accounts in the voting structure
+					let translated_voting = Pallet::<T>::translate_voting_accounts(voting.clone());
```
muharem: `translate_voting_accounts` is not that big, would be nicer to just repeat that logic here and not use it

pallets/ah-migrator/src/multisig.rs
```diff
@@ -61,29 +61,49 @@ impl<T: Config> Pallet<T> {
...
+
+		// Translate the details account (derived multisig account) if present.
+		// NOTE: There are instances where we expect the translation to be a no-op. It's acceptable
+		// to retain it for now and remove it later if we determine it is consistently a no-op.
+		let translated_details = multisig
+			.details
+			.as_ref()
+			.map(|details_account| Self::translate_account_rc_to_ah(details_account.clone()));
+
+		log::debug!(target: LOG_TARGET, "Integrating multisig {} -> {}, deposit: {:?}, details: {:?} -> {:?}",
```
muharem: may be we better do not do this `{} -> {}` translation log everywhere and only include the translated account into the log. 
but will have this translation log within `translate_account_rc_to_ah` only when an account is actually translated. this would cut a lot of noise. 
additionally we would need to make sure there is some log before or after that would help us to identify for what that translation was applied.

pallets/ah-migrator/src/staking/nom_pools.rs
muharem: this should be gone in dev branch. you need to merge dev branch, or the conflicts were resolved in wrong way.

pallets/ah-migrator/src/staking/nom_pools.rs
muharem: lets please not use macro here

pallets/ah-migrator/src/staking/nom_pools.rs
```diff
@@ -177,6 +249,51 @@ impl<T: Config> crate::types::AhMigrationCheck for NomPoolsMigrator<T> {
...
+						let translated_account =
+							Pallet::<T>::translate_account_rc_to_ah(account_id.clone());
+						PoolMembers { member: (translated_account, member.clone()) }
+					},
+					BondedPools { pool: (pool_id, pool) } => {
+						let mut translated_pool = pool.clone();
+						translate_bonded_pool_accounts!(
+							translated_pool,
+							Pallet::<T>::translate_account_rc_to_ah
+						);
```
muharem: lets please rewrite the mapping for the test and use reuse what we use within the actual implementation

pallets/ah-migrator/src/xcm_translation.rs
muharem: yes, your understanding is right. we only need to take care of (2) and only consider one junction - `X1(Junction:: AccountId32)`, if anything else we should ignore.

pallets/ah-migrator/src/multisig.rs
ggwpez: Yea just logging the final acc is fine. We have events anyway and in the trnaslate function we could also add a log if we want.

pallets/ah-migrator/src/conviction_voting.rs
sigurpol: Done

pallets/ah-migrator/src/multisig.rs
sigurpol: Done

pallets/ah-migrator/src/referenda.rs
sigurpol: done

pallets/ah-migrator/src/staking/nom_pools.rs
sigurpol: Removed the macro

pallets/ah-migrator/src/staking/nom_pools.rs
sigurpol: Fair, something went wrong with me rebasing. Now it's gone also here :)

pallets/ah-migrator/src/staking/nom_pools.rs
sigurpol: indeed, not needed anymore, removed

pallets/ah-migrator/src/staking/nom_pools.rs
sigurpol: done

pallets/ah-migrator/src/xcm_translation.rs
sigurpol: Implemented 2.

pallets/ah-migrator/src/staking/bags_list.rs
sigurpol: Added in an ad-hoc commit so we can easily revert if the integration test fails 🤞

pallets/ah-migrator/src/referenda.rs
sigurpol: Fair! Reworked.

pallets/ah-migrator/src/staking/bags_list.rs
sigurpol: CI's usual `ahm-test-polkadot` is happy with that, so we should be good

pallets/ah-migrator/src/referenda.rs
```diff
@@ -81,6 +81,67 @@ pub type AhReferendumStatusOf<T, I> = ReferendumStatus<
...
+				});
+				let translated_decision = decision_deposit.map(|mut deposit| {
+					deposit.who = Self::translate_account_rc_to_ah(deposit.who.clone());
+					deposit
+				});
+				ReferendumInfo::Rejected(block, translated_submission, translated_decision)
+			},
+			ReferendumInfo::Cancelled(block, submission_deposit, decision_deposit) => {
+				let translated_submission = submission_deposit.map(|mut deposit| {
+					deposit.who = Self::translate_account_rc_to_ah(deposit.who.clone());
```
ggwpez, line 121: Looks clean now 😄

pallets/ah-migrator/src/staking/bags_list.rs
```diff
@@ -121,8 +182,15 @@ impl<T: Config> crate::types::AhMigrationCheck for BagsListMigrator<T> {
...
 		// Assert storage "VoterList::ListBags::ah_post::consistent"
 		assert_eq!(
-			rc_pre_payload, ah_messages,
+			rc_pre_translated, ah_messages,
 			"Bags list data mismatch: Asset Hub data differs from original Relay Chain data"
 		);
+
+		// Run bags-list pallet integrity check
+		#[cfg(feature = "try-runtime")]
+		<pallet_bags_list::Pallet<T, pallet_bags_list::Instance1> as frame_election_provider_support::SortedListProvider<
```
ggwpez, line 191: Note to self: Remove the feature gate in the SDK so that we can use it here as well.

ggwpez (approved):
LGTM

pallets/ah-migrator/src/xcm_translation.rs
```diff
@@ -0,0 +1,102 @@
...
+	pub fn translate_beneficiary_location(
+		location: VersionedLocation,
+	) -> Result<VersionedLocation, Error<T>> {
+		// Convert to latest version for unified processing
+		let latest_location: Location = location.clone().try_into().map_err(|_| {
+			log::error!(
+				target: LOG_TARGET,
+				"Failed to convert VersionedLocation to latest version"
+			);
+			Error::<T>::FailedToConvertType
```
muharem: may be we log, but return what we have received? not worth giving up the migration of a treasury spend if we failed the version conversion

pallets/ah-migrator/src/xcm_translation.rs
```diff
@@ -0,0 +1,102 @@
...
+					let original_version = location.identify_version();
+					return VersionedLocation::from(translated_location)
+						.into_version(original_version)
+						.map_err(|_| {
+							log::error!(
+								target: LOG_TARGET,
+								"Failed to convert back to original XCM version {}",
+								original_version
+							);
+							Error::<T>::FailedToConvertType
```
muharem: probably none of this errors worth giving up the spend migration. we can log and return the original location. or we do it on a caller side if error is returned.

pallets/ah-migrator/src/xcm_translation.rs
```diff
@@ -0,0 +1,102 @@
...
+		})?;
+
+		// Check if this is exactly X1(AccountId32) pattern
+		if latest_location.parents == 0 && latest_location.interior.len() == 1 {
+			if let Some(junction) = latest_location.interior.first() {
+				if let Some(id) = junction.get_account_id32() {
+					// This is X1(AccountId32), translate it
+					let account_id = T::AccountId::decode(&mut &id[..])
+						.map_err(|_| Error::<T>::FailedToConvertType)?;
+					let translated_account = Self::translate_account_rc_to_ah(account_id);
```
muharem: if translated account is equal to the original, we can return early.

pallets/ah-migrator/src/conviction_voting.rs
```diff
@@ -61,20 +64,42 @@ impl<T: Config> Pallet<T> {
 		class: alias::ClassOf<T>,
 		voting: alias::VotingOf<T>,
 	) {
-		log::debug!(target: LOG_TARGET, "Processing VotingFor record for: {:?}", &account_id);
-		alias::VotingFor::<T>::insert(account_id, class, voting);
+		// Translate the voter account from RC to AH format
+		let translated_account = Self::translate_account_rc_to_ah(account_id.clone());
+
+		log::debug!(target: LOG_TARGET, "Processing VotingFor record for: {} -> {}",
```
muharem: do we still wanna log both account ids?

muharem: can we log only final account id everywhere? there are more places where we still have `{} -> {}`.

sigurpol: Done, hopefully everywhere this time :)

pallets/ah-migrator/src/xcm_translation.rs
sigurpol: Done

pallets/ah-migrator/src/xcm_translation.rs
sigurpol: Done

pallets/ah-migrator/src/xcm_translation.rs
sigurpol: done

PR #808, opened 2025-07-06
[AHM] Migration idempotency

muharem:
Migration idempotency

This PR introduces support for running the migration multiple times from any stage within the range `[AccountsMigrationOngoing { last_key: None }, MigrationDone]`. 

Changes:
- default values in StorageValues are excluded from migration data.
- redundant data transfers and actions that could occur during repeated runs are filtered or disabled to minimize unnecessary logs, events and message confirmaitons.
- proxy announcements are removed after being migrated from RC.
- treasury::LastSpendPeriod storage value is migrated.
- no checking balance correction on subsequent runs.
- added tests to verify migration idempotency.

Subsequent runs should not start from `AccountsMigrationInit` or call `start_migration` on Asset Hub, as these step records the total issuances and captures the checking account balances used for the checking account correction on Asset Hub as the last step of the migration. This part is not currently idempotent.

While it could be made idempotent - for example, by moving the checking balance correction to a separate migration stage - such a change would require rewriting the correction logic and its tests. Given the limited impact, current priorities, and the fact that the balance correction likely needs a full redesign anyway, this is not considered necessary at the moment.


integration-tests/ahm/src/tests.rs
```diff
@@ -470,13 +484,38 @@ async fn migration_works_time() {
...
 		rc_block_end - rc_block_start,
 		ah_block_end - ah_block_start
 	);
+
+	// run the migration again to check its idempotent
+
+	// Set the initial migration stage from env var if set.
+	set_initial_migration_stage(
+		&mut rc,
+		Some(RcMigrationStage::AccountsMigrationOngoing { last_key: None }),
```
muharem, line 500: we do not wanna run the accounts init that includes the balances record initialization since it is not idempotent

pallets/ah-migrator/src/lib.rs
```diff
@@ -187,8 +187,6 @@ pub enum MigrationStage {
 	Pending,
 	/// Migrating data from the Relay Chain.
 	DataMigrationOngoing,
-	/// Migrating data from the Relay Chain is completed.
-	DataMigrationDone,
```
muharem, line 191: this variant is not used

pallets/ah-migrator/src/referenda.rs
```diff
@@ -221,14 +221,20 @@ impl<T: Config> Pallet<T> {
...
-		});
-		track_queue.into_iter().for_each(|(track_id, queue)| {
-			let queue = BoundedVec::<_, T::MaxQueued>::defensive_truncate_from(queue);
-			TrackQueue::<T, ()>::insert(track_id, queue);
-		});
+		if let Some(referendum_count) = referendum_count {
+			ReferendumCount::<T, ()>::put(referendum_count);
+		}
+		if deciding_count.len() > 0 {
+			deciding_count.iter().for_each(|(track_id, count)| {
```
ggwpez, line 228: Why the length check? If its empty then just nothing happens anyway?

pallets/rc-migrator/src/accounts.rs
```diff
@@ -675,6 +685,11 @@ impl<T: Config> AccountsMigrator<T> {
 	///
 	/// Should be executed once before the migration starts.
 	pub fn obtain_rc_accounts() -> Weight {
+		if RcAccounts::<T>::count() > 0 {
```
ggwpez, line 688: I wonder if there are cases where we would want to re-run this indexing?  
But we could still add an extrinsic to the ah-ops pallet, if needed.

pallets/rc-migrator/src/lib.rs
```diff
@@ -1705,9 +1719,22 @@ pub mod pallet {
 					pallet_staking_async_ah_client::Pallet::<T>::on_migration_end();
 
 					// Send finish message to AH, TODO: weight
-					let tracker = RcMigratedBalance::<T>::get();
-					let data = MigrationFinishedData {
-						rc_balance_kept: tracker.kept,
+					let data = if RcMigratedBalance::<T>::exists() {
+						let tracker = if cfg!(feature = "std") {
+							// we should keep this value for the tests.
```
ggwpez, line 1724: Instead we could also filter for the event?

pallets/rc-migrator/src/staking/message.rs
```diff
@@ -445,21 +445,57 @@ impl<T: pallet_staking::Config> StakingMigrator<T> {
...
-			min_validator_count: MinimumValidatorCount::<T>::take(),
-			min_nominator_bond: MinNominatorBond::<T>::take(),
-			min_validator_bond: MinValidatorBond::<T>::take(),
-			min_active_stake: MinimumActiveStake::<T>::take(),
-			min_commission: MinCommission::<T>::take(),
+			validator_count: if ValidatorCount::<T>::exists() {
+				Some(ValidatorCount::<T>::take())
+			} else {
+				None
+			},
```
ggwpez:
```suggestion
			validator_count: ValidatorCount::<T>::exists().then(ValidatorCount::<T>::take),
```
A bit shorter, but its fine like this as well 😋

pallets/rc-migrator/src/staking/message.rs
```diff
@@ -469,24 +505,49 @@ impl<T: pallet_staking_async::Config> StakingMigrator<T> {
 	pub fn put_values(values: AhStakingValuesOf<T>) {
 		use pallet_staking_async::*;
 
-		ValidatorCount::<T>::put(&values.validator_count);
+		if let Some(validator_count) = values.validator_count {
+			ValidatorCount::<T>::put(validator_count);
+		}
```
ggwpez:
```suggestion
		values.validator_count.map(ValidatorCount::<T>::put);
```
🙈

pallets/rc-migrator/src/staking/nom_pools.rs
```diff
@@ -71,14 +71,27 @@ pub enum NomPoolsStage<AccountId> {
...
 	pub max_pools: Option<u32>,
 	pub max_pool_members: Option<u32>,
 	pub max_pool_members_per_pool: Option<u32>,
 	pub global_max_commission: Option<Perbill>,
-	pub last_pool_id: u32,
+	pub last_pool_id: Option<u32>,
+}
+
+impl<Balance> NomPoolsStorageValues<Balance> {
+	pub fn is_empty(&self) -> bool {
```
ggwpez, line 85: Tthere is no harm in sending an empty message, or?

pallets/rc-migrator/src/treasury.rs
```diff
@@ -168,8 +174,10 @@ impl<T: Config> PalletMigration for TreasuryMigrator<T> {
 				},
 				// TODO: with new sdk version
```
ggwpez: We can actually do this now

ggwpez (commented):
The treasury `LastSpendPeriod` would be good to add

pallets/ah-migrator/src/referenda.rs
muharem: Yes, the `DecidingCount` is `ValueQuery`, no reason to insert empty vec, it can only rewrite a previous correct value.

pallets/rc-migrator/src/accounts.rs
muharem: its hard to imagine, bc the list should not change and even if it would we can just ignore it, the impact gonna be very low. but I might add the option to run it manually

pallets/rc-migrator/src/lib.rs
muharem: we need this in post_check, this event probably wont be available anymore. but I also do not like this much, might look for a different solution

pallets/rc-migrator/src/staking/nom_pools.rs
muharem: no harm. but it crates some false logs/events/counts/ and not unnecessary waitings for confirmations from AH.

pallets/ah-migrator/src/referenda.rs
ggwpez: But the iterator will just do nothing if its empty? Sorry for DQ

pallets/rc-migrator/src/lib.rs
ggwpez: Okay i see. I thought it was for the integration tests.

ggwpez (approved):

PR #812, opened 2025-07-08
[AHM] Async staking config

ggwpez:
WIP


system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -21,9 +21,81 @@
...
+
+parameter_types! {
+	/// Number of election pages that we operate upon. 32 * 6s block = 192s = 3.2min snapshots
+	pub Pages: u32 = 32;
+
+	/// Compatible with Polkadot, we allow up to 22_500 nominators to be considered for election
+	pub MaxElectingVoters: u32 = 22_500;
+
+	/// Maximum number of validators that we may want to elect. 1000 is the end target.
+	pub const MaxValidatorSet: u32 = 1000;
```
ggwpez: @kianenigma I copied these from Westend, please make adjustments and directly push to the branch, thanks.

sigurpol: Some values seems indeed  off e.g. SignedValidationPhase (I believe we want to be able to validate up to 8 good solutions). AFAIK the closest approximation to a source of truth are the comments per chain in this file https://github.com/paritytech/polkadot-sdk/blob/d1ccd6b789d7e19bd9ca80be03df77eed231d306/substrate/frame/staking-async/runtimes/parachain/src/staking.rs but @kianenigma  can definitely confirm

kianenigma: I will push a commit with fixes to this branch soon.

Cargo.toml
```diff
@@ -72,40 +72,40 @@ path = "system-parachains/asset-hubs/asset-hub-polkadot"
...
 git = "https://github.com/paritytech/polkadot-sdk"
-branch = "oty-2503-paseo"
+branch = "oty-tmp2"
 
 [workspace.dependencies.assets-common]
 version = "*"
 default-features = false
 git = "https://github.com/paritytech/polkadot-sdk"
-branch = "oty-2503-paseo"
+branch = "oty-tmp2"
```
kianenigma: We discussed this in the call as well, I see in the oty-tmp2 that the versions of the staking async + epmb that is being deployed here are pretty [old](https://github.com/paritytech/polkadot-sdk/commits/oty-tmp2/substrate/frame/election-provider-multi-block).. should work fine, but I do wish we had time to properly put 2507 on Paseo :(

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/call_filter.rs
```diff
@@ -86,9 +86,13 @@ pub fn call_allowed_status(call: &<Runtime as frame_system::Config>::RuntimeCall
 		ForeignAssets(..) => ON,
 		Indices(..) => OFF,
 		MessageQueue(..) => ON, // TODO think about this
+		MultiBlockElection(..) => ON,
+		MultiBlockElectionSigned(..) => ON,
+		MultiBlockElectionUnsigned(..) => ON,
+		MultiBlockElectionVerifier(..) => ON,
 		Multisig(..) => OFF,
-		NominationPools(..) => OFF,
 		Nfts(..) => ON,
```
kianenigma, line 95: are you sure?

sigurpol: + possibly silly slightly related question: is there a specific reason why DelegateStaking is not included in this list?

kianenigma: It is not currently in the runtime, otherwise the match statement would fail.

We missed this pallet in westend as well, I am sure we will not this time. 

@ggwpez? :D

ggwpez: Delegated staking pallet is deployed. But it does not seem to have any calls? Only used through the `DelegationInterface`?

ggwpez: I wasnt sure if they should be enabled. Should they or not?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -21,9 +21,89 @@
...
+		VoterIndex = u32,
+		// allows up to 64k validators
+		TargetIndex = u16,
+		Accuracy = sp_runtime::PerU16,
+		MaxVoters = VoterSnapshotPerBlock,
+	>(16)
+);
+
+ord_parameter_types! {
+	pub const PolkadotStakingMiner: AccountId = AccountId::from(hex_literal::hex!("b65991822483a6c3bd24b1dcf6afd3e270525da1f9c8c22a4373d1e1079e236a"));
```
ggwpez: And we need this or is it only for testnet?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -49,10 +129,12 @@ parameter_types! {
 type VoterBagsListInstance = pallet_bags_list::Instance1;
 impl pallet_bags_list::Config<VoterBagsListInstance> for Runtime {
 	type RuntimeEvent = RuntimeEvent;
-	type ScoreProvider = nom_pools::StakingMock;
+	type ScoreProvider = Staking;
 	type WeightInfo = (); // TODO weights::pallet_bags_list::WeightInfo<Runtime>;
 	type BagThresholds = BagThresholds;
 	type Score = sp_npos_elections::VoteWeight;
+	// only for paseo, use 0 for Polkadot.
+	type MaxAutoRebagPerBlock = ConstU32<5>;
```
ggwpez, line 143: Why should it be different?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -68,5 +150,291 @@ impl pallet_delegated_staking::Config for Runtime {
...
+	type Solver = <Runtime as multi_block::unsigned::Config>::OffchainSolver;
+	type Pages = Pages;
+	type Solution = NposCompactSolution16;
+	type VoterSnapshotPerBlock = <Runtime as multi_block::Config>::VoterSnapshotPerBlock;
+	type TargetSnapshotPerBlock = <Runtime as multi_block::Config>::TargetSnapshotPerBlock;
+}
+
+// FAIL-CI wrong use the one from the relay
+pub struct EraPayout;
+impl pallet_staking_async::EraPayout<Balance> for EraPayout {
```
ggwpez, line 292: I will factor this out into some common crate.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -68,5 +150,291 @@ impl pallet_delegated_staking::Config for Runtime {
...
+
+		(to_stakers.saturated_into(), to_treasury.saturated_into())
+	}
+}
+
+parameter_types! {
+	// Six sessions in an era (6 hours).
+	pub const SessionsPerEra: SessionIndex = prod_or_fast!(6, 2);
+	/// Duration of a relay session in our blocks. Needs to be hardcoded per-runtime.
+	pub const RelaySessionDuration: BlockNumber = 1 * HOURS;
```
ggwpez: Did you also check these values @kianenigma? I did not put TODOs everywhere, but the stuff is all from Westend.

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/call_filter.rs
sigurpol: (you are right about delegated staking not exposing call, just double-checked. Next time I'll do at lest a grep per `#\[pallet::call\]" before bothering  you 😓 ). It's even written in the documentation that `"does not expose any
  dispatchable calls but is written with a vision to expose them in the future." 😄

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: We need it for all testnets incl Paseo, but not for Polkadot Kusama relay ofc.

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/call_filter.rs
sigurpol: And @kianenigma ,can you confirm we want NominationPallets and also all MultiBlock* pallets off during migration ?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: We should not enable it in Polkadot at first, since it is not tested in any prod/testnet yet.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: ahh good point, not, let me check the Paseo configs.

kianenigma:
Also, will there be a PR to paseo RC at some point? or should I open one? some of the staking configs will have to go there.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: ede7463152fe4989e6527e76b92cce57ac2e2630

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/call_filter.rs
kianenigma: Yup, they should be off.

ggwpez:
We will run our translation script that converts Polkadot to Paseo runtimes and then propose that as MR. But the diff should be auto-generated. Donal is looking into it.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
ggwpez: I will add a Paseo feature-gate then so that we dont accidentally do it in prod.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
ggwpez: Actually no since its for staking_async and just copy&paste is simpler for now.

relay/polkadot/src/weights/pallet_bags_list.rs
```diff
@@ -62,6 +62,30 @@ impl<T: frame_system::Config> pallet_bags_list::WeightInfo for WeightInfo<T> {
...
+	/// Storage: `VoterList::ListNodes` (r:11 w:11)
+	/// Proof: `VoterList::ListNodes` (`max_values`: None, `max_size`: Some(154), added: 2629, mode: `MaxEncodedLen`)
+	/// Storage: `Staking::Bonded` (r:10 w:0)
+	/// Proof: `Staking::Bonded` (`max_values`: None, `max_size`: Some(72), added: 2547, mode: `MaxEncodedLen`)
+	/// Storage: `Staking::Ledger` (r:10 w:0)
+	/// Proof: `Staking::Ledger` (`max_values`: None, `max_size`: Some(1091), added: 3566, mode: `MaxEncodedLen`)
+	fn on_idle() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `4777`
+		//  Estimated: `512390`
```
ggwpez, line 82: big

kianenigma:
> These configs will be used for Polkadot and Paseo.

Some details like deposits and so on will definitely be different between Polkadot and Paseo. So I am reviewing this only for Paseo. We def need to revisit before Polkadot. 

I am okay with this and will review one last time before approving, but in general I think it will be much easier to have the final review of this when it is in the form of a PR in the paseo runtime, not here. 

For example, if this is meant to be the final final code, we need to: 

1. Add ah-client to RC
2. Use it as session manager and offence handler
3. make sure the pallet index of ah-client and rc-client is correct in both runtime 

(all of the above is missing now)

I can fix all of this, but then the Polkadot runtime will be in a state where the next upgrade for it has to be AHM. Is this the intention?

kianenigma (approved):

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -49,10 +130,14 @@ parameter_types! {
 type VoterBagsListInstance = pallet_bags_list::Instance1;
 impl pallet_bags_list::Config<VoterBagsListInstance> for Runtime {
 	type RuntimeEvent = RuntimeEvent;
-	type ScoreProvider = nom_pools::StakingMock;
+	type ScoreProvider = Staking;
 	type WeightInfo = (); // TODO weights::pallet_bags_list::WeightInfo<Runtime>;
```
muharem: how about generating weights for all staking related pallets? I am closing the rest in another PR.
we cannot leave this todos for longer

ggwpez: Yea doing it now.

relay/polkadot/src/lib.rs
```diff
@@ -486,19 +486,22 @@ impl_opaque_keys! {
 }
 
 impl pallet_session::Config for Runtime {
+	type Currency = Balances;
+	type KeyDeposit = ();
```
Ank4n: Optional: If we are planning to set key deposit post AHM in Polkadot, we should consider settting it pre AHM itself just to decouple this change from AHM. In any case, we should definitely set it on Paseo first.

relay/polkadot/src/lib.rs
```diff
@@ -1509,6 +1519,119 @@ impl pallet_delegated_staking::Config for Runtime {
...
 
+impl pallet_staking_async_ah_client::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type CurrencyBalance = Balance;
+	type AssetHubOrigin =
+		frame_support::traits::EitherOfDiverse<EnsureRoot<AccountId>, EnsureAssetHub>;
+	type AdminOrigin = EnsureRoot<AccountId>;
+	type SessionInterface = Self;
+	type SendToAssetHub = StakingXcmToAssetHub;
+	type MinimumValidatorSetSize = ConstU32<1>;
```
Ank4n: This should be a bigger number for Polkadot. I guess it could be either the actual expected validator size (600) or at least 2/3rd of it (400).

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/call_filter.rs
```diff
@@ -98,6 +102,8 @@ pub fn call_allowed_status(call: &<Runtime as frame_system::Config>::RuntimeCall
 		Referenda(..) => OFF,
 		Scheduler(..) => ON,
 		Session(..) => OFF,
+		Staking(..) => OFF,
+		StakingRcClient(..) => ON,
```
Ank4n: Is it intentional to keep this ON?

Ank4n: Actually, it makes sense to be `ON`. This pallet only receives xcm messages from RC origin, and since RC might potentially send a bunch of offence messages at the end of migration, it's safer to avoid potential timing issues by keeping it enabled. 

Wonder if that was your thinking as well, might just leave a comment in the code about the reasoning.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -21,9 +21,90 @@
...
+		20 * MINUTES,
+		4 * MINUTES
+	);
+
+	/// Offchain miner shall mine at most 4 pages.
+	pub storage MinerPages: u32 = 4;
+
+	/// 30m for unsigned phase.
+	pub storage UnsignedPhase: u32 = prod_or_fast!(
+		30 * MINUTES,
```
Ank4n, line 56: Curious why the unsigned phase is longer than the signed phase?

Just checking if this timing breakdown is accurate (might be good to add as a comment):
4m (Snapshot) + 20m (Signed) + 26m (SV) + 30m (Unsigned) + 4m (Export) ≈ 2 hours?

Given Polkadot has 4-hour sessions, it might make sense to extend the signed phase to account for potential high on-chain activity (e.g., smart contracts).

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -21,9 +21,90 @@
...
+	);
+
+	/// Allow OCW miner to at most run 4 times in the entirety of the 10m Unsigned Phase.
+	pub OffchainRepeat: u32 = UnsignedPhase::get() / 4;
+
+	/// Compatible with Polkadot, we allow up to 22_500 nominators to be considered for election
+	pub storage MaxElectingVoters: u32 = 22_500;
+
+	/// Always equal to `staking.maxValidatorCount`.
+	pub storage TargetSnapshotPerBlock: u32 = MaxValidatorSet::get(); // TODO @kianenigma I changed this from 2000
```
Ank4n: The target snapshot is single page for now, so it should be 2000 I think.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -21,9 +21,90 @@
...
+
+	/// Number of nominators per page of the snapshot, and consequently number of backers in the solution.
+	pub VoterSnapshotPerBlock: u32 = MaxElectingVoters::get() / Pages::get();
```
Ank4n: Minor: we should probably round up this number?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -36,7 +117,7 @@ impl pallet_fast_unstake::Config for Runtime {
 	type BatchSize = frame_support::traits::ConstU32<16>;
 	type Deposit = FastUnstakeDeposit;
 	type ControlOrigin = EnsureRoot<AccountId>;
-	type Staking = nom_pools::StakingMock;
+	type Staking = Staking;
 	type MaxErasToCheckPerBlock = ConstU32<1>;
 	// TODO: use weights::pallet_fast_unstake::WeightInfo<Runtime> instead of ()
 	type WeightInfo = pallet_ah_migrator::MaxOnIdleOrInner<AhMigrator, ()>;
```
Ank4n: Is this correct?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -68,5 +154,295 @@ impl pallet_delegated_staking::Config for Runtime {
...
+
+impl pallet_staking_async::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type Filter = ();
+	type OldCurrency = Balances;
+	type Currency = Balances;
+	type CurrencyBalance = Balance;
+	type RuntimeHoldReason = RuntimeHoldReason;
+	type CurrencyToVote = sp_staking::currency_to_vote::SaturatingCurrencyToVote;
+	type RewardRemainder = ();
```
Ank4n:
```suggestion
	type RewardRemainder = ResolveTo<TreasuryAccount, Balances>;
```

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -68,5 +154,295 @@ impl pallet_delegated_staking::Config for Runtime {
...
+	type Slash = ();
```
Ank4n:
```suggestion
	type Slash = ResolveTo<TreasuryAccount, Balances>;
```

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -21,9 +21,90 @@
...
+	/// Translates to "no limit" as of now.
+	pub MaxBackersPerWinner: u32 = VoterSnapshotPerBlock::get();
+
+	/// Total number of backers per winner across all pages.
+	///
+	/// Translates to "no limit" as of now.
+	pub MaxBackersPerWinnerFinal: u32 = MaxElectingVoters::get();
+
+	/// Size of the exposures. This should be small enough to make the reward payouts feasible.
+	pub MaxExposurePageSize: u32 = 512;
```
Ank4n, line 89: Do we want to reduce this number?

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/call_filter.rs
ggwpez: Yea I was assuming that RC will send some calls to it, which is why i kept it on.  
I can put a comment.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
ggwpez: Ah yea... this is a bit weird since we need to disable the `on_idle` of the fast-unstake pallet during the migration.  
Will def add a comment, thanks for pointing it out. But we also still need to use the correct weight into instead of `()` 😆

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: Let's take this to https://github.com/paritytech/polkadot-sdk/pull/8802 where we are spec-ing these things out?

relay/polkadot/src/lib.rs
kianenigma: Yeah okay, let's set it to a small value so we can try it, it is a good idea.

relay/polkadot/src/lib.rs
kianenigma: Paseo seems to have 150 validators. Let's set it to 50.

(not relevant for Polkadot ofc)

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: Yeah similarly I reviewed this stuff with the intention of "paseo" in mind, not Polkadot 😅

2000 is better. I think it was changed maybe for benchmarking. Let's revert please.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: ah yeah, a great nit actually.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: I think it will be okay once `()` is replaced.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: Yeah Paseo actually has a treasury, let's do this.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: I wouldn't do it now to avoid further complications, 512 will work fine weight-wise, and my opinion is to minimize the moving pieces during AHM :) 

both polkadot and paseo use 512.

relay/polkadot/src/lib.rs
ggwpez: Okay then I feature-gate 50 for Paseo and let it 400 for Polkadot as we will re-check before production anyway.

relay/polkadot/src/lib.rs
ggwpez: Setting it to 1 DOLLARS now.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
ggwpez: done now

relay/polkadot/Cargo.toml
```diff
@@ -113,9 +115,6 @@ sp-debug-derive = { workspace = true }
 # just for the coretime migration
 polkadot-parachain-primitives = { workspace = true }
 
-# just for use with zombie-bite to test migration
-pallet-sudo = { workspace = true, optional = true }
```
kianenigma, line 117: @ggwpez this breaks the flow of ahm-dry-run. Is there a fallback? 

https://github.com/paritytech/ahm-dryrun/blob/81c9aa9f526e9568bc08f3a65371e26fcff3cee9/zombie-bite-scripts/helpers.ts#L137-L163

CC @pepoviola @x3c41a

kianenigma: So ZN, if possible, should set //Alice as the manager here and then it should wokr again. 

https://github.com/polkadot-fellows/runtimes/blob/22116f7d02c220db4f7187c6967dbd6bf89274cf/pallets/rc-migrator/src/lib.rs#L702-L707

kianenigma: I think this stuff happens deep in the ZB engine.

kianenigma: I don't see any ways to set storage at the top level TS codes of ahm-dryrun repo, probably when you call `zombie-bite <args>` it will override some storages from the the chains that it has warp-sync-ed before, changing session keys and so on.

but be aware that hard-coding this manager stuff in ZB will also make it not usable in general sense. ZB needs a dictionary file (JSON/YAML, similar to CS) that allows you to override custom storage items upon startup.

and you would pass it next to the wasm files here: https://github.com/paritytech/ahm-dryrun/blob/81c9aa9f526e9568bc08f3a65371e26fcff3cee9/zombie-bite-scripts/orchestrator.ts#L61-L71

pepoviola: Hi @kianenigma, we create the `overrides/injects` key/value files [here](https://github.com/pepoviola/zombie-bite/blob/main/src/overrides.rs#L8). In this case we override the _sudo key_ only if is already present.
I remember we add the sudo pallet to use that key to in the startMIgration _tx_ ( can be called with manager key?).

> but be aware that hard-coding this manager stuff in ZB will also make it not usable in general sense. ZB needs a dictionary file (JSON/YAML, similar to CS) that allows you to override custom storage items upon startup.

Yes, this is a very hardcoded version for AHM. The general purpose one will accept a json as input for inject/overrides keys in the _relaychain_ and _paras_.

Thx!

PR #816, opened 2025-07-16
[AHM] Create new response query on resend and tests

muharem:
Create new response query on resend and tests

With this PR, when an XCM message resend is triggered (e.g., if the message failed to be processed on Asset Hub), we generate a new response query instead of reusing the old one. To achieve this, we no longer store the original message appendix but instead attach a new appendix with a fresh query ID upon resending. Additionally, we introduce a new storage item to avoid duplicating the full XCM message for each new response query.


pallets/rc-migrator/src/lib.rs
```diff
@@ -812,16 +823,46 @@ pub mod pallet {
...
+			let new_query_id = pallet_xcm::Pallet::<T>::new_notify_query(
+				asset_hub_location.clone(),
+				<T as Config>::RuntimeCall::from(receive_notification_call),
+				frame_system::Pallet::<T>::block_number() + T::XcmResponseTimeout::get(),
+				Location::here(),
+			);
+
+			let xcm_with_report = {
+				let mut xcm = xcm.clone();
+				xcm.inner_mut().push(SetAppendix(Xcm(vec![ReportTransactStatus(
```
bkontur, line 871: iirc, if you append `SetAppendix` at the end as the last instruction and any of the previous instructions fail, then the appendix won't be triggered, and `ReportTransactStatus` won't send anything to `destination: Location::parent()`, right?

I think that if you also want `SetAppendix` to be triggered in case of an error, you need to append it after `UnpaidExecution` or at least before `Transact`.

But I still don't fully know the complete flow involving sending, resend_xcm, and ReportTransactStatus handling - maybe the order of `SetAppendix` isn't actually an issue.

bkontur: Actually, when the encoded call in `Transact` fails, it doesn't stop or return an error - the next instruction is still executed. So adding `SetAppendix` doesn’t make any difference in this case, you can just append `ReportTransactStatus` directly.

As far as I know, adding `SetAppendix` makes sense before a critical instruction where you want to trigger some XCM regardless of whether it succeeds or fails.

integration-tests/ahm/src/tests.rs
```diff
@@ -832,3 +832,227 @@ fn test_account_references() {
...
+		ump_messages
+	});
+
+	// process the acknowledgement message from AH in the RC.
+	rc.execute_with(|| {
+		rc_now = rc_now + 1;
+		frame_system::Pallet::<RcRuntime>::reset_events();
+		frame_system::Pallet::<RcRuntime>::set_block_number(rc_now);
+
+		enqueue_ump((ump_messages, 0u32.into()));
```
ggwpez, line 949: Can we also check the content of this UMP?

pallets/rc-migrator/src/lib.rs
muharem: It make sure it executed after the rest of the message is processed and also makes sure it will be executed even for some reason the execution of the rest of the messages exited with an error. 
Do you see any problem with using `SetAppendix`?

integration-tests/ahm/src/tests.rs
muharem: asserting the event would make sense

pallets/rc-migrator/src/lib.rs
bkontur: > Do you see any problem with using SetAppendix?

No, not at all. Just to clarify - are we expecting to add any more instructions after `SetAppendix` in the future? Technically, using `SetAppendix` as the last instruction might not be necessary (we usually add it before critical instructions). From what I see in the current [`rc_migrator` config](https://github.com/polkadot-fellows/runtimes/blob/43f3fe308da7a0c914974942eaea05120950d17e/relay/polkadot/src/lib.rs#L1729), it looks like only `SetTopic` gets appended via `send_xcm`. So unless we're planning to add more later, we might be able to skip `SetAppendix(Xcm` here and save a few bytes and decoding effort - but it is just a very ultra nit.

muharem: if its ultra nit (= I would not touch it, just not worth adjusting the code everywhere, running tests again, etc.

integration-tests/ahm/src/tests.rs
muharem: added event assertion

ggwpez (approved):

PR #818, opened 2025-07-17
[AHM] Optional migration manager account id

muharem:
Optional migration manager account id

Adds the possibility to set a specific account id as a manager for the migrators pallets. The manager has the similar to migrators `AdminOrigin` (`ManagerOrigin` before) privileges except that it can not set the manager account id via `set_manager` call. The account id later can be an off-chain multisig set by Governance or Fellowship to manage the migration process.

### DIFF
--- a/pallets/rc-migrator/src/lib.rs
+++ b/pallets/rc-migrator/src/lib.rs
@@ -466,7 +466,7 @@ pub mod pallet {
 		/// The origin that can perform permissioned operations like setting the migration stage.
 		///
 		/// This is generally root, Asset Hub and Fellows origins.
-		type ManagerOrigin: EnsureOrigin<<Self as frame_system::Config>::RuntimeOrigin>;
+		type AdminOrigin: EnsureOrigin<<Self as frame_system::Config>::RuntimeOrigin>;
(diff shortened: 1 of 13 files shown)


pallets/rc-migrator/src/lib.rs
```diff
@@ -866,6 +885,20 @@ pub mod pallet {
...
+		///
+		/// The manager has the similar to [`Config::AdminOrigin`] privileges except that it
+		/// can not set the manager account id via `set_manager` call.
+		#[pallet::call_index(7)]
+		#[pallet::weight(T::RcWeightInfo::set_manager())]
+		pub fn set_manager(origin: OriginFor<T>, new: Option<T::AccountId>) -> DispatchResult {
+			<T as Config>::AdminOrigin::ensure_origin(origin)?;
+			let old = Manager::<T>::get();
+			Manager::<T>::set(new.clone());
+			Self::deposit_event(Event::ManagerSet { old, new });
```
ggwpez, line 899: Could check that it is setting it to something new, but okay.

ggwpez (approved):

PR #819, opened 2025-07-17
[AHM] Type safe Hold and Freeze-Reason translation

muharem:
Make Hold and Reserve reason translation type safe.    
The trick is to create a "portable" format for the Hold and Reserve reason and then translate it on both sides; RC -> Portable -> AH.

### DIFF
--- a/pallets/rc-migrator/src/types.rs
+++ b/pallets/rc-migrator/src/types.rs
@@ -583,7 +583,7 @@ impl<T: Encode> XcmBatchAndMeter<T> {
 	TypeInfo,
 	MaxEncodedLen,
 )]
-pub enum RcHoldReason {
+pub enum PortableHoldReason {
 	Preimage(pallet_preimage::HoldReason),
 	Staking(pallet_staking::HoldReason),
 	StateTrieMigration(pallet_state_trie_migration::HoldReason),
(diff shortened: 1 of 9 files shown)


pallets/rc-migrator/src/types.rs
```diff
@@ -546,6 +571,57 @@ impl<T: Encode> XcmBatchAndMeter<T> {
...
+	DecodeWithMemTracking,
+	Decode,
+	Clone,
+	PartialEq,
+	Eq,
+	RuntimeDebug,
+	TypeInfo,
+	MaxEncodedLen,
+)]
+pub enum PortableHoldReason {
```
re-gius, line 586: Are these only the ones for Polkadot? On Kusama we have some other old reasons like for [pallet-nis](https://github.com/paritytech/polkadot-sdk/blob/13950eeb99e2ab13adfde8a89fe53f3c6cd74296/substrate/frame/nis/src/lib.rs#L469), maybe we can add a TODO for Kusama here

re-gius (approved):
LGTM

pallets/rc-migrator/src/types.rs
ggwpez: My current plan is to only focus on Polkadot and do it 100% correct, ignoring Kusama.  

In this case the compiler would also fail since the Kusama -> Portable translation match is not exhaustive. So I think TODO is not needed :)

PR #824, opened 2025-07-23
check-migrations: Cache & Download snapshot only once per day

bkchr:
Before we always redownloaded the snapshot every time running the job. Now the snapshots are generated 1am every day and then used from there.

### DIFF
--- a/.github/workflows/check-migrations.yml
+++ b/.github/workflows/check-migrations.yml
@@ -90,7 +90,7 @@ jobs:
 
           for i in {1..10}; do
             echo "Snapshot creation attempt $i/10"
-            if ~/.cargo/bin/try-runtime create-snapshot $URI_ARGS -- snapshot.raw; then
+            if try-runtime create-snapshot $URI_ARGS -- snapshot.raw; then
               if [ -f "snapshot.raw" ] && [ -s "snapshot.raw" ]; then
                 echo "Snapshot created successfully"
                 break
@@ -209,7 +209,15 @@ jobs:
           EXTRA_FLAGS: ${{ env.EXTRA_FLAGS }}
           CHECKS: ${{ env.CHECKS }}
         run: |
-          ~/.cargo/bin/try-runtime --version
+          try-runtime --version
+
+          if [ -n "${{ matrix.runtime.build_extra_features }}" ]; then
+            FEATURES="--features try-runtime,${{ matrix.runtime.build_extra_features }}"
+          else
+            FEATURES="--features try-runtime"
+          fi
+          echo "Setting features: ${FEATURES}"
+          cargo build --profile production -p ${{ matrix.runtime.package }} $FEATURES -q --locked
 
           PACKAGE_NAME=${{ matrix.runtime.package }}
           RUNTIME_BLOB_NAME=$(echo $PACKAGE_NAME | sed 's/-/_/g').compact.compressed.wasm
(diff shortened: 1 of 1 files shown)


.github/workflows/check-migrations.yml
```diff
@@ -113,31 +172,45 @@ jobs:
...
       - name: Run ${{ matrix.runtime.name }} Runtime Checks
         #uses: "paritytech/try-runtime-gha@v0.2.0"
         env:
           EXTRA_FLAGS: ${{ env.EXTRA_FLAGS }}
           CHECKS: ${{ env.CHECKS }}
         run: |
-          cargo install -q --git https://github.com/paritytech/try-runtime-cli --tag v0.8.0 --locked && try-runtime --version
-
-          cargo build --profile production -p ${{ matrix.runtime.package }} --features try-runtime -q --locked
+          try-runtime --version
```
bkontur, line 212: Isn't `try-runtime` installed only for `generate-snapshots`?
Don't we need to pass it also via "Download Snapshot" or something?

.github/workflows/check-migrations.yml
```diff
@@ -113,31 +172,45 @@ jobs:
...
+          if [ ! -f "snapshot.raw" ] || [ ! -s "snapshot.raw" ]; then
+            echo "No valid snapshot found. This job requires pre-generated snapshots from the nightly job."
+            echo "Please wait for the nightly snapshot generation or run it manually via workflow_dispatch."
+            exit 1
+          else
+            echo "Using snapshot from $(date -r snapshot.raw)"
+          fi
+
       - name: Run ${{ matrix.runtime.name }} Runtime Checks
         #uses: "paritytech/try-runtime-gha@v0.2.0"
```
bkontur:
```suggestion
```

.github/workflows/check-migrations.yml
```diff
@@ -19,6 +21,62 @@ concurrency:
...
+              exit 1
+            fi
+            sleep 10
+          done
+
+      - name: Cache snapshot
+        uses: actions/cache@v4
+        with:
+          path: snapshot.raw
+          key: try-runtime-snapshot-${{ matrix.runtime.name }}
```
ggwpez: In the SDK we use the current day as the cache key `try-runtime-snapshot-${{ matrix.network }}-${{ steps.get-date.outputs.today }}` and if that keys does not exist, then it creates a new snap.  
This prevents us downloading snapshots on days where we dont need them. But I guess otherwise its identical.

.github/workflows/check-migrations.yml
```diff
@@ -19,6 +21,62 @@ concurrency:
...
+        uses: dtolnay/rust-toolchain@master
+        with:
+          targets: "wasm32-unknown-unknown"
+          components: "rust-src"
+          toolchain: "${{env.RUST_STABLE_VERSION}}"
+
+      - name: Generate snapshot for ${{ matrix.runtime.name }}
+        run: |
+          echo "Generating snapshot for ${{ matrix.runtime.name }}..."
+          cargo install -q --git https://github.com/paritytech/try-runtime-cli --tag v0.8.0 --locked
```
ggwpez: There is a binary, I assume building it is a bit slow: https://github.com/paritytech/polkadot-sdk/blob/1756db06b1d7794a23b7d069d76716a64d0524cc/.github/workflows/check-runtime-migration.yml#L86-L89

.github/workflows/check-migrations.yml
bkchr: Yeah I know. I decided against this. Thought that this way we always have it available. I thought it would be better to improve the days when we need it :D Good question to think about on what we should optimize for :D

bkontur: looks like the day stuff fixes what AI review complains about: https://github.com/polkadot-fellows/runtimes/pull/1058

sigurpol:
@bkchr can you re-run the failed migration job (see https://github.com/polkadot-fellows/runtimes/issues/1057#issuecomment-3784526869)  :bow: ?

eskimor (approved):

PR #827, opened 2025-07-28
[AHM] Tests work for Paseo

muharem:
Tests work for Paseo


integration-tests/ahm/src/tests.rs
```diff
@@ -109,6 +110,16 @@ pub type RcPolkadotChecks = (
 	ProxyWhaleWatching,
 );
 
+// Checks that are specific to Paseo.
+#[cfg(feature = "paseo")]
+pub type RcRuntimeSpecificChecks = (
+	pallet_rc_migrator::multisig::MultisigMigrationChecker<Polkadot>,
+	pallet_rc_migrator::bounties::BountiesMigrator<Polkadot>,
+	pallet_rc_migrator::treasury::TreasuryMigrator<Polkadot>,
```
ggwpez, line 118: Instead of duplicating them, can you maybe put them into `RcPaseoChecks` and always include that?  
Otherwise we can forget to add a check twice

PR #829, opened 2025-07-29
[AHM] Account Translation mapping

ggwpez:
Change:
- Add test that auto-generates Rust file with account translation maps
- Use account translation maps to translate the accounts
- Fix child bounties pallet to also use account translation
- Fix some tests to make them work with Paseo snapshot
- Run Paseo snapshot tests in CI
- Add account translation convenience functions to several types
- Emit events upon account translation
- Add `SortByEncoded` trait to sort un-sortable collections easily (testing only)


pallets/ah-migrator/src/sovereign_account_translation.rs
```diff
@@ -0,0 +1,267 @@
+// Manuall stuff
+use hex_literal::hex;
+use sp_runtime::AccountId32;
+
+// AUTOGENERATED BELOW
+// RC snap path: ../../paseo.snap
+/// List of RC para to AH sibl sovereign account translation sorted by RC account.
+pub const SOV_TRANSLATIONS: &[(AccountId32, AccountId32)] = &[
```
ggwpez: @seadanda we maybe want to re-generate this for Paseo, although i might just merge all networks into one (Paseo, Polkadot & Kusama).  
Just for now its simpler like this to keep it per network.

pallets/ah-migrator/src/sovereign_account_translation.rs
```diff
@@ -0,0 +1,131 @@
+// Manuall stuff
+use hex_literal::hex;
+use sp_runtime::AccountId32;
+
+// AUTOGENERATED BELOW
+// RC snap path: ../../polkadot.snap
```
ggwpez, line 23: `SKIP_WASM_BUILD=1 SNAP_RC="../../polkadot.snap" RUST_LOG="debug" ct polkadot-integration-tests-ahm find_translatable_accounts -r -- --nocapture  --ignored` for posterity.

integration-tests/ahm/src/accounts_translation_works.rs
```diff
@@ -0,0 +1,79 @@
...
+
+type RelayRuntime = polkadot_runtime::Runtime;
+type AssetHubRuntime = asset_hub_polkadot_runtime::Runtime;
+
+pub struct AccountTranslationWorks;
+
+pub const TRANSLATIONS: &[(AccountId32, AccountId32)] = &[
+	// para 2034: 5Ec4AhPbkXX97KXMcf9v9SkRNG4Gyc3VhcMMuQe9QXfAHnrC ->
+	// 5Eg2fntQqFi3EvFWAf71G66Ecjjah26bmFzoANAeHFgj9Lia
+	(
```
ggwpez, line 33: I will add some more here; this is just for sanity checking that we dont test `f(x) = f(x)` again.

pallets/ah-migrator/src/account_translation.rs
```diff
@@ -0,0 +1,89 @@
...
+			.or_else(|| Self::maybe_derived_translate(&account))
+		else {
+			return account;
+		};
+
+		log::debug!(
+			"Translated account: {} -> {}",
+			&account.to_polkadot_ss58(),
+			&new.to_polkadot_ss58()
+		);
```
muharem: I think can be info

pallets/ah-migrator/src/account_translation.rs
```diff
@@ -0,0 +1,89 @@
...
+					.defensive()
+			})
+			.ok()
+			.flatten()
+			.cloned()
+		else {
+			return None;
+		};
+
+		Self::deposit_event(crate::Event::AccountTranslatedParachainSovereign {
```
muharem, line 59: 👍

integration-tests/ahm/src/accounts_translation_works.rs
muharem: you can also add some on pre_check

pallets/ah-migrator/src/sovereign_account_translation.rs
```diff
@@ -0,0 +1,390 @@
...
+	(
+		AccountId32::new(hex!("7061726100000000000000000000000000000000000000000000000000000000")),
+		AccountId32::new(hex!("7369626c00000000000000000000000000000000000000000000000000000000")),
+	),
+	// para 2048: 5Ec4AhNmG8sxwXqncL8VfyMnPftV1mJcwwpdokxfDhZ91zK2 ->
+	// 5Eg2fnsaLs4s58ZwAL5anchbe9ZnjBMj1bU54iVA6Rahsb2S
+	(
+		AccountId32::new(hex!("7061726100080000000000000000000000000000000000000000000000000000")),
+		AccountId32::new(hex!("7369626c00080000000000000000000000000000000000000000000000000000")),
+	),
```
muharem: we can also have a test that asserts hex to ss58. in std we can convert AccountId32 to ss58. the ss58 can be just third item here. that would be simpler to verify for reviewers

integration-tests/ahm/src/accounts_translation_works.rs
ggwpez: I added an RC pre check that the account exists. But for AH pre it is different since sometimes people already accidentally sent funds to it so the account exists.

pallets/ah-migrator/src/sovereign_account_translation.rs
ggwpez: I put a test now

muharem (approved):

integration-tests/ahm/src/tests.rs
```diff
@@ -297,54 +301,165 @@ fn sovereign_account_translation() {
...
+			)
+			.unwrap();
 
-	rc.execute_with(|| {
-		for para_id in paras_registrar::Paras::<Polkadot>::iter_keys().collect::<Vec<_>>() {
-			let rc_acc =
-				xcm_builder::ChildParachainConvertsVia::<ParaId, AccountId32>::convert_location(
-					&Location::new(0, Junction::Parachain(para_id.into())),
-				)
+		let (ah_para_sibl, found_para_id) =
```
muharem, line 334: can be also sanity checked with `SiblingParachainConvertsVia<Sibling, AccountId>` result

ggwpez: Hm this seems to produce the wrong account ID actually, it is the same as the child one, see: https://github.com/paritytech/polkadot-sdk/blob/1dbaeba9044ef491847a8204da98cd9e74534cdb/polkadot/xcm/xcm-builder/src/location_conversion.rs#L302-L314 

Will ask in XCM chat, but for now it seems wrong.

PR #833, opened 2025-08-01
Run try-state check for all pallets post AHM rust tests

kianenigma:


integration-tests/ahm/src/tests.rs
```diff
@@ -173,6 +175,41 @@ pub type AhRuntimeSpecificChecks = (
 	pallet_rc_migrator::crowdloan::CrowdloanMigrator<AssetHub>,
 );
 
+pub struct PalletsTryStateCheck;
```
ggwpez: There is an adjacent check.rs file

PS: I am just adding the `try_decode_entire_state` since we are currently missing it.

integration-tests/ahm/src/tests.rs
```diff
@@ -173,6 +175,43 @@ pub type AhRuntimeSpecificChecks = (
...
+	}
+
+	#[cfg(not(feature = "try-runtime"))]
+	fn post_check(_: Self::RcPrePayload, _: Self::AhPrePayload) {
+		// nada
+	}
+	#[cfg(feature = "try-runtime")]
+	fn post_check(_: Self::RcPrePayload, _: Self::AhPrePayload) {
+		use frame_support::traits::TryState;
+		let res = asset_hub_polkadot_runtime::AllPalletsWithSystem::try_state(
```
muharem: does it make sense to run it for RC to?

kianenigma: good point, will add.

kianenigma:
We need to remove pallet-fast-unstake from PAH runtime. @ggwpez this CI check must have already been broken before I would imagine, no? 

LMK if you want help with removing the pallet.

PR #835, opened 2025-08-04
[AHM] Record migration timings

muharem:
Record migration timings.

Closes: https://github.com/paritytech/polkadot-sdk/issues/9292

This PR records the start and end times of the migration in dedicated storage items. The start is defined as the moment when the extrinsics of the migrating pallets are locked; the end is when those locks are released.

The PR also updates the trigger for the `AssetHubMigrationStarted` event - it is now emitted upon transitioning to the `WaitingForAh` state instead of `Starting`. This change ensures the event accurately reflects the point at which the extrinsics become locked.

### DIFF
--- a/integration-tests/ahm/src/tests.rs
+++ b/integration-tests/ahm/src/tests.rs
@@ -53,11 +53,13 @@ use frame_system::pallet_prelude::BlockNumberFor;
 use pallet_ah_migrator::{
 	sovereign_account_translation::{DERIVED_TRANSLATIONS, SOV_TRANSLATIONS},
 	types::AhMigrationCheck,
-	AhMigrationStage as AhMigrationStageStorage, MigrationStage as AhMigrationStage,
+	AhMigrationStage as AhMigrationStageStorage, MigrationEndBlock as AhMigrationEndBlock,
+	MigrationStage as AhMigrationStage, MigrationStartBlock as AhMigrationStartBlock,
 };
 use pallet_rc_migrator::{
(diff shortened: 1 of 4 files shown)


integration-tests/ahm/src/tests.rs
```diff
@@ -653,11 +665,24 @@ async fn migration_works_time() {
...
+	let new_ah_block_end = ah.execute_with(frame_system::Pallet::<AssetHub>::block_number);
+
+	rc.execute_with(|| {
+		assert_eq!(RcMigrationStartBlock::<Polkadot>::get(), Some(rc_block_start + 1));
+		assert_eq!(RcMigrationEndBlock::<Polkadot>::get(), Some(new_rc_block_end));
+	});
+
+	ah.execute_with(|| {
+		assert_eq!(AhMigrationStartBlock::<AssetHub>::get(), Some(ah_block_start + 1));
+		assert_eq!(AhMigrationEndBlock::<AssetHub>::get(), Some(new_ah_block_end));
```
ggwpez, line 684: Not sure how to handle the idempotent case; but I guess there is no easy answer here.

muharem: yes. but if we re run migration, most probably we wont start it from the beginning. we better do not unlock anything and start from some of the stages that wont pass `WaitingForAh`. in this case we only gonna later set the end of the migration which will be correct.

ggwpez: Ah yea, i forgot that we wanted to have this cooloff stage where we can rerun some stages. 👍

TarikGul:
The structure looks really great. Thanks @muharem

PR #838, opened 2025-08-05
[AHM] Fix migration tests for pallet-preimage

re-gius:
During migration, some preimages may be unrequested and therefore deleted as a side effect. This PR fixes preimage tests to deal with preimage changes performed during migration.

There is also a fix to the method mapping RC calls to AH calls: we now unrequest preimages migrated to Asset Hub if they were unrequested on the Relay Chain. Without this fix, unrequested preimages from RC became requested on AH.

### DIFF
--- a/pallets/ah-migrator/src/call.rs
+++ b/pallets/ah-migrator/src/call.rs
@@ -21,6 +21,10 @@ pub type BoundedCallOf<T> =
 	Bounded<<T as frame_system::Config>::RuntimeCall, <T as frame_system::Config>::Hashing>;
 
 impl<T: Config> Pallet<T> {
+	// Maps a RC BoundedCall to an AH BoundedCall.
+	// The preimage from the RC BoundedCall is unrequested on the relay chain, then re-created on
+	// Asset Hub with translated parameters. Notice that the unrequest operation may delete the
+	// preimage hash from storage.
 	pub fn map_rc_ah_call(
 		rc_bounded_call: &BoundedCallOf<T>,
 	) -> Result<BoundedCallOf<T>, Error<T>> {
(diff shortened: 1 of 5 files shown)


pallets/ah-migrator/src/call.rs
```diff
@@ -21,6 +21,9 @@ pub type BoundedCallOf<T> =
 	Bounded<<T as frame_system::Config>::RuntimeCall, <T as frame_system::Config>::Hashing>;
 
 impl<T: Config> Pallet<T> {
+	// Maps a RC BoundedCall to an AH BoundedCall.
+	// This function also unrequests the preimage if it was requested. As a side effect, it may
+	// delete the preimage hash from storage.
```
muharem: this comment does not make it clear? we unrequest if it was requested? this is not informative.

pallets/ah-migrator/src/call.rs
```diff
@@ -49,6 +56,13 @@ impl<T: Config> Pallet<T> {
 			Error::<T>::FailedToBoundCall
 		})?;
 
+		if let Some(hash) = ah_bounded_call.lookup_hash() {
+			if unrequested_preimage {
+				log::debug!(target: LOG_TARGET, "Unrequesting preimage with hash {:?} which was unrequested on RC", hash);
+				T::Preimage::unrequest(&hash);
```
muharem: did you check what this will result? we just bound/saved the preimage for the call above, and here we wanna remove it?

pallets/ah-migrator/src/call.rs
re-gius: Right, improved in https://github.com/polkadot-fellows/runtimes/pull/838/commits/2ee804b138e685add4ca0dcd2a7dc4a036836e92

pallets/ah-migrator/src/call.rs
re-gius: This doesn't work, I reverted the change and just accepted that requested preimages on the RC coming from scheduled calls may become requested on AH. With the current Polkadot snapshot it only happens for one scheduled call.
In general, since scheduled calls are mapped to different scheduled calls on AH and preimage hashes may differ, it's not correct to assert that all preimage hashes should be migrated correctly to AH. I excluded scheduled calls from preimage checks in https://github.com/polkadot-fellows/runtimes/pull/838/commits/2ee804b138e685add4ca0dcd2a7dc4a036836e92.

PR #841, opened 2025-08-06
[AHM] Post migration cool off stage

muharem:
Post migration cool off stage

Adds the Cool-Off stage at the end of the migration, with a configurable duration. This stage provides time to verify that the migration was executed correctly before unlocking the extrinsics of the migrated pallets.

### DIFF
--- a/pallets/rc-migrator/src/lib.rs
+++ b/pallets/rc-migrator/src/lib.rs
@@ -197,8 +197,8 @@ pub enum MigrationStage<
 		///
 		/// The block number at which we notify the Asset Hub about the start of the migration and
 		/// move to `WaitingForAH` stage. After we receive the confirmation, the Relay Chain will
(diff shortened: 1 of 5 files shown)


pallets/rc-migrator/src/lib.rs
```diff
@@ -197,27 +197,17 @@ pub enum MigrationStage<
...
-	WaitingForAh {
-		/// The block number at which the cool-off period will end.
-		///
-		/// After the cool-off period ends, the Relay Chain will start to send the migration data
-		/// to the Asset Hub.
-		cool_off_end: BlockNumber,
-	},
-	CoolOff {
+	WaitingForAh,
+	PreMigrationCoolOff {
```
ggwpez: This is more like a `Warmup`?

muharem: True!

PR #844, opened 2025-08-07
Moves single block migrations from frame_executive to frame_system config

RomarQ:
Based on PR https://github.com/paritytech/polkadot-sdk/pull/1781 and [PRDoc](https://github.com/paritytech/polkadot-sdk/blob/beb9030b249cc078b3955232074a8495e7e0302a/prdoc/1.9.0/pr_1781.prdoc#L29), the new way for providing the single block migrations should be through `SingleBlockMigrations` in `frame_system::Config`. Providing them from `frame_executive::Executive` is still supported, but from what I understood is or will be deprecated.

> `SingleBlockMigrations` this is the new way of configuring migrations that run in a single block. Previously they were defined as last generic argument of Executive. This shift is brings all central configuration about migrations closer into view of the developer (migrations that are configured in Executive will still work for now but is deprecated).

Will also open a PR on polkadot-sdk side to add a deprecation warning.

### DIFF
--- a/relay/kusama/src/lib.rs
+++ b/relay/kusama/src/lib.rs
@@ -239,7 +239,7 @@ impl frame_system::Config for Runtime {
 	type SS58Prefix = SS58Prefix;
 	type OnSetCode = ();
 	type MaxConsumers = frame_support::traits::ConstU32<16>;
-	type SingleBlockMigrations = ();
+	type SingleBlockMigrations = SingleBlockMigrations;
 	type MultiBlockMigrator = ();
 	type PreInherents = ();
 	type PostInherents = ();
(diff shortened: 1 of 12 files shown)


RomarQ:
> It looks like some pallets are actually missing migrations. Can you please add them so that the check will be green?
> 
> Thanks for the MR!

Yes, it requires a fix on the polkadot-sdk side, will open a PR there later today or tomorrow:

image

RomarQ:
Should be fixed once we merge https://github.com/paritytech/polkadot-sdk/pull/9451 and the changed are backported. Will change the PR status to DRAFT for now.

bkchr (approved):

PR #847, opened 2025-08-11
[AHM] Filter staking calls before migration starts

ggwpez:
We currently only have call filter for the stages during and after them migration - not before. This MR adds a third entry to the filter tuple that can check in the stage before the migration starts.

Staking is disabled in this stage and a few other pallets. Indices pallet was also forbidden on AH now to avoid insert conflicts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1106,7 +1106,7 @@ dependencies = [
 [[package]]
 name = "asset-test-utils"
 version = "23.0.0"
-source = "git+https://github.com/paritytech/polkadot-sdk?branch=oty-2503-paseo#8d13ffdf16077a5c8ab7c65feb0b714b5fe1e252"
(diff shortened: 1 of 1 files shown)


system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/call_filter.rs
```diff
@@ -120,5 +137,67 @@ pub fn call_allowed_status(call: &<Runtime as frame_system::Config>::RuntimeCall
...
+pub fn call_allowed_before_migration(
+	call: &<Runtime as frame_system::Config>::RuntimeCall,
+) -> bool {
+	use RuntimeCall::*;
+	const ON: bool = true;
+	const OFF: bool = false;
+
+	match call {
+		// Disabled to avoid state insert conflicts.
+		Staking(..) => OFF,
```
ggwpez, line 153: @Ank4n disabling Staking now on AH before the migration starts.

muharem (approved):

PR #851, opened 2025-08-13
[AHM] Send xcm extrinsic within migrator pallets

muharem:
Send xcm extrinsic within migrator pallets

During migration, the pallet xcm send extrinsic's router will filter all messages from the RC to the AH and vice versa, in order to prevent users from spamming the message queues used for migration.
In this PR, we add a send extrinsic to the migrator pallets that uses a different router without filtering. This will allow the migration manager or governance to send xcm messages over those channels during migration.


pallets/ah-migrator/src/lib.rs
```diff
@@ -1001,6 +1013,60 @@ pub mod pallet {
...
+
+			let origin_location = <T as pallet_xcm::Config>::SendXcmOrigin::ensure_origin(origin)?;
+			let interior: Junctions =
+				origin_location.clone().try_into().map_err(|_| Error::<T>::InvalidOrigin)?;
+			let dest = Location::try_from(*dest).map_err(|()| Error::<T>::BadXcmVersion)?;
+			let mut message: Xcm<()> =
+				(*message).try_into().map_err(|()| Error::<T>::BadXcmVersion)?;
+
+			if interior != Junctions::Here {
+				message.0.insert(0, DescendOrigin(interior.clone()));
```
ggwpez, line 1037: Why is this needed?

ggwpez (approved):

pallets/ah-migrator/src/lib.rs
muharem: otherwise regardless the calling origin/location a message gonna be sent as from the root

pallets/ah-migrator/src/lib.rs
```diff
@@ -1001,6 +1013,60 @@ pub mod pallet {
...
+				)?;
+			// send
+			let message_id = <T as Config>::SendXcm::deliver(ticket).map_err(|error| {
+				log::error!(
+					target: LOG_TARGET,
+					"XCM send failed with error: {:?}; destination: {:?}; message: {:?}",
+					error, dest, message
+				);
+				Error::<T>::XcmError
+			})?;
```
bkontur, line 1060: Just a nit, it could be simplified with just `xcm::latest::send_xcm` instead of `validate_send + SendXcm::deliver`, but ok, if you just want to have two different logs `XCM validation` vs `XCM send` (not sure if worth in this case)

```suggestion
			let message_id =
				send_xcm::<<T as Config>::SendXcm>(dest.clone(), message.clone()).map_err(
					|error| {
						log::error!(
							target: LOG_TARGET,
							"XCM send failed with error: {:?}; destination: {:?}; message: {:?}",
							error, dest, message
						);
						Error::<T>::XcmError
					},
				)?;
```

muharem: I just wanna have it exactly as pallet xcm send, as I mentioned in the doc

PR #852, opened 2025-08-13
Improve Era Scheduling

Ank4n:
Original issue described here: https://hackmd.io/@ak0n/ahm-schedule

We already have warm up period so we can reuse it.

Needs some ahm test runs before merging. We can also do this after Paseo.

## What
Block staking election and then wait at least 1 session (warm up) before starting the migration.

Q: Should I add a hard constraint, or is documentation enough (imo this is fine) to ensure it’s set to more than one session? 

## Why
Once election for next era is complete, the validator set is queued in `pallet-session`. If we at this point freeze `pallet-staking` for migration, we lose some crucial validator set information to correctly pay era payout rewards (and risk minor data corruption).

### DIFF
--- a/integration-tests/ahm/src/tests.rs
+++ b/integration-tests/ahm/src/tests.rs
@@ -759,7 +759,7 @@ async fn scheduled_migration_works() {
 			DispatchTime::At(start),
 			DispatchTime::At(warm_up_end),
 			cool_off_end,
-			Default::default(),
+			true, // Ignore the staking era check
 		));
 		assert_eq!(
 			RcMigrationStageStorage::<Polkadot>::get(),
--- a/pallets/rc-migrator/src/benchmarking.rs
+++ b/pallets/rc-migrator/src/benchmarking.rs
@@ -79,7 +79,7 @@ pub mod benchmarks {
 		let cool_off = DispatchTime::<BlockNumberFor<T>>::After(20u32.into());
 
 		#[extrinsic_call]
-		_(RawOrigin::Root, start, warm_up, cool_off);
+		_(RawOrigin::Root, start, warm_up, cool_off, true);
 
 		assert_last_event::<T>(
 			Event::StageTransition {


pallets/rc-migrator/src/lib.rs
```diff
@@ -808,14 +808,20 @@ pub mod pallet {
...
 		///   `DispatchTime` calculated at the moment of the transition to the warm-up stage.
 		/// - `cool_off`: The block number at which the post migration cool-off period will end. The
 		///   `DispatchTime` calculated at the moment of the transition to the cool-off stage.
 		///
+		/// Note: If the staking election for next era is already complete, and the next
+		/// validator set is queued in `pallet-session`, we want to avoid starting the data migration
+		/// at this point as it can lead to some missed validator rewards. To address this, we
+		/// stop staking election at the start of migration and must wait atleast 1 session (set via
+		/// warm_up) before starting the data migration.
+		///
```
ggwpez, line 835: >Q: Should I add a hard constraint, or is documentation enough (imo this is fine) to ensure it’s set to more than one session?

Given that the session duration is different on Paseo, Kusama and Polkadot; i think it can be helpful to have a check in here that checks that the warmup duration is at least one (or two) session.  
And then a boolean argument like `ignore_warmup_length_check` to skip it in testing.

Ank4n: I could also make the `warm_up` arg a u32 (duration in sessions), where 0 means no warm-up (for tests). For Paseo, Kusama, and Polkadot, we would use the same value. Wdyt?

ggwpez: Yea would also work, no strong pref. cc @muharem

muharem: Polkadot session is 4 hours, right? is not it too long for warm up?

ggwpez: Will change it now to do the ForcingNone in the scheduled stage to avoid locking all extrinsics for 4-8 hrs.

Ank4n: Yeah, we should not block pallet extrinsics for 4-8 hour. The aim is only to block staking election for 4-8 hours and everything as usual. 

So does it mean we should add another stage? Like a prep window?

Ank4n:
LGTM, thanks @ggwpez for finishing it.

Ank4n:
@kianenigma Would be great to get your review as well.

kianenigma:
with delay: LGTM!

PR #854, opened 2025-08-15
[AHM] Migrate Pallet Recovery

ggwpez:
Changes:
- Migrate pallet-recovery

Currently untested since we need to integrate AHM into Kusama first. Same for weights...

### DIFF
--- a/pallets/ah-migrator/src/lib.rs
+++ b/pallets/ah-migrator/src/lib.rs
@@ -332,6 +332,11 @@ pub mod pallet {
 			Self::RcPalletsOrigin,
 			<<Self as frame_system::Config>::RuntimeOrigin as OriginTrait>::PalletsOrigin,
 		>;
+		
+		/// Config for Kusama pallets.
+		#[cfg(feature = "kusama")]
+		type KusamaConfig: pallet_recovery::Config;
+
 		/// Preimage registry.
 		type Preimage: QueryPreimage<H = <Self as frame_system::Config>::Hashing> + StorePreimage;
 		/// Convert a Relay Chain Call to a local AH one.
--- a/pallets/ah-migrator/src/lib.rs
+++ b/pallets/ah-migrator/src/lib.rs
@@ -55,6 +55,8 @@ pub mod types;
 pub mod vesting;
 pub mod xcm_config;
 pub mod xcm_translation;
+#[cfg(feature = "kusama")]
+pub mod recovery;
 
 pub use pallet::*;
 pub use pallet_rc_migrator::{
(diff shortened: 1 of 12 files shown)


integration-tests/ahm/src/recovery_test.rs
```diff
@@ -0,0 +1,94 @@
...
+impl RcMigrationCheck for RecoveryDataMigrated {
+	type RcPrePayload = TestData;
+
+	fn pre_check() -> Self::RcPrePayload {
+		TestData {
+			recoverable: pallet_recovery::Recoverable::<RcRuntime>::iter()
+				.map(|(who, config)| (who, config.into_portable()))
+				.collect(),
+			active_recoveries: pallet_recovery::ActiveRecoveries::<RcRuntime>::iter()
+				.map(|(w1, w2, config)| (w1, w2, config.into_portable()))
```
muharem, line 44: It might be better to not use portable but original type to make sure some bug are not inherited from the portable and into mappings. but not a blocker

pallets/ah-migrator/src/benchmarking.rs
```diff
@@ -935,6 +935,40 @@ pub mod benchmarks {
 		);
 	}
 
+	/*#[benchmark]
```
muharem: should we remove?

pallets/ah-migrator/src/lib.rs
```diff
@@ -961,7 +963,7 @@ pub mod pallet {
 
 		#[cfg(feature = "kusama-ahm")]
 		#[pallet::call_index(26)]
-		#[pallet::weight(T::AhWeightInfo::receive_staking_messages(messages.len() as u32))] // TODO @ggwpez weight
+		#[pallet::weight(T::AhWeightInfo::receive_recovery_messages(messages.len() as u32))]
```
muharem: should we use estimated?

pallets/rc-migrator/src/recovery.rs
```diff
@@ -257,9 +257,8 @@ impl<T: Config> PalletMigration for RecoveryMigrator<T> {
 				}
 			}
 
-			// TODO @ggwpez weight
 			if T::MaxAhWeight::get()
-				.any_lt(T::AhWeightInfo::receive_vesting_schedules((messages.len() + 1) as u32))
+				.any_lt(T::AhWeightInfo::receive_recovery_messages((messages.len() + 1) as u32))
```
muharem: should we use estimated?

pallets/ah-migrator/src/lib.rs
ggwpez: yea done

integration-tests/ahm/src/recovery_test.rs
ggwpez: Generally agreeing, but not sure how much effort is justified here. I will check how much it is.

muharem: not a blocker again

PR #858, opened 2025-08-20
add action to clean disk space before checkout

pepoviola:

CI: Add a step to clean disk space before checkout.

cc: @bkontur


bkontur:
@pepoviola what about those errors? Do we need to clean the docker also?
```
0 upgraded, 0 newly installed, 0 to remove and 1 not upgraded.
/_work/fellowship-001/_temp/6f861543-40dc-4aaf-b69b-24463c72fc55.sh: line 131: 5.18671e+09: syntax error: invalid arithmetic operator (error token is ".18671e+09")
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
/_work/fellowship-001/_temp/6f861543-40dc-4aaf-b69b-24463c72fc55.sh: line 141: 5.18671e+09: syntax error: invalid arithmetic operator (error token is ".18671e+09")
/_work/fellowship-001/_temp/6f861543-40dc-4aaf-b69b-24463c72fc55.sh: line 163: free: command not found
```

bkontur (commented):
I think it’s okay, but my only concern is that `jlumbroso/free-disk-space` is quite old (last release 10/2023) and seems unmaintained. Could we use a newer action instead, for example: [https://github.com/marketplace/actions/disk-space-reclaimer](https://github.com/marketplace/actions/disk-space-reclaimer)?

Also we use `actions/checkout` for other pipelines (check-migrations.yml, clippy.yml, fmt.yml, ..), shouldn't we also change those?

pepoviola:
> > Hi @bkontur, thanks for review it. [disk-space-reclaimer](https://github.com/insightsengineering/disk-space-reclaimer/releases) last release is also from 2023. We use `jlumbroso/free-disk-space` in the past (and keep using) in `zombienet-sdk` and also in `polkadot-sdk` repos. Also, I only included this action when we use the gh-hosted runners since our runners doesn't have the space issue.
> > Thx!
> 
> @pepoviola Ok, ok, and what about other jobs: (check-migrations.yml, clippy.yml, fmt.yml, ..)?, Should we also add it there too?

Hi @bkontur, I don't think is needed in fmt.yml but we can add it in clippy and check-migrations. Do we had failures there?
Thx!

PR #860, opened 2025-08-21
Added GetParachainInfo runtime api

bkontur:
Not sure if this is really needed now, please, give here more context @iulianbarbu

### DIFF
--- a/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
+++ b/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
@@ -2024,6 +2024,12 @@ pallet_revive::impl_runtime_apis_plus_revive!(
 		}
 	}
 
+	impl cumulus_primitives_core::GetParachainInfo<Block> for Runtime {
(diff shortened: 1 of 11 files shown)


iulianbarbu:
The API should be added to simplify chain specs (by removing the para_id field from there). Otherwise, as long as we're using stable2506-X node to run the runtimes (not master or unstable2507 yet), we should be fine with or without these APIs. 

All in all it doesn't hurt to merge them. stable2509 nodes will be able to use the new runtime API, so chain specs can omit the para_id at that time, if nodes are started from a state where runtime already contains the `GetParachainInfo` API (e.g. after the runtime upgrade).

I will backport [paritytech/polkadot-sdk/#9201](https://github.com/paritytech/polkadot-sdk/pull/9201) to polkadot-sdk/unstable2507 to have a node that supports runtimes without `GetParachainInfo` (since unstable2507 released a not fully functional node as it seems, and also the runtime API in cumulus_primitives_core), but I also need to test it a bit before that.

PR #862, opened 2025-08-24
[AHM] Integrate migration for Kusama

muharem:
Integrate migration for Kusama

The PR introduces migrating pallets to the Kusama Asset Hub runtime. Additionally adjusting and integrating the migrator pallets for the Kusama Relay and Asset Hub chains.

PR #863, opened 2025-08-25
Update Kusama submission deposit as per ref#558

olanod:
A recent [WFC](https://kusama.subsquare.io/referenda/558) passed to adjust submission deposit to prevent spams specially on the WFC track. 

A missing part of the WFC is to add a new extrinsic to the PolkadotSDK to "collect submission deposits", a permissionless extrinsic that would enable transferring the reserved deposit of failed or timed out referenda to a well-known account as the deposits are blocked forever in the user's account, kind of like a limbo state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+### Changed
+
+- Update Kusama submission deposit to 3.333KSM as per ref#558
+
 ## [1.7.0] 22.08.2025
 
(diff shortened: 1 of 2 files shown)


bkontur:
@olanod When do you plan/want to release this on-chain live? Because of upcoming AHM and governance migration to the AssetHub, I suggest to adjust this value also for migrated OpenGov on AssetHub here: https://github.com/polkadot-fellows/runtimes/pull/862

cc: @muharem

olanod:
> @olanod When do you plan/want to release this on-chain live? Because of upcoming AHM and governance migration to the AssetHub, I suggest to adjust this value also for migrated OpenGov on AssetHub here: #862
> 
> cc: @muharem

I don't think we are in a rush so it makes sense for the change to go directly into the migration

PR #872, opened 2025-09-01
[AHM] Fix derived account translation script

ggwpez:
Add missing parachain sovereign and derived accounts to translation map.  
Will need to merge with Kusama and re-run before the migration. Maybe also add a permissionless extrinsic to add more if someone created a new account right before the migration.


integration-tests/ahm/src/tests.rs
```diff
@@ -338,7 +339,7 @@ async fn find_translatable_accounts() {
 	// Para ID -> (RC sovereign, AH sovereign)
 	let mut sov_translations = BTreeMap::<u32, (AccountId32, AccountId32)>::new();
 	// Para ID -> (RC derived, index, AH derived)
-	let mut derived_translations = BTreeMap::<u32, (AccountId32, u16, AccountId32)>::new();
+	let mut derived_translations = Vec::<(ParaId, AccountId32, u16, AccountId32)>::new();
```
ggwpez, line 342: Bug was here

PR #874, opened 2025-09-02
Add migration for updating Kusama hrmpChannelMaxCapacity

serban300:
Equivalent of Polkadot referenda https://polkadot.subsquare.io/referenda/1740


bkchr:
As for Polkadot, this should be a normal referendum. Please prepare it and I can post it on chain for you.

serban300:
> As for Polkadot, this should be a normal referendum. Please prepare it and I can post it on chain for you.

Ok, thanks ! Will do

PR #875, opened 2025-09-02
[AHM] Society pallet migration

muharem:
Society pallet migration


pallets/ah-migrator/src/lib.rs
```diff
@@ -971,6 +984,18 @@ pub mod pallet {
 			Self::do_receive_recovery_messages(messages).map_err(Into::into)
 		}
 
+		#[cfg(feature = "kusama-ahm")]
+		#[pallet::call_index(27)]
+		#[pallet::weight(T::AhWeightInfo::receive_staking_messages(messages.len() as u32))] // TODO @muharem weight
```
ggwpez: It was a bit difficult to get it working with a feature-gate for the recovery pallet, maybe we just estimate it here.

muharem: its a good idea, lets have it just estimated

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1327,6 +1327,47 @@ impl pallet_recovery::Config for Runtime {
...
 
+parameter_types! {
+	pub const SocietyPalletId: PalletId = PalletId(*b"py/socie");
+}
+
+// TODO @muharem
+pub struct MockRandomness;
+impl Randomness<Hash, BlockNumber> for MockRandomness {
+	fn random(_subject: &[u8]) -> (Hash, BlockNumber) {
+		(Hash::zero(), 0u32)
```
ggwpez: TODO

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1327,6 +1327,47 @@ impl pallet_recovery::Config for Runtime {
...
+	type Currency = Balances;
+	type Randomness = system_parachains_common::randomness::RelayChainOneEpochAgoWithoutBlockNumber<
+		Runtime,
+		cumulus_primitives_core::relay_chain::BlockNumber,
+	>;
+	type GraceStrikes = ConstU32<10>;
+	type PeriodSpend = ConstU128<{ 500 * QUID }>;
+	type VotingPeriod = pallet_ah_migrator::LeftOrRight<
+		AhMigrator,
+		ConstU32<{ u32::MAX - 2 * RC_DAYS }>, // disable rotation `on_initialize` during migration
```
ggwpez: \- 2 Days because of overflow? Then maybe take 1000 days or something?

system-parachains/common/src/randomness.rs
ggwpez, line 1: Did you copy this from somewhere or write it new? Maybe a link would help from where it is

pallets/rc-migrator/src/society.rs
```diff
@@ -0,0 +1,718 @@
...
+				if messages.is_empty() {
+					return Err(Error::OutOfWeight);
+				} else {
+					break;
+				}
+			}
+
+			// TODO replace `receive_vesting_schedules` by actual ah weight func
+			if T::MaxAhWeight::get()
+				.any_lt(T::AhWeightInfo::receive_vesting_schedules((messages.len() + 1) as u32))
```
ggwpez: Some estimated values wouldbe fine

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
muharem: right. I updated now with the comment

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
muharem: removed

system-parachains/common/src/randomness.rs
muharem: this is proposed by @gui1117, left an authorship comment there

system-parachains/common/src/randomness.rs
```diff
@@ -0,0 +1,94 @@
...
+	for RelayChainOneEpochAgoWithoutBlockNumber<T, BlockNumber>
+where
+	T: cumulus_pallet_parachain_system::Config,
+	BlockNumber: From<u32>,
+{
+	fn random(subject: &[u8]) -> (T::Hash, BlockNumber) {
+		// Defensive fallback used if the `well_known_keys::ONE_EPOCH_AGO_RANDOMNESS` key
+		// is missing or absent from the validation data. This situation is unexpected,
+		// as the key should always be present.
+		let defensive_fallback = || (T::Hashing::hash(subject), 0.into());
```
gui1117, line 46: ~Maybe the fallback can use the parachain parent state root? it would be safer IIRC it is easily accessible.~

EDIT: I misread, it is using the relay chain state.

system-parachains/common/src/randomness.rs
```diff
@@ -0,0 +1,94 @@
...
+		let trie_backend = TrieBackendBuilder::new(db, relay_parent_storage_root).build();
+
+		let Ok(Some(random)) = trie_backend
+			.storage(relay_chain::well_known_keys::ONE_EPOCH_AGO_RANDOMNESS)
+			.inspect_err(|e| {
+				log::error!(
+					target: LOG_TARGET,
+					"Failed to lookup `well_known_keys::ONE_EPOCH_AGO_RANDOMNESS` from trie: {e}"
+				);
+			})
```
gui1117, line 79: This error happens when the proof `validation_data` cannot prove the value for `ONE_EPOCH_AGO_RANDOMNESS`. The trie nodes in the `validation_data` are missing or invalid and thus doesn't prove anything about `ONE_EPOCH_AGO_RANDOMNESS` in the relay chain.

If we fallback to some insecure randomness I think we should panic here instead.

The thing is collators control the part of the relay chain trie that they put in the validation data. If they intentionally maliciously decide not to include the nodes to read into `ONE_EPOCH_AGO_RANDOMNESS` then they produce a valid block with an insecure randomness. So the collator can decide to force the randomness from the defensive fallback. (even if we fallback to parachain parent state root, if 2 collators cooperate they together decide the randomness).

Either we panic here so that a collator must include the necessary trie nodes to read `ONE_EPOCH_AGO_RANDOMNESS` (which omni-node does by default). Or we must return an option so that the implementation can wait for a non-malicious collator.

(In my implementation I had decided to wait for a non-malicious collator)

system-parachains/common/src/randomness.rs
```diff
@@ -0,0 +1,94 @@
...
+				"`well_known_keys::ONE_EPOCH_AGO_RANDOMNESS` is none; cannot fetch randomness"
+			);
+			return defensive_fallback();
+		};
+
+		let mut subject = subject.to_vec();
+		subject.reserve(VRF_RANDOMNESS_LENGTH);
+		subject.extend_from_slice(&random);
+
+		(T::Hashing::hash(&subject[..]), 0.into())
```
gui1117, line 92: The block number is "the time in the past since when it was determinable by chain observers."
I think the real value is the last block number of the past epoch. But there should be the real value written in Babe.
That said 0 is safe, but any correct user of randomness would not be able to use this randomness given it is only valid for events happening before genesis.

I think if you want the best randomness but with least delay you should use `ParentBlockRandomness` which is also in Babe, slightly less good than `ONE_EPOCH_AGO_RANDOMNESS` but still completely fine IIRC.

gui1117: In general I feel like ignoring this value is worse than just using the parachain parent block root hash.

Example you want to use this randomness for an election, people register, it gives them an order and then you draw using this randomness the index for the winner.
If the randomness is from last epoch and people can still register for the election, then the election result is justa deterministic function of number of people registered, and people can decide to register additional people in a specific order to be sure to be elected. Which is not good.

PR #879, opened 2025-09-04
[AHM] Free 'Any' proxies for pure accounts

ggwpez:
Give pure accounts their `Any` proxies for free on the Relay Chain even after the migration. Otherwise its bad UX to recover access to them through remote proxies pallet only.


pallets/rc-migrator/src/lib.rs
```diff
@@ -1237,7 +1245,7 @@ pub mod pallet {
 					Self::transition(MigrationStage::AccountsMigrationInit);
 				},
 				MigrationStage::AccountsMigrationInit => {
-					let weight = AccountsMigrator::<T>::obtain_rc_accounts();
+					let weight = AccountsMigrator::<T>::obtain_rc_accounts().saturating_add(AccountsMigrator::<T>::obtain_free_proxy_candidates());
```
muharem: this stage getting quite big. can we check its not topp big already?

muharem: may be separate stage for pure proxies? 
it actually can be at PureProxyInit stage

pallets/rc-migrator/src/proxy.rs
```diff
@@ -88,12 +88,35 @@ impl<T: Config> PalletMigration for ProxyProxiesMigrator<T> {
...
+						let mut free_proxies: BoundedVec<_, _> = proxies
+							.into_iter()
+							.filter(|p| T::PureProxyFreeVariants::contains(&p.proxy_type))
+							.collect::<Vec<_>>()
+							.defensive_truncate_into();
+						let deposit: BalanceOf<T> = Zero::zero();
+						log::debug!(target: LOG_TARGET, "Pure account {} gets {} proxies for free: {:?}", acc.to_ss58check(), free_proxies.len(), free_proxies);
+
+						if !free_proxies.is_empty() {
+							pallet_proxy::Proxies::<T>::insert(&acc, (free_proxies, deposit));
```
muharem: does not this break idempotency of the proxy stage?

pallets/rc-migrator/src/lib.rs
ggwpez: Yea I can make another stage.

pallets/rc-migrator/src/proxy.rs
ggwpez: If we run it twice then we just filter the Any proxies again, but it will again return the Any proxies.  
I just checked the output after adding a log and it is inserting the exact same stuff in the second idempotent run.

pallets/rc-migrator/src/lib.rs
ggwpez: put it in the init now.

muharem (approved):

PR #893, opened 2025-09-08
Bump unstable2507

bkontur:


sigurpol (approved):
LGTM, and indeed, staking-async is not used in main so its bump to 0.4.0 needs to be handled separately

kianenigma (approved):

PR #896, opened 2025-09-08
[AHM] Max data messages per block limit

muharem:
Introduces max data messages per block limit and closes few TODOs.

### DIFF
--- a/pallets/rc-migrator/src/lib.rs
+++ b/pallets/rc-migrator/src/lib.rs
@@ -137,7 +137,7 @@ pub const MAX_XCM_SIZE: u32 = 50_000;
 pub const MAX_ITEMS_PER_BLOCK: u32 = 1600;
 
 /// The maximum number of XCM messages that can be sent in a single block.
-pub const MAX_XCM_MSG_PER_BLOCK: u32 = 20;
+pub const MAX_XCM_MSG_PER_BLOCK: u32 = 10;
 
 /// Out of weight Error. Can be converted to a pallet error for convenience.
 pub struct OutOfWeightError;


pallets/rc-migrator/src/child_bounties.rs
```diff
@@ -133,13 +142,24 @@ impl<T: Config> PalletMigration for ChildBountiesMigrator<T> {
...
 				log::info!(
+					target: LOG_TARGET,
 					"Maximum number of items ({:?}) to migrate per block reached, current batch size: {}",
 					MAX_ITEMS_PER_BLOCK,
 					messages.len()
 				);
 				break;
 			}
 
+			if messages.batch_count() >= MAX_XCM_MSG_PER_BLOCK {
```
ggwpez, line 153: We could also have made a function like `XcmBatchAndMeter.is_full` and then no need to duplicate this check code everywhere 😅   
But i guess its fine now.

pallets/rc-migrator/src/lib.rs
```diff
@@ -136,6 +136,9 @@ pub const MAX_XCM_SIZE: u32 = 50_000;
 /// and Asset Hub.
 pub const MAX_ITEMS_PER_BLOCK: u32 = 1600;
 
+/// The maximum number of XCM messages that can be sent in a single block.
+pub const MAX_XCM_MSG_PER_BLOCK: u32 = 20;
```
ggwpez: This is the resolution that we check with? So if our limit is 50 and this is set to 20 then we can at most overshoo by 20 and have 70 messages in the queue?

pallets/rc-migrator/src/child_bounties.rs
muharem: without much of refactoring we can only replace `messages.batch_count() >= MAX_XCM_MSG_PER_BLOCK` by `XcmBatchAndMeter.is_full`. I think its not much difference.

system-parachains/asset-hubs/asset-hub-kusama/src/ah_migration/call_filter.rs
```diff
@@ -136,7 +136,7 @@ pub fn call_allowed_status(
 		NftFractionalization(..) => OFF,
 		Recovery(..) => OFF,
 		MultiBlockMigrations(..) => OFF, // has not calls
-		Revive(..) => OFF,               // TODO: OFF or ON?
+		Revive(..) => ON,
```
ggwpez: I think contracts should be OFF during the migration. They could interfere with other pallets through XCM Transact.

pallets/rc-migrator/src/lib.rs
muharem: right. but its still far from the limits of the queue

system-parachains/asset-hubs/asset-hub-kusama/src/ah_migration/call_filter.rs
muharem: but its still gonna use general executor and filters? but I am gonna disable anyway

PR #897, opened 2025-09-08
[AHM] Update sovereign account translation maps

ggwpez:
Re-run the sovereign account indexing script and merge Polkadot and Kusama outputs.  
We can merge them since the calculation is the same - regardless of the runtime.

### DIFF
--- a/integration-tests/ahm/src/tests.rs
+++ b/integration-tests/ahm/src/tests.rs
@@ -43,7 +43,6 @@ use super::{
 	multisig_test::MultisigsAccountIdStaysTheSame,
 	proxy::{ProxyBasicWorks, ProxyWhaleWatching},
 };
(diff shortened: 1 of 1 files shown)


pallets/ah-migrator/src/sovereign_account_translation.rs
```diff
@@ -639,6 +1282,21 @@ pub const SOV_TRANSLATIONS: &[((AccountId32, &'static str), (AccountId32, &'stat
 			"5Eg2fntP2fDQZquEmzoTXSLbk3mqPmhVss1QUugaxcP174FB",
 		),
 	),
+	// para 2281
```
ggwpez, line 1285: We missed to refresh and merge those for Paseo, so now we have Kusama and Polkadot merged in one.

PR #898, opened 2025-09-08
Implement 'Hard Pressure' Issuance Model

PolkadotDom:
With [Ref 1710 - Hard Pressure Capped & Stepped Supply Schedule](https://polkadot.subsquare.io/referenda/1710) having passed, we have here an implementation.

---

The code itself can be reasoned about in roughly 3 parts:
- A general stepped curve, asymptotic to some defined amount by moving a percentage closer to that amount per step.
- The era_payout() usage of this stepped curve.
- New tests for the hard pressure inflation model, as well as some old tests for the previous inflation model that didn't make their way over to the AHM branch.

### DIFF
--- a/system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
+++ b/system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
@@ -484,6 +484,8 @@ impl pallet_staking_async_rc_client::Config for Runtime {
 	type AHStakingInterface = Staking;
 	type SendToRelayChain = StakingXcmToRelayChain;
 	type MaxValidatorSetRetries = ConstU32<64>;
+	// export validators as soon as election results are ready.
+	type ValidatorSetExportSession = ConstU32<0>;
 }
 
 #[derive(Encode, Decode)]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+- AH Polkadot - A new stepped curve primitive. Used for the 'Hard Pressure' inflation changes from [Ref 1710](https://polkadot.subsquare.io/referenda/1710) ([#898](https://github.com/polkadot-fellows/runtimes/pull/898)).
+
+### Changed
+- AH Polkadot - Inflation now follows the 'Hard Pressure' schedule from [Ref 1710](https://polkadot.subsquare.io/referenda/1710) ([#898](https://github.com/polkadot-fellows/runtimes/pull/898)).
+
+## [2.0.4] 18.12.2025
+
 ### Changed
 
 - asset-hub-polkadot: remove treasury burns ([#1028](https://github.com/polkadot-fellows/runtimes/pull/1028))
@@ -13,7 +21,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - asset-hub-polkadot: increase the base deposit for a signed NPoS solution from 4 to 100 DOT, significantly mitigating the risk of transaction spamming ([#1022](https://github.com/polkadot-fellows/runtimes/pull/1022))
 - Adjust OpenGov parameters based on WFC 1701 ([polkadot-fellows/runtimes/pull/873](https://github.com/polkadot-fellows/runtimes/pull/873))
 - asset-hub-kusama: update to latest version of pallet-revive with EVM backend ([polkadot-fellows/runtimes/pull/1029](https://github.com/polkadot-fellows/runtimes/pull/1029))
-- AH Polkadot - Inflation now follows the 'Hard Pressure' schedule from [Ref 1710](https://polkadot.subsquare.io/referenda/1710) ([#898](https://github.com/polkadot-fellows/runtimes/pull/898)).
+- Enable Elastic Scaling on Assethub Kusama ([polkadot-fellows/runtimes/pull/1037](https://github.com/polkadot-fellows/runtimes/pull/1037))
+- Enable buffered validator selection on Assethub Kusama ([polkadot-fellows/runtimes/pull/1037](https://github.com/polkadot-fellows/runtimes/pull/1037))
 
 ### Added
 
@@ -22,7 +31,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - asset-hub-kusama: use inmemorydb weights([polkadot-fellows/runtimes/pull/918](https://github.com/polkadot-fellows/runtimes/pull/918))
 - Add fast runtime feature for Encointer Kusama([#1021](https://github.com/polkadot-fellows/runtimes/pull/1021))
 - asset-hub-polkadot: use inmemorydb weights([polkadot-fellows/runtimes/pull/918](https://github.com/polkadot-fellows/runtimes/pull/1025))
-- AH Polkadot - A new stepped curve primitive. Used for the 'Hard Pressure' inflation changes from [Ref 1710](https://polkadot.subsquare.io/referenda/1710) ([#898](https://github.com/polkadot-fellows/runtimes/pull/898)).
 
 ## [2.0.3] 21.11.2025
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3750,9 +3750,9 @@ dependencies = [
 
 [[package]]
 name = "cumulus-pallet-parachain-system"
-version = "0.24.0"
+version = "0.24.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5d5b26ee5416df1e6e13b1fd0a1780a682e26ad5789dbed5c8ed54a0436f496b"
+checksum = "ffb02cf667494e2987f88dc904b80bf27e132222cf924b7e0c50e5bfbb893d04"
 dependencies = [
  "bytes",
  "cumulus-pallet-parachain-system-proc-macro",
@@ -3967,13 +3967,15 @@ dependencies = [
 
 [[package]]
 name = "cumulus-test-relay-sproof-builder"
-version = "0.23.0"
+version = "0.23.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "04c89420792d10c9628ad8adb5eac4b7ae5e479b150da450a1420dbef83cb7ac"
+checksum = "8f4dfff163c88093b298b83ff63453be5eb56854fc9ea9dff65f4c309de609b4"
 dependencies = [
  "cumulus-primitives-core",
  "parity-scale-codec",
  "polkadot-primitives",
+ "sp-consensus-babe",
+ "sp-core 38.1.0",
  "sp-runtime",
  "sp-state-machine",
  "sp-trie",
@@ -10478,9 +10480,9 @@ dependencies = [
 
 [[package]]
 name = "pallet-staking-async"
-version = "0.7.0"
+version = "0.7.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "163316fdf72b1bc1683ec426a94f8f91ae4a4362e19137625dd3d2be1a8819ac"
+checksum = "e075a86fb995955dd5a2426ae283147bd87c805dcc55fe35f97f8a24f6c5a621"
 dependencies = [
  "frame-benchmarking",
  "frame-election-provider-support",
@@ -10524,9 +10526,9 @@ dependencies = [
 
 [[package]]
 name = "pallet-staking-async-rc-client"
-version = "0.6.0"
+version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "273b006de13412fd8e4f5d689316edb39d1313871994a573b4e0da774a5f0b6b"
+checksum = "c864cf98b8ffcb7aa6dbebf1501ee082fa7ab11845ab69dd91d6acb08343720b"
 dependencies = [
  "frame-support",
  "frame-system",
@@ -10868,9 +10870,9 @@ dependencies = [
 
 [[package]]
 name = "parachains-runtimes-test-utils"
-version = "27.0.0"
+version = "27.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "2eba3dfe5a0f8ef4ca7da6fcade53086535566b13fd593f6699f04de27b053f9"
+checksum = "e1b479747a59d62704b32c0ae6d30d3da575b18c329ec8b33899027b7086c664"
 dependencies = [
  "cumulus-pallet-parachain-system",
  "cumulus-pallet-xcmp-queue",
@@ -18983,9 +18985,9 @@ dependencies = [
 
 [[package]]
 name = "xcm-emulator"
-version = "0.24.0"
+version = "0.24.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a04959e80d5626a25f61950f92bf12739b741c4cb6e28a1f172880db175782d2"
+checksum = "51dc3044122d10846d15aabc7b381a7591dd1a7a9841ad243f5c4256fe0e9386"
 dependencies = [
  "array-bytes 6.2.3",
  "cumulus-pallet-parachain-system",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ pallet-ah-migrator = { path = "pallets/ah-migrator", default-features = false }
 pallet-rc-migrator = { path = "pallets/rc-migrator", default-features = false }
 pallet-ah-ops = { path = "pallets/ah-ops", default-features = false }
 pallet-election-provider-multi-block = { version = "0.5.0", default-features = false }
-pallet-staking-async = { version = "0.7.0", default-features = false }
+pallet-staking-async = { version = "0.7.1", default-features = false }
 hex = { version = "0.4.3", default-features = false }
 rand = { version = "0.9.2" }
 impl-trait-for-tuples = { version = "0.2.3", default-features = false }
@@ -188,7 +188,7 @@ pallet-xcm-bridge-hub = { version = "0.20.0", default-features = false }
 pallet-xcm-bridge-hub-router = { version = "0.22.0", default-features = false }
 parachain-info = { version = "0.24.0", default-features = false, package = "staging-parachain-info" }
 parachains-common = { version = "26.0.0", default-features = false }
-parachains-runtimes-test-utils = { version = "27.0.0" }
+parachains-runtimes-test-utils = { version = "27.0.1" }
 paste = { version = "1.0.14" }
 penpal-emulated-chain = { path = "integration-tests/emulated/chains/parachains/testing/penpal" }
 penpal-runtime = { version = "0.34.0" }
@@ -266,7 +266,7 @@ system-parachains-common = { path = "system-parachains/common", default-features
 tokio = { version = "1.45.0" }
 xcm = { version = "20.0.0", default-features = false, package = "staging-xcm" }
 xcm-builder = { version = "24.0.0", default-features = false, package = "staging-xcm-builder" }
-xcm-emulator = { version = "0.24.0" }
+xcm-emulator = { version = "0.24.1" }
 xcm-executor = { version = "23.0.0", default-features = false, package = "staging-xcm-executor" }
 xcm-runtime-apis = { version = "0.11.0", default-features = false }
 anyhow = { version = "1.0.82" }
--- a/relay/kusama/src/lib.rs
+++ b/relay/kusama/src/lib.rs
@@ -188,7 +188,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: alloc::borrow::Cow::Borrowed("kusama"),
 	impl_name: alloc::borrow::Cow::Borrowed("parity-kusama"),
 	authoring_version: 2,
-	spec_version: 2_000_003,
+	spec_version: 2_000_004,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 26,
--- a/relay/polkadot/src/lib.rs
+++ b/relay/polkadot/src/lib.rs
@@ -177,7 +177,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: alloc::borrow::Cow::Borrowed("polkadot"),
 	impl_name: alloc::borrow::Cow::Borrowed("parity-polkadot"),
 	authoring_version: 0,
-	spec_version: 2_000_003,
+	spec_version: 2_000_004,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 26,
--- a/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
+++ b/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
@@ -109,7 +109,9 @@ use system_parachains_constants::{
 	},
 	kusama::{
 		consensus::{
-			async_backing::UNINCLUDED_SEGMENT_CAPACITY, BLOCK_PROCESSING_VELOCITY,
+			elastic_scaling::{
+				BLOCK_PROCESSING_VELOCITY, RELAY_PARENT_OFFSET, UNINCLUDED_SEGMENT_CAPACITY,
+			},
 			RELAY_CHAIN_SLOT_DURATION_MILLIS,
 		},
 		currency::*,
@@ -145,7 +147,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("statemine"),
 	impl_name: Cow::Borrowed("statemine"),
 	authoring_version: 1,
-	spec_version: 2_000_003,
+	spec_version: 2_000_004,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 15,
@@ -850,7 +852,7 @@ impl cumulus_pallet_parachain_system::Config for Runtime {
 	type ConsensusHook = ConsensusHook;
 	type WeightInfo = weights::cumulus_pallet_parachain_system::WeightInfo<Runtime>;
 	type SelectCore = cumulus_pallet_parachain_system::DefaultCoreSelector<Runtime>;
-	type RelayParentOffset = ConstU32<0>;
+	type RelayParentOffset = ConstU32<RELAY_PARENT_OFFSET>;
 }
 
 type ConsensusHook = cumulus_pallet_aura_ext::FixedVelocityConsensusHook<
@@ -2313,7 +2315,7 @@ pallet_revive::impl_runtime_apis_plus_revive_traits!(
 
 	impl cumulus_primitives_core::RelayParentOffsetApi<Block> for Runtime {
 		fn relay_parent_offset() -> u32 {
-			0
+			RELAY_PARENT_OFFSET
 		}
 	}
 
--- a/system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
+++ b/system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
@@ -412,8 +412,8 @@ impl pallet_staking_async::Config for Runtime {
 	type EventListeners = (NominationPools, DelegatedStaking);
 	// Note used; don't care.
 	type MaxInvulnerables = frame_support::traits::ConstU32<20>;
-	type PlanningEraOffset =
-		pallet_staking_async::PlanningEraOffsetOf<Self, RelaySessionDuration, ConstU32<10>>;
+	// This will start election for the next era as soon as an era starts.
+	type PlanningEraOffset = ConstU32<6>;
 	type RcClientInterface = StakingRcClient;
 	type MaxEraDuration = MaxEraDuration;
 	type WeightInfo = weights::pallet_staking_async::WeightInfo<Runtime>;
@@ -425,6 +425,8 @@ impl pallet_staking_async_rc_client::Config for Runtime {
 	type AHStakingInterface = Staking;
 	type SendToRelayChain = StakingXcmToRelayChain;
 	type MaxValidatorSetRetries = ConstU32<64>;
+	// Export elected validator set at end of session 4.
+	type ValidatorSetExportSession = ConstU32<4>;
 }
 
 #[derive(Encode, Decode)]
--- a/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
+++ b/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
@@ -193,7 +193,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_name: Cow::Borrowed("statemint"),
 	spec_name: Cow::Borrowed("statemint"),
 	authoring_version: 1,
-	spec_version: 2_000_003,
+	spec_version: 2_000_004,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 15,
--- a/system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
+++ b/system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
@@ -179,7 +179,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("bridge-hub-kusama"),
 	impl_name: Cow::Borrowed("bridge-hub-kusama"),
 	authoring_version: 1,
-	spec_version: 2_000_003,
+	spec_version: 2_000_004,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 5,
--- a/system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
+++ b/system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
@@ -186,7 +186,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("bridge-hub-polkadot"),
 	impl_name: Cow::Borrowed("bridge-hub-polkadot"),
 	authoring_version: 1,
-	spec_version: 2_000_003,
+	spec_version: 2_000_004,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 4,
--- a/system-parachains/collectives/collectives-polkadot/src/lib.rs
+++ b/system-parachains/collectives/collectives-polkadot/src/lib.rs
@@ -130,7 +130,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("collectives"),
 	impl_name: Cow::Borrowed("collectives"),
 	authoring_version: 1,
-	spec_version: 2_000_003,
+	spec_version: 2_000_004,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 7,
--- a/system-parachains/constants/src/kusama.rs
+++ b/system-parachains/constants/src/kusama.rs
@@ -33,6 +33,23 @@ pub mod consensus {
 		/// the relay chain.
 		pub const UNINCLUDED_SEGMENT_CAPACITY: u32 = 3;
 	}
+
+	/// Parameters enabling elastic scaling functionality.
+	pub mod elastic_scaling {
+		/// Build with an offset of 1 behind the relay chain.
+		pub const RELAY_PARENT_OFFSET: u32 = 1;
+
+		/// The upper limit of how many parachain blocks are processed by the relay chain per
+		/// parent. Limits the number of blocks authored per slot. This determines the minimum
+		/// block time of the parachain:
+		/// `RELAY_CHAIN_SLOT_DURATION_MILLIS/BLOCK_PROCESSING_VELOCITY`
+		pub const BLOCK_PROCESSING_VELOCITY: u32 = 3;
+
+		/// Maximum number of blocks simultaneously accepted by the Runtime, not yet included
+		/// into the relay chain.
+		pub const UNINCLUDED_SEGMENT_CAPACITY: u32 =
+			(3 + RELAY_PARENT_OFFSET) * BLOCK_PROCESSING_VELOCITY;
+	}
 }
 
 /// Constants relating to KSM.
--- a/system-parachains/coretime/coretime-kusama/src/lib.rs
+++ b/system-parachains/coretime/coretime-kusama/src/lib.rs
@@ -153,7 +153,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("coretime-kusama"),
 	impl_name: Cow::Borrowed("coretime-kusama"),
 	authoring_version: 1,
-	spec_version: 2_000_003,
+	spec_version: 2_000_004,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,
--- a/system-parachains/coretime/coretime-polkadot/src/lib.rs
+++ b/system-parachains/coretime/coretime-polkadot/src/lib.rs
@@ -150,7 +150,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("coretime-polkadot"),
 	impl_name: Cow::Borrowed("coretime-polkadot"),
 	authoring_version: 1,
-	spec_version: 2_000_003,
+	spec_version: 2_000_004,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 0,
--- a/system-parachains/encointer/src/lib.rs
+++ b/system-parachains/encointer/src/lib.rs
@@ -146,7 +146,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("encointer-parachain"),
 	impl_name: Cow::Borrowed("encointer-parachain"),
 	authoring_version: 1,
-	spec_version: 2_000_003,
+	spec_version: 2_000_004,
 	impl_version: 1,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 4,
--- a/system-parachains/gluttons/glutton-kusama/src/lib.rs
+++ b/system-parachains/gluttons/glutton-kusama/src/lib.rs
@@ -94,7 +94,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("glutton"),
 	impl_name: Cow::Borrowed("glutton"),
 	authoring_version: 1,
-	spec_version: 2_000_003,
+	spec_version: 2_000_004,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,
--- a/system-parachains/people/people-kusama/src/lib.rs
+++ b/system-parachains/people/people-kusama/src/lib.rs
@@ -174,7 +174,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("people-kusama"),
 	impl_name: Cow::Borrowed("people-kusama"),
 	authoring_version: 1,
-	spec_version: 2_000_003,
+	spec_version: 2_000_004,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 1,
--- a/system-parachains/people/people-polkadot/src/lib.rs
+++ b/system-parachains/people/people-polkadot/src/lib.rs
@@ -152,7 +152,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("people-polkadot"),
 	impl_name: Cow::Borrowed("people-polkadot"),
 	authoring_version: 1,
-	spec_version: 2_000_003,
+	spec_version: 2_000_004,
 	impl_version: 0,
 	apis: RUNTIME_API_VERSIONS,
 	transaction_version: 0,


acatangiu:
> The community is asking for swift execution on this and is hoping to slate it into a 1.8.1 release.
This PR points towards the AHM branch for simplicity, but will be closed in favor of one that points towards main when AHM is merged.

I do not like bundling this in the same release as AHM. It simply hasn't been tested enough to drop it live on Polkadot. I would bring it after AHM as a dedicated patch release. We need to test these params properly with the new async staking code.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -262,14 +398,42 @@ impl pallet_staking_async::EraPayout<Balance> for EraPayout {
...
+		let yearly_emission = if relay_block_num < march_14_2026 {
+			// TI at the time of execution of [Referendum 1139](https://polkadot.subsquare.io/referenda/1139), block hash: `0x39422610299a75ef69860417f4d0e1d94e77699f45005645ffc5e8e619950f9f`.
+			let fixed_total_issuance: u128 = 15_011_657_390_566_252_333;
+			let fixed_inflation_rate = FixedU128::from_rational(8, 100);
+			fixed_inflation_rate.saturating_mul_int(fixed_total_issuance)
+		} else {
+			// Extrapolated total issuance at time of first stepping. Based on a per era emission of ~328_774.593.
+			// Note: This deviates a bit from [Ref 1710's](https://polkadot.subsquare.io/referenda/1710) initial
+			// extrapolation of 1_676_733_867.
+			let march_14_2026_ti = 1_676_562_737u128 * UNITS;
```
kianenigma: Can't we rewrite this code such that it actually looks at the current TI, our target TI, and then inflate based on that?

kianenigma: Alternatively, this code can use a runtime storage item (either in a pallet, or a floating one), that is initially `None`, and mutated to `Some()` the first time this code path is executed. In this, can store the accurate actual TI at the time of this code kicking in.

PolkadotDom: Yes, in agreeance with the `Some()` approach. I couldn't find any examples of a floating storage item, if you know of one. Will look deeper into it.

Turning to stateful vs stateless - afaict the arguments here really come down to what you're trying to optimize for. With the stateful approach, you're going to end up a bit closer to the target TI as the period emissions would be adapting to the small discrepancies in era duration and thus era emission. With stateless you're going to get high certainty around era emission and emission schedule indefinitely into the future, as it's fixed.

In the end both approaches are going to be, what I would think is, nominally close in both era emission and ending issuance, so I just opted for the conceptually simpler and easier to use API. And I suppose I had a bias towards static emission scheduling.

Very open to opinions or rebuts on all this.

PolkadotDom: I suppose if we really wanted to do this right we would decouple inflation from staking 🤔 I know you had envisioned an inflation pallet a while back.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -252,6 +254,140 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+	use scale_info::TypeInfo;
+	use sp_arithmetic::traits::Saturating;
+
+	/// The step type for the stepped curve.
+	#[derive(PartialEq, Eq, sp_core::RuntimeDebug, TypeInfo, Clone)]
+	pub enum Step<V> {
+		/// Move towards a desired value by a percentage of the remaining difference at each step.
+		///
+		/// Step size will be (target_total - current_value) * pct.
+		RemainingPct(V, Perbill),
```
ggwpez:
```suggestion
		RemainingPct {
				/// The asymptotic limit that will never be reached.
				limite: V,
				/// This percentage multiplied with the remaining amount is the next step size.
				increase: Perbill,
		}
```

Tuples can easily mess stuff up, especially when someone uses the same types like `V = Perbill`.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -252,6 +254,140 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+	}
+
+	/// A stepped curve.
+	///
+	/// Steps every `period` from the `initial_value` as defined by `step`.
+	/// First step from `initial_value` takes place at `start` + `period`.
+	#[derive(PartialEq, Eq, sp_core::RuntimeDebug, TypeInfo, Clone)]
+	pub struct SteppedCurve<P, V> {
+		/// The starting point for the curve.
+		pub start: P,
```
ggwpez: X or Y? I guess X, but maybe put it in the comment.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -252,6 +254,140 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+	}
+
+	impl<P, V> SteppedCurve<P, V>
+	where
+		P: AtLeast32BitUnsigned + Copy,
+		V: AtLeast32BitUnsigned + Copy + From<P>,
+	{
+		/// Creates a new `SteppedCurve`.
+		pub fn new(start: P, end: Option<P>, initial_value: V, step: Step<V>, period: P) -> Self {
+			Self { start, end, initial_value, step, period }
```
ggwpez: `try_new` maybe and sanity checking like `end >= start`.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -252,6 +254,140 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+		/// Step size will be (target_total - current_value) * pct.
+		RemainingPct(V, Perbill),
+	}
+
+	/// A stepped curve.
+	///
+	/// Steps every `period` from the `initial_value` as defined by `step`.
+	/// First step from `initial_value` takes place at `start` + `period`.
+	#[derive(PartialEq, Eq, sp_core::RuntimeDebug, TypeInfo, Clone)]
+	pub struct SteppedCurve<P, V> {
```
ggwpez:
```suggestion
	pub struct SteppedCurve<X, Y> {
```
Maybe this would help with the terminology to be a bit more mathematical.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -252,6 +254,140 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+				return V::zero();
+			}
+
+			// If the period is zero, the value never changes.
+			if self.period.is_zero() {
+				return V::zero();
+			}
+
+			// Determine the effective point for calculation, capped by the end point if it exists.
+			let _effective_point = self.end.map_or(point, |e| point.min(e));
```
ggwpez: unused?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -252,6 +254,140 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+
+			// Calculate how many full periods have passed.
+			let num_periods = (point - self.start) / self.period;
```
ggwpez, line 321: Just dividing can lead to precision loss. I think using `Fixed*64` for all calculations would guarantee us precision.  
Not sure if we even need generic types, Fixing `X` and `Y` to some `Fixed*` type would ensure that any user gets correct precision like in the [referenda curves](https://github.com/paritytech/polkadot-sdk/blob/56234513d1d1b3cc9fb85fcc1a9735ab9df22ef2/substrate/frame/referenda/src/types.rs#L431).

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -252,6 +254,140 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+				return val_prev.saturating_sub(val_curr);
+			}
+		}
+
+		/// Evaluate the curve at a given point.
+		pub fn evaluate(&self, point: P) -> V {
+			let initial = self.initial_value;
+
+			// If the point is before the curve starts, return the initial value.
+			if point < self.start {
```
ggwpez:
```suggestion
			if point <= self.start {
```
?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -252,6 +254,140 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+
+			if val_curr >= val_prev {
+				return val_curr.saturating_sub(val_prev);
+			} else {
+				return val_prev.saturating_sub(val_curr);
+			}
+		}
+
+		/// Evaluate the curve at a given point.
+		pub fn evaluate(&self, point: P) -> V {
```
ggwpez:
```suggestion
		pub fn evaluate(&self, num_periods: usize) -> V {
```
Given that this is a discrete function, i think it could make sense to keep it that way and call it directly with discrete `num_periods` instead of up- and down-sampling the point.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -252,6 +254,140 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+		pub end: Option<P>,
+		/// The initial value of the curve at the `start` point.
+		pub initial_value: V,
+		/// The change to apply at the end of each `period`.
+		pub step: Step<V>,
+		/// The duration of each step.
+		pub period: P,
+	}
+
+	impl<P, V> SteppedCurve<P, V>
```
ggwpez: Not sure if we even need generic types, Fixing `X` and `Y` to some `Fixed*` type would ensure that any user gets correct precision like in the [referenda curves](https://github.com/paritytech/polkadot-sdk/blob/56234513d1d1b3cc9fb85fcc1a9735ab9df22ef2/substrate/frame/referenda/src/types.rs#L431).

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
ggwpez: I dont really mind if we fix it or not, IMHO it should be fine to hard-code since our current inflation formula is quite stable.

Otherwise for free-floating storage you can use `storage_alias(verbatim)` (I think we only use it once in the SDK).

system-parachains/asset-hubs/asset-hub-polkadot/tests/tests.rs
```diff
@@ -954,3 +954,291 @@ fn governance_authorize_upgrade_works() {
...
+		.execute_with(|| {
+			let (to_stakers, to_treasury) = <staking::EraPayout as EraPayout<Balance>>::era_payout(
+				123,                                 // ignored
+				456,                                 // ignored
+				(36525 * MILLISECONDS_PER_DAY) / 10, // 10 years
+			);
+			let initial_ti: i128 = 15_011_657_390_566_252_333;
+			let projected_total_issuance = (to_stakers as i128 + to_treasury as i128) + initial_ti;
+
+			// In 2034, there will be about 2.7 billion DOT in existence.
```
ggwpez: This is testing the old formula now. Maybe add an `old_formula_` prefix or something to the test so that it wont confuse us.

system-parachains/asset-hubs/asset-hub-polkadot/tests/tests.rs
```diff
@@ -954,3 +954,291 @@ fn governance_authorize_upgrade_works() {
...
+			Some( PersistedValidationData {
+				parent_head: vec![].into(),
+				relay_parent_number: n,
+				max_pov_size: Default::default(),
+				relay_parent_storage_root: Default::default(),
+			})
+		);
+	}
+
+	const TWO_YEAR_STEP_RATE: f64 = 0.2628;
```
ggwpez: Maybe we can express this as fraction with `Perbill::from_rational(2_628, 10_000)`?  
Floats will otherwise have small errors already when defining the constants and then its downhill with the precision from there 🙈 😆

system-parachains/asset-hubs/asset-hub-polkadot/tests/tests.rs
```diff
@@ -954,3 +954,291 @@ fn governance_authorize_upgrade_works() {
...
+				456, // ignored
+				MILLISECONDS_PER_DAY,
+			);
+			let two_year_rate = Perbill::from_float(TWO_YEAR_STEP_RATE);
+			let first_period_emission = two_year_rate * (TARGET_TI - MARCH_TI);
+			assert_relative_eq!(
+				(to_stakers as f64 + to_treasury as f64) * 365.25 * 2.0,
+				(first_period_emission * UNITS) as f64,
+				max_relative = 0.00001
+			);
```
ggwpez:
```suggestion
			);
			assert_relative_eq!((first_period_emission * UNITS) as f64, 123....); 
```
Some sanity check values please for human verification. Just to check that the order of magnitude looks good.

system-parachains/asset-hubs/asset-hub-polkadot/tests/tests.rs
```diff
@@ -954,3 +954,291 @@ fn governance_authorize_upgrade_works() {
...
+				123, // ignored
+				456, // ignored
+				MILLISECONDS_PER_DAY,
+			);
+
+			// TI has converged on asymptote. Payout is zero.
+			assert_eq!(to_stakers + to_treasury, 0);
+		});
+	}
+
```
ggwpez: Can you add one more test please that integrates under the curve until +100 years or something far in the future?
All emission values summed up plus the initial one should be within our 21B limit, just to check that it never exceeds the limit.

ggwpez (commented):
Yes looks good, i put some comments. Only major comment would be if we need to have generics on the curve.  
I think using `FixedU64` would give us better precision and remove footguns of accidentally using low-precision types with the curve.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: I'll just remove the end option since we don't use it here. Will change in sdk PR.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: Either works. Will add the `=`. Conceptually better I think.

system-parachains/asset-hubs/asset-hub-polkadot/tests/tests.rs
PolkadotDom: Going to use `pre_march_2026_formula_` as technically still using until then 👍

system-parachains/asset-hubs/asset-hub-polkadot/tests/tests.rs
PolkadotDom: done

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: 🤔 .. Yeah I think you're right 👍 For this bespoke usage it works but will be better with `Fixed` for sdk.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: I see what you're saying.. but it feels like a leaky abstraction / confusing api design. Not too worried about the down-sampling personally.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: Wavered a bit back and forth on this, but ultimately realized when extrapolating the TI on March 14, 2026 we hadn't taken into account any burns from now until then. Which would be hard to forecast considering it's mainly coretime and treasury burn. So let's just store the value to be more precise.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -299,14 +433,56 @@ impl pallet_staking_async::EraPayout<Balance> for EraPayout {
...
 		let relative_era_len =
 			FixedU128::from_rational(era_duration_millis.into(), MILLISECONDS_PER_YEAR.into());
 
-		// TI at the time of execution of [Referendum 1139](https://polkadot.subsquare.io/referenda/1139), block hash: `0x39422610299a75ef69860417f4d0e1d94e77699f45005645ffc5e8e619950f9f`.
-		let fixed_total_issuance: i128 = 5_216_342_402_773_185_773;
-		let fixed_inflation_rate = FixedU128::from_rational(8, 100);
-		let yearly_emission = fixed_inflation_rate.saturating_mul_int(fixed_total_issuance);
+		// Branch based off the 12AM 14th March 2026 initial stepping date -[Ref 1710](https://polkadot.subsquare.io/referenda/1710?tab=votes_bubble).
+		let relay_block_num = <RelaychainDataProvider::<Runtime> as BlockNumberProvider>::current_block_number();
+		let march_14_2026: BlockNumber = 30_362_493; // https://polkadot.subscan.io/block/30362493
```
PolkadotDom, line 442: This block number appears to be a moving target due to fluctuations in block time. Wondering if we should use timestamp.

PolkadotDom: I feel RC block number is higher security, but the value is a moving target due empirically. If anyone wants to argue for timestamp..

PolkadotDom:
Please note to reviewers thus far-- The target branch has been changed to main.

relay/polkadot/constants/src/lib.rs
```diff
@@ -53,6 +53,7 @@ pub mod time {
 	pub const HOURS: BlockNumber = MINUTES * 60;
 	pub const DAYS: BlockNumber = HOURS * 24;
 	pub const WEEKS: BlockNumber = DAYS * 7;
+	pub const YEARS: BlockNumber = (DAYS * 36525) / 100;
```
ggwpez, line 56: We should have a tests that this does not do rounding. AFAIK it evenly divides but here it is not obvious.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -284,6 +291,147 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+			}
+
+			// If the period is zero, the value never changes.
+			if self.period.is_zero() {
+				return FixedU128::zero();
+			}
+
+			// Calculate how many full periods have passed, saturate.
+			let num_periods =
+				(point - self.start).checked_div(&self.period).unwrap_or(FixedU128::max_value());
```
ggwpez: nit: with `if let Ok(from_start) = point.checked_sub(self.start) else` you can do it underflow safe and use that in the `if` above instead of `point <= self.start`.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -284,6 +291,147 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+		/// Ex. In period 4, the last step taken was 10 -> 7, it would return 3.
+		pub fn last_step_size(&self, point: FixedU128) -> FixedU128 {
+			// No step taken yet.
+			if point <= self.start {
+				return FixedU128::zero();
+			}
+
+			// If the period is zero, the value never changes.
+			if self.period.is_zero() {
+				return FixedU128::zero();
```
ggwpez:
```suggestion
				return Zero::zero();
```
nit

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -284,6 +291,147 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+				self.start.saturating_add(num_periods.saturating_mul(self.period));
+
+			// Evaluate the curve at those two points.
+			let val_prev = self.evaluate(prev_period_point);
+			let val_curr = self.evaluate(curr_period_point);
+
+			if val_curr >= val_prev {
+				val_curr.saturating_sub(val_prev)
+			} else {
+				val_prev.saturating_sub(val_curr)
```
ggwpez: Can it ever go down? Maybe if the `target` is below the `initial_value`?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -284,6 +291,147 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+			// If the period is zero, the value never changes.
+			if self.period.is_zero() {
+				return initial;
+			}
+
+			// Calculate how many full periods have passed, downsampled to usize.
+			let num_periods =
+				(point - self.start).checked_div(&self.period).unwrap_or(FixedU128::max_value());
+			let num_periods_usize =
+				(num_periods.into_inner() / FixedU128::DIV).saturated_into::<usize>();
```
ggwpez:
```suggestion
				(num_periods.into_inner() / FixedU128::DIV).saturated_into::<u128>();
```
Why not a concrete type size? usize is a bit vague.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -284,6 +291,147 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+
+			// No periods have passed.
+			if num_periods_usize.is_zero() {
+				return initial;
+			}
+
+			match self.step {
+				Step::RemainingPct { target: asymptote, pct: percent } => {
+					// asymptote +/- diff(asymptote, initial_value) * (1-percent)^num_periods.
+					let ratio = FixedU128::one().saturating_sub(FixedU128::from(percent));
```
ggwpez: Huh, is there also a `FixedU128::from_percent`? Or maybe `from_rational(percent, 100)` is clearer.  
Or directly `FixedU128::from_rational(100.saturating_sub(percent), 100)`, but maybe that is too much.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
ggwpez: Maybe you can put it into the parameters pallet? I hope we deployed it to PAH now.  
Then we can soft-lock these values in and still adjust with governance alter.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -284,6 +291,147 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+		pub initial_value: Y,
+		/// The change to apply at the end of each `period`.
+		pub step: Step<Y>,
+		/// The duration of each step.
+		pub period: X,
+	}
+
+	impl SteppedCurve<FixedU128, FixedU128> {
+		/// Creates a new `SteppedCurve`.
+		pub fn new(
```
ggwpez: sorry I did not go through all discussions, but i guess this functions is always safe now and we dont need a `try_from`?

ggwpez (commented):
Curve code looks good now, did not check the exact values yet.  
Will wait for another reviewer on the Curve code

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -284,6 +291,147 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+// To be replaced after that is merged and available.
+pub mod temp_curve {
+	use super::*;
+	use scale_info::TypeInfo;
+	use sp_arithmetic::traits::Saturating;
+	use sp_runtime::traits::One;
+
+	/// The step type for the stepped curve.
+	#[derive(PartialEq, Eq, sp_core::RuntimeDebug, TypeInfo, Clone)]
+	pub enum Step<Y> {
```
kianenigma: Enum with one variant? 🙈

kianenigma: should be a struct I guess.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -284,6 +291,147 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+			pct: Perbill,
+		},
+	}
+
+	/// A stepped curve.
+	///
+	/// Steps every `period` from the `initial_value` as defined by `step`.
+	/// First step from `initial_value` takes place at `start` + `period`.
+	#[derive(PartialEq, Eq, sp_core::RuntimeDebug, TypeInfo, Clone)]
+	pub struct SteppedCurve<X, Y> {
```
kianenigma: esp for the version of this curve that lives in this repo, it is questionable why we have to be generic at all, and even if, why would `X != Y`?

ggwpez: I also commented that just using `FixedU128` is fine for me.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: I'd preferably like to keep the logic inline, as the non-underflow is guaranteed. But will change if not satisfied with the security.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: Yes thats correct-- If it's a down stepping curve.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: Was an artifact of `saturating_pow` only taking `usize`. I'll pass it through u32 first for determinism 👍

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: Looks like theres a `from_perbill`.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: Yeah we didn't need the option so figured easier to just remove.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: Yes it's a bit silly. It's an attempt to keep parity with [#9556](https://github.com/paritytech/polkadot-sdk/pull/9556)

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: Upon reflection it was my impression that we were trying to match the bitcoin halvings, but looks like those are non deterministic time wise as well, so can probably just leave.

Concerning parameters pallet - I'd say let's stray from adding any governance to this as it potentially ruins the optics of it being ossified.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: Not sure domain and ranged need to be confined to the same set, but yes fair on the generics. Have removed.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
ggwpez: For the asset hub migration this is actually quite annoying, since Polkadot Relay chain has a clock drift of about 11.7 minutes per day. If we now try to project it outwards, we could be off by about 27 hrs.  

I think that is okay, but if it drifts more then we should be ready to do another runtime upgrade to adjust it. Otherwise it will be difficult to do a launch party 😆

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -293,14 +441,59 @@ impl pallet_staking_async::EraPayout<Balance> for EraPayout {
...
+
+			// Start date of the curve is set two years prior, thus ensuring first step in March,
+			// 2026.
+			let two_years_before_march =
+				FixedU128::saturating_from_integer(march_14_2026 - (2 * RC_YEARS));
+			let relay_block_fp = FixedU128::saturating_from_integer(relay_block_num);
+			let step_duration = FixedU128::saturating_from_integer(2 * RC_YEARS);
+
+			// Pct change towards target TI at each step.
+			let two_year_rate = Perbill::from_rational(2_628u32, 10_000u32);
```
ggwpez: More of a cosmetic, but this could be set to the 1314 mentioned in the ref, when using step_duration of 1 year, or?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: I've calculated a new block number as follows-- find the amount of time until March 14th 12pm. Divide by a block time of 6.04875 seconds. Add this to current block time.

Target time is now midday to allow for variance in time drift. Block time is based off 11.7 minutes extra a day.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: Step duration of a year would mean the emitted amount per era changes each year instead of every two. I figured it was easier to lean into the linearity of the function over adding complexity elsewhere.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
ggwpez: Fine for the moment, but from experience I can tell you that some other dev in the fufture will change the code three lines above an break the invariant :P

system-parachains/asset-hubs/asset-hub-polkadot/tests/tests.rs
```diff
@@ -953,3 +953,374 @@ fn governance_authorize_upgrade_works() {
 		RuntimeOrigin,
 	>(GovernanceOrigin::Location(RelayChainLocation::get())));
 }
+
+#[cfg(test)]
+mod inflation_tests {
```
kianenigma: can be merged with `inflation_sanity_check` in `staking/mod.rs`

system-parachains/asset-hubs/asset-hub-polkadot/tests/tests.rs
```diff
@@ -953,3 +953,374 @@ fn governance_authorize_upgrade_works() {
...
+		ExtBuilder::<Runtime>::default().build().execute_with(|| {
+			// Pre-march.
+			pallet_balances::pallet::TotalIssuance::<Runtime, ()>::set(MARCH_TI);
+			<staking::EraPayout as EraPayout<Balance>>::era_payout(0, 0, MILLISECONDS_PER_DAY);
+			assert!(March2026TI::get().is_none());
+
+			// Post-march.
+			set_relay_number(MARCH_14_2026);
+			<staking::EraPayout as EraPayout<Balance>>::era_payout(0, 0, MILLISECONDS_PER_DAY);
+			assert!(March2026TI::get() == Some(MARCH_TI));
```
kianenigma:
```suggestion
			assert_eq!(March2026TI::get(), Some(MARCH_TI));
```

system-parachains/asset-hubs/asset-hub-polkadot/tests/tests.rs
```diff
@@ -953,3 +953,374 @@ fn governance_authorize_upgrade_works() {
...
+
+			// No change on subsequent call.
+			set_relay_number(MARCH_14_2026 + 2 * RC_YEARS);
+			pallet_balances::pallet::TotalIssuance::<Runtime, ()>::set(MARCH_TI + 1);
+			<staking::EraPayout as EraPayout<Balance>>::era_payout(0, 0, MILLISECONDS_PER_DAY);
+			assert!(March2026TI::get() == Some(MARCH_TI));
```
kianenigma:
```suggestion
			assert_eq!(March2026TI::get(), Some(MARCH_TI));
```

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -287,6 +293,147 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+
+			match self.step {
+				Step::RemainingPct { target: asymptote, pct: percent } => {
+					// asymptote +/- diff(asymptote, initial_value) * (1-percent)^num_periods.
+					let ratio = FixedU128::one().saturating_sub(FixedU128::from_perbill(percent));
+					let scale = ratio.saturating_pow(num_periods_u32 as usize);
+
+					if initial >= asymptote {
+						let diff = initial.saturating_sub(asymptote);
+						asymptote.saturating_add(diff.saturating_mul(scale))
```
ggwpez: Is there a test for a decreasing curve? I think we currently only test with the exact Polkadot parameters.  
The possibility of decreasing curve could also be removed 🤷‍♂️

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -297,14 +444,60 @@ impl pallet_staking_async::EraPayout<Balance> for EraPayout {
...
 
-		// TI at the time of execution of [Referendum 1139](https://polkadot.subsquare.io/referenda/1139), block hash: `0x39422610299a75ef69860417f4d0e1d94e77699f45005645ffc5e8e619950f9f`.
-		let fixed_total_issuance: i128 = 15_011_657_390_566_252_333;
-		let fixed_inflation_rate = FixedU128::from_rational(8, 100);
-		let yearly_emission = fixed_inflation_rate.saturating_mul_int(fixed_total_issuance);
+		// Branch based off the 12AM 14th March 2026 initial stepping date -[Ref 1710](https://polkadot.subsquare.io/referenda/1710?tab=votes_bubble).
+		let relay_block_num =
+			<RelaychainDataProvider<Runtime> as BlockNumberProvider>::current_block_number();
+		// Calculated assuming a 11.7 minute per day time drift (A block time of 6.04875 seconds).
+		let march_14_2026: BlockNumber = 30_349_908; // https://polkadot.subscan.io/block/30362493
```
sigurpol: [nitpick] comment / url mismatch:
  - Constant is 30_349_908 
  - Comment URL points to 30362493 

according to subscan prediction, the closest to 12:00AM UTC seems to be https://polkadot.subscan.io/block/30349762

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
sigurpol: Strong +1 for simplifying the code by removing the decreasing curve logic entirely.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -287,6 +293,147 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
 	type TargetSnapshotPerBlock = <Runtime as multi_block::Config>::TargetSnapshotPerBlock;
 }
 
+// Wittled copy of the code from https://github.com/paritytech/polkadot-sdk/pull/9556
```
sigurpol: This can be removed because we are making this PR self-consistent AFAIK - and also the name temp_curve can be replaced as well

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -287,6 +293,147 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+			}
+		}
+	}
+}
+
+// Holds the TI from March 14, 2026
+#[storage_alias(verbatim)]
+pub type March2026TI = StorageValue<Runtime, Balance, OptionQuery>;
+
 // We cannot re-use the one from the relay since that is for pallet-staking and will be removed soon
```
sigurpol: [nitpick] since we are now postAHM, we can probably remove/update the comment

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -297,14 +444,60 @@ impl pallet_staking_async::EraPayout<Balance> for EraPayout {
...
+				current_ti
+			});
+
+			// The calculated TI used in [Ref 1710's](https://polkadot.subsquare.io/referenda/1710).
+			let march_14_2026_ti = FixedU128::saturating_from_integer(starting_ti);
+			let target_ti = FixedU128::saturating_from_integer(2_100_000_000u128 * UNITS);
+
+			// Start date of the curve is set two years prior, thus ensuring first step in March,
+			// 2026.
+			let two_years_before_march =
```
sigurpol: [very debatable] we are hard-coding the 2y logic in many places here. Would it make sense to make the 2-year period configurable instead ?
E.g. adding to  dynamic_params::staking_election module (in lib.rs ~line 1258):
```rust  
/// Step period in years for inflation curve adjustments.
  #[codec(index = 7)]
  pub static StepPeriodYears: u32 = 2;
```
and  use it in era_payout() e.g. something like 
```rust
  let step_period_years = StepPeriodYears::get();
  let step_duration = FixedU128::saturating_from_integer(
      u128::from(step_period_years) * RC_YEARS
  );
```

One benefit could be that in the future we could adjust granularity w/o runtime upgrade. One could think the same of the step rate (26.28%) (or even the target TI...) but maybe it's an overkill

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: I also don't like the `.unwrap_or(FixedU128::max_value())`, it should be `unwrap_or_default()`, issuing zero steps in case of any errors.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: The sentiment is that we change inflation and not touch it again, so I suggest we move more things to `pub const: _ = ... ` as associated types of `EraPayout`, but not make them exposed to OpenGov. 

The start time is the only reasonable one IMO, so we can adjust the block time drift.

sigurpol: `not touch it again`  sounds a bit on the optimistic side. But  it's more than reasonable to assume `very very rarely if ever`  and if that ever happens, then it's probably ok going for `const` now.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: Ah good catch 👍 And yes I need to adjust the time again. Block drift seems to be much more than 11 min day cited.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: I've removed the possibility of a decreasing curve. Curve now default to a zero curve if parameters are somehow wrong, and there's a test to make sure the default is in fact a zero curve `default_curve_is_zero_curve`

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
ggwpez: Subscan does not take clock drift into account. It will show the wrong date / time.

It is not stable, yes. Sometimes I measure the last 7 day average and it was 11.7 last time I checked.  
But it can definitely drift either way and possibly we should re-measure before publishing the release.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
ggwpez: OpenGov voted to enshrine this logic into JAM, so the sentiment is indeed to not modify it again.  
Ofc OpenGov can overturn itself at any time, but at least for now we should not make it adjustable 😆

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: Oh interesting it can go either way 🤔 Okay sounds good on the pre-release readjust.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/stepped_curve.rs
```diff
@@ -0,0 +1,369 @@
...
+impl SteppedCurve {
+	/// Creates a new `SteppedCurve`.
+	pub fn try_new(
+		start: FixedU128,
+		initial_value: FixedU128,
+		step: RemainingPct,
+		period: FixedU128,
+	) -> Result<Self, &'static str> {
+		if step.target < initial_value {
+			return Err("step.target must be >= initial_value");
```
sigurpol, line 81: It would be probably good to add a small UT to test the invariant preventing decreasing curves, something like 

```rust
  #[test]
  fn try_new_rejects_target_less_than_initial() {
      let result = SteppedCurve::try_new(
          FixedU128::zero(),
          FixedU128::from_u32(100),  // initial
          RemainingPct {
              target: FixedU128::from_u32(50),  // target < initial
              pct: Perbill::from_percent(10)
          },
          FixedU128::from_u32(10),
      );
      assert!(result.is_err());
      assert_eq!(result.unwrap_err(), "step.target must be >= 
  initial_value");
  }
```

Since we are here, maybe good testing also the edge case target == initial value and verify it should result in zero emission at all periods.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/stepped_curve.rs
```diff
@@ -0,0 +1,369 @@
...
+		let scale = ratio.saturating_pow(num_periods_u32 as usize);
+
+		// Asymptote can only be greater than or equal to previous, enforced in `try_new`.
+		let diff = asymptote.saturating_sub(initial);
+		asymptote.saturating_sub(diff.saturating_mul(scale))
+	}
+}
+
+#[cfg(test)]
+mod stepped_curve_tests {
```
sigurpol, line 164: [less important than https://github.com/polkadot-fellows/runtimes/pull/898/files#r2517553324] maybe another UT to add is testing at period boundary (e.g. block N-1 and N gives different results at period boundary, but ok, we can live w/o)

PolkadotDom: I believe `first_step_occurs_at_start_plus_period` covers this

sigurpol: true!

sigurpol (approved):
LGTM (but my approval doesn't count if not for moral support 😅 )

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -287,36 +293,81 @@ impl multi_block::unsigned::miner::MinerConfig for Runtime {
...
+		// Start date of the curve is set two years prior, thus ensuring first step in March,
+		// 2026.
+		let two_years_before_march =
+			FixedU128::saturating_from_integer(Self::HARD_CAP_START - (2 * RC_YEARS));
+		let relay_block_fp = FixedU128::saturating_from_integer(relay_block_num);
+		let step_duration = FixedU128::saturating_from_integer(2 * RC_YEARS);
+
+		let two_year_rate = Self::BI_ANNUAL_RATE;
+
+		let ti_curve = SteppedCurve::try_new(
```
kianenigma:
```suggestion
		let Ok(ti_curve) = SteppedCurve::try_new(..) else { return 0 }
```

is clearer than a comment saying `// Default curve is zero curve, in case target < start.`.

kianenigma: Similarly, `last_step_size` and `evaluate` can use `Result<_>` and instead of any saturation, we simply mint nothing.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -332,6 +383,41 @@ impl EraPayout {
...
+		let relay_block_num =
+			<RelaychainDataProvider<Runtime> as BlockNumberProvider>::current_block_number();
+
+		let yearly_emission = if relay_block_num < Self::HARD_CAP_START {
+			Self::yearly_before_hard_cap()
+		} else {
+			Self::yearly_after_hard_cap(relay_block_num)
+		};
+
+		let era_emission = relative_era_len.saturating_mul_int(yearly_emission);
```
kianenigma:
```suggestion
		let era_emission = relative_era_len.saturating_mul_int(yearly_emission).max(Self::MAX_ERA_EMISSION);
```

As a final safety measure, let's add the following:

- We can calculate what is the rough ballpark daily inflation with the current fixed model.
- Hardcode one example of such value here.
- We know that if we have a prolonged era, for any reason, this emission amount becomes more. But as long as we don't have prolonged eras, emission will only go down.
- Therefore we multiply this value by, say, 7, meaning that up to 7 days of emission can be squeezed into one. This will be `Self::MAX_ERA_EMISSION`
- Again, we know that inflation is only ever going down in both current and future model, so this value will never get in the way.

Maybe I am too paranoid, but I just want to make sure that a silly math bug will never ever cause us to issue an unreasonable amount.

Note that we already have this in `pallet-staking-async`; even if an era takes 30 days, the `era_duration_millis` that is passed to this function is capped by `MaxEraDuration` ([here](https://github.com/paritytech/polkadot-sdk/blob/4a09391b10e515f3f5bfda278222107a6404ff1c/substrate/frame/staking-async/src/session_rotation.rs#L774-L795) -- currently 36h, but we plan to increase it as the system is in production for longer). This is more or less a second layer to this check, ensuring that regardless of any bug in: 

1. `era_duration_millis`
2. or the curve math here

The system is never inflating a crazy amount.

sigurpol: sounds a tiny but reasonable change, in the end two independent safeguards are better than one 😜 
@kianenigma , if we want to over-engineer it a bit... we could make it dynamic and use ErasValidatorReward from last N (7 probably good enough) eras to calculate the cap (avg or max multiplied for 150% or so to be on the safe side)

kianenigma: Yeah, something like last 7 days average multiplied by 1.5 sounds good to me. But since it is only going down, we can hardcode it for now and no need to keep track of the average onchain.

PolkadotDom: Have added a cap according to [AH Block 10469901](https://assethub-polkadot.subscan.io/event/10469901-6)

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
PolkadotDom: Updated the try_new code. Regarding the `Result<_>` change-- I would think it isn't necessary due to the math converging to 0 in all cases of saturation.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -1052,6 +1040,25 @@ mod tests {
...
+
+			// Simulate an era that lasted 100 years (anomalous).
+			let anomalous_duration = 36525 * MILLISECONDS_PER_DAY;
+			let (to_stakers, to_treasury) = EraPayout::era_payout(
+				123, // ignored
+				456, // ignored
+				anomalous_duration,
+			);
+
+			// Capped at MAX_ERA_EMISSION (2 * UNITS)
```
sigurpol: [nitpick] should we remove the (2 * UNITS ) and just leave Capped at MAX_ERA_EMISSION? The actual value would be around ~2.3M DOT but I believe MAX_ERA_EMISSION is expressive enough

PolkadotDom: Artifact from writing. Good catch

system-parachains/asset-hubs/asset-hub-polkadot/src/weights/pallet_staking_async.rs
```diff
@@ -290,11 +290,11 @@ impl<T: frame_system::Config> pallet_staking_async::WeightInfo for WeightInfo<T>
...
 		//  Estimated: `6248 + n * (2520 ±0)`
-		// Minimum execution time: 150_928_000 picoseconds.
-		Weight::from_parts(162_392_374, 0)
+		// Minimum execution time: 6_097_619_000 picoseconds.
+		Weight::from_parts(11_384_471_641, 0)
 			.saturating_add(Weight::from_parts(0, 6248))
-			// Standard Error: 39_584
-			.saturating_add(Weight::from_parts(5_149_263, 0).saturating_mul(n.into()))
+			// Standard Error: 10_755_226
+			.saturating_add(Weight::from_parts(39_604_011, 0).saturating_mul(n.into()))
```
ggwpez: Is the benchmark server not ref HW?

sigurpol: I am assuming we end up correctly on `runs-on: benchmark` as per https://github.com/polkadot-fellows/runtimes/blob/main/.github/workflows/cmd.yml#L223  so seems that one/some of the runners with benchmark label are virtualized and not ref HW (e.g ibp-gb)

PR #899, opened 2025-09-09
Fix randomness specification and warn about its unsafeness in its name

gui1117:
The implementation of `RelayChainOneEpochAgoWithoutBlockNumber` has some errors:
A malicious collator can force the fallback by omitting some data from the validation data. nowhere is it enforced to put those data. As long as a collator can freely force the fallback I think the type should reflect that this is unsafe.

I don't really have an opinion about using it or not, but I don't want to introduce this unsafe type and later we forgot and use it in polkadot.

Also I think we should return the valid time for this randomness by maybe adding some well known keys and make the omni-node forward them into the validation data. In its current form any correct usage of this type will be stuck because it says the randomness is only valid for action that happens before block 0.

@muharem @ggwpez

### DIFF
--- a/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
+++ b/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
@@ -1561,10 +1561,11 @@ parameter_types! {
 impl pallet_society::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type Currency = Balances;
-	type Randomness = system_parachains_common::randomness::RelayChainOneEpochAgoWithoutBlockNumber<
-		Runtime,
-		cumulus_primitives_core::relay_chain::BlockNumber,
-	>;
+	type Randomness =
+		system_parachains_common::randomness::RelayChainOneEpochAgoWithoutBlockNumberWarningUnsafe<
+			Runtime,
+			cumulus_primitives_core::relay_chain::BlockNumber,
+		>;
 	type GraceStrikes = ConstU32<10>;
 	type PeriodSpend = ConstU128<{ 500 * QUID }>;
 	type VotingPeriod = ConstU32<{ 5 * RC_DAYS }>;
--- a/system-parachains/common/src/randomness.rs
+++ b/system-parachains/common/src/randomness.rs
@@ -26,24 +26,29 @@ pub const LOG_TARGET: &str = "runtime::randomness";
 /// VRF output length for per-slot randomness.
 pub const VRF_RANDOMNESS_LENGTH: usize = 32;
 
-// The initial version of the `Randomness` implementation proposed by @gui1117
 /// Provides randomness from the Relay Chain VRF from one epoch ago, but does not include the block
 /// number indicating when this randomness was generated or became observable to chain observers.
 ///
+/// WARNING: A malicious collator can omit the `well_known_keys::ONE_EPOCH_AGO_RANDOMNESS` key and
+/// value in the validation data, causing this implementation to fall back to randomness from the
+/// current relay chain state root.
+///
 /// WARNING: This implementation does not return the block number associated with the randomness,
 /// because this information is not available in the validation data.
-pub struct RelayChainOneEpochAgoWithoutBlockNumber<T, BlockNumber>(PhantomData<(T, BlockNumber)>);
+pub struct RelayChainOneEpochAgoWithoutBlockNumberWarningUnsafe<T, BlockNumber>(
+	PhantomData<(T, BlockNumber)>,
+);
 
 impl<T, BlockNumber> Randomness<T::Hash, BlockNumber>
-	for RelayChainOneEpochAgoWithoutBlockNumber<T, BlockNumber>
+	for RelayChainOneEpochAgoWithoutBlockNumberWarningUnsafe<T, BlockNumber>
 where
 	T: cumulus_pallet_parachain_system::Config,
 	BlockNumber: From<u32>,
 {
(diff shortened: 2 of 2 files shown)


system-parachains/common/src/randomness.rs
```diff
@@ -81,17 +86,29 @@ where
 			.inspect_err(|e| {
 				log::error!(
 					target: LOG_TARGET,
-					"Failed to lookup `well_known_keys::ONE_EPOCH_AGO_RANDOMNESS` from trie: {e}"
+					"Failed to lookup `well_known_keys::ONE_EPOCH_AGO_RANDOMNESS` from the \
+					validation data. The node may have maliciously omitted it. Error: {e}"
```
gui1117, line 90: we could also panic here, to enforce the collator to construct correctly the validation data.

ggwpez: I did not know that the collator can just omit RC proof data?! But sounds like a valid reason to panic, yes.  
WDYT @skunert?

skunert: The keys listed here will be included by the default implementation: https://github.com/paritytech/polkadot-sdk/blob/6060499f9a807406e13449561b0fc603d9aaeedc/cumulus/client/parachain-inherent/src/lib.rs#L103-L103

But there is no specific enforcement of the shape of this proof. Some things are for sure mandatory, because the inherent in `parachain-system` tries to access them. If that fails, the block will not be build.

However it looks like `ONE_EPOCH_AGO_RANDOMNESS` is not being used by `parachain-system`, so you could still build blocks without it. Idk when exactly this codepath here is called, if its inside an extrinsics it will just get aborted right?

gui1117: The potential usage is inside `on_initialize` for society.

But panicking still sounds ok to me.

(In general I agree it would be ideal to move those operations in some tasks, so we could avoid bricking the STF on wrong implementation.)

gui1117:
I think the original PR this one is based on, has been changed, not sure, but I can't find it.

EDIT: it was merged on master and simply ignored by comments...

gui1117:
We should merge this PR. We should not let in the codebase a type `RelayChainOneEpochAgoWithoutBlockNumber` that let a collator skip it if they want to. This is unsafe.

ggwpez (approved):
I still think that we should panic to prevent a malicious author from exploiting this but its already an improvement to stop others from blindly using this.

system-parachains/common/src/randomness.rs
```diff
@@ -26,24 +26,29 @@ pub const LOG_TARGET: &str = "runtime::randomness";
 /// VRF output length for per-slot randomness.
 pub const VRF_RANDOMNESS_LENGTH: usize = 32;
 
-// The initial version of the `Randomness` implementation proposed by @gui1117
 /// Provides randomness from the Relay Chain VRF from one epoch ago, but does not include the block
 /// number indicating when this randomness was generated or became observable to chain observers.
 ///
+/// WARNING: A malicious collator can omit the `well_known_keys::ONE_EPOCH_AGO_RANDOMNESS` key and
+/// value in the validation data, causing this implementation to fall back to randomness from the
+/// current relay chain state root.
```
bkchr, line 34: I know that you talked about it, but we should really panic. If the key isn't present, something bad is going on and we should end the block.

gui1117: ok let's panic, I'll rework the PR

gui1117: wait but this seems not good, this is used in `on_initialize` so the `ValidationData` is maybe already removed (parachain system removes it in its own `on_initialize`, so depending on the ordering of pallets).

So potentially the previous block could be responsible for the panic in the next block (if the validation data isn't removed and the well known key wasn't included)

I think we should hook into something like `ConsensusHook` to fetch the well known key and update it. Or we should make those key directly written in parachain-system.

Personally I would just read those well-known keys in parachain-system and make them available. randomness is very useful.

skunert: Hmm couldn't you use [OnSystemEvent](https://github.com/paritytech/polkadot-sdk/blob/2c5feefbfe33b2488d64ce1035587cb150769b9a/cumulus/pallets/parachain-system/src/lib.rs#L1900) to get this and store it somewhere for your usage here to prevent the ordering issue?

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1561,10 +1561,11 @@ parameter_types! {
 impl pallet_society::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type Currency = Balances;
-	type Randomness = system_parachains_common::randomness::RelayChainOneEpochAgoWithoutBlockNumber<
-		Runtime,
-		cumulus_primitives_core::relay_chain::BlockNumber,
-	>;
+	type Randomness =
+		system_parachains_common::randomness::RelayChainOneEpochAgoWithoutBlockNumberWarningUnsafe<
```
muharem, line 1565: I remember that we indeed considered a situation where a malicious collator does not include the randomness. But specifically for the society pallet and its use cases it was the best what we could get at the moment.

muharem (approved):

muharem:
@gui1117 are we merging this as it is or we panic?

gui1117:
For individuality I wrote a new pallet that lookup the relay chain randomness: `indiv-pallet-relay-randomness`, it plugs into the consensus hook. In the future we can migrate to it.

bkchr:
@gui1117 but do we still want to merge this? If not, please close.

PR #877, opened 2025-09-03
[AHM] OpenGov for Kusama AH

karolk91:
Merging https://github.com/polkadot-fellows/runtimes/pull/776 to https://github.com/polkadot-fellows/runtimes/pull/856

## TODO

- [x] waiting for @muharem confirmation on several comments bellow
- [ ] @karolk91 fix either here or follow-up: https://github.com/polkadot-fellows/runtimes/pull/626/files#r2322701735

PR #856, opened 2025-08-18
THE ASSET HUB MIGRATION (Please Review)

ggwpez:
A high-level look of the migration is this:

Screenshot 2025-09-17 at 18 23 57

You can see the main components being the `pallet_rc_migrator` and `pallet_ah_migrator`. The RC migrator pallet takes data out of the RC storage and enqueues it for AH through XCM::Transact calls to the AH migrator pallet.  
Each pallet on the RC side has a migrator module that handles the removal of the old data and creation of the messages for AH. AH then processes these messages through the MQ pallet and acknowledges back to RC that they were processed (not shown in the graphic).

There is rate-limiting code in place to prevent it from overloading the DMP queue or Asset Hub.

## For Reviewers

The review comments are being addressed in https://github.com/polkadot-fellows/runtimes/pull/915.  
We need eyes on the Core part that is described below.

## Core STM

The main stage transitions look like this. The only manual part is the scheduling call by the fellowship that kicks off the procedure:

* `Pending` wait for Fellowship to call `schedule_migration(when)` -> Scheduled
* `Scheduled` waiting for scheduled block -> `WaitingForAH`
* `WaitingForAH` AH replies that it is ready -> `WarmUp`
* `WarmUp`  waiting for warm-up end -> `Starting`
* `Starting` doing some startup initialization -> `DataMigration`
* `DataMigration` MAIN PART **all pallet migrations finish** -> `CoolOff`
* `CoolOff` waiting for cool-off to end -> `SignalMigrationFinish`
* `SignalMigrationFinish` send signal to AH that we are done -> `MigrationDone`
* `MigrationDone` finished, unlock all Pallets.

### State Machine

The core State Machine lives in the `RcMigrator` pallet in its `on_initialize`:

https://github.com/polkadot-fellows/runtimes/blob/8ef1cc739d756b109dc93eaf3b9da00e3dcf1383/pallets/rc-migrator/src/lib.rs#L1338-L1342

On each block the Rc Migrator checks the current state and possibly progresses to the next one. It will only do a single state transition per block to reduce complexity of the code. Generally, there are three states per pallet (called [*stages*](https://github.com/polkadot-fellows/runtimes/blob/8ef1cc739d756b109dc93eaf3b9da00e3dcf1383/pallets/rc-migrator/src/lib.rs#L190-L389)  in the code). Some pallets have more then three, but it is generally still a multiple of three and always follows the pattern of `PalletInit`, `PalletOngoing { cursor }`, `PalletDone`:

[...]

PR #900, opened 2025-09-09
[integration-tests][ahm] Governance origins propagation

karolk91:
Aim of this PR is to cover scenarios which verify correct propagation of various governance related origins between Asset Hub and other chains

- [x] FellowshipAdmin
- [x] GeneralAdmin
- [x] StakingAdmin
- [x] Treasurer

relates to: https://github.com/polkadot-fellows/runtimes/pull/877#pullrequestreview-318375764,  https://github.com/polkadot-fellows/runtimes/pull/626#discussion_r2297664290 

## TODO

- [x] ~~@kianenigma had a very good point [here](https://github.com/polkadot-fellows/runtimes/pull/856#discussion_r2335683239), we have integration `authorize_upgrade` tests for `WhitelistedCall` track, but we need to also cover `Root` track, it means, that OpenGov on AH `Root` track can directly upgrade relaychain or system parachains (OriginKind::Superuser executes `pallet_xcm::send(root, xcm)`)(without need for Collectives whitelisting)~~
- [x] OpenGov on AH `Root` track root propagation over RC (I think you already prepared https://github.com/paritytech/polkadot-sdk/pull/9595)
- [x] (nice-to-have) do not use `force_batch` when not needed (see @kianenigma [comment](https://github.com/polkadot-fellows/runtimes/pull/856#discussion_r2335674696))


integration-tests/emulated/tests/governance/polkadot/src/open_gov_on_relay.rs
```diff
@@ -229,3 +229,67 @@ fn relaychain_can_authorize_upgrade_for_system_chains() {
 		)
 	});
 }
+
+#[test]
+fn relaychain_fellowsip_admin_can_induct_on_collectives() {
```
bkontur: @karolk91 I think AH scenario's are enough, because after AHM the RC way won't work

integration-tests/emulated/tests/governance/polkadot/src/open_gov_on_asset_hub.rs
```diff
@@ -299,3 +299,63 @@ fn assethub_can_authorize_upgrade_for_system_chains() {
...
+		AssetHubPolkadot,
+		CollectivesPolkadot,
+		pallet_core_fellowship::Instance1,
+	>(
+		AssetHubPolkadot::sibling_location_of(<CollectivesPolkadot as Parachain>::para_id()),
+		account_to_induct.clone(),
+		None,
+	);
+
+	let fellows_origin: AssetHubOrigin = FellowshipAdmin.into();
```
bkontur: either `fellowship_admin` or `fellowship_admin_origin`, because `fellows_origin` is used already in the code `Fellows` as Rank3+ member, so let's avoid any confusion:
```
        /// Origin commanded by Polkadot Fellows (3rd Dan fellows or greater).
		Fellows,
```


```suggestion
	let fellowship_admin: AssetHubOrigin = FellowshipAdmin.into();
```

integration-tests/emulated/tests/governance/polkadot/src/open_gov_on_relay.rs
karolk91: removed in https://github.com/polkadot-fellows/runtimes/pull/900/commits/b8a04c73a11cf6304c7f6eb9aab6e9e9cf07b546

integration-tests/emulated/tests/governance/polkadot/src/open_gov_on_asset_hub.rs
karolk91: This was some work in progress and I've actually wanted to also introduce "bad" scenarios - so I have changed this since then and its now `ok_origin` and `bad_origin`

integration-tests/emulated/helpers/src/lib.rs
```diff
@@ -450,3 +453,344 @@ macro_rules! test_parachain_is_trusted_teleporter {
...
+			min_rank,
+		}
+		.into();
+
+	build_xcm_send_call::<SourceChain, DestChain>(dest, fallback_max_weight, call, OriginKind::Xcm)
+}
+
+/// Builds a `pallet_xcm::send` call to promote Fellowship member,
+/// wrapped in an unpaid XCM `Transact` with `OriginKind::Xcm`.
+pub fn build_xcm_send_fellowship_promote_member<SourceChain, DestChain, Instance>(
```
bkontur, line 574: @karolk91 very nice tests so far, very nit, maybe if we don't reuse for both Kusama/Polkadot, it should just go directly to the kusama-governance tests

karolk91: I haven't yet verified which of these also apply to Westend. My intention is to upstream these to polkadot-sdk as part of common utils thats why I'm putting this already in some temporary common place

PR #901, opened 2025-09-09
[AHM] Align AH Proxy Types

ggwpez:
Align PAH and KAH proxy types with the relay.

You can review by opening the Relay chain proxy definition side-by-side. The intention is to have Asset Hub proxy types be 1:1 equivalent to the ones on the Relay.

### DIFF
--- a/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
+++ b/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
@@ -626,34 +626,96 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
 	fn filter(&self, c: &RuntimeCall) -> bool {
 		match self {
 			ProxyType::Any => true,
-			ProxyType::Auction | ProxyType::ParaRegistration => false, // Only for remote proxy
 			ProxyType::NonTransfer => matches!(
 				c,
-				RuntimeCall::System(_) |
-					RuntimeCall::ParachainSystem(_) |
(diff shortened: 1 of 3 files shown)


system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -775,13 +807,7 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
...
 			(ProxyType::Assets, ProxyType::AssetOwner) => true,
 			(ProxyType::Assets, ProxyType::AssetManager) => true,
-			(
-				ProxyType::NonTransfer,
-				ProxyType::Collator |
-				ProxyType::Governance |
-				ProxyType::Staking |
-				ProxyType::NominationPools,
-			) => true,
+			(ProxyType::NonTransfer, _) => true,
```
muharem, line 814: does this make the non transfer a superset for everything like any?

ggwpez: Yea, for everything except `Any`. It is like that in the Relay apparently. Do you think it should be different?

muharem: its mostly same, but there are disjoints like `NonTransfer` vs `AssetOwner`. may be its even the only

ggwpez: True. Will exclude asset stuff.

PR #887, opened 2025-09-06
[AHM] Fresh weights

bkontur:
**Actual diff:** https://weights.tasty.limo/compare?repo=runtimes&threshold=10&path_pattern=relay%2F*%2Fsrc%2Fweights%2F**%2F*.rs%2Csystem-parachains%2F**%2Fweights%2F**%2F*.rs&method=asymptotic&ignore_errors=true&unit=time&old=oty-dev-asset-hub-migration-2506&new=bko-on-oty-dev-asset-hub-migration-2506-weights

## Keep up-to-date AssetHubKusama with `oty-dev-asset-hub-migration-2506`
Based on the commit `4111ecff5610a4803e9526600bd2bae0edc2e5a7`

- [x] 'polkadot_runtime_common::claims'
- [x] 'pallet_nfts', 'frame_system_extensions'
- [x] 'pallet_scheduler'
- [x] 'pallet_referenda'
- [x] 'pallet_staking_async'
- [x] 'pallet_utility', 
- [x] 'pallet_transaction_payment', 
- [x] 'pallet_bounties', 
- [x] 'pallet_collator_selection', 
- [x] 'pallet_uniques', 
- [x] 'pallet_xcm', 
- [x] 'pallet_election_provider_multi_block_verifier', 
- [x] 'pallet_nft_fractionalization', 
- [x] 'pallet_timestamp', 
- [x] 'cumulus_pallet_xcmp_queue', 
- [x] 'pallet_xcm_benchmarks::generic', 
- [x] 'pallet_asset_conversion', 
- [x] 'pallet_ah_migrator', 
- [x] 'pallet_whitelist', 
- [x] 'pallet_migrations', 
- [x] 'pallet_session',
- [x] 'pallet_vesting', 
- [x] 'pallet_asset_conversion_tx_payment',
- [x] 'pallet_message_queue', 
- [x] 'pallet_asset_rate', 
- [x] 'pallet_ah_ops', 
- [x] 'pallet_society', 
- [x] 'pallet_treasury', 
- [x] 'pallet_remote_proxy', 
- [x] 'pallet_multisig', 
- [x] 'pallet_election_provider_multi_block',
- [x] 'pallet_assets',
- [x] 'pallet_child_bounties',
- [x] 'pallet_indices', 
- [x] 'pallet_bags_list', 
- [x] 'frame_system', 
- [x] 'cumulus_pallet_parachain_system', 
- [x] 'pallet_proxy', 
[...]

PR #902, opened 2025-09-09
[AHM] Unused benchmarks cleanup to get compile

bkontur:
Fresh rc-migrator weights do not compile here: https://github.com/polkadot-fellows/runtimes/pull/887

### DIFF
--- a/.github/scripts/cmd/cmd.py
+++ b/.github/scripts/cmd/cmd.py
@@ -90,15 +90,7 @@ if args.command == 'bench':
     # loop over remaining runtimes to collect available pallets
     for runtime in runtimesMatrix.values():
         print(f'-- compiling the runtime {runtime["name"]}')
-        features = "runtime-benchmarks"
-        # TODO: temporary hack for AHM - start...
-        if runtime["name"] in ["kusama", "asset-hub-kusama"]:
-            features += ",kusama-ahm"
-        if runtime["name"] in ["polkadot", "asset-hub-polkadot"]:
-            features += ",polkadot-ahm"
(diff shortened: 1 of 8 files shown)


.github/scripts/cmd/cmd.py
```diff
@@ -90,7 +90,15 @@
     # loop over remaining runtimes to collect available pallets
     for runtime in runtimesMatrix.values():
         print(f'-- compiling the runtime {runtime["name"]}')
-        os.system(f"cargo build -p {runtime['package']} --profile {profile} -q --features runtime-benchmarks")
+        features = "runtime-benchmarks"
+        # TODO: temporary hack for AHM - start...
+        if runtime["name"] in ["kusama", "asset-hub-kusama"]:
+            features += ",kusama-ahm"
```
bkontur: @ggwpez is this relevant fix for benchmarking? We could use runtime-matrix.json to provide additional features per runtime, but I think it is not that worth, because we will remove this after AHM, wdyt?

ggwpez: We should put it in the matrix to ensure that we build the runtimes correctly. We can manually verify that the release was build correctly before publishing it but just in case. (assuming it will not break the CI).

bkontur: > We should put it in the matrix to ensure that we build the runtimes correctly. We can manually verify that the release was build correctly before publishing it but just in case. (assuming it will not break the CI).

Ok, I revert this here and I will create separate PR for this

ggwpez (approved):

PR #903, opened 2025-09-10
[AHM] Ensure `kusama-ahm` and `polkadot-ahm` features

bkontur:
## TODO
- [x] test bench bot

### DIFF
--- a/pallets/ah-migrator/src/sovereign_account_translation.rs
+++ b/pallets/ah-migrator/src/sovereign_account_translation.rs
@@ -16,6 +16,7 @@
 
 //! Account translation maps for sovereign accounts and their derived accounts.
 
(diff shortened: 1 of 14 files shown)


pallets/ah-migrator/src/sovereign_account_translation.rs
```diff
@@ -16,6 +16,7 @@
 
 //! Account translation maps for sovereign accounts and their derived accounts.
 
+#[cfg(feature = "polkadot-ahm")]
```
pepoviola, line 19: @bkontur / @ggwpez this make integration test fails to compile for `kusama-ahm`, if you are agree I will add a pr to remove it. Thx!

PR #904, opened 2025-09-10
[AHM] Enable check-migrations jobs

bkontur:
TODO: just to see if anything is not working

### DIFF
--- a/.github/workflows/check-migrations.yml
+++ b/.github/workflows/check-migrations.yml
@@ -5,12 +5,9 @@ name: Check Migrations
 
 on:
   push:
-    branches: ["never"]
-  # Disabled on AHM PRs
-  #push:
-  #  branches: ["main", "release-*"]
-  #pull_request:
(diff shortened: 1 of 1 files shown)


.github/workflows/check-migrations.yml
```diff
@@ -5,12 +5,9 @@ name: Check Migrations
...
   push:
-    branches: ["never"]
-  # Disabled on AHM PRs
-  #push:
-  #  branches: ["main", "release-*"]
-  #pull_request:
-  #workflow_dispatch:
+    branches: ["main", "release-*"]
+  pull_request:
+  workflow_dispatch:
```
bkontur, line 10: @ggwpez Should/could we enable this now for AHM branch? Or maybe just to disable for AssetHubs or something?

ggwpez: I think we should keep the check enabled, given that we want to merge it into master soon.

PR #906, opened 2025-09-11
[AHM] Support migration of treasury asset acquisition calls.

muharem:
Support migration of treasury asset acquisition calls.

There are at least three known treasury spends that differ from regular ones (already ). These are significant spends intended to swap treasury DOT assets for USDT/USDC on Hydration (listed below). Such calls should be mapped either while they are under referendum in the referenda pallet, or once they are scheduled via the scheduler.

This PR adds mapping support for the calls from the listed referendas. Both active referenda containing such calls and scheduled calls will be mapped during the migration and function correctly on Asset Hub.

Additionally, this PR introduces an operational call that allows anyone to transfer assets from the pre-migration treasury account on Asset Hub to the post-migration treasury account once migration is complete. This is necessary because the Hydration swap calls send funds to the current (pre migration) Asset Hub treasury account, and we intentionally avoid decoding and mapping Hydration calls for simplicity.

### Known limitation:
The calls use a pure proxy account on Hydration controlled by the Relay Chain to store, swap, and access treasury funds. After migration, remaining executions would be commanded by Asset Hub, which is not yet registered as a proxy of the pure proxy. To resolve this, Asset Hub must be added as a proxy of the Hydration pure proxy.

Referendums with the spends:
- https://polkadot.subsquare.io/referenda/1729
- https://polkadot.subsquare.io/referenda/1728
- https://polkadot.subsquare.io/referenda/1501

### DIFF
--- a/pallets/ah-ops/src/lib.rs
+++ b/pallets/ah-ops/src/lib.rs
@@ -45,7 +45,8 @@ use cumulus_primitives_core::ParaId;
 use frame_support::{
 	pallet_prelude::*,
 	traits::{
-		fungible::{Inspect, InspectFreeze, Mutate, MutateFreeze, MutateHold, Unbalanced},
+		fungible::{InspectFreeze, Mutate, MutateFreeze, MutateHold, Unbalanced},
+		fungibles::{Inspect as FungiblesInspect, Mutate as FungiblesMutate},
 		tokens::Preservation,
 		Defensive, LockableCurrency, ReservableCurrency,
 	},
@@ -88,6 +89,10 @@ pub mod pallet {
 			+ Unbalanced<Self::AccountId>
 			+ ReservableCurrency<Self::AccountId, Balance = u128>
 			+ LockableCurrency<Self::AccountId, Balance = u128>;
+
+		/// Fungibles registry type.
+		type Fungibles: FungiblesInspect<Self::AccountId, Balance = u128>
+			+ FungiblesMutate<Self::AccountId, Balance = u128>;
 
 		/// Access the block number of the Relay Chain.
 		type RcBlockNumberProvider: BlockNumberProvider<BlockNumber = BlockNumberFor<Self>>;
@@ -316,6 +321,7 @@ pub mod pallet {
 		})]
 		pub fn transfer_to_post_migration_treasury(
 			origin: OriginFor<T>,
+			asset_id: Box<<T::Fungibles as FungiblesInspect<T::AccountId>>::AssetId>,
 		) -> DispatchResultWithPostInfo {
 			ensure_signed(origin)?;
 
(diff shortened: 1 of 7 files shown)


integration-tests/ahm/src/tests.rs
```diff
@@ -1617,3 +1617,48 @@ fn bifrost_addresses_are_in_translation_map() {
...
+				Ok(bounded) => Bounded::Inline(bounded),
+				Err(unbounded) => {
+					let len = unbounded.len() as u32;
+					Bounded::Lookup {
+						hash: pallet_preimage::Pallet::<AssetHub>::note(unbounded.into())
+							.expect("failed to note call"),
+						len,
+					}
+				},
+			};
```
ggwpez, line 1648: There should be some `pallet_preimage::Pallet::<AssetHub>::bound` convenience function to do this. Not sure how exactly it was called.

pallets/ah-ops/src/lib.rs
```diff
@@ -187,6 +198,10 @@ pub mod pallet {
 		NotSovereign,
 		/// Internal error, please bug report.
 		InternalError,
+		/// The Asset Hub migration is not completed.
+		MigrationNotCompleted,
+		/// The balance is zero.
```
ggwpez, line 208:
```suggestion
		/// The balance of the native token and all assets is zero.
```
?

PS: Nvm, we just care about native asset I guess.

pallets/ah-ops/src/lib.rs
```diff
@@ -289,6 +304,38 @@ pub mod pallet {
...
+
+			ensure!(T::MigrationCompletion::get(), Error::<T>::MigrationNotCompleted);
+
+			let pre_migration_account = T::TreasuryPreMigrationAccount::get();
+			let post_migration_account = T::TreasuryPostMigrationAccount::get();
+
+			let balance = <T as Config>::Currency::balance(&pre_migration_account);
+			ensure!(balance > 0, Error::<T>::ZeroBalance);
+
+			<T as Config>::Currency::transfer(
```
ggwpez: Why we are not transferring fungibles?

integration-tests/ahm/src/tests.rs
muharem: I can bound the RuntimeCall type, but not the bytes. if I do the bytes, it adds something in front on encoding

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/mod.rs
```diff
@@ -351,6 +399,102 @@ impl RcToAhCall {
...
+							target: LOG_TARGET,
+							"Failed to reanchor assets: {:?}",
+							err,
+						);
+					})?;
+
+				Ok(RuntimeCall::PolkadotXcm(
+					pallet_xcm::Call::<Runtime>::limited_reserve_transfer_assets {
+						dest: Box::new(dest.into()),
+						beneficiary,
```
ggwpez, line 492: Is the `beneficiary` relative to the destination or why does it not need re-anchoring?

muharem: right, beneficiary relative to the destination. I will leave the comment

PR #907, opened 2025-09-12
Few fixes to KAH Staking Configs + Preliminary PAH Configs

kianenigma:
Would be happy to approve the base branch after this.


system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
```diff
@@ -554,7 +550,7 @@ impl frame_support::traits::OnRuntimeUpgrade for InitiateStakingAsync {
 			// * Move to using `on_poll`
 			//
 			// After which this limit can be increased again.
-			pallet_staking_async::ValidatorCount::<Runtime>::put(2500);
+			pallet_staking_async::MaxValidatorsCount::<Runtime>::put(2500);
```
kianenigma: Man, this would have been a bad mistake 🙈 @Ank4n @sigurpol

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -280,18 +317,21 @@ impl pallet_staking_async::EraPayout<Balance> for EraPayout {
...
-	pub const MaxControllersInDeprecationBatch: u32 = 751;
+	pub const MaxControllersInDeprecationBatch: u32 = 512;
 	// alias for 16, which is the max nominations per nominator in the runtime.
 	pub const MaxNominations: u32 = <NposCompactSolution16 as frame_election_provider_support::NposSolution>::LIMIT as u32;
-	pub const MaxEraDuration: u64 = RelaySessionDuration::get() as u64 * RELAY_CHAIN_SLOT_DURATION_MILLIS as u64 * SessionsPerEra::get() as u64;
+	/// This is the upper bound on how much we are willing to inflate per era. We also emit a
+	/// warning event in case an era is longer than this amount.
+	///
+	/// Under normal conditions, this upper bound is never needed, and eras would be 24h each exactly. Yet, since this is the first deployment of pallet-staking-async, there might be misconfiguration, so we allow up to 12h more in each era.
+	pub const MaxEraDuration: u64 = 36 * (1000 * 60 * 60);
```
kianenigma: sanity check plz

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
```diff
@@ -514,9 +512,9 @@ impl frame_support::traits::OnRuntimeUpgrade for InitiateStakingAsync {
 		if Self::needs_init() {
 			use pallet_election_provider_multi_block::verifier::Verifier;
 			// set parity staking miner as the invulnerable submitter in `multi-block`.
-			// https://kusama.subscan.io/account/GtGGqmjQeRt7Q5ggrjmSHsEEfeXUMvPuF8mLun2ApaiotVr
+			// https://kusama.subscan.io/account/FyrGiYDGVxg5UUpN3qR5nxKGMxCe5Ddfkb3BXjxybG6j8gX
 			let acc = hex_literal::hex!(
-				"bea06e6ad606b2a80822a72aaae84a9a80bec27f1beef1880ad4970b72227601"
+				"96a6df31a112d610277c818fd9a8443d265fb5ab83cba47c5e89cff16cf9e011"
```
sigurpol: so you have double-checked with our devops friends that this is the one we are going to use, right?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
sigurpol: what is the rationale of having 3h extra on Kusama and 12h here? Just a reasonable buffer and no more than that?

kianenigma: It is 1/2 of the corresponding era. 

Actually I had a hard time deciding this, because `MaxEraDuration` is doing two things:

- Cap the inflation by this amount
- Emit an event

Ideally, I wanted to have two different values: 

* `expected_era_duration` + `small_epsilon` should emit the event that _something is wrong_
* `expected_era_duration` + `large_buffer` should be the cap for inflation.

So for example if we have a bug and an era is 3 days long, we still inflate in accordance to 3 days, but emit events.

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
kianenigma: yup, I checked again the ones in devops, and they have more funds + the old miner keys likely have more compromised seed accounts, as they have been in use for many years, so I reverted back to using the parity-controlled accounts

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: Sanity checked this. Here are two consecutive eras in Polkadot RC: 

```
{
  index: 1,927
  start: 1,757,518,578,001
}
{
  index: 1,928
  start: 1,757,604,978,001
}
```

The diff is 86400000. Which is exactly 24 * 1000 * 60 * 60. 

So our value here is fine.

relay/polkadot/src/lib.rs
```diff
@@ -676,7 +667,7 @@ impl pallet_bags_list::Config<VoterBagsListInstance> for Runtime {
 	#[cfg(feature = "runtime-benchmarks")]
 	type MaxAutoRebagPerBlock = ConstU32<5>;
 	#[cfg(not(feature = "runtime-benchmarks"))]
-	type MaxAutoRebagPerBlock = AutoRebagNumber;
+	type MaxAutoRebagPerBlock = ConstU32<0>;
```
ggwpez:
```suggestion
	type MaxAutoRebagPerBlock = ();
```
Nit

relay/polkadot/src/lib.rs
```diff
@@ -1559,10 +1550,11 @@ impl pallet_staking_async_ah_client::Config for Runtime {
 	type AdminOrigin = EnsureRoot<AccountId>;
 	type SessionInterface = Self;
 	type SendToAssetHub = StakingXcmToAssetHub;
-	type MinimumValidatorSetSize = ConstU32<400>; // TODO @kianenigma
+	// Polkadot RC currently has 600 validators. 500 minimum for now.
+	type MinimumValidatorSetSize = ConstU32<500>;
 	type UnixTime = Timestamp;
 	type PointsPerBlock = ConstU32<20>;
-	type MaxOffenceBatchSize = ConstU32<50>;
+	type MaxOffenceBatchSize = ConstU32<32>;
```
ggwpez, line 1557: Do we need to rebench for this or is it fine?

relay/polkadot/src/lib.rs
```diff
@@ -1630,12 +1622,12 @@ impl pallet_staking_async_ah_client::SendToAssetHub for StakingXcmToAssetHub {
...
-			xcm_config::XcmRouter,
-			AssetHubLocation,
-			pallet_staking_async_rc_client::SessionReport<AccountId>,
-			SessionReportToXcm,
-		>::split_then_send(session_report, Some(8));
+		// TODO: after https://github.com/paritytech/polkadot-sdk/pull/9619, use `XCMSender::send` and handle error
+		let message = SessionReportToXcm::convert(session_report);
+		let dest = AssetHubLocation::get();
+		let _ = xcm::prelude::send_xcm::<xcm_config::XcmRouter>(dest, message).inspect_err(|err| {
+			log::error!(target: "runtime::ah-client", "Failed to send relay session report: {:?}", err);
```
ggwpez, line 1629: Is this defensive?

relay/polkadot/src/lib.rs
```diff
@@ -1658,9 +1650,12 @@ impl pallet_staking_async_ah_client::SendToAssetHub for StakingXcmToAssetHub {
 				.into(),
 			},
 		]);
-		if let Err(err) = send_xcm::<xcm_config::XcmRouter>(AssetHubLocation::get(), message) {
-			log::error!(target: "runtime::ah-client", "Failed to send relay offence message: {:?}", err);
-		}
+		// TODO: after https://github.com/paritytech/polkadot-sdk/pull/9619, use `XCMSender::send` and handle error
+		let _ = send_xcm::<xcm_config::XcmRouter>(AssetHubLocation::get(), message).inspect_err(
+			|err| {
+				log::error!(target: "runtime::ah-client", "Failed to send relay offence message: {:?}", err);
```
ggwpez, line 1656: ditto

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
```diff
@@ -33,25 +33,32 @@ use sp_staking::SessionIndex;
...
-// * Some of the parameters are defined in `dynamic_params` block, prefixed by `params`
 // * The EPMB pallets only use local block times. They can one day be moved to use the relay-chain
 //   block, based on how the core-count and fast-blocks evolve, they might benefit from moving to
 //   relay-chain blocks. As of now, the duration of all phases are more about "POV" than "time", so
 //   AH deciding to use 2 or 3 cores is not an issue, as its PoV capacity also increases. The signed
 //   and unsigned phase are more about "time", yet the values used here are generous and should
 //   leave plenty of time for solution mining and submission.
 parameter_types! {
+	// alias for the ones backed by parameters-pallet.
+	pub MaxSignedSubmissions: u32 = dynamic_params::staking_election::MaxSignedSubmissions::get();
```
ggwpez: Hm do you think these `get`s will be re-evaluated every time or it will just use the default value?  
I dont think we tested this before?

But you can just do a `use dynamic_params::staking_election::MaxSignedSubmissions;` to have it handy.

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
```diff
@@ -238,15 +240,15 @@ impl multi_block::signed::CalculateBaseDeposit<Balance> for GeometricDeposit {
...
+	pub SignedDepositPerPage: Balance = system_para_deposit(1, NposCompactSolution24::max_encoded_len() as u32);
 	/// Bailing is rather disincentivized, as it can allow attackers to submit bad solutions, but
 	/// get away with it last minute. We only return 25% of the deposit in case someone bails. In
 	/// Polkadot, this value will be lower or simply zero.
 	pub BailoutGraceRatio: Perbill = Perbill::from_percent(25);
 	/// Invulnerable miners will pay this deposit only.
 	pub InvulnerableFixedDeposit: Balance = UNITS;
 	/// Being ejected is already paid for by the new submitter replacing you; no need to charge deposit.
-	pub EjectGraceRatio: Perbill = Perbill::from_percent(0);
+	pub EjectGraceRatio: Perbill = Perbill::from_percent(100);
```
ggwpez, line 249: This means no punishment for bad solutions, or?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -23,48 +23,37 @@ pub mod nom_pools;
...
-	/// 20 mins for signed phase.
-	pub storage SignedPhase: u32 = prod_or_fast!(
-		20 * MINUTES,
-		4 * MINUTES
+	// stuff aliased to `parameters` pallet.
+	pub SignedValidationPhase: u32 = prod_or_fast!(
+		Pages::get() * crate::dynamic_params::staking_election::MaxSignedSubmissions::get(),
+		Pages::get()
 	);
+	pub SignedPhase: u32 = crate::dynamic_params::staking_election::SignedPhase::get();
```
ggwpez: ditto

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -380,12 +420,12 @@ pub struct StakingXcmToRelayChain;
...
-			xcm_config::XcmRouter,
-			RelayLocation,
-			rc_client::ValidatorSetReport<Self::AccountId>,
-			ValidatorSetToXcm,
-		>::split_then_send(report, Some(8));
+		// TODO: after https://github.com/paritytech/polkadot-sdk/pull/9619, use `XCMSender::send`
+		let message = ValidatorSetToXcm::convert(report);
+		let dest = RelayLocation::get();
+		let _ = crate::send_xcm::<xcm_config::XcmRouter>(dest, message).inspect_err(|err| {
+			log::error!(target: "runtime::ah-client", "Failed to send validator set report: {:?}", err);
```
ggwpez, line 422: defensive maybe

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -416,3 +456,273 @@ where
...
+	fn needs_init() -> bool {
+		// A good proxy whether this pallet is initialized or not is that no invulnerable is set in
+		// `epmb::signed`. The rest are more fuzzy or are inaccessble.
+		multi_block::signed::Invulnerables::<Runtime>::get().is_empty()
+	}
+}
+
+impl frame_support::traits::OnRuntimeUpgrade for InitiateStakingAsync {
+	fn on_runtime_upgrade() -> Weight {
+		if Self::needs_init() {
```
ggwpez:
```suggestion
		if !Self::needs_init() {
				return;
```
Nit for less indentation.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -416,3 +456,273 @@ where
...
+			// set parity staking miner as the invulnerable submitter in `multi-block`.
+			// https://polkadot.subscan.io/account/16ciP5rjt4Yqivi1SWCGh7XsA8BDguV4tnTuyr937u2NME6h
+			let acc = hex_literal::hex!(
+				"f86a0e73c498fa0c135fae2e66da58346e777a6687cc7f7d234b0cb09c021232"
+			);
+			if let Ok(bounded) = BoundedVec::<AccountId, _>::try_from(vec![acc.into()]) {
+				multi_block::signed::Invulnerables::<Runtime>::put(bounded);
+			}
+
+			// set the minimum score for the election, as per the kusama RC state.
```
ggwpez:
```suggestion
			// Set the minimum score for the election, as per the Polkadot RC state.
```

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -416,3 +456,273 @@ where
...
+			// The last example, at block [27721215](https://polkadot.subscan.io/event/27721215-0) being:
+			//
+			// * minimal_stake: 10907549130714057 (1.28x the minimum)
+			// * sum_stake: 8028519336725652293 (2.44x the minimum)
+			// * sum_stake_squared: 108358993218278434700023844467997545 (0.4 the minimum, the lower
+			//   the better)
+			let minimum_score = sp_npos_elections::ElectionScore {
+				minimal_stake: 8474057820699941,
+				sum_stake: 3276970719352749444,
+				sum_stake_squared: 244059208045236715654727835467163294,
```
ggwpez: OT: These are basically our POW Parameters? But we have to fix lower bounds still it seems.

relay/polkadot/src/lib.rs
kianenigma: Yes, but a lot of things need re-benchmarking in P-AC and P-AH, but since this code path is never reached for now, no problem. 

I still hold the opinion that we should do a separate release for PAHM, in which we re-benchmark everything, among other fixes. What I have done in PAH for now is preminiary.

relay/polkadot/src/lib.rs
kianenigma: nope, this is the lack of https://github.com/paritytech/polkadot-sdk/pull/9619 not giving us the ability to handle the error. One more reason why we'd do a 1.8.1 for PAHM.

relay/polkadot/src/lib.rs
kianenigma: same

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
kianenigma: ah yeah, what I am doing here is kinda stupid code-wise, I am optimizing for being able to read all parameters in one place. I will change it, `use` makes sense.

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
kianenigma: nope, it just means if a better solution comes in, and the qeueue is full, and you are pushed out of the queue (all during submission, not validation), you get your deposit back. This is fine because the new submitter pays the fee for ejecting you, no point in penalizing the one being ejected.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: fixed.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: Not applicable, this can legit fail, and defensive in prod would do the same as the log error here.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: nice, fixed.

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: exactly, it is the worst validator set that we accept.

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
ggwpez: I see. Sounds good since it allows more people to play around with the miner :)

relay/polkadot/src/lib.rs
ggwpez: Yea separate release is a better idea, true.

PR #916, opened 2025-09-19
Fix pallet_election_provider_multi_block_signed::register_eject benchmark

sigurpol:
This PR reverts the workaround introduced in  #910 and bumps EPMB pallet version to `0.3.2` to include SDK fix https://github.com/paritytech/polkadot-sdk/pull/9778

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9002,9 +9002,9 @@ dependencies = [
 
 [[package]]
 name = "pallet-election-provider-multi-block"
-version = "0.3.1"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


bkontur (approved):
let's wait for test pipelines and merge

sigurpol:
ok, not mergeable see https://github.com/polkadot-fellows/runtimes/actions/runs/17853585308/job/50767471947?pr=916

The multiplier could have made the test pass but we need a more robust solution in the SDK.
I think (cc @kianenigma ) that to be on the very safe side, the funded_account()  fn should calculate deposit using the max queue lenght and not current size so we are sure to be fully covered.

sigurpol:
A better solution might come from https://github.com/paritytech/polkadot-sdk/pull/9787 🤞

Issue #800, opened 2025-07-04
assethub: Use memorydb read/write weights


alexggh:
Post changes and optimisation from https://github.com/paritytech/polkadot-sdk/issues/6131, after stable2506 is integrated in the runtime assethub-kusama and assethub-polkadot should be modified to use the read/write weights measured in validation context.

Examples of PR that changed westend: https://github.com/paritytech/polkadot-sdk/pull/8359.

- [x] Asset-hub-kusama 
- [x] Asset-hub-polkadot

bkchr:
We basically just use a different set of weights that should be closer to reality.

PR #918, opened 2025-09-22
asset-hub-kusama: use inmemorydb weights

alexggh:
Modify asset-hub-kusama to use the read/write costs computed in validate block on validator reference hardware, since these are weights that guarantee the blocks will pass validation and make it on the relay chain.

This were obtained with the benchmark created here: https://github.com/paritytech/polkadot-sdk/pull/8069.

Part of: https://github.com/polkadot-fellows/runtimes/issues/800

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,11 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
-- asset-hub-kusama: use inmemorydb weights([polkadot-fellows/runtimes/pull/918](https://github.com/polkadot-fellows/runtimes/pull/918))
-
-### Added
-
 - Add `cumulus-pallet-weight-reclaim` to all Kusama system parachains for storage weight reclaim functionality ([#941](https://github.com/polkadot-fellows/runtimes/pull/941))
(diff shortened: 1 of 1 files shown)


athei (approved):

bkchr (approved):

alexggh:
@ggwpez @muharem, now that kusama migration has happened, can I move forward with this PR ?

system-parachains/asset-hubs/asset-hub-kusama/src/weights/inmemorydb_weights.rs
```diff
@@ -0,0 +1,108 @@
+//! THIS FILE WAS AUTO-GENERATED USING THE SUBSTRATE BENCHMARK CLI VERSION 49.1.0
+//! DATE: 2025-09-10 (Y/M/D)
+//! HOSTNAME: `versi-developer-0`, CPU: `Intel(R) Xeon(R) CPU @ 2.60GHz`
```
bkontur, line 3: @alexggh we should probably also adjust cmd bench bot and re-trigger with other benchmarks/weights when doing releases, right?

alexggh: These read/write weights, does not affect the results of the computed  per-extrinsics weights, so no need to re-run them.

Yes, bench bot should be updated to also re-run these benchmarks every now and then and there is an issue for that https://github.com/polkadot-fellows/runtimes/issues/992.

PR #919, opened 2025-09-22
[AHM] Clippy and Review comments

ggwpez:
We love clippy 🙏


integration-tests/ahm/src/tests.rs
```diff
@@ -364,6 +365,8 @@ async fn find_translatable_accounts() {
 	write_account_translation_map(&sov_translations, &derived_translations);
 }
 
+#[cfg(all(feature = "polkadot-ahm", feature = "kusama-ahm"))]
```
pepoviola, line 368: I think we run the tests with only one of those.

ggwpez: Yea this is only for re-generating the sovereign account translation map, not really a test. So it should be fine.

pallets/rc-migrator/src/types.rs
```diff
@@ -397,7 +392,11 @@ impl<Status: MigrationStatus, Left: TypedGet, Right: Get<Left::Type>> Get<Left::
 	for LeftIfFinished<Status, Left, Right>
 {
 	fn get() -> Left::Type {
-		Status::is_ongoing().then(|| Left::get()).unwrap_or_else(|| Right::get())
+		if Status::is_ongoing() {
```
muharem:
```suggestion
		if Status::is_finished() {
```

pallets/ah-migrator/src/benchmarking.rs
```diff
@@ -550,10 +558,7 @@ pub mod benchmarks {
 	fn receive_conviction_voting_messages(n: Linear<1, 255>) {
 		let create_conviction_vote = |n: u8| -> RcConvictionVotingMessageOf<T> {
 			let class = <T as pallet_conviction_voting::Config>::Polls::classes()
-				.iter()
-				.cycle()
-				.skip(n as usize)
-				.next()
+				.get(n as usize)
```
muharem: would not this return none without `cycle`?

muharem: should we run benchmarks locally?

ggwpez: Yea true, need to use modulo first or revert.

ggwpez: reverted

PR #929, opened 2025-09-25
Simple multisig for cancelling and pausing migration

kianenigma:
- Uses unsigned txs (suggestion by @ggwpez) in case any of the signers don't have funds (or any other limitation imposed mid migration). 
- Allows anything that was guarded by `Manager` account to be dispatched, using a keyless account that is been artificially fed into `RuntimeOrigin::Signed` (suggestion by @muharem)
- Signing + submission can easily happen via PJS. Will test with a live dev chain to confirm.

Test: 

* first, edit the code so that the we use the testing keyring for live chain as well.
* Then: 

```
cargo br -p staging-kusama-runtime
chain-spec-builder create --para-id 777 --relay-chain kusama -t development  --runtime ./target/release/wbuild/staging-kusama-runtime/staging_kusama_runtime.wasm  named-preset development
polkadot --chain chain_spec.json --tmp --force-authoring --alice --unsafe-force-node-key-generation
```

### DIFF
--- a/pallets/rc-migrator/src/lib.rs
+++ b/pallets/rc-migrator/src/lib.rs
@@ -1305,20 +1305,21 @@ pub mod pallet {
 		#[pallet::weight({ Weight::from_parts(10_000_000, 1000) })]
 		pub fn vote_cancel(
 			origin: OriginFor<T>,
-			// payload is only interpreted as "aye"
-			payload: VotePayload,
+			payload: BailVote,
 			_sig: sp_core::sr25519::Signature,
 		) -> DispatchResult {
 			let _ = ensure_none(origin);
 
+			ensure!(CancelRound::<T>::get() == payload.round, "RoundStale");
 			let mut votes = CancelVotes::<T>::get();
-			ensure!(!votes.contains(&payload), "Duplicate");
-			votes.push(payload);
+			ensure!(!votes.contains(&payload.who), "Duplicate");
+			votes.push(payload.who);
 
 			if votes.len() >= 3 {
 				Self::transition(MigrationStage::Pending);
 				Self::deposit_event(Event::MigrationCancelled);
 				CancelVotes::<T>::kill();
+				CancelRound::<T>::mutate(|r| *r += 1);
 			} else {
 				CancelVotes::<T>::put(votes);
 			}
@@ -1330,23 +1331,50 @@ pub mod pallet {
 		#[pallet::weight({ Weight::from_parts(10_000_000, 1000) })]
 		pub fn vote_pause(
 			origin: OriginFor<T>,
-			// payload is only interpreted as "aye"
-			payload: VotePayload,
+			payload: BailVote,
 			_sig: sp_core::sr25519::Signature,
 		) -> DispatchResult {
 			let _ = ensure_none(origin);
 
+			ensure!(PauseRound::<T>::get() == payload.round, "RoundStale");
 			let mut votes = PauseVotes::<T>::get();
-			ensure!(!votes.contains(&payload), "Duplicate");
-			votes.push(payload);
+			ensure!(!votes.contains(&payload.who), "Duplicate");
+			votes.push(payload.who);
 
 			if votes.len() >= 3 {
 				let pause_stage = RcMigrationStage::<T>::get();
 				Self::transition(MigrationStage::MigrationPaused);
 				Self::deposit_event(Event::MigrationPaused { pause_stage });
 				PauseVotes::<T>::kill();
+				PauseRound::<T>::mutate(|r| *r += 1);
 			} else {
 				PauseVotes::<T>::put(votes);
+			}
+
+			Ok(())
+		}
+
+		#[pallet::call_index(15)]
+		#[pallet::weight({ Weight::from_parts(10_000_000, 1000) })]
+		pub fn vote_force_set_stage(
+			origin: OriginFor<T>,
+			payload: Box<ForceSetStageVote<T>>,
+			_sig: sp_core::sr25519::Signature,
+		) -> DispatchResult {
+			let _ = ensure_none(origin);
+
+			ensure!(ForceSetStageRound::<T>::get() == payload.round, "RoundStale");
+			let mut votes_for_stage = ForceSetStageVotes::<T>::get(&payload.stage);
+			ensure!(!votes_for_stage.contains(&payload.who), "Duplicate");
+			votes_for_stage.push(payload.who);
+
+			if votes_for_stage.len() >= 3 {
+				Self::transition(payload.stage);
+				// clear any votes, either in winning or losing stages.
+				let _ = ForceSetStageVotes::<T>::clear(u32::MAX, None);
+				ForceSetStageRound::<T>::mutate(|r| *r += 1);
+			} else {
+				ForceSetStageVotes::<T>::insert(payload.stage, votes_for_stage);
 			}
 
 			Ok(())
@@ -1363,29 +1391,66 @@ pub mod pallet {
 		TypeInfo,
 		sp_core::DecodeWithMemTracking,
 	)]
-	pub struct VotePayload {
+	pub struct BailVote {
 		pub(crate) who: sp_core::sr25519::Public,
+		pub(crate) round: u32,
 	}
 
-	impl VotePayload {
-		pub fn from(who: sp_core::sr25519::Public) -> Self {
-			Self { who }
+	impl BailVote {
+		pub fn from(who: sp_core::sr25519::Public, round: u32) -> Self {
+			Self { round, who }
 		}
-	}
 
-	impl core::convert::AsRef<sp_core::sr25519::Public> for VotePayload {
-		fn as_ref(&self) -> &sp_core::sr25519::Public {
-			&self.who
+		pub fn encode_with_bytes_wrapper(&self) -> Vec<u8> {
+			[b"<Bytes>", &*self.encode(), b"</Bytes>"].concat()
 		}
 	}
 
 	#[pallet::storage]
 	#[pallet::unbounded]
-	pub type CancelVotes<T: Config> = StorageValue<_, Vec<VotePayload>, ValueQuery>;
+	pub type CancelVotes<T: Config> = StorageValue<_, Vec<sp_core::sr25519::Public>, ValueQuery>;
+	#[pallet::storage]
+	pub type CancelRound<T: Config> = StorageValue<_, u32, ValueQuery>;
 
 	#[pallet::storage]
 	#[pallet::unbounded]
-	pub type PauseVotes<T: Config> = StorageValue<_, Vec<VotePayload>, ValueQuery>;
+	pub type PauseVotes<T: Config> = StorageValue<_, Vec<sp_core::sr25519::Public>, ValueQuery>;
+	#[pallet::storage]
+	pub type PauseRound<T: Config> = StorageValue<_, u32, ValueQuery>;
+
+	#[derive(
+		Encode,
+		Decode,
+		DebugNoBound,
+		CloneNoBound,
+		PartialEqNoBound,
+		EqNoBound,
+		TypeInfo,
+		sp_core::DecodeWithMemTracking,
+	)]
+	#[scale_info(skip_type_params(T))]
+	pub struct ForceSetStageVote<T: Config> {
+		pub(crate) round: u32,
+		pub(crate) who: sp_core::sr25519::Public,
+		pub(crate) stage: MigrationStageOf<T>,
+	}
+
+	impl<T: Config> ForceSetStageVote<T> {
+		pub fn new(who: sp_core::sr25519::Public, round: u32, stage: MigrationStageOf<T>) -> Self {
+			Self { who, round, stage }
+		}
+
+		pub fn encode_with_bytes_wrapper(&self) -> Vec<u8> {
+			[b"<Bytes>", &*self.encode(), b"</Bytes>"].concat()
+		}
+	}
+
+	#[pallet::storage]
+	pub type ForceSetStageRound<T: Config> = StorageValue<_, u32, ValueQuery>;
+	#[pallet::storage]
+	#[pallet::unbounded]
+	pub type ForceSetStageVotes<T: Config> =
+		StorageMap<_, Twox64Concat, MigrationStageOf<T>, Vec<sp_core::sr25519::Public>, ValueQuery>;
 
 	#[pallet::validate_unsigned]
 	impl<T: Config> ValidateUnsigned for Pallet<T> {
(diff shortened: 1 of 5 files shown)


relay/kusama/src/lib.rs
```diff
@@ -3148,6 +3179,156 @@ sp_api::impl_runtime_apis! {
...
+	use pallet_rc_migrator::VotePayload;
+	use sp_core::Pair;
+	use sp_io::TestExternalities;
+	use sp_runtime::traits::{Dispatchable, ValidateUnsigned};
+
+	use super::*;
+
+	#[test]
+	fn test_calls() {
+		let alice = sp_keyring::Sr25519Keyring::Alice.pair();
```
kianenigma: only issue is that while the signatures generated here work well, whatever the PJS sign page is doing is not compatible, even though itself will detect the signature as valid (with the right crypto type)

Image

The downside of this is that we cannot use this existing UI to sign. Members of the multisig must have their seed phrase at hand, and sign the message manually with a bit of Rust code.

ggwpez: We need to wrap the payload into `<bytes>data</bytes>` tags, PJS does it always even without telling you.  
I am trying it locally with that.

pallets/rc-migrator/src/lib.rs
```diff
@@ -1295,11 +1296,122 @@ pub mod pallet {
...
+		type Call = Call<T>;
+
+		fn validate_unsigned(_source: TransactionSource, call: &Self::Call) -> TransactionValidity {
+			use sp_runtime::traits::Verify;
+			match call {
+				// note: payload can be empty, but it is better to ask the signer to revel
+				// themselves, so we don't have to check against all of `MultisigMembers`.
+				Call::vote_cancel { payload, sig } | Call::vote_pause { payload, sig } =>
+					if T::MultisigMembers::get().contains(&payload.who) &&
+						sig.verify(&payload.encode()[..], &payload.who)
```
ggwpez: If we ever sign something then we should add a prefix and nonce to the pallet, otherwise someone can replay it.

kianenigma: True, I added a replay protection within a round where a same signer cannot be repeated, but it can still be reused. I will add a simple `rount` counter that is incremented, and each signature is only valid in one round.

sigurpol: or something like this maybe  instead of having only round counter?
```rust
pub struct VotePayload {
    pub who: sp_core::sr25519::Public,
    pub action: &'static str,    // "CANCEL" or "PAUSE"   - no cross-action replay
    pub round: u32,              //  Migration round number - no cross-migration replay
    // maybe having also nonce is an overkill?!?
    // pub nonce: u32,             //  Unique per vote
}
```

relay/kusama/src/lib.rs
```diff
@@ -1908,6 +1908,41 @@ const AH_MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
...
+			"HnMAUz7r2G8G3hB27SYNyit5aJmh2a5P4eMdDtACtMFDbam", // Eve
+		]
+	} else {
+		vec![
+			"FFFF3gBSSDFSvK2HBq4qgLH75DHqXWPHeCnR1BSksAMacBs", // basti
+			"FcxNWVy5RESDsErjwyZmPCW6Z8Y3fbfLzmou34YZTrbcraL", // gav
+			"EGVQCe73TpFyAZx5uKfE1222XfkT3BSKozjgcqzLBnc5eYo", // sahwn
+			"HL8bEp8YicBdrUmJocCAWVLKUaR2dd1y6jnD934pbre3un1", // kian/ksm
+			"G5MVrgFmBaYei8N6t6DnDrb8JE53wKDkazLv5f46wVpi14y", // alex
+			"Ea6jhP5gF4r7NqhkEoAXJDgSgYpNQNaTYU6gPsrEGfctaKR", // oliver
```
ggwpez:
```suggestion
			"F2jgWXy7X8GQ2ykf1UGrsCXRERZvjEcd2aDf39fMf3BWVy6", // oliver
```

My hot account is this one since i always vote through a Proxy on fellowship stuff: https://kusama.subscan.io/account/F2jgWXy7X8GQ2ykf1UGrsCXRERZvjEcd2aDf39fMf3BWVy6

pallets/rc-migrator/src/lib.rs
```diff
@@ -1295,11 +1307,120 @@ pub mod pallet {
...
+			ensure!(ManagerMultisigRound::<T>::get() == payload.round, "RoundStale");
+			let mut votes_for_call = ManagerMultisigs::<T>::get(&payload.call);
+			ensure!(!votes_for_call.contains(&payload.who), "Duplicate");
+			votes_for_call.push(payload.who.clone());
+
+			if votes_for_call.len() >= T::MultisigThreshold::get() as usize {
+				let origin: <T as Config>::RuntimeOrigin =
+					frame_system::RawOrigin::Signed(Self::manager_multisig_id()).into();
+				let call = payload.call.clone();
+				let res = call.dispatch_bypass_filter(origin);
```
ggwpez: Not sure if we actually need to bypass the filter? The RcMigrator pallet should not be disabled

pallets/rc-migrator/src/lib.rs
```diff
@@ -1295,11 +1307,120 @@ pub mod pallet {
...
+				ManagerMultisigs::<T>::insert(payload.call, votes_for_call);
+			}
+
+			Ok(())
+		}
+	}
+
+	impl<T: Config> Pallet<T> {
+		fn manager_multisig_id() -> T::AccountId {
+			let pallet_id = PalletId(*b"rcmigmts");
```
ggwpez, line 1363: So we use this instead of a pallet origin? okay

pallets/rc-migrator/src/lib.rs
```diff
@@ -1295,11 +1307,120 @@ pub mod pallet {
...
+			[b"<Bytes>", &*self.encode(), b"</Bytes>"].concat()
+		}
+	}
+
+	#[pallet::storage]
+	#[pallet::unbounded]
+	pub type ManagerMultisigs<T: Config> = StorageMap<
+		_,
+		Twox64Concat,
+		<T as Config>::RuntimeCall,
```
ggwpez, line 1404: I thought it could be just a hash of the call, but i guess they should be rather small? Otherwise you will have very long keys with `Twox64Concat`

pallets/rc-migrator/src/lib.rs
```diff
@@ -1295,11 +1307,120 @@ pub mod pallet {
...
+	pub type ManagerMultisigRound<T: Config> = StorageValue<_, u32, ValueQuery>;
+
+	#[pallet::validate_unsigned]
+	impl<T: Config> ValidateUnsigned for Pallet<T> {
+		type Call = Call<T>;
+
+		fn validate_unsigned(_source: TransactionSource, call: &Self::Call) -> TransactionValidity {
+			use sp_runtime::traits::Verify;
+			match call {
+				Call::vote_manager_multisig { payload, sig } => {
```
ggwpez: Nit: let .. else for less indentation

pallets/rc-migrator/src/lib.rs
kianenigma: Is it particularly bad to have a map key to be too long? frame storage will hash it under the hood anyways. I can hash it, it just degrades the debug-ability a bit I'd say.

pallets/rc-migrator/src/lib.rs
```diff
@@ -1295,11 +1307,120 @@ pub mod pallet {
...
+		pub fn new(
+			who: sp_core::sr25519::Public,
+			call: <T as Config>::RuntimeCall,
+			round: u32,
+		) -> Self {
+			Self { who, call, round }
+		}
+
+		pub fn encode_with_bytes_wrapper(&self) -> Vec<u8> {
+			[b"<Bytes>", &*self.encode(), b"</Bytes>"].concat()
```
ggwpez:
```suggestion
			(b"<Bytes>", self, b"</Bytes>").encode()
```
Sorry for the nit comments 🙈

pallets/rc-migrator/src/lib.rs
kianenigma: you're right, this can be normal dispatch.

pallets/rc-migrator/src/lib.rs
```diff
@@ -1295,11 +1307,128 @@ pub mod pallet {
...
+					frame_system::RawOrigin::Signed(Self::manager_multisig_id()).into();
+				let call = payload.call.clone();
+				let res = call.dispatch(origin);
+				let _ = ManagerMultisigs::<T>::clear(u32::MAX, None);
+				Self::deposit_event(Event::ManagerMultisigDispatched {
+					res: res.map(|_| ()).map_err(|e| e.error),
+				});
+				ManagerMultisigRound::<T>::mutate(|r| *r += 1);
+			} else {
+				ManagerMultisigs::<T>::insert(payload.call, votes_for_call);
```
ggwpez, line 1354: One event here would also help with debugging, i am trying it now with papi

pallets/rc-migrator/src/lib.rs
```diff
@@ -1295,11 +1307,128 @@ pub mod pallet {
...
+			let mut votes_for_call = ManagerMultisigs::<T>::get(&payload.call);
+			ensure!(!votes_for_call.contains(&payload.who), "Duplicate");
+			votes_for_call.push(payload.who);
+
+			if votes_for_call.len() >= T::MultisigThreshold::get() as usize {
+				let origin: <T as Config>::RuntimeOrigin =
+					frame_system::RawOrigin::Signed(Self::manager_multisig_id()).into();
+				let call = payload.call.clone();
+				let res = call.dispatch(origin);
+				let _ = ManagerMultisigs::<T>::clear(u32::MAX, None);
```
muharem, line 1345: can you explain this? should no be just ..remove(&payload.call) ?

relay/kusama/src/lib.rs
```diff
@@ -1908,6 +1908,38 @@ const AH_MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
...
+			"12HWjfYxi7xt7EvpTxUis7JoNWF7YCqa19JXmuiwizfwJZY2", // muharem
+			"121dd6J26VUnBZ8BqLGjANWkEAXSb9mWq1SB7LsS9QNTGFvz", // adrian
+			"12pRzYaysQz6Tr1e78sRmu9FGB8gu8yTek9x6xwVFFAwXTM8", // robK
+			"FcJnhk4i1bfuN9E2B6yMnL8h97ogtuL7e4ZpqnYgvj9moQy", // donal
+		]
+	};
+
+	addresses
+		.into_iter()
+		.filter_map(|ss| sp_runtime::AccountId32::from_ss58check(ss).ok())
```
muharem, line 1946: I did not know we can convert ss58 in runtime

pallets/rc-migrator/src/lib.rs
ggwpez: I think of the Concat it does concat the full encode call again. Should not break stuff, just unusual.

pallets/rc-migrator/src/lib.rs
muharem: there can be only one multisig call registered?

kianenigma: Yeah my assumption for simplicity was that there could only be one "competition" at the time, as soon as something has enough signs to dispatch, we erase everything, and bump the round number, making any existing signature invalid. 

I don't have an exact scenario in mind, but I worry if we don't do this, we won't be as resilient against replay. 

Note that this deletion is unbounded and can be dangerous, but since only fellowship members can add calls, it is fine.

relay/kusama/src/lib.rs
kianenigma: yeah, and I should actually add a sanity check to the tests that ensures all of these `.ok()`s work. I tested it manually and it is fine though.

pallets/rc-migrator/src/lib.rs
```diff
@@ -621,6 +626,12 @@ pub mod pallet {
 		///
 		/// This configuration can be overridden by a storage item [`AhUmpQueuePriorityConfig`].
 		type AhUmpQueuePriorityPattern: Get<(BlockNumberFor<Self>, BlockNumberFor<Self>)>;
+
+		/// Members of an multisig that can be submit unsigned txs and act as the manager.
```
joepetrowski:
```suggestion
		/// Members of a multisig that can be submit unsigned txs and act as the manager.
```

pallets/rc-migrator/src/lib.rs
```diff
@@ -1295,11 +1307,128 @@ pub mod pallet {
 			);
 
 			Self::transition(MigrationStage::Pending);
-
 			Self::deposit_event(Event::MigrationCancelled);
 
 			Ok(Pays::No.into())
 		}
+
+		/// Vote on on behalf of any of the members in `MultisigMembers`.
```
joepetrowski:
```suggestion
		/// Vote on behalf of any of the members in `MultisigMembers`.
```

joepetrowski (approved):

joepetrowski:
bot merge

pallets/rc-migrator/src/lib.rs
```diff
@@ -1055,7 +1070,7 @@ pub mod pallet {
 		#[pallet::call_index(4)]
 		#[pallet::weight(T::RcWeightInfo::resend_xcm())]
 		pub fn resend_xcm(origin: OriginFor<T>, query_id: u64) -> DispatchResultWithPostInfo {
-			Self::ensure_admin_or_manager(origin)?;
+			ensure_root(origin)?;
```
gui1117, line 1073: I don't know what is this change, but I assume it is a fix.

relay/kusama/src/lib.rs
```diff
@@ -1908,6 +1908,47 @@ const AH_MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
...
```
gui1117, line 1946: you could also have done `defensive_ok` or some similar function that panic in testing.

relay/kusama/src/lib.rs
```diff
@@ -1908,6 +1908,47 @@ const AH_MAXIMUM_BLOCK_WEIGHT: Weight = Weight::from_parts(
...
+			"FFFF3gBSSDFSvK2HBq4qgLH75DHqXWPHeCnR1BSksAMacBs", // basti
+			"FcxNWVy5RESDsErjwyZmPCW6Z8Y3fbfLzmou34YZTrbcraL", // gav
+			"HL8bEp8YicBdrUmJocCAWVLKUaR2dd1y6jnD934pbre3un1", // kian
+			"F2jgWXy7X8GQ2ykf1UGrsCXRERZvjEcd2aDf39fMf3BWVy6", // oliver
+			"GcDZZCVPwkPqoWxx8vfLb4Yfpz9yQ1f4XEyqngSH8ygsL9p", // joe
+			"12HWjfYxi7xt7EvpTxUis7JoNWF7YCqa19JXmuiwizfwJZY2", // muharem
+			"121dd6J26VUnBZ8BqLGjANWkEAXSb9mWq1SB7LsS9QNTGFvz", // adrian
+			"12pRzYaysQz6Tr1e78sRmu9FGB8gu8yTek9x6xwVFFAwXTM8", // robK
+			"FcJnhk4i1bfuN9E2B6yMnL8h97ogtuL7e4ZpqnYgvj9moQy", // donal
+		]
```
gui1117, line 1941: We know that those key are public key from sr25519, isn't it? because ed25519 and ecdsa are also valid signature for transaction but not for this multisig module.

pallets/rc-migrator/src/lib.rs
gui1117: In the trie, long keys are stored inside the nodes, but it is true that I don't know the performance, and I don't know if there is some implicit max length, I am not aware of any though.

pallets/rc-migrator/src/lib.rs
```diff
@@ -1295,11 +1309,131 @@ pub mod pallet {
...
+		///
+		/// Unsigned extrinsic, requiring the `payload` to be signed.
+		///
+		/// Upon each call, a new entry is created in `ManagerMultisigs` map the `payload.call` to
+		/// be dispatched. Once `MultisigThreshold` is reached, the entire map is deleted, and we
+		/// move on to the next round.
+		///
+		/// The round system ensures that signatures from older round cannot be reused.
+		#[pallet::call_index(13)]
+		#[pallet::weight({ Weight::from_parts(10_000_000, 1000) })]
```
gui1117, line 1327: The weight should include the weight of the call in case it gets dispatch no? It is only used on the relay chain AFAICT so it is fine anyway to go beyond the block weight limits AFAIK

muharem: fine in this case, bc its allowed only for that multisig

relay/kusama/src/lib.rs
ggwpez: Oh damn, yes need to check this.

acatangiu: how many of this list have we confirmed to be using a "hot wallet" or something that can sign arbitrary payloads in PJS?

should we remove the unconfirmed ones?

ggwpez: I pinged everyone in the MR and am currently changing it to support MultiSignature

ggwpez: https://github.com/polkadot-fellows/runtimes/pull/936

PR #934, opened 2025-09-26
[AHM] Schedule migration on runtime upgrade

ggwpez:
Schedule the AHM on runtime upgrade so we can avoid doing another fellowship call.  
Exact values will be set in the subsequent runtime release MR.

Changes:
- Factor scheduling logic out into `do_schedule_migration`
- Add `KickoffAhm` to Kusama RC runtime
- Put a placeholder for the actual scheduling values that we can fill in later

### DIFF
--- a/relay/kusama/src/lib.rs
+++ b/relay/kusama/src/lib.rs
@@ -2194,20 +2194,20 @@ pub mod migrations {
 	use pallet_rc_migrator::{MigrationStage, MigrationStartBlock, RcMigrationStage};
 
 	/// Unreleased migrations. Add new ones here:
-	pub type Unreleased = (KickoffAhm<Runtime>,);
+	pub type Unreleased = (KickOffAhm<Runtime>,);
 
 	/// Migrations/checks that do not need to be versioned and can run on every update.
 	pub type Permanent = pallet_xcm::migration::MigrateToLatestXcmVersion<Runtime>;
 
-	/// Kickoff the Asset Hub Migration.
-	pub struct KickoffAhm<T>(pub core::marker::PhantomData<T>);
-	impl<T: pallet_rc_migrator::Config> OnRuntimeUpgrade for KickoffAhm<T> {
+	/// Kick off the Asset Hub Migration.
+	pub struct KickOffAhm<T>(pub core::marker::PhantomData<T>);
+	impl<T: pallet_rc_migrator::Config> OnRuntimeUpgrade for KickOffAhm<T> {
 		fn on_runtime_upgrade() -> Weight {
 			if MigrationStartBlock::<T>::exists() ||
 				RcMigrationStage::<T>::get() != MigrationStage::Pending
 			{
 				// Already started or scheduled
-				log::info!("KickoffAhm: Asset Hub Migration already started or scheduled");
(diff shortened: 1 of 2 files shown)


sigurpol (approved):

relay/kusama/src/lib.rs
```diff
@@ -2190,12 +2190,43 @@ pub type Migrations = (migrations::Unreleased, migrations::Permanent);
...
 	/// Unreleased migrations. Add new ones here:
-	pub type Unreleased = ();
+	pub type Unreleased = (KickoffAhm<Runtime>,);
 
 	/// Migrations/checks that do not need to be versioned and can run on every update.
 	pub type Permanent = pallet_xcm::migration::MigrateToLatestXcmVersion<Runtime>;
+
+	/// Kickoff the Asset Hub Migration.
+	pub struct KickoffAhm<T>(pub core::marker::PhantomData<T>);
+	impl<T: pallet_rc_migrator::Config> OnRuntimeUpgrade for KickoffAhm<T> {
```
joepetrowski: Sorry 🤓 
```suggestion
	pub type Unreleased = (KickOffAhm<Runtime>,);

	/// Migrations/checks that do not need to be versioned and can run on every update.
	pub type Permanent = pallet_xcm::migration::MigrateToLatestXcmVersion<Runtime>;

	/// Kick off the Asset Hub Migration.
	pub struct KickOffAhm<T>(pub core::marker::PhantomData<T>);
	impl<T: pallet_rc_migrator::Config> OnRuntimeUpgrade for KickOffAhm<T> {
```

relay/kusama/src/lib.rs
```diff
@@ -2190,12 +2190,43 @@ pub type Migrations = (migrations::Unreleased, migrations::Permanent);
...
+		fn on_runtime_upgrade() -> Weight {
+			if MigrationStartBlock::<T>::exists() ||
+				RcMigrationStage::<T>::get() != MigrationStage::Pending
+			{
+				// Already started or scheduled
+				log::info!("KickoffAhm: Asset Hub Migration already started or scheduled");
```
joepetrowski:
```suggestion
				log::info!("KickOffAhm: Asset Hub Migration already started or scheduled");
```

relay/kusama/src/lib.rs
```diff
@@ -2190,12 +2190,43 @@ pub type Migrations = (migrations::Unreleased, migrations::Permanent);
...
+				DispatchTime::After(/* START */),
+				DispatchTime::After(/* WARM UP */),
+				DispatchTime::After(/* COOL OFF */),
+				Default::default(),
+			);
+			if let Err(e) = result {
+				log::error!("KickoffAhm: Failed to schedule Asset Hub Migration: {:?}", e);
+			} else {
+				log::info!("KickoffAhm: Scheduled Asset Hub Migration");
+			}*/
```
joepetrowski, line 2225:
```suggestion
			if let Err(e) = result {
				log::error!("KickOffAhm: Failed to schedule Asset Hub Migration: {:?}", e);
			} else {
				log::info!("KickOffAhm: Scheduled Asset Hub Migration");
			}*/
```

PR #939, opened 2025-09-30
Don't migrate staking era forcing info to AH

kianenigma:
Doing it quick and dirty way here, just ignoring it on the recipient, to minimize code change.


ggwpez (approved):
I think @franciscoaguirre asked if the default value is sane on the AH side.

joepetrowski (approved):

sigurpol (approved):

Issue #494, opened 2024-10-24
Enable PoV reclaim in all system chains


kianenigma:
as described in https://paritytech.github.io/polkadot-sdk/master/polkadot_sdk_docs/guides/enable_pov_reclaim/index.html

ggwpez:
@skunert do you think we can go ahead with this, or is something missing?

skunert:
At this time I think our proof weight story is not as accurate as we would like it to be.

- https://github.com/paritytech/polkadot-sdk/issues/5229: This will be mitigated by having a proper base-extrinsic weight benchmarked, which will be possible after https://github.com/paritytech/polkadot-sdk/pull/5891 and https://github.com/paritytech/polkadot-sdk/pull/6040
- Then there is a more recent discovery here: https://github.com/paritytech/polkadot-sdk/issues/6020. This one is not really reclaim specific and also needs some changes to our benchmarking code. I think @bkchr is working on this (at least he assigned himself).

These issues lead to inaccuracies but we have safeguards in place to prevent overshooting the limits:
1. Collators are still using pov_limit / 2 as the client-side size limit
2. https://github.com/paritytech/polkadot-sdk/issues/5281

However, these are workarounds while we have the proper fixes in progress. I expect the situation to get a lot better in the next releases, so if we can we should hold off a bit longer.

kianenigma:
Thanks for the update @skunert

This is not super time critical, but we probably want it in place before the AHM is done. 

@seadanda short term, I highly suggest using PoV reclaim + its workarounds in `ah-next` runtime that you are working on, so that we can more accurately experiment with the throughput empirically.

seadanda:
@skunert how's the progress on this? Ideally we'd enable PoV reclaim on AH before we deploy AHM, especially for staking on a parachain. Having a quick read around it looks like the your first point can be mitigated only after the release of stable2503, is this correct?

If we're happy with the current workarounds + safeguards I'd suggest sticking with that and adding PoV reclaim to AH-kusama before the migration on Kusama (release ~end of April), and trying to benchmark the base extrinsic weight properly before the AHM on Polkadot (Q3). Does that sound reasonable?

skunert:
[...]

PR #941, opened 2025-10-01
Add storage weight reclaim to Kusama system chains

skunert:
This PR adds weight reclaim to Kusama system chains. Part of #494.

Originally we were hoping for quick fixing of https://github.com/paritytech/polkadot-sdk/issues/6020, but it has proven to be quite tricky. The risk of overshooting in reality is low though, since we still have configurable training wheels in place (PoV size capped at 80% by default).


skunert:
/cmd bench --runtime asset-hub-kusama bridge-hub-kusama coretime-kusama --pallet cumulus_pallet_weight_reclaim

bkchr (approved):

system-parachains/coretime/coretime-kusama/src/weights/cumulus_pallet_weight_reclaim.rs
```diff
@@ -0,0 +1,67 @@
...
+	/// Proof: `System::AllExtrinsicsLen` (`max_values`: Some(1), `max_size`: Some(4), added: 499, mode: `MaxEncodedLen`)
+	fn storage_weight_reclaim() -> Weight {
+		// Proof Size summary in bytes:
+		//  Measured:  `24`
+		//  Estimated: `1533`
+		// Minimum execution time: 13_270_000 picoseconds.
+		Weight::from_parts(14_210_000, 0)
+			.saturating_add(Weight::from_parts(0, 1533))
+			.saturating_add(T::DbWeight::get().reads(3))
+			.saturating_add(T::DbWeight::get().writes(2))
```
acatangiu: why is it reading and writing state for coretime, but not for AH and BH?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Added
 
+- Add `cumulus-pallet-weight-reclaim` to all Kusama system parachains for storage weight reclaim functionality
```
acatangiu: why not add it to Encointer too?

skunert: Good catch! Done.

system-parachains/coretime/coretime-kusama/src/weights/cumulus_pallet_weight_reclaim.rs
skunert: Was looking into this, the problem was a misconfig in coretime. These storage items are not supposed to be benched here, because they are touched in every block and therefore part of the whitelisted keys in frame_system. But coretime was for some reason (probably oversight) not using the whitelisted keys from the pallets. Fixed in https://github.com/polkadot-fellows/runtimes/pull/941/commits/66025c133ed866821d9bee45d17af46fc55fb92b

CHANGELOG.md
```diff
@@ -60,6 +60,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
```
bkchr: Needs to be moved.

lexnv (approved):

PR #945, opened 2025-10-07
[AHM]: manually fix proxy tests for whales

sigurpol:
The whale watching integration test is fragile and requires adjustment e.g. each time a whale creates or removes a new proxy. This is one of the cases.


sigurpol:
> Yes, I will disable the Kusama tests.

Make sense - dropping this one then. Ty

PR #964, opened 2025-10-13
Run all jobs on the self hosted machines

bkchr:
For benchmarking we use a special machine that reports the `benchmark` tag.


.github/workflows/changelog.yml
```diff
@@ -19,7 +19,7 @@ jobs:
   # Job required by "confirmChangelogChecksPassed"
   verify-changelog-updated:
     name: Verify that Changelog is Updated
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
```
xlc, line 22: do we actually want to use self-hosted runner for lightweight tasks?

bkchr: Yes. We run out of github actions time already.

.github/workflows/test.yml
```diff
@@ -319,14 +321,30 @@ jobs:
...
+        run: |
+          mkdir -p ./logs
+          for pod in $(docker ps --format '{{.Names}}' --filter name="zombie-");do
+            docker logs $pod > "./logs/${pod}.log" 2>&1;
+            docker stop $pod;
+            docker rm $pod;
+          done;
+
+      - name: Upload_logs
+        uses: actions/upload-artifact@v4
```
bkchr:
```suggestion
      - name: Collect_logs
        if: success() || failure()
        run: |
          mkdir -p ./logs
          for pod in $(docker ps --format '{{.Names}}' --filter name="zombie-");do
            docker logs $pod > "./logs/${pod}.log" 2>&1;
            docker stop $pod;
            docker rm $pod;
          done;

      - name: Upload_logs
        if: success() || failure()
        uses: actions/upload-artifact@v4
```

.github/workflows/test.yml
```diff
@@ -274,7 +276,7 @@ jobs:
 
   # Job required by "confirmTestPassed"
   zombienet-smoke:
-    runs-on: ubuntu-latest
+    runs-on: [self-hosted, benchmark]
```
bkchr:
```suggestion
    runs-on: ibp-ch
```

PR #966, opened 2025-10-13
Adds Fulu mainnet fork version

claravanstaden:


system-parachains/bridge-hubs/bridge-hub-polkadot/src/bridge_to_ethereum_config.rs
```diff
@@ -237,6 +241,10 @@ parameter_types! {
 			version: hex!("05000000"),
 			epoch: 0,
 		}
+		fulu: Fork {
+			version: hex!("06000000"),
+			epoch: 50000000,
```
bkchr, line 246: Why is the epoch here not 0?

claravanstaden: Because the test fixtures rely on the fork version being `0x06000000` (Electra).

PR #969, opened 2025-10-15
[AHM] Final Staking Configs for PAHM

kianenigma:
- Put staking pallet in a different order -- order of on-init is based in the order, not pallet index
- re-calculate minimum staking score based on election data, not reusing what is in the polkadot RC now
- Validator set size check in RC should be strict as it is on Kusama.
- few comment updates

### DIFF
--- a/relay/polkadot/src/lib.rs
+++ b/relay/polkadot/src/lib.rs
@@ -1552,7 +1552,8 @@ impl ah_client::Config for Runtime {
 	type AdminOrigin = EnsureRoot<AccountId>;
 	type SessionInterface = Self;
 	type SendToAssetHub = StakingXcmToAssetHub;
-	// Polkadot RC currently has 600 validators. Note: this has to be updated with AH validator count increasing.
+	// Polkadot RC currently has 600 validators. Note: this has to be updated with AH validator
+	// count increasing.
 	type MinimumValidatorSetSize = ConstU32<600>;
 	type UnixTime = Timestamp;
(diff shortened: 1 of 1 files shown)


system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -471,20 +471,29 @@ impl frame_support::traits::OnRuntimeUpgrade for InitiateStakingAsync {
...
+		// * sum_stake: 8028519336725652293 (1.49x the minimum)
+		// * sum_stake_squared: 108358993218278434700023844467997545 (0.57 the minimum, the lower
+		//   the better)
 		let minimum_score = sp_npos_elections::ElectionScore {
-			minimal_stake: 8474057820699941,
-			sum_stake: 3276970719352749444,
-			sum_stake_squared: 244059208045236715654727835467163294,
+			minimal_stake: 7895552765679931,
+			sum_stake: 5655838551978860651,
+			sum_stake_squared: 187148285683372481445131595645808873,
```
ggwpez, line 496: Maybe someone else from staking can sanity check those. But we will go ahead with merging.

sigurpol: checking now

sigurpol: checked against the magic script, we are good :)

sigurpol (approved):

PR #977, opened 2025-10-20
[AHM] Fix manager multisig

ggwpez:
Changes:
- Set the set of members for the PAHM multisig.
- Move tests from Kusama to Polkadot runtime.
- Do not ignore the return value of `ensure_none` 🤪 
- Rerun the unsigned validation logic inside the extrinsic.
- Limit the number of votes per candidate and round to 10 to prevent spam.

### DIFF
--- a/pallets/rc-migrator/src/lib.rs
+++ b/pallets/rc-migrator/src/lib.rs
@@ -1322,7 +1322,7 @@ pub mod pallet {
 			payload: Box<ManagerMultisigVote<T>>,
 			sig: sp_runtime::MultiSignature,
 		) -> DispatchResult {
-			let _ = ensure_none(origin)?;
+			ensure_none(origin)?;
 
 			Self::do_validate_unsigned(&payload, &sig)
 				.map_err(|_| Error::<T>::UnsignedValidationFailed)?;
@@ -1399,10 +1399,15 @@ pub mod pallet {
 		}
 	}
 
+	/// The multisig AccountIDs that votes to execute a specific call.
 	#[pallet::storage]
 	#[pallet::unbounded]
 	pub type ManagerMultisigs<T: Config> =
 		StorageMap<_, Twox64Concat, <T as Config>::RuntimeCall, Vec<AccountId32>, ValueQuery>;
+
+	/// The current round of the multisig voting.
+	///
+	/// Votes are only valid for the current round.
 	#[pallet::storage]
 	pub type ManagerMultisigRound<T: Config> = StorageValue<_, u32, ValueQuery>;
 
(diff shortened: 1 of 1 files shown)


relay/polkadot/src/lib.rs
```diff
@@ -1735,6 +1735,60 @@ parameter_types! {
...
+	let addresses = vec![
+		"13fvj4bNfrTo8oW6U8525soRp6vhjAFLum6XBdtqq9yP22E7", // Basti
+		"16SDAKg9N6kKAbhgDyxBXdHEwpwHUHs2CNEiLNGeZV55qHna", // Gav
+		"1eTPAR2TuqLyidmPT9rMmuycHVm9s9czu78sePqg2KHMDrE",  // Kian
+		"16a357f5Sxab3V2ne4emGQvqJaCLeYpTMx3TCjnQhmJQ71DX", // Oliver
+		"15DCWHQknBjc5YPFoVj8Pn2KoqrqYywJJ95BYNYJ4Fj3NLqz", // Joe
+		"12HWjfYxi7xt7EvpTxUis7JoNWF7YCqa19JXmuiwizfwJZY2", // Muharem
+		"121dd6J26VUnBZ8BqLGjANWkEAXSb9mWq1SB7LsS9QNTGFvz", // Adrian
+		"12pRzYaysQz6Tr1e78sRmu9FGB8gu8yTek9x6xwVFFAwXTM8", // RobK
+		"142zGifFwRrDbFLJD7LvbyoHQAqDaXeHjkxJbUVwmDYBD7Gf", // Donal
```
ggwpez, line 1785: Need to check those again. I picked them from the P-Collectives Fellowship list.

pallets/rc-migrator/src/lib.rs
```diff
@@ -1314,21 +1320,32 @@ pub mod pallet {
 		pub fn vote_manager_multisig(
 			origin: OriginFor<T>,
 			payload: Box<ManagerMultisigVote<T>>,
-			_sig: sp_runtime::MultiSignature,
+			sig: sp_runtime::MultiSignature,
 		) -> DispatchResult {
-			let _ = ensure_none(origin);
+			ensure_none(origin)?;
+
+			Self::do_validate_unsigned(&payload, &sig)
```
muharem, line 1327: why do we perform this twice?

muharem (approved):

pallets/rc-migrator/src/lib.rs
ggwpez: I think the first time only happens in TX pool, not in the block. So here we do it also in the block.  
But it is good practice anyway. It would have prevented the issue we had as well.

kianenigma (approved):

PR #983, opened 2025-10-21
Add an XCM mapper for calls during AHM

franciscoaguirre:
Added an XCM mapper so we can map `XcmPallet::execute()` calls from the Relay Chain to Asset Hub.

Individual instructions are mapped by reanchoring the locations and assets involved.

### DIFF
--- a/system-parachains/asset-hubs/asset-hub-polkadot/Cargo.toml
+++ b/system-parachains/asset-hubs/asset-hub-polkadot/Cargo.toml
@@ -308,6 +308,7 @@ std = [
 	"frame-system-rpc-runtime-api/std",
 	"frame-system/std",
 	"frame-try-runtime?/std",
+	"hex/std",
 	"kusama-runtime-constants/std",
 	"log/std",
 	"pallet-ah-ops/std",
@@ -392,7 +393,6 @@ std = [
 	"xcm-executor/std",
 	"xcm-runtime-apis/std",
 	"xcm/std",
-	"hex/std"
 ]
 
 # Set timing constants (e.g. session period) to faster versions to speed up testing.


system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/mod.rs
```diff
@@ -408,9 +411,49 @@ impl RcToAhCall {
 					);
 				})?;
 
+				// Convert the versioned XCM to latest version for DescendOrigin conversion
+				let xcm: xcm::latest::Xcm<()> = (*message).try_into().map_err(|err| {
```
ggwpez: Why did we not do this for Kusama? Was it missing?

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/xcm_mapping.rs
```diff
@@ -0,0 +1,196 @@
...
+///
+/// If `DescendOrigin` is used, it has to be mapped to `AliasOrigin` since
+/// Asset Hub's location is one lower than the Relay.
+/// For `AliasOrigin` to work, the recipient chain needs to comply with the
+/// suggested aliasing rules in RFC#122:
+/// https://github.com/polkadot-fellows/RFCs/blob/main/text/0122-alias-origin-on-asset-transfers.md#suggested-aliasing-rules.
+pub fn reanchor_xcm_for_send(
+	xcm: Xcm<()>,
+	_ah_location: &Location,
+	_universal_location: &InteriorLocation,
```
ggwpez: Why pass them if not needed?

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/xcm_mapping.rs
```diff
@@ -0,0 +1,196 @@
...
+	ah_location: &Location,
+	universal_location: &InteriorLocation,
+) -> Result<Instruction<()>, ()> {
+	use Instruction::*;
+
+	// We only map a particular subset of instructions, to keep the logic minimal.
+	match instruction {
+		WithdrawAsset(assets) => {
+			let reanchored_assets =
+				assets.reanchored(ah_location, universal_location).map_err(|err| {
```
ggwpez, line 52: This also sorts them? Or do these not need to be sorted?

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/xcm_mapping.rs
```diff
@@ -0,0 +1,196 @@
...
+			let reanchored_asset =
+				asset.reanchored(ah_location, universal_location).map_err(|err| {
+					log::error!(target: LOG_TARGET, "Failed to reanchor asset: {err:?}");
+				})?;
+			Ok(PayFees { asset: reanchored_asset })
+		},
+		DepositAsset { assets, beneficiary } => {
+			let reanchored_beneficiary = if is_local_account(&beneficiary) {
+				// Local accounts (parents: 0, AccountId32/AccountKey20) don't need reanchoring.
+				beneficiary
```
ggwpez, line 67: Would it error or just be a no-op when calling the reanchor function on them?

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/xcm_mapping.rs
```diff
@@ -0,0 +1,196 @@
...
+			};
+			Ok(DepositAsset { assets, beneficiary: reanchored_beneficiary })
+		},
+		DepositReserveAsset { assets, dest, xcm } => {
+			let reanchored_dest =
+				dest.reanchored(ah_location, universal_location).map_err(|err| {
+					log::error!(target: LOG_TARGET, "Failed to reanchor dest: {err:?}");
+				})?;
+			// The nested xcm is already from the perspective of the dest, so no reanchoring needed.
+			Ok(DepositReserveAsset { assets, dest: reanchored_dest, xcm })
```
ggwpez: We dont reanchor the assets?

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/xcm_mapping.rs
```diff
@@ -0,0 +1,196 @@
...
+	}
+}
+
+fn is_local_account(location: &Location) -> bool {
+	// Check if this is a local account (parents: 0, AccountId32/AccountKey20)
+	if location.parents != 0 {
+		return false;
+	}
+
+	match location.interior.first() {
```
ggwpez, line 103: Just to be 100% maybe check that there is only one interior location?

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/xcm_mapping.rs
```diff
@@ -0,0 +1,196 @@
...
+fn convert_descend_to_alias_origin(instruction: Instruction<()>) -> Result<Instruction<()>, ()> {
+	use Instruction::*;
+
+	match instruction {
+		DescendOrigin(interior) => {
+			// Convert `DescendOrigin` to `AliasOrigin` from Asset Hub.
+			// `DescendOrigin` by the Relay Chain operated from one parent up,
+			// so we need to add one parent	to the location when converting to `AliasOrigin`.
+			let alias_location = Location::new(1, interior);
+			Ok(AliasOrigin(alias_location))
```
ggwpez: So asset hub pretends to be Relay chain origin? Basically an "AscendOrigin"?

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/xcm_mapping.rs
```diff
@@ -0,0 +1,196 @@
...
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	use crate::ah_migration::{RcRuntimeCall, RcToAhCall, RcXcmCall, RuntimeCall};
+	use xcm::prelude::*;
+
+	#[test]
+	fn map_xcm_execute() {
```
ggwpez, line 137: I think this ref could be a good test. You can load the raw data through PJS and then just past it as Hex string in here and decode: https://polkadot.subsquare.io/referenda/1770

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/mod.rs
franciscoaguirre: `DescendOrigin` is not the most common instruction/pattern. Probably there was not a need for this.

franciscoaguirre: However, I decided it was too complicated to map this pattern. I'll advise any new proposal to use `AliasOrigin` directly instead. Since it would end up being mapped to AH anyway.

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/xcm_mapping.rs
franciscoaguirre: Removed this in the latest version

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/xcm_mapping.rs
franciscoaguirre: [It does](https://github.com/paritytech/polkadot-sdk/blob/de740e1892f47e91c356858324f6e3e4b0f1be60/polkadot/xcm/src/v5/asset.rs#L744)

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/xcm_mapping.rs
franciscoaguirre: It would result in the wrong thing. All accounts are moving to AH as part of the migration, so if we were referencing a local account before (on Relay), we want it to keep on being a local account (on Asset Hub).

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/xcm_mapping.rs
franciscoaguirre: `assets` is an `AssetFilter`. More often than not it's a `Wild` filter that doesn't need to be reanchored.
I've added a function that will reanchor them if they are a `Definite` filter.

system-parachains/asset-hubs/asset-hub-polkadot/src/ah_migration/xcm_mapping.rs
franciscoaguirre: Basically but scrapped the idea as it was getting complicated. It'd be better if new refs don't use this

PR #988, opened 2025-10-23
Resolve conflicts

claravanstaden:
@clangenb there are a couple of Encointer errors:

```
error[E0599]: no function or associated item named `convert` found for struct `EncointerBalanceConverter` in the current scope
   --> /Users/claravanstaden/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/pallet-encointer-balances-20.2.1/src/impl_fungibles.rs:27:29
    |
27  |     EncointerBalanceConverter::convert(balance)
    |                                ^^^^^^^ function or associated item not found in `EncointerBalanceConverter`
    |
note: there are multiple different versions of crate `sp_runtime` in the dependency graph
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -942,11 +942,11 @@ dependencies = [
  "kusama-emulated-chain",
  "parachains-common",
  "penpal-emulated-chain",
- "polkadot-parachain-primitives",
+ "polkadot-parachain-primitives 19.0.0",
  "sp-core 38.1.0",
  "sp-keyring",
- "staging-xcm",
- "staging-xcm-builder",
+ "staging-xcm 19.0.0",
+ "staging-xcm-builder 23.0.0",
 ]
 
 [[package]]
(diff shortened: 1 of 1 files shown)


clangenb:
The encointer errors are most certainly because of duplicate dep issues still, and I believe that the encointer crates are on the correct versions.

But if you believe everything is correct on your end, I can look into it.

Cargo.lock
```diff
@@ -1927,47 +1927,82 @@ name = "bp-bridge-hub-polkadot"
...
 dependencies = [
  "bp-bridge-hub-cumulus",
- "bp-header-chain",
- "bp-messages",
- "bp-polkadot-core",
- "bp-runtime",
+ "bp-header-chain 0.23.0",
+ "bp-messages 0.23.0",
+ "bp-polkadot-core 0.23.0",
+ "bp-runtime 0.23.0",
```
clangenb, line 1933: You are introducing many duplicate dependencies. Is this on purpose?

Cargo.lock
```diff
@@ -1977,15 +2012,33 @@ dependencies = [
...
+dependencies = [
+ "bp-header-chain 0.22.0",
+ "bp-polkadot-core 0.22.0",
+ "bp-runtime 0.22.0",
+ "frame-support 42.0.0",
+ "impl-trait-for-tuples",
+ "parity-scale-codec",
+ "scale-info",
+ "sp-core 38.1.0",
+ "sp-runtime 43.0.0",
```
clangenb, line 2029: I think it is not on purpose. For example: this new duplicate dependency uses the old sp-runtime v43.0.0, whereas the current one is supposed to be v44.0.0.

I believe encointer's error will go away once the duplicate sp-runtime dependency has been resolved, aka sp-runtime v43.0.0 has been completely eliminated from the workspace.

PR #996, opened 2025-11-04
Fix depencencies

clangenb:
This PR fixes the dependency mess that we still have in the update branch.

The issues with encointer's dependencies was that we only introduced patch bumps for the stable202509 upgrade to keep being aligned with the polkadot-sdk releases (we messed up the rhythm, as the unstable branch was a major release for us). The patch bump did not propagate to our innermost dependencies in the lock file, as it was deemed compatible with the ones explicitly defined in the root toml.

So this PR fixes (I believe all) of the duplicate substrate dependencies by running

1.  `cargo update -p encointer-crates-at-wrong-version`
2. `psvm --version polkadot-stable2509-1`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,8 +9,8 @@ license = "GPL-3.0-only"                                        # TODO <https://
 pallet-ah-migrator = { path = "pallets/ah-migrator", default-features = false }
 pallet-rc-migrator = { path = "pallets/rc-migrator", default-features = false }
 pallet-ah-ops = { path = "pallets/ah-ops", default-features = false }
-pallet-election-provider-multi-block = { version = "0.3.2", default-features = false }
-pallet-staking-async = { version = "0.6.2", default-features = false }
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -231,7 +231,7 @@ snowbridge-inbound-queue-primitives = { version = "0.5.0", default-features = fa
 snowbridge-runtime-common = { version = "0.17.0", default-features = false }
 snowbridge-runtime-test-common = { version = "0.19.0" }
 snowbridge-system-runtime-api = { version = "0.16.0", default-features = false }
-snowbridge-system-v2-runtime-api = { version = "0.4.0", default-features = false }
+snowbridge-system-v2-runtime-api = { version = "0.5.0", default-features = false }
```
clangenb, line 234: @claravanstaden `psvm --version polkadot-stable2509-1` yielded this. Can you confirm that this is ok?

claravanstaden: This is 100%, thank you!

PR #1011, opened 2025-11-14
fix zombienet test for `dind` environment 

pepoviola:

Fix zombienet test to run in `dind` environment.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4155,9 +4155,9 @@ dependencies = [
  "parity-scale-codec",
  "polkadot-primitives",
  "sp-consensus-babe",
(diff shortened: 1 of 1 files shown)


Cargo.lock
skunert, line 1: Why do we have so many changes here? Did you update by accident or is it purely because of the `zombienet-sdk` update?

bkchr:
@pepoviola please update this today after we have the runtime release finished

Cargo.lock
pepoviola: I think is related to the updates in the latest version of zombienet. I didn't run an _update_ here.

skunert (approved):

PR #1012, opened 2025-11-19
add review bot workflow

xlc:
Integrate a generic review bot for new PR. It can also be dispatched manually in actions.

Future tasks:
- command bot integration so non github org users can request review
- polkadot-sdk specific prompts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,11 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## [Unreleased]
+## [2.0.3] 21.11.2025
 
 ### Added
 
-- Set Ethereum Fulu fork epoch ([#1005](https://github.com/polkadot-fellows/runtimes/pull/1005)).
+- Set Ethereum Fulu fork epoch ([#1005](https://github.com/polkadot-fellows/runtimes/pull/1005))
 
 ### Changed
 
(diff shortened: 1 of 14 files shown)


bkchr (approved):

pandres95 (approved):
LGT🤖

claravanstaden (approved):

.github/workflows/ai-review.yml
```diff
@@ -0,0 +1,43 @@
...
+
+concurrency:
+  group: review-pr-${{ github.event.inputs.pr_number || github.event.number }}
+  cancel-in-progress: true
+
+jobs:
+  review:
+    if: github.event_name == 'workflow_dispatch' || contains(fromJson('["OWNER","MEMBER","COLLABORATOR"]'), github.event.pull_request.author_association)
+    timeout-minutes: 10
+    runs-on: ubuntu-latest
```
bkontur, line 25: Recently, Basti switched all the runners to self-hosted: https://github.com/polkadot-fellows/runtimes/pull/964

```suggestion
    runs-on: self-hosted
```

.github/workflows/ai-review.yml
```diff
@@ -0,0 +1,43 @@
...
+          fetch-depth: 0
+
+      - name: Process PR
+        uses: polka-codes/action@v2
+        with:
+          pr_number: ${{ github.event.inputs.pr_number || github.event.number }}
+          review: true
+        env:
+          POLKA_API_PROVIDER: openrouter
+          POLKA_MODEL: google/gemini-2.5-pro
```
PolkadotDom, line 39: 3.0 preview?

sigurpol: if [this](https://ai.google.dev/gemini-api/docs/models) doesn't lie, the name seems `gemini-3-pro-preview`  in case 😄  (but then, google/gemini-2.5-pro is also valid even if the bot doesn't seem to agree...)

xlc: There is some compatibility issue which pending on some upstream lib changes. I think all the blockers are fixed in upstream but still need to verify.

.github/workflows/ai-review.yml
xlc: but do we have to make all workflows to use self-hosted? IMO public repo have unlimited free action minutes, and this workflow doesn't need beefy machine to run.

PR #1015, opened 2025-11-21
Update install-deps.sh

senseless:
Add yarn to CI dependencies

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,11 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## [Unreleased]
+## [2.0.3] 21.11.2025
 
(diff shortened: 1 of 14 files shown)


.github/install-deps.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 sudo apt update
-sudo apt install --assume-yes openssl pkg-config g++ make cmake protobuf-compiler libssl-dev libclang-dev libudev-dev git lz4
+sudo apt install --assume-yes openssl pkg-config g++ make cmake protobuf-compiler libssl-dev libclang-dev libudev-dev git lz4 yarn
```
pepoviola: I think you need to setup the apt repository to install yarn (https://classic.yarnpkg.com/lang/en/docs/install/#debian-stable). Also, if the runner already have `nodejs` we can just run `npm install --global yarn` since I think most of those deps aren't needed for the `ecossytem-test`.

PR #1020, opened 2025-11-28
update ai review

xlc:
~~Use google/gemini-3-pro-preview~~
Use prompts from https://github.com/polka-codes/prompts

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Add pay module implementation to system_parachains_common for use in Asset Hub Polkadot/Kusama treasury with XCM V3 to V5 conversion support for asset transactions ([#1019](https://github.com/polkadot-fellows/runtimes/pull/1019))
 - asset-hub-kusama: use inmemorydb weights([polkadot-fellows/runtimes/pull/918](https://github.com/polkadot-fellows/runtimes/pull/918))
 - Add fast runtime feature for Encointer Kusama([#1021](https://github.com/polkadot-fellows/runtimes/pull/1021))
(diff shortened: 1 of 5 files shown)


xlc:
gemini 3 pro still not working with openrouter https://github.com/OpenRouterTeam/ai-sdk-provider/issues/239#issuecomment-3590779460

.github/workflows/ai-review.yml
```diff
@@ -2,7 +2,7 @@ name: AI Review PR
 
 on:
   pull_request:
-    types: [opened]
+    types: [opened, ready_for_review]
```
bkchr, line 5: Here we want to have this for every pr? Who is paying for this? :D

xlc: it costs somewhere between $0.5 to $2 per review, we have like 3 PR per week? so that will be <$10 per week
I will say the time saved by AI review significantly outweighs the cost.
Right now it is using my personal openrouter api key. I will ask fellowship treasury reimbursement when the total cost is >$50.

I also have a budget limit so it can't go crazy
 
Screenshot 2025-12-12 at 11 21 05 AM

.github/workflows/ai-review.yml
```diff
@@ -20,7 +20,7 @@ concurrency:
 
 jobs:
   review:
-    if: github.event_name == 'workflow_dispatch' || contains(fromJson('["OWNER","MEMBER","COLLABORATOR"]'), github.event.pull_request.author_association)
+    if: github.event_name == 'workflow_dispatch' || contains(fromJson('["OWNER","MEMBER","COLLABORATOR","CONTRIBUTOR"]'), github.event.pull_request.author_association)
```
xlc, line 23: github have a funny requirement that only mark the author as MEMBER if their org association visibility is public (and that defaults to private)
so currently most of PR requires manual trigger

PR #1022, opened 2025-12-02
asset-hub-polkadot: increase the base deposit for a signed NPoS solution from 4 to 100 DOT

sigurpol:
On Polkadot AH,  we are currently allowing `MaxSignedSumbissions=16` to be submitted during `Signed` phase of a NPoS election.
The last submission  will cost `4 * 2**15 = 131072 DOT` and the sum of all 16 solutions will cost `4 * (2**16 - 1) = 262140 DOT`, which makes a transaction spamming attack  expensive but not prohibitive. 

We now increase the base deposit from 4 to 100 DOT increasing the cost of the attack by factor 25. This  doesn't disincentivize community participation but significantly reduces the spamming attack vector risk to minimum.

It's worth mentioning that an attacker, aside from wasting a considerable amount of DOT, wouldn't gain anything from the attack. If no valid submission is made during the Signed phase, we will select the solution produced by OCW during the Unsigned phase. Therefore, the net result of the attack would be a slightly suboptimal but still decent solution compared to the one offered by the legitimate miner.

Note that for Kusama the deposit is set to `0.1 * 4^N` with again at most 16 solutions, meaning the spamming attack cost is `0.1 * 4 ^15 = 107374182.4 KSM` so already insanely high, no changes are needed for Kusama.

Fix https://github.com/paritytech-secops/srlabs_findings/issues/510.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+
+### Changed
+
 - asset-hub-polkadot: increase the base deposit for a signed NPoS solution from 4 to 100 DOT, significantly mitigating the risk of transaction spamming ([#1022](https://github.com/polkadot-fellows/runtimes/pull/1022))
 
 ### Added


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+- asset-hub-polkadot: increase the base deposit for a signed NPoS solution from 4 to 100 DOT, significantly mitigating the risk of transaction spamming ([#1022](https://github.com/polkadot-fellows/runtimes/pull/1022))
```
ggwpez, line 12:
```suggestion

### Changed

- asset-hub-polkadot: increase the base deposit for a signed NPoS solution from 4 to 100 DOT, significantly mitigating the risk of transaction spamming ([#1022](https://github.com/polkadot-fellows/runtimes/pull/1022))
```

ggwpez (approved):

PR #1025, opened 2025-12-04
asset-hub-polkadot: use inmemorydb weights

alexggh:
Modify asset-hub-polkadot to use the read/write costs computed in validate block on validator reference hardware, since these are weights that guarantee the blocks will pass validation and make it on the relay chain.

This were obtained with the benchmark created here: https://github.com/paritytech/polkadot-sdk/pull/8069.

Part of: https://github.com/polkadot-fellows/runtimes/issues/800


AndreiEres (approved):

claravanstaden (approved):
@alexggh perfect!

PR #1029, opened 2025-12-10
Upgrade to unstable2507

TorstenStueber:
This upgrades polkadot-sdk crates to the latest version of `unstable2507`. The main change is the following backport: https://github.com/paritytech/polkadot-sdk/pull/10552.

This is an important step for upgrading the Kusama AH runtime to the latest version of pallet-revive with:
- EVM backend
- Ethereum block storage
- New gas mapping

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -14,7 +14,7 @@ jobs:
     if: ${{ github.event.issue.pull_request && startsWith(github.event.comment.body, '/merge') }}
     steps:
       - name: Get the GitHub handle of the fellows
-        uses: paritytech/get-fellows-action@v1.3.1
+        uses: paritytech/get-fellows-action@v1.3.2
         timeout-minutes: 5
         id: fellows
       - name: Generate a token
--- a/.github/workflows/cmd.yml
+++ b/.github/workflows/cmd.yml
@@ -21,7 +21,7 @@ jobs:
 
       - name: Load fellows
         id: load-fellows
-        uses: paritytech/get-fellows-action@v1.3.1
+        uses: paritytech/get-fellows-action@v1.3.2
         timeout-minutes: 5
 
   reject-non-fellows:
--- a/.github/workflows/review-bot.yml
+++ b/.github/workflows/review-bot.yml
@@ -13,6 +13,7 @@ permissions:
 jobs:
   review-approvals:
     runs-on: self-hosted
+    timeout-minutes: 5
     steps:
       - name: Extract content of artifact
         id: number
@@ -26,7 +27,7 @@ jobs:
           app-id: ${{ secrets.REVIEW_APP_ID }}
           private-key: ${{ secrets.REVIEW_APP_KEY }}
       - name: "Evaluates PR reviews and assigns reviewers"
-        uses: paritytech/review-bot@v2.7.1
+        uses: paritytech/review-bot@v2.7.2
         with:
           repo-token: ${{ secrets.GITHUB_TOKEN }}
           team-token: ${{ steps.team_token.outputs.token }}
--- a/.github/workflows/review-trigger.yml
+++ b/.github/workflows/review-trigger.yml
@@ -28,7 +28,7 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
       - name: Get the GitHub handle of the fellows
-        uses: paritytech/get-fellows-action@v1.3.1
+        uses: paritytech/get-fellows-action@v1.3.2
         timeout-minutes: 5
         id: fellows
         # Require new reviews when the author is pushing and he is not a fellow


system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
```diff
@@ -329,11 +328,8 @@ parameter_types! {
 
 impl pallet_transaction_payment::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
-	type OnChargeTransaction = impls::tx_payment::FungiblesAdapter<
```
TorstenStueber, line 332: Reasoning for this change: Using `FungiblesAdapter` (plural) for `NativeAndAsset` and `DotLocation` is the same as using `FungibleAdapter` (singular) with `Balances` because `NativeAndAsset` will just forward all calls to `Balances` for `DotLocation`.

Cargo.toml
```diff
@@ -9,8 +9,8 @@ license = "GPL-3.0-only"                                        # TODO <https://
 pallet-ah-migrator = { path = "pallets/ah-migrator", default-features = false }
 pallet-rc-migrator = { path = "pallets/rc-migrator", default-features = false }
 pallet-ah-ops = { path = "pallets/ah-ops", default-features = false }
-pallet-election-provider-multi-block = { version = "0.3.4", default-features = false }
-pallet-staking-async = { version = "0.6.2", default-features = false }
+pallet-election-provider-multi-block = { version = "0.5.0", default-features = false }
+pallet-staking-async = { version = "0.7.0", default-features = false }
```
ggwpez, line 13: @kianenigma anything to do when bumping these?

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -302,7 +303,7 @@ impl pallet_transaction_payment::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type OnChargeTransaction =
 		pallet_transaction_payment::FungibleAdapter<Balances, ResolveTo<StakingPot, Balances>>;
-	type WeightToFee = WeightToFee;
+	type WeightToFee = pallet_revive::evm::fees::BlockRatioFee<333333333, 10_815_700_000_000, Self>;
```
ggwpez: Where do these numbers come from?

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1173,20 +1174,24 @@ impl pallet_remote_proxy::Config for Runtime {
 
 parameter_types! {
 	pub const DepositPerItem: Balance = system_para_deposit(1, 0);
+	pub const DepositPerChildTrieItem: Balance = system_para_deposit(1, 0) / 100;
```
ggwpez: Why / 100?

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1173,20 +1174,24 @@ impl pallet_remote_proxy::Config for Runtime {
...
 	pub const DepositPerByte: Balance = system_para_deposit(0, 1);
 	pub CodeHashLockupDepositPercent: Perbill = Perbill::from_percent(30);
+	pub const MaxEthExtrinsicWeight: FixedU128 = FixedU128::from_rational(9, 10);
```
ggwpez, line 1186:
```suggestion
	pub const MaxEthExtrinsicWeight: FixedU128 = FixedU128::from_percent(90);
```

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1204,8 +1209,12 @@ impl pallet_revive::Config for Runtime {
 	type CodeHashLockupDepositPercent = CodeHashLockupDepositPercent;
 	type ChainId = ConstU64<420_420_418>;
 	type NativeToEthRatio = ConstU32<1_000_000>; // 10^(18 - 12) Eth is 10^18, Native is 10^12.
-	type EthGasEncoder = ();
 	type FindAuthor = <Runtime as pallet_authorship::Config>::FindAuthor;
+	type AllowEVMBytecode = ConstBool<true>;
+	type FeeInfo = pallet_revive::evm::fees::Info<Address, Signature, EthExtraImpl>;
+	type MaxEthExtrinsicWeight = MaxEthExtrinsicWeight;
+	type DebugEnabled = ConstBool<false>;
```
ggwpez, line 1223: I think debug stuff can be feature gated behind some `debug` feature to never accidentally land in prod.

TorstenStueber: @athei what are your thoughts on this?

athei: This is not what features are meant for. We should only use them for additive things that are **harmless** when activated (i.e just add compilation time) when activated. Reasoning is that they can be implicitly activated through unification. Unlikely for a leave dependency but still.

We should rather rename the debug variable to `UnsafeDebug` to make clear that nobody should flip that switch. In the meantime a comment here will also help.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1173,20 +1174,24 @@ impl pallet_remote_proxy::Config for Runtime {
...
 	type RuntimeEvent = RuntimeEvent;
 	type RuntimeCall = RuntimeCall;
+	type RuntimeOrigin = RuntimeOrigin;
 	type DepositPerItem = DepositPerItem;
+	type DepositPerChildTrieItem = DepositPerChildTrieItem;
 	type DepositPerByte = DepositPerByte;
-	type WeightPrice = pallet_transaction_payment::Pallet<Self>;
 	// TODO(#840): use `weights::pallet_revive::WeightInfo` here
-	type WeightInfo = pallet_revive::weights::SubstrateWeight<Self>;
+	type WeightInfo = weights::pallet_revive::WeightInfo<Runtime>;
```
athei: Please roll back. No need to copy paste the weights. I will try to generate the weights. And if it still doesn't work we want to leave it as-is anyways.

system-parachains/asset-hubs/asset-hub-kusama/src/weights/pallet_revive.rs
athei: Just delete.

athei, line 1: Just delete. No need to copy paste. We either use one generated by the bot here. Or we stick buy referencing them from polkadot-sdk (leave as-is).

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
athei: Contracts can potentially create a lot of small storage items. During testing we needed to lower the costs to make contracts viable. However, I think the deposits were configured too highly on the test runtimes. Those are the costs right now on AH-KSM:
```
item deposit: 0.00666666666  KSM
kb deposit:   0.000341332992 KSM
```

athei: The current costs seem fine. Let's just keep it safe in order to move fast. We can lower later. Refunds are pro-rata to lowering or increasing is both fine.

```suggestion
	pub const DepositPerChildTrieItem: Balance = system_para_deposit(1, 0);
```

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
TorstenStueber: This encodes the same fraction as the one in the `WeightToFeePolynomial` used in `WeightToFee`, which is defined as: 
- numerator: `super::currency::CENTS`
  - this is `1_000_000_000_000 / 30 / 100` = `333_333_333`
- denominator: `100 * Balance::from(ExtrinsicBaseWeight::get().ref_time())`
  - this is `100 * 1_000 * 108_157` = `10_815_700_000`

TorstenStueber: I corrected the value that was here before, it was wrongly calculated.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
TorstenStueber: This method does not exist for `FixedU128`, only for the `Per...` types and `Perbill::from_percent(90).into()` cannot be used as it is not a `const` function.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
TorstenStueber: We set it to `system_para_deposit(1, 0) / 10` now. 

Reasoning:
1) This only affects the deposit cost of child trie items. All the persistent storage of a contract is stored in per-contract child tries. Therefore, the danger of bloating the main trie and slowing the constructing of the PoV is greatly reduced.

2) Smart contracts use their storage much more liberally and frequently. The cost of storing a single state variable today when using `DepositPerChildTrieItem = system_para_deposit(1, 0)` would be 4 times higher than the cost on Ethereum, taking the greatly different valuations of KSM and ETH into account!

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -302,7 +303,7 @@ impl pallet_transaction_payment::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type OnChargeTransaction =
 		pallet_transaction_payment::FungibleAdapter<Balances, ResolveTo<StakingPot, Balances>>;
-	type WeightToFee = WeightToFee;
+	type WeightToFee = pallet_revive::evm::fees::BlockRatioFee<333333333, 10_815_700_000, Self>;
```
acatangiu, line 312: can you add a comment here explaining the used `WeightToFee`? I'm sure it will not age well without clear docs/code-comments..

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1173,18 +1174,22 @@ impl pallet_remote_proxy::Config for Runtime {
...
+	type Balance = Balance;
 	type Currency = Balances;
 	type RuntimeEvent = RuntimeEvent;
 	type RuntimeCall = RuntimeCall;
+	type RuntimeOrigin = RuntimeOrigin;
 	type DepositPerItem = DepositPerItem;
+	type DepositPerChildTrieItem = DepositPerChildTrieItem;
 	type DepositPerByte = DepositPerByte;
-	type WeightPrice = pallet_transaction_payment::Pallet<Self>;
 	// TODO(#840): use `weights::pallet_revive::WeightInfo` here
```
acatangiu, line 1199: with the new `pallet-revive` version you should be able to fix this too, right?

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
athei: > 2\. would be 4 times higher than the cost on Ethereum, taking the greatly different valuations of KSM and ETH into account!

On ETH the costs are not refundable though. But I agree that we need to reduce the costs. And / 10 is already pretty conservative.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
ggwpez: Yea true, the feature would need to be named something like `revive-unsafe-debug` to not accidentally be unified in the workspace.  
So its okay for now.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
ggwpez: Maybe you can add a sanity check test to ensure that the fee of a simple transaction is sane?  
And please put the calculation into a comment above the code so that we can reproduce it later.

Cargo.toml
```diff
@@ -198,47 +200,47 @@ polkadot-emulated-chain = { path = "integration-tests/emulated/chains/relays/pol
...
+snowbridge-pallet-outbound-queue = { version = "0.17.0", default-features = false }
+snowbridge-pallet-outbound-queue-v2 = { version = "0.6.0", default-features = false }
+snowbridge-pallet-system = { version = "0.17.0", default-features = false }
+snowbridge-pallet-system-v2 = { version = "0.6.0", default-features = false }
+snowbridge-pallet-system-frontend = { version = "0.6.0", default-features = false }
+snowbridge-inbound-queue-primitives = { version = "0.6.0", default-features = false }
+snowbridge-runtime-common = { version = "0.18.0", default-features = false }
+snowbridge-runtime-test-common = { version = "0.20.0" }
+snowbridge-system-runtime-api = { version = "0.17.0", default-features = false }
+snowbridge-system-v2-runtime-api = { version = "0.6.0", default-features = false }
```
acatangiu, line 238: @claravanstaden please confirm the snowbridge pallet versions bump

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
ggwpez: Sounds good, thanks!

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
athei: I believe the benchmark is still not fixed. Also we need `solc` to be installed on the benchmarking bot. We don't have time to get this done right now.

Cargo.toml
```diff
@@ -198,47 +200,47 @@ polkadot-emulated-chain = { path = "integration-tests/emulated/chains/relays/pol
...
-snowbridge-runtime-common = { version = "0.16.0", default-features = false }
-snowbridge-runtime-test-common = { version = "0.18.0" }
-snowbridge-system-runtime-api = { version = "0.15.0", default-features = false }
-snowbridge-system-v2-runtime-api = { version = "0.4.0", default-features = false }
+snowbridge-outbound-queue-runtime-api = { version = "0.17.0", default-features = false }
+snowbridge-outbound-queue-v2-runtime-api = { version = "0.6.0", default-features = false }
+snowbridge-outbound-queue-primitives = { version = "0.6.0", default-features = false }
+snowbridge-pallet-ethereum-client = { version = "0.17.0", default-features = false }
+snowbridge-pallet-inbound-queue = { version = "0.17.0", default-features = false }
+snowbridge-pallet-inbound-queue-v2 = { version = "0.6.0", default-features = false }
```
claravanstaden:
```suggestion
snowbridge-pallet-inbound-queue-v2 = { version = "0.6.1", default-features = false }
```

Cargo.toml
claravanstaden: All of the version bumps are just because of crate dependency changes.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
TorstenStueber: I added a comment with explanation.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
TorstenStueber: I added a comment with explanation.

ggwpez (approved):

Cargo.toml
Ank4n: No changes needed for this bump.

claravanstaden (approved):

Cargo.toml
```diff
@@ -89,106 +90,108 @@ kusama-runtime = { path = "relay/kusama", package = "staging-kusama-runtime" }
...
-pallet-salary = { version = "27.0.0", default-features = false }
-pallet-scheduler = { version = "43.0.0", default-features = false }
-pallet-session = { version = "42.0.0", default-features = false }
-pallet-session-benchmarking = { version = "42.0.0", default-features = false }
-pallet-society = { version = "42.2.0", default-features = false }
-pallet-staking = { version = "42.0.0", default-features = false }
-pallet-delegated-staking = { version = "9.0.0", default-features = false }
-pallet-staking-async-ah-client = { version = "0.5.0", default-features = false }
-pallet-staking-async-rc-client = { version = "0.5.0", default-features = false }
+pallet-revive = { version = "0.11.1", default-features = false }
```
Ank4n: @TorstenStueber Is it okay to bump pallet-revive to this?

acatangiu (approved):

bkchr (approved):

PR #1016, opened 2025-11-21
[AHM] Remove migrator pallets (mostly)

ggwpez:
Changes:
- Undeploy AH Migrator pallet
- Delete 99% code of RC migrator pallet. Only thing kept is storage definition for posterior investigation.
- Add Runtime API for accessing the migration start and end block. Both on RC and AH: `AssetHubMigrationApi::MigrationStartBlock()` (and `End` resp.). This is needed for indexers. 

TODO:
- [x] Remove AH pallet storage

Remaining RC storage until final pallet cleanup:
```pre
52 K RcMigrator (647 keys, key: 46 K, value: 5.53 K, compressed: 1.86 K)
├── 36 K PureProxyCandidatesMigrated (498 keys, key: 36 K, compressed_key: 1.22 K, value: 498 B, compressed_value: 75 B)
├── 14 K RcAccounts (133 keys, key: 9.58 K, compressed_key: 405 B, value: 4.88 K, compressed_value: 113 B)
├── 168 B ManagerVotesInCurrentRound (2 keys, key: 160 B, compressed_key: 89 B, value: 8, compressed_value: 8)
├── 160 B ManagerMultisigs (2 keys, key: 94 B, compressed_key: 54 B, value: 66 B, compressed_value: 40 B)
├── 64 B RcMigratedBalanceArchive (1 keys, key: 32 B, compressed_key: 39 B, value: 32 B, compressed_value: 23 B)
├── 38 B Settings (1 keys, key: 32 B, compressed_key: 39 B, value: 6, compressed_value: 10 B)
├── 37 B WarmUpPeriod (1 keys, key: 32 B, compressed_key: 39 B, value: 5, compressed_value: 9)
├── 37 B CoolOffPeriod (1 keys, key: 32 B, compressed_key: 39 B, value: 5, compressed_value: 9)
├── 36 B UnprocessedMsgBuffer (1 keys, key: 32 B, compressed_key: 39 B, value: 4, compressed_value: 8)
├── 36 B MigrationStartBlock (1 keys, key: 32 B, compressed_key: 39 B, value: 4, compressed_value: 8)
├── 36 B MigrationEndBlock (1 keys, key: 32 B, compressed_key: 39 B, value: 4, compressed_value: 8)
├── 36 B ManagerMultisigRound (1 keys, key: 32 B, compressed_key: 39 B, value: 4, compressed_value: 8)
[...]

PR #1036, opened 2025-12-16
[AHM] Migrate missing stellaswap account

ggwpez:
The Asset Hub migration translated specific account IDs such that they would stay accessible after the migration.  
We missed two such accounts that were generated in a specific recursive manner and are fixing this now.

## Changes

Add an extrinsic `AhOps::translate_para_sovereign_child_to_sibling_derived` that can be called with the `old` and `new` account and a derivation proof (para-id and derivation-path). This then moves all data from the `old` account to the `new` one.  
The old and new accounts can be calculated just from the para-id and derivation-path but we still pass them in to ensure correctness of usage. The extrinsic can be called by the fellowship or root.

## What Is Migrated?

- Native DOT Balance
- Assets: USDC, USDT, DED, TSN
- Staked balances (unbonding balances are immediately unbonded and bonded amounts will be rebonded and start earning in the next era)

Not migrated would be vested transfers (the accounts have none) or similar.

## Translated Accounts

The translated accounts belong to Stellaswap and are derived from the Moonbeam Sovereign child account:

- Moonbeam Child 2004:  13YMK2eZbf9AyGhewRs6W6QTJvBSM5bxpnTD8WgeDofbg8Q1
	- Index 5: 14vDXpWfcSRPn8eWPKt2Xc8KN57tGNnAo7Z8M2C8kpkJav5q
	- Index 5/1: 1zAWXSCmRTR9ZkRXZXeHZftj1J6rnDe8BLXV8UJ2S2exCvL (**old 1**)
	- Index 5/2: 14KQD8dRoT3q2fCbCC49bFjU1diFu1d516tYuGmSUMmEoGNa  (**old 2**)

- Moonbeam Sibl 2004
	- 13cKp89NgPL56sRoVRpBcjkGZPrk4Vf4tS6ePUD96XhAXozG
	- Index 5: 12gb2DBw5HfpmUyBKCHxJWwGxMyXbUgf6a7bagNduHZC5S9z
	- Index 5/1: 13GWAfgWAKLGm8AsKLn5pDbDyMHfShFgtMFEqM4TRNhXbSea  (**new 1**)
	- Index 5/2: 123oqim7B24XzwB1hC4Fh7LGwbTas3QmxL6v6sVd95eTD5ee  (**new 2**)

The two accounts that will be translated are therefore:
- 1zAWXSCmRTR9ZkRXZXeHZftj1J6rnDe8BLXV8UJ2S2exCvL to 13GWAfgWAKLGm8AsKLn5pDbDyMHfShFgtMFEqM4TRNhXbSea
- 14KQD8dRoT3q2fCbCC49bFjU1diFu1d516tYuGmSUMmEoGNa to 123oqim7B24XzwB1hC4Fh7LGwbTas3QmxL6v6sVd95eTD5ee

You can see the snapshot test doing this here https://github.com/polkadot-fellows/runtimes/blob/f254eee1fdb071b3d2f5d8cfe6c909b9e156448e/system-parachains/asset-hubs/asset-hub-polkadot/src/remote_tests.rs#L56-L61


pallets/ah-ops/src/lib.rs
```diff
@@ -494,6 +556,169 @@ pub mod pallet {
...
+			}
+			pallet_balances::Pallet::<T>::ensure_upgraded(&from); // prevent future headache
+
+			// Force unstake. The actual function is private, so we use the call:
+			pallet_staking_async::Pallet::<T>::force_unstake(
+				frame_system::Origin::<T>::Root.into(),
+				from.clone(),
+				0,
+			)
+			.map_err(|_| Error::<T>::FailedToForceUnstake)?;
```
ggwpez: Not sure what to do here. There is no force_bond or force_stake?

pallets/ah-ops/src/lib.rs
```diff
@@ -494,6 +556,169 @@ pub mod pallet {
...
+
+			Self::deposit_event(Event::SovereignMigrated {
+				para_id,
+				from: from.clone(),
+				to: to.clone(),
+				derivation_path,
+			});
+
+			Ok(())
+		}
```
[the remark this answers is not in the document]
ggwpez: This is not how it works.

pallets/ah-ops/src/lib.rs
```diff
@@ -511,3 +736,23 @@ pub fn derivative_account_id<AccountId: Encode + Decode>(who: AccountId, index:
...
+	account
+}
+
+/// Backward mapping from https://github.com/paritytech/polkadot-sdk/blob/74a5e1a242274ddaadac1feb3990fc95c8612079/substrate/frame/balances/src/types.rs#L38
+pub fn map_lock_reason(reasons: LockReasons) -> LockWithdrawReasons {
+	match reasons {
+		LockReasons::All => LockWithdrawReasons::TRANSACTION_PAYMENT | LockWithdrawReasons::RESERVE,
+		LockReasons::Fee => LockWithdrawReasons::TRANSACTION_PAYMENT,
+		LockReasons::Misc => LockWithdrawReasons::TIP,
+	}
```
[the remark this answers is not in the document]
ggwpez: We already did this in the real migration.

pallets/ah-ops/src/lib.rs
```diff
@@ -344,6 +381,31 @@ pub mod pallet {
...
+		/// witness data to ensure correct usage.
+		#[pallet::call_index(4)]
+		#[pallet::weight(Weight::from_parts(100_000_000, 9000)
+				.saturating_add(T::DbWeight::get().reads_writes(2, 2)))]
+		pub fn translate_para_sovereign_child_to_sibling_derived(
+			origin: OriginFor<T>,
+			para_id: u16,
+			derivation_path: Vec<u16>,
+			old_account: T::AccountId,
+			new_account: T::AccountId,
```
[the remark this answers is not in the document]
ggwpez: Its root. I will bump it.

pallets/ah-ops/src/lib.rs
```diff
@@ -494,6 +556,169 @@ pub mod pallet {
...
+			raw[0..4].copy_from_slice(b"para");
+			raw[4..6].copy_from_slice(&id.encode());
+			raw.into()
+		}
+
+		pub fn para_sov_sibling(id: u16) -> AccountId32 {
+			let mut raw = [0u8; 32];
+			raw[0..4].copy_from_slice(b"sibl");
+			raw[4..6].copy_from_slice(&id.encode());
+			raw.into()
```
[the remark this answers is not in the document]
ggwpez: Para ID type is not u32 as its definition suggests.

pallets/ah-ops/src/lib.rs
Ank4n: `bond` is as good as a hypothetical `force_bond`. Don't think its a good idea to force an account to be exposed in the active era. Once you bond, they will be exposed in the next era.

pallets/ah-ops/src/tests.rs
```diff
@@ -131,3 +134,96 @@ fn sovereign_account_translation() {
...
+
+		crate::Pallet::<AssetHub>::do_translate_para_sovereign_child_to_sibling_derived(
+			2004,
+			derivation_path.clone(),
+			child_5_2.clone(),
+			sibl_5_2.clone(),
+		)
+		.unwrap();
+	});
+}
```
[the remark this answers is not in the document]
ggwpez: Yes we have the remote_externalities test for this.

pallets/ah-ops/src/lib.rs
```diff
@@ -207,12 +224,38 @@ pub mod pallet {
...
+		/// The account has already been translated.
+		AlreadyTranslated,
 	}
 
 	#[pallet::event]
 	#[pallet::generate_deposit(pub(crate) fn deposit_event)]
 	pub enum Event<T: Config> {
-		/// Some lease reserve could not be unreserved and needs manual cleanup.
+		/// A hold was released.
+		HoldReleased {
```
Ank4n: Wouldn't this (changing index for enum variant) break parsing existing events?

pallets/ah-ops/src/lib.rs
```diff
@@ -344,6 +389,31 @@ pub mod pallet {
...
+		///
+		/// Uses the same derivation path on the sibling. The old and new account arguments are only
+		/// witness data to ensure correct usage. Can only be called once per account.
+		#[pallet::call_index(4)]
+		#[pallet::weight(Weight::from_parts(100_000_000, 9000)
+				.saturating_add(T::DbWeight::get().reads_writes(20, 20)))]
+		pub fn translate_para_sovereign_child_to_sibling_derived(
+			origin: OriginFor<T>,
+			para_id: u16,
+			derivation_path: Vec<u16>,
```
Ank4n, line 392: We should bound this since the extrinsic can be called by anyone and we iterate over it (each iteration does a blake 256 hash).

pallets/ah-ops/src/lib.rs
```diff
@@ -494,6 +564,201 @@ pub mod pallet {
...
+			let missing = <T as Config>::Currency::unreserve(&from, unnamed_reserve);
+			defensive_assert!(missing == 0, "Should have unreserved the full amount");
+
+			// Set consumer refs to zero
+			let consumers = frame_system::Pallet::<T>::consumers(&from);
+			frame_system::Account::<T>::mutate(&from, |acc| {
+				acc.consumers = 0;
+			});
+			// We dont handle sufficients and there should be none
+			ensure!(frame_system::Pallet::<T>::sufficients(&from) == 0, Error::<T>::InternalError);
```
Ank4n: this can be moved at the beginning of the function.

pallets/ah-ops/src/lib.rs
```diff
@@ -494,6 +564,201 @@ pub mod pallet {
...
+				}
+			}
+
+			// Apply consumer refs
+			frame_system::Account::<T>::mutate(&to, |acc| {
+				acc.consumers += consumers;
+			});
+
+			// Reapply the holds
+			for hold in &holds {
```
Ank4n: This can create some inconsistencies, no? The pallet that placed the hold likely keeps some state referencing the old account, and we are not updating the pallet state. Maybe we should just skip applying the holds (and freezes) to the new account (except staking).

pallets/ah-ops/src/lib.rs
```diff
@@ -344,6 +389,31 @@ pub mod pallet {
...
+		#[pallet::weight(Weight::from_parts(100_000_000, 9000)
+				.saturating_add(T::DbWeight::get().reads_writes(20, 20)))]
+		pub fn translate_para_sovereign_child_to_sibling_derived(
+			origin: OriginFor<T>,
+			para_id: u16,
+			derivation_path: Vec<u16>,
+			old_account: T::AccountId,
+			new_account: T::AccountId,
+		) -> DispatchResult {
+			ensure_signed_or_root(origin)?;
```
Ank4n: Maybe we should just limit this to only known account/s that needs to be migrated? Since its a permissionless call, it increases the attack surface (and need for  a proper audit).

pallets/ah-ops/src/lib.rs
ggwpez: Yea even if it does not break stuff, it would be better to not change existing indices. Will change.

pallets/ah-ops/src/lib.rs
ggwpez: Yes not sure about this yet. Will talk about it with Donal, but it could make sense to restrict to root or whitelist the accounts that can be migrated.

pallets/ah-ops/src/lib.rs
ggwpez: hm yes.

ggwpez: Changed it now to only transfer native tokens and assets (and the staking).

pallets/ah-ops/src/lib.rs
ggwpez: I set it to a permissioned origin with Fellowship or Root now.

pallets/ah-ops/src/lib.rs
```diff
@@ -494,6 +553,153 @@ pub mod pallet {
...
+			Ok(())
+		}
+
+		/// Sovereign child account of a parachain (normally on the relay chain).
+		pub fn para_sov_child(id: u16) -> AccountId32 {
+			let mut raw = [0u8; 32];
+			raw[0..4].copy_from_slice(b"para");
+			raw[4..6].copy_from_slice(&id.encode());
+			raw.into()
+		}
```
franciscoaguirre, line 694: These could actually use the proper converter types but since it's a one-off it's fine

pallets/ah-ops/src/tests.rs
```diff
@@ -131,3 +134,92 @@ fn sovereign_account_translation() {
...
+		let derivation_path = vec![5, 2];
+
+		crate::Pallet::<AssetHub>::translate_para_sovereign_child_to_sibling_derived(
+			RuntimeOrigin::root(),
+			2004,
+			derivation_path.clone(),
+			child_5_2.clone(),
+			sibl_5_2.clone(),
+		)
+		.unwrap();
```
franciscoaguirre: `assert_ok` instead?

system-parachains/asset-hubs/asset-hub-polkadot/src/remote_tests.rs
```diff
@@ -0,0 +1,91 @@
...
+use super::*;
+use frame_support::traits::fungible::Inspect as FungibleInspect;
+use remote_externalities::{Builder, Mode, OfflineConfig};
+use sp_runtime::AccountId32;
+use std::{env::var, str::FromStr};
+
+#[tokio::test]
+async fn moonbeam_stellaswap_translation() {
+	sp_tracing::try_init_simple();
+	let Some(state_snapshot) = var("SNAP").map(|s| s.into()).ok() else {
```
franciscoaguirre, line 31: Should this file be committed if we don't commit the snapshot file? Maybe mention the block height from which the snapshot was taken? Adds a little more context for people finding this code later.

ggwpez: Yea this is mostly for debugging... we have more snapshot tests in the runtimes file but I can add a comment.

Will remove it here: https://github.com/polkadot-fellows/runtimes/pull/1016

PR #1037, opened 2025-12-16
asset-hub-kusama: Prepare ElasticScaling with 3 cores on asset hub kusama and enable buffered election

lexnv:
This PR configures the AssetHubKusama runtime to utilize elastic scaling with 3 cores.

- the runtime is prepared for elastic scaling with expected 2s block times
- 3 cores will be assigned to AHK when we have sufficient confidence in testing stacks
- AHK should function similarly with expected 6s block times until cores are assigned

The PR also enables buffered export of the validator set to the Relay Chain to handle block time changes from elastic scaling in KAH more gracefully. Without this, era lengths can shrink when multiple cores are used. 

See https://github.com/paritytech/polkadot-sdk/issues/10142.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ pallet-ah-migrator = { path = "pallets/ah-migrator", default-features = false }
 pallet-rc-migrator = { path = "pallets/rc-migrator", default-features = false }
 pallet-ah-ops = { path = "pallets/ah-ops", default-features = false }
 pallet-election-provider-multi-block = { version = "0.5.0", default-features = false }
-pallet-staking-async = { version = "0.7.0", default-features = false }
+pallet-staking-async = { version = "0.7.1", default-features = false }
 hex = { version = "0.4.3", default-features = false }
 rand = { version = "0.9.2" }
 impl-trait-for-tuples = { version = "0.2.3", default-features = false }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -89,106 +89,108 @@ kusama-runtime = { path = "relay/kusama", package = "staging-kusama-runtime" }
...
-pallet-salary = { version = "27.0.0", default-features = false }
-pallet-scheduler = { version = "43.0.0", default-features = false }
-pallet-session = { version = "42.0.0", default-features = false }
-pallet-session-benchmarking = { version = "42.0.0", default-features = false }
-pallet-society = { version = "42.2.0", default-features = false }
-pallet-staking = { version = "42.0.0", default-features = false }
-pallet-delegated-staking = { version = "9.0.0", default-features = false }
-pallet-staking-async-ah-client = { version = "0.5.0", default-features = false }
-pallet-staking-async-rc-client = { version = "0.5.0", default-features = false }
+pallet-revive = { version = "0.11.1", default-features = false }
```
Ank4n: Checked with @TorstenStueber and 0.9.0 is the latest version we can use if we don't want to include all the backported revive changes in https://github.com/paritytech/polkadot-sdk/pull/10552

https://github.com/paritytech/polkadot-sdk/blob/f7c06aee7e9a4aeaf264dd71e9fe470645e8b5c3/Plan.toml#L2002

PR #1041, opened 2025-12-18
ci: Test for metadata-hash feature more robustly

lexnv:
Sometimes the CI is wrongly detecting the metadata-hash feature as being enabled for glutton based runtimes:
- https://github.com/polkadot-fellows/runtimes/actions/runs/20337480608/job/58441235613#step:8:336

To improve this check:
- grep is used in quiet mode to avoid stdout/stderr pollution for our checks
- glutton runtimes are entirely skipped for the metadata-hash feature

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -267,7 +267,7 @@ jobs:
           FEATURES=""
           # Find out if the `metadata-hash` feature exists for the given package.
           # Exclude glutton runtimes as they don't support metadata-hash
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -266,7 +266,8 @@ jobs:
 
           FEATURES=""
           # Find out if the `metadata-hash` feature exists for the given package.
-          if cargo metadata --format-version=1 | jq '.packages[] | select(.name=="${{ matrix.runtime.package }}") | .features' | grep metadata-hash; then
+          # Exclude glutton runtimes as they don't support metadata-hash
```
bkchr, line 269: Ser, the point of thks if is to filter out packages that don't have the feature. So, we should fix this and not ignoring particular packages. I remember having fixed this at some point, but maybe my memory is not correct.

PR #1043, opened 2025-12-22
Install solc/resolc via `install-deps`

bkchr:
It is better to use the already existing infrastructure instead of adding another way we to install dependencies.

This pull request also ensures we are forwarding return values in `/cmd` spawned processes to e.g. fail on errors.

### DIFF
--- a/.github/install-deps.sh
+++ b/.github/install-deps.sh
@@ -1,7 +1,5 @@
 #!/usr/bin/env bash
 
-set -eu
-
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 source "$SCRIPT_DIR/env"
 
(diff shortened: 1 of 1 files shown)


xlc (approved):
I still don't fully understand why solc is needed for runtimes?

sigurpol:
> I still don't fully understand why solc is needed for runtimes?

AFAIK (`~0`), it is only needed for benchmarks / tests , not for production builds. This PR aims to fix couple of issues found while running benchmarks see https://github.com/polkadot-fellows/runtimes/actions/runs/20395561412/job/58610645713
1. compilation fails due to missing `solc` compiler in the logs above
2. ...but despite the epic fail, we return success 😓

athei (approved):

sigurpol (approved):

acatangiu (approved):

athei:
Apparently, `cargo clippy --all-features --all-targets` does also add the dependency. This is what broke the CI here, too.

PR #1045, opened 2026-01-02
Bridge-Hubs: Add proxy pallet

bkchr:


system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -533,6 +534,110 @@ impl pallet_utility::Config for Runtime {
...
+		}
+	}
+
+	fn is_superset(&self, o: &Self) -> bool {
+		match (self, o) {
+			(x, y) if x == y => true,
+			(ProxyType::Any, _) => true,
+			(_, ProxyType::Any) => false,
+			(ProxyType::NonTransfer, ProxyType::Collator) => true,
+			_ => false,
```
ggwpez, line 611: At least from the definition, `NonTransfer` is also a superset of `CancelProxy`.

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
```diff
@@ -544,6 +545,110 @@ impl pallet_utility::Config for Runtime {
...
+		}
+	}
+
+	fn is_superset(&self, o: &Self) -> bool {
+		match (self, o) {
+			(x, y) if x == y => true,
+			(ProxyType::Any, _) => true,
+			(_, ProxyType::Any) => false,
+			(ProxyType::NonTransfer, ProxyType::Collator) => true,
+			_ => false,
```
ggwpez, line 622: ditto

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -533,6 +534,113 @@ impl pallet_utility::Config for Runtime {
...
+					RuntimeCall::CollatorSelection(_) |
+					RuntimeCall::Session(_) |
+					RuntimeCall::Utility(_) |
+					RuntimeCall::Multisig(_) |
+					RuntimeCall::Proxy(_) |
+					RuntimeCall::BridgeRelayers(pallet_bridge_relayers::Call::register { .. }) |
+					RuntimeCall::BridgeRelayers(pallet_bridge_relayers::Call::deregister { .. }) |
+					RuntimeCall::BridgeRelayers(
+						pallet_bridge_relayers::Call::claim_rewards { .. }
+					)
```
clangenb, line 587: These are the only three calls in the pallet_bridge_relayers, so listing them explicitly is the defensive approach in case there is going to be a transferring extrinsic in the future. I guess this is ok, but why chose this approach for this pallet, while not for all the other pallets?

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -533,6 +534,113 @@ impl pallet_utility::Config for Runtime {
...
+					RuntimeCall::Timestamp(_) |
+					RuntimeCall::CollatorSelection(_) |
+					RuntimeCall::Session(_) |
+					RuntimeCall::Utility(_) |
+					RuntimeCall::Multisig(_) |
+					RuntimeCall::Proxy(_) |
+					RuntimeCall::BridgeRelayers(pallet_bridge_relayers::Call::register { .. }) |
+					RuntimeCall::BridgeRelayers(pallet_bridge_relayers::Call::deregister { .. }) |
+					RuntimeCall::BridgeRelayers(
+						pallet_bridge_relayers::Call::claim_rewards { .. }
```
claravanstaden, line 586: What determines the necessity to add extrinsic here? Wondering if Snowbridge extrinsic should also be listed here.

bkchr: I'm open to add whatever you like that isn't transferring funds :D

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
bkchr: There is also `claim_rewards_to`. Which would result in way to transfer funds away, just rewards, but still

claravanstaden (approved):

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
clangenb: Ah, sorry my bad. I didn't see that. 🙈

Issue #1010, opened 2025-11-13
Rework PET job in `test.yml` workflow


rockbmb:
This repository contains a testing workflow in `./.github/workflows/test.yml`, which includes the Polkadot Ecosystem Test suite (presently disabled).

The PET job in this workflow is currently as follows:

https://github.com/polkadot-fellows/runtimes/blob/c1a537c4a2b2e55624c491703f5ebbc8c8f33888/.github/workflows/test.yml#L334-L339

This is not ideal, because with PET having recently gone beyond 100+ individual test suites, a single failure, which can happen for various reasons (RPC timeouts are frequent), will cause the above step to be retried from the beginning.

To mitigate this, each individual test suite should run independently of the others, and the timeout/retry mechanisms should apply individually.

PR #1060, opened 2026-01-23
Use GH runners for small jobs + disable ecosystem tests

bkchr:
This pull request changes the CI pipeline to use GH runners for small jobs to use the big machines for the heavier jobs. Also this disables the ecosystem tests because they are taking right now multiple hours. Alexanxdre will work on fixing them.

### DIFF
--- a/.github/workflows/check-migrations.yml
+++ b/.github/workflows/check-migrations.yml
@@ -52,11 +52,12 @@ jobs:
       - name: Checkout sources
         uses: actions/checkout@v4
 
-      - name: Try to restore snapshot from cache
+      - name: Check if snapshot exists in cache
         id: cache-restore
         if: github.event_name != 'schedule'
         uses: actions/cache/restore@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
skunert, line 1: This one is considered a smaller job? Not sure how the tests usually run.

.github/workflows/test.yml
```diff
@@ -351,7 +351,8 @@ jobs:
   ecosystem-tests:
     needs: [runtime-matrix, build-runtimes]
     runs-on: self-hosted
-    if: github.event_name == 'pull_request'
+    # TODO: Disabled for now
```
rockbmb, line 354: 👍

Will address this in #1010.

.github/workflows/test.yml
bkchr: Only certain jobs are running on the GH runners. The actual heave jobs then use the self hosted runners.

PR #951, opened 2025-10-09
Upgrade to `stable2512`

pandres95:
Co-authors: @johandroid (Candidate), @PolkadotDom (Rank 1).

## Requests

> [!NOTE]
> I'll be adding some additional requests whenever they start appearing

- [x] Update Encointer dependencies to `stable2512` (@clangenb / @brenzi).
- [x] ~Set the epoch value for the `fulu` chain (@claravanstaden / @bkontur)~ (already done in #1005)

PR #1065, opened 2026-01-26
Preparation for Release 2.1.0

sigurpol:
🚨🚨🚨 **THIS IS NOT THE FINAL  RELEASE PR** 🚨🚨🚨

This branch aims to prepare `Release 2.1.0` for  Polkadot (see  [release tracker](https://github.com/orgs/polkadot-fellows/projects/1/views/3)),  which is expected to land on chain no later than March 14th. We will release on Kusama 1 week before.

Notables changes that will be merged **after** this PR is merged:

Main content of this PR:
- bump Polkadot SDK from 2507 to 2512
  -  It builds on top of @pandres95 's awesome work in #951  - thanks Pablo and everyone who has contributed there :bow:

Some of key changes:
-   Upgrade SDK to `stable2512` (currently `stable2512-2-rc1`
-   Introduction of the first version of DAP pallet
-  All staking changes needed for [DAP phase 1](https://forum.polkadot.network/t/the-roadmap-for-the-dynamic-allocation-pool-dap/16511): session keys on AH, nominator fast-unbondanble and not slashable via config, no more treasury burns, slash burns redirected to DAP.  All these changes are Polkadot-only

## Notes

In retrospect, this PR should have solely focused on bumping the SDK to 2512. The scope became excessive, and I apologize for this.

## TODO:

In the scope of this PR:
- [x] review breaking changes coming with the bump to  sdk `2512`
- [x]  re-run benchmarks also for `pallet_staking_async` and  `pallet_staking_async_rc_client` (brand new on KAH and PAH!)

**After this PR is merged and before creating 2.1.0 release PR**,  we also still want the following PRs to land:

- @acatangiu 's [PR](https://github.com/sigurpol/runtimes/pull/5):  Full integration of SDK #9948 in fellowship runtimes for 2.1.0 release
- @dhirajs0's [PR](https://github.com/sigurpol/runtimes/pull/4) to support multi assets bounties

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Add hyperbridge to parachains header root ([#1073](https://github.com/polkadot-fellows/runtimes/pull/1073))
 
+### Changed
+
+- Set hard cap enactment block number to relay block number [30_354_008](https://polkadot.subscan.io/block/30354008) ([#1076](https://github.com/polkadot-fellows/runtimes/pull/1076))
+
 ## [2.0.5] 13.01.2026
 
 ### Added
--- a/system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
+++ b/system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
@@ -319,9 +319,9 @@ impl EraPayout {
 	// Taken from [AH Block 10469901](https://assethub-polkadot.subscan.io/event/10469901-6).
 	const PRE_HARD_CAP_DAILY_EMISSION: Balance = 328797u128 * UNITS;
 
-	// Calculated assuming a 11.7 minute per day time drift (A block time of 6.04875 seconds).
-	// https://polkadot.subscan.io/block/30349908
-	const HARD_CAP_START: BlockNumber = 30_349_908;
+	// Should be around 14th March 2026 noon UTC assuming 0.5-1% clock time drift.
+	// https://polkadot.subscan.io/block/30354008
+	const HARD_CAP_START: BlockNumber = 30_354_008;
 
 	// The hard issuance cap ratified in Referendum 1710.
 	const HARD_CAP_TARGET: Balance = 2_100_000_000u128 * UNITS;
@@ -878,7 +878,7 @@ mod tests {
 		}));
 	}
 
-	const MARCH_14_2026: RC_BlockNumber = 30_349_908;
+	const MARCH_14_2026: RC_BlockNumber = 30_354_008;
 	// The March 14, 2026 TI used for calculations in [Ref 1710](https://polkadot.subsquare.io/referenda/1710).
 	const MARCH_TI: u128 = 1_676_733_867 * UNITS;
 	const TARGET_TI: u128 = 2_100_000_000 * UNITS;


integration-tests/emulated/chains/parachains/assets/asset-hub-polkadot/src/lib.rs
```diff
@@ -40,6 +40,7 @@ decl_test_parachains! {
 			LocationToAccountId: asset_hub_polkadot_runtime::xcm_config::LocationToAccountId,
 			ParachainInfo: asset_hub_polkadot_runtime::ParachainInfo,
 			MessageOrigin: cumulus_primitives_core::AggregateMessageOrigin,
+			DigestProvider: AuraDigestProvider,
```
sigurpol: @skunert can you verify that these changes around AuraDigestProvider here and for the other chains  are correct?  I have used WAH on SDK as reference

pallets/remote-proxy/src/tests.rs
```diff
@@ -233,6 +233,8 @@ fn remote_proxy_works() {
 				who: 1,
 				proxy_type: ProxyType::Any,
 				disambiguation_index: 0,
+				at: <Test as pallet_proxy::Config>::BlockNumberProvider::current_block_number(),
```
sigurpol, line 236: @bkchr  can you please verify that changes in remote-proxy tests are correct?

relay/kusama/src/lib.rs
```diff
@@ -2802,6 +2786,13 @@ sp_api::impl_runtime_apis! {
 			let nodes = leaves.into_iter().map(|leaf|mmr::DataOrHash::Data(leaf.into_opaque_leaf())).collect();
 			pallet_mmr::verify_leaves_proof::<mmr::Hashing, _>(root, nodes, proof)
 		}
+
+		fn generate_ancestry_proof(
+			prev_block_number: BlockNumber,
+			best_known_block_number: Option<BlockNumber>,
+		) -> Result<mmr::AncestryProof<mmr::Hash>, mmr::Error> {
+			Mmr::generate_ancestry_proof(prev_block_number, best_known_block_number)
```
sigurpol, line 2780: @Lederstrumpf  can you double check if these changes are correct ? Seems to be aligned with what we are doing in the SDK for Westend for example (SDK 9295)

relay/polkadot/src/lib.rs
```diff
@@ -2916,7 +2909,7 @@ sp_api::impl_runtime_apis! {
 		}
 
 		fn execute_block(
-			block: Block,
+			block: <Block as BlockT>::LazyBlock,
```
sigurpol, line 2905: @serban300 can you verify that the changes here and in many many files where we are  introducing lazy block are correct ? I have used SDK  / Westend as reference (e.g. SDK 9480)

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
```diff
@@ -374,8 +375,15 @@ parameter_types! {
...
 
-	/// Maximum numbers that we prune from pervious eras in each `prune_era` tx.
+	/// Maximum numbers that we prune from previous eras in each `prune_era` tx.
 	pub MaxPruningItems: u32 = 100;
+
+	/// Unlike Polkadot, Kusama nominators are expected to be slashable and do not
+	/// support fast unbonding. Consequently, AreNominatorSlashable is intended to
+	/// remain set to true and should not be modified via governance.
+	/// NominatorFastUnbondDuration value below is therefore ignored.
+	pub const NominatorFastUnbondDuration: sp_staking::EraIndex = 2;
```
sigurpol: sanity check @Ank4n , @kianenigma  please 🙏

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -355,6 +355,7 @@ impl pallet_assets::Config<TrustBackedAssetsInstance> for Runtime {
 	type CallbackHandle = pallet_assets::AutoIncAssetId<Runtime, TrustBackedAssetsInstance>;
 	type AssetAccountDeposit = AssetAccountDeposit;
 	type RemoveItemsLimit = frame_support::traits::ConstU32<1000>;
+	type ReserveData = ();
```
sigurpol, line 359: @acatangiu  can you double check that `type ReserveData = ()` is correct here and below? I have taken SDK 9948 and related WAH changes as reference

system-parachains/asset-hubs/asset-hub-kusama/tests/tests.rs
```diff
@@ -408,10 +409,19 @@ fn receive_reserve_asset_deposited_dot_from_asset_hub_polkadot_fees_paid_by_pool
...
 
 	let foreign_asset_id_location = Location::new(2, [GlobalConsensus(Polkadot)]);
+	let reserve_location = Location::new(2, [GlobalConsensus(Polkadot), Parachain(1000)]);
+	let foreign_asset_reserve_data =
+		ForeignAssetReserveData { reserve: reserve_location, teleportable: false };
 	let foreign_asset_id_minimum_balance = 1_000_000_000;
 	// sovereign account as foreign asset owner (can be whoever for this scenario)
 	let foreign_asset_owner = LocationToAccountId::convert_location(&Location::parent()).unwrap();
-	let foreign_asset_create_params =
+	let foreign_asset_create_params = (
```
sigurpol, line 418: @acatangiu  same here - more or less copy-pasted from WAH changes when you added reserves info stuff

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -437,13 +443,19 @@ parameter_types! {
 	pub const SessionsPerEra: SessionIndex = prod_or_fast!(6, 1);
 	pub const RelaySessionDuration: BlockNumber = prod_or_fast!(4 * RC_HOURS, RC_MINUTES);
 	pub const BondingDuration: sp_staking::EraIndex = 28;
+	/// Nominators are expected to be slashable and support fast unbonding
+	/// depending on AreNominatorSlashable storage value, as set by governance.
+	/// NominatorFastUnbondDuration value below is ignored if nominators are slashable.
+	pub const NominatorFastUnbondDuration: sp_staking::EraIndex = 2;
```
sigurpol, line 450: @Ank4n , @kianenigma  sanity check please 🙇

system-parachains/bridge-hubs/bridge-hub-polkadot/Cargo.toml
```diff
@@ -123,8 +123,12 @@ snowbridge-outbound-queue-runtime-api = { workspace = true }
 snowbridge-outbound-queue-v2-runtime-api = { workspace = true }
 snowbridge-inbound-queue-primitives = { workspace = true }
 snowbridge-runtime-common = { workspace = true }
+snowbridge-pallet-inbound-queue-fixtures = { workspace = true, optional = true }
+snowbridge-pallet-inbound-queue-v2-fixtures = { workspace = true, optional = true }
 
 [dev-dependencies]
+# TODO: Remove when https://github.com/paritytech/polkadot-sdk/issues/10962 is closed
```
sigurpol, line 130: @pandres95 , is anyone working on this ?

CHANGELOG.md
```diff
@@ -4,15 +4,21 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## [Unreleased]
+## [2.0.7] 13.02.2026
```
sigurpol: We will have to fix the date when time comes...

system-parachains/bridge-hubs/bridge-hub-polkadot/primitives/src/lib.rs
```diff
@@ -166,6 +166,8 @@ pub mod snowbridge {
 	parameter_types! {
 		/// Should match the `ForeignAssets::create` index on Asset Hub.
 		pub const CreateAssetCall: [u8;2] = [53, 0];
+		/// Should match the `ForeignAssets::set_reserves` index on Asset Hub.
```
sigurpol, line 169: @claravanstaden could you kindly verify that changes here and in general on bridge-hub make sense? Not the ones you introduced, but mine - I trust myself even less than usual here 😄

claravanstaden: @sigurpol this is correct, the set_reserves config is right since [ForeignAssets = `53`](https://github.com/sigurpol/runtimes/blob/sigurpol-release-2_0_6/system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs#L1529) and the call index in [pallet assets is `33`](https://github.com/paritytech/polkadot-sdk/blob/stable2512/substrate/frame/assets/src/lib.rs#L1870).

sigurpol: thanks for checking :heart:

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
```diff
@@ -446,40 +472,54 @@ pub enum RelayChainRuntimePallets {
 
 #[derive(Encode, Decode)]
 pub enum AhClientCalls {
-	// index of `fn validator_set` in `staking-async-ah-client`. It has only one call.
+	// index of `fn validator_set` in `staking-async-ah-client`.
 	#[codec(index = 0)]
 	ValidatorSet(rc_client::ValidatorSetReport<AccountId>),
+	// index of `fn set_keys_from_ah` in `staking-async-ah-client`.
+	#[codec(index = 3)]
+	SetKeys { stash: AccountId, keys: Vec<u8> },
```
sigurpol, line 483: @Ank4n , @kianenigma  please double check the session keys on AH bits here and in asset-hub-polkadot 🙏

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -111,7 +111,7 @@ impl sp_runtime::traits::Get<u32> for RebagIffMigrationDone {
 			pallet_ah_migrator::MigrationEndBlock::<Runtime>::get()
 				.is_some_and(|n| frame_system::Pallet::<Runtime>::block_number() > n + 1)
 		{
-			5
+			10
```
sigurpol, line 114: @Ank4n , @kianenigma . Same as AutoRebagNumber in Westend.  Both 5 and 10 are arbitrary. 5 let the benchmark on bags-list::on_idle fail  though 😄 
```
Expected exactly 5 rebagged nodes, found 4                                    
  [VoterList]Error during rebagging: List
```
   This is a  bug triggered by MaxAutoRebagPerBlock = 5.
  The benchmark computes:                
  - rebag_budget = MaxAutoRebagPerBlock::get()  // 5 for Kusama/Polkadot, 10 for Westend
  - pending_count = rebag_budget / 3            // 1 for budget=5, 3 for budget=10

 Then for pending nodes:
```rust
  if i % 7 != 0 {
      T::ScoreProvider::set_score_of(&pending_node, pending_score);
  }
```

With pending_count=1, only i=0 runs, and 0 % 7 == 0  so the only pending node gets no score set. When on_idle tries to rebag it, it fails ("Error during rebagging: List") => Result: 4 instead of 5.

Aligned the behavior with WAH AutoRebagNumber.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -2922,6 +2931,21 @@ mod tests {
 		assert!(ProxyType::StakingOperator
 			.filter(&RuntimeCall::Staking(pallet_staking_async::Call::kick { who: vec![] })));
 
+		// - StakingRcClient::set_keys (session key management on Asset Hub)
```
sigurpol, line 2979: @kianenigma , @Ank4n  in https://github.com/polkadot-fellows/runtimes/pull/1065/changes/73fdf84a39fd2f7dce7eff3d31eb6b3614ab10d6 I have made StakingOperator AH only and removed the whole RC configuration (not a breaking change since it was not shipped yet. Please verify it makes sense. It should be aligned with runtime changes in westend in https://github.com/paritytech/polkadot-sdk/pull/10980

yrong:
> Runtime wise, is it just matter of bumping crate version or do they require changes in runtime config?

Yes, only bumps the crate version; there are no changes to the runtime configuration and no migrations are required.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -714,7 +717,11 @@ impl InstanceFilter<RuntimeCall> for ProxyType {
 				RuntimeCall::Staking(pallet_staking_async::Call::validate { .. }) |
 					RuntimeCall::Staking(pallet_staking_async::Call::chill { .. }) |
 					RuntimeCall::Staking(pallet_staking_async::Call::kick { .. }) |
```
Ank4n, line 721: should this proxy be allowed to kick nominators? I think the operator should not be able to do this.

sigurpol: I think so. The current implementation - beside setting/purging keys, allows validate / chill / kick because I assumed that all of these are "operational" calls that don't move funds, destination rewards or validator's bonded stake.  But ofc debatable. Asked @kianenigma  and Michalis for their opinion too.

sigurpol: From Michalis:
> I think we should allow it. In white label deals you typically want only the one nomination and having either the stash or the operator kick any excess nominations is valid operationally. After all the assumption here is always that the StakingOperator won't act maliciously in any case, so allowing it for them is not a problem IMO.

kianenigma: I can see that there are two definitions for operator:

I (big bag entity) do everything, you (operator) just run the node
I (big bag entity) do nothing other than giving you my bag, you (operator) do the rest

atm we have chosen 2. I would look at the security of it: In both cases, the operator can screw the bag provider, in option 2 it is somewhat easier to do this. I think it if fair to say that everyone should trust their staking operator. I invite others to also think if there is a security difference that I might have missed here.

So the option chosen here seems fine.

sigurpol: I don't necessarily agree with the fact we are doing 2. honestly,  we are still preventing nominate and set_payee, we are not allowing operator to touch funds. What kick adds to the equation (in my mind) is just the option to handle the use case validate{blocked:true} + kick(), allowing some nominators hygiene and no more than that

kianenigma:
/cmd bench --runtime asset-hub-kusama --pallet pallet_staking_async

kianenigma:
/cmd bench --runtime asset-hub-polkadot --pallet pallet_staking_async

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
sigurpol: we discussed offline with Michalis (W3F) and we have decided to ship this variant (chill + kick + validate + keys for 2.1.0) - we will re-assess later if needed.

sigurpol:
> @skunert would you mind to help about the failures we are seeing in all integration-tests e.g. https://github.com/polkadot-fellows/runtimes/actions/runs/21755806769/job/62765666400?pr=1065 ?
> 
> All tests are failing here:
> 
> ```shell
> Parachain slot must match relay-derived slot:
>   parachain_slot=Slot(2), derived_from_relay_slot=Slot(1),
>   velocity=3, relay_chain_slot=Slot(2)
> ```
> 


I believe a potential fix might consist of two parts:
- replace AuraDigestProvider in the emulated tests with an ad-hoc Digest provider that adjusts the Aura slot for chains where the parachain slot duration differs from relay chain slot duration
-  change this line https://github.com/paritytech/polkadot-sdk/blob/stable2512/cumulus/xcm/xcm-emulator/src/lib.rs#L729-L729 in xcm_emulator with something like `now: *pallet_aura::CurrentSlot::<$runtime::Runtime>::get() * slot_duration`

CHANGELOG.md
```diff
@@ -4,15 +4,21 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## [Unreleased]
+## [2.1.0] 13.02.2026
```
kianenigma: Just FYI if needed you can also merge this and trigger the release separately. We usually do them separate. Just an option that you have.

relay/kusama/src/lib.rs
```diff
@@ -1371,20 +1365,7 @@ impl InstanceFilter<RuntimeCall> for TransparentProxyType {
 					RuntimeCall::Utility(..) |
 					RuntimeCall::FastUnstake(..) |
 					RuntimeCall::VoterList(..) |
-					RuntimeCall::NominationPools(..) |
```
kianenigma, line 1374: These changes were never deployed right? else removing a proxy type is a hassle for existing proxies.

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
kianenigma: > NominatorFastUnbondDuration value below is therefore ignored.

Correct, But still I don't see a reason to set it 2, no?


```suggestion
	pub const NominatorFastUnbondDuration: sp_staking::EraIndex = BondingDuration::get();
```

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
```diff
@@ -417,15 +429,40 @@ impl pallet_staking_async::Config for Runtime {
...
 impl pallet_staking_async_rc_client::Config for Runtime {
 	type RelayChainOrigin = EnsureRoot<AccountId>;
 	type AHStakingInterface = Staking;
 	type SendToRelayChain = StakingXcmToRelayChain;
 	type MaxValidatorSetRetries = ConstU32<64>;
-	// Export elected validator set at the end of the 4th session of the era.
-	type ValidatorSetExportSession = ConstU32<4>;
+	type ValidatorSetExportSession = ValidatorSetExportSession;
+	type RelayChainSessionKeys = RelayChainSessionKeys;
+	type Balance = Balance;
```
kianenigma, line 454: Suggestion: you can use chopsticks to fork Ah and relay, and test setting session keys once. It will be the ultimate test for this.

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
```diff
@@ -417,15 +429,40 @@ impl pallet_staking_async::Config for Runtime {
...
+	// |---------------------|---------|------------|-----------|
+	// | grandpa             | Ed25519 | 32 bytes   | 64 bytes  |
+	// | babe                | Sr25519 | 32 bytes   | 64 bytes  |
+	// | para_validator      | Sr25519 | 32 bytes   | 64 bytes  |
+	// | para_assignment     | Sr25519 | 32 bytes   | 64 bytes  |
+	// | authority_discovery | Sr25519 | 32 bytes   | 64 bytes  |
+	// | beefy               | ECDSA   | 33 bytes   | 65 bytes  |
+	// Buffer for SCALE encoding overhead and future expansions.
+	type MaxSessionKeysLength = ConstU32<256>;
+	type WeightInfo = ();
```
kianenigma: needs accurate weights?

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
kianenigma: checked: https://docs.rs/crate/pallet-staking-async-ah-client/0.7.1/source/src/lib.rs#634

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
```diff
@@ -417,15 +429,40 @@ impl pallet_staking_async::Config for Runtime {
 	type MaxEraDuration = MaxEraDuration;
 	type WeightInfo = weights::pallet_staking_async::WeightInfo<Runtime>;
 	type MaxPruningItems = MaxPruningItems;
+	type NominatorFastUnbondDuration = NominatorFastUnbondDuration;
+}
+
+// Must match Kusama relay chain's `SessionKeys` structure for encoding/decoding compatibility.
+sp_runtime::impl_opaque_keys! {
+	pub struct RelayChainSessionKeys {
```
kianenigma, line 437: Review notes: On the relay chain, these are assigned to the corresponding pallets (e.g. `GrandPa`).

We need to look at the `impl BoundToRuntimeAppPublic` for each pallet, and make sure the right type is used.

Some are tricky, please double check. For example, for beefy we should use `ecdsa_crypto`.

Note: All that matters here is that these keys are decode-able -> have the same length.

relay/kusama/src/lib.rs
sigurpol: correct, they were never part of any release - they were meant to be the decent-but-not-amazing fallback solution in case we didn't have time to add support for session keys on AH

CHANGELOG.md
sigurpol: yeah, maybe not a bad idea - let's see how many contributions we expect as well beside the multi-asset bounties PR already opened and in review against this branch

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
```diff
@@ -417,15 +429,40 @@ impl pallet_staking_async::Config for Runtime {
...
+	// | Key                 | Crypto  | Public Key | Signature |
+	// |---------------------|---------|------------|-----------|
+	// | grandpa             | Ed25519 | 32 bytes   | 64 bytes  |
+	// | babe                | Sr25519 | 32 bytes   | 64 bytes  |
+	// | para_validator      | Sr25519 | 32 bytes   | 64 bytes  |
+	// | para_assignment     | Sr25519 | 32 bytes   | 64 bytes  |
+	// | authority_discovery | Sr25519 | 32 bytes   | 64 bytes  |
+	// | beefy               | ECDSA   | 33 bytes   | 65 bytes  |
+	// Buffer for SCALE encoding overhead and future expansions.
+	type MaxSessionKeysLength = ConstU32<256>;
```
kianenigma, line 464: Sorry I missed this in the pallet review: This is a bit redundant and can be removed. The session keys here are never stored in the storage, so don't need a bound. We don't need to bound transaction inputs or things that are just passed in the runtime via traits.

What needs bounding is only storage. 
For transaction inputs, we do need to `Box<>` things with a large static size though, as the stack/staic-size of an `enum` (in this case `RuntimeCall`) in Rust is the size of the largest variant.

(I really need to start channeling my review notes to a claude skill that can replace me 🙈)

kianenigma: No need to backport or anything, you can fix it later. Removing bound and/or adding a `Box` is non-breaking.

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
```diff
@@ -438,40 +475,54 @@ pub enum RelayChainRuntimePallets {
...
+		};
+		rc_client::build_transact_xcm(encoded_call)
 	}
 }
 
 parameter_types! {
 	pub RelayLocation: Location = Location::parent();
+	/// Conservative RC execution cost for set/purge keys operations.
+	/// ~3x of Kusama relay benchmarked session set/purge_keys (~61-62M ref_time, ~16538 proof).
+	pub RemoteKeysExecutionWeight: Weight = Weight::from_parts(190_000_000, 50_000);
```
kianenigma, line 517: 16k proof for a storage item that sets a ~200byte value in storage? interesting.. 

Probably not a big deal, but made me raise an eyebrow.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -869,8 +879,7 @@ impl cumulus_pallet_parachain_system::Config for Runtime {
 	type CheckAssociatedRelayNumber = RelayNumberMonotonicallyIncreases;
 	type ConsensusHook = ConsensusHook;
 	type WeightInfo = weights::cumulus_pallet_parachain_system::WeightInfo<Runtime>;
-	type SelectCore = cumulus_pallet_parachain_system::DefaultCoreSelector<Runtime>;
-	type RelayParentOffset = ConstU32<RELAY_PARENT_OFFSET>;
+	type RelayParentOffset = ConstU32<0>;
```
kianenigma: Should be double-checked by the original author of the sdk-pr.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -1703,6 +1710,7 @@ pub type BlockId = generic::BlockId<Block>;
 pub type TxExtension = cumulus_pallet_weight_reclaim::StorageWeightReclaim<
 	Runtime,
 	(
+		frame_system::AuthorizeCall<Runtime>,
```
kianenigma, line 1715: AFAIK this will be needed for later integrate with people chain, but not strictly now. @georgepisaltu correct?

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
kianenigma: If it was never deployed, then it is all good. If it was deployed, we better keep it and think of a better strategy.

system-parachains/asset-hubs/asset-hub-kusama/tests/tests.rs
```diff
@@ -1029,3 +1039,165 @@ fn staking_proxy_can_manage_staking_operator() {
...
+
+#[test]
+fn session_keys_are_compatible_between_ah_and_rc() {
+	use asset_hub_kusama_runtime::staking::RelayChainSessionKeys;
+	use sp_runtime::traits::OpaqueKeys;
+
+	// Verify the key type IDs match in order.
+	// This ensures that when keys are encoded on AssetHub and decoded on Kusama (or vice versa),
+	// they map to the correct key types.
+	assert_eq!(
```
kianenigma, line 1198: Great!

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
kianenigma: The root of the issue is ultimately that the benchmark for `auto_rebag` is too dependant on its state and configs (how many nodes exists, how many can be rebagged, what is the rebah budget). Happy with the parameter here, not a big difference if benchmarks are happy.

The benchmark measures the weight for the entirety of the on-idle call, so as long as the count in production and benchmarking is hte same, we are good. 

The system should be able to easily handle 10, and there is a weight backstop as well.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
acatangiu: needs more changes from that PR (including a migration) - I will prepare a PR against your branch

sigurpol:
> > Command "bench --runtime asset-hub-polkadot --pallet pallet_staking_async" has finished ✅ [See logs here](https://github.com/polkadot-fellows/runtimes/actions/runs/21900595186)
> > Details
> > Subweight results:
> > File	Extrinsic	Old	New	Change [%]
> > system-parachains/asset-hubs/asset-hub-polkadot/src/weights/pallet_staking_async.rs	process_offence_queue	804.77us	14.58ms	+1711.28
> > system-parachains/asset-hubs/asset-hub-polkadot/src/weights/pallet_staking_async.rs	force_apply_min_commission	197.26us	2.74ms	+1286.90
> 
> This is **unexpected** since e.g. `force_apply_min_commission` consists in two read + 1 write. I can reproduce the issue running on my Ubuntu desktop machine too. Moreover, same benchmark takes ms on PAH/KAH but ~55µs on WAH. SDK 2512 include two fixes aiming to avoid test setup to leak in benchmark results [paritytech/polkadot-sdk#10802](https://github.com/paritytech/polkadot-sdk/pull/10802) + [paritytech/polkadot-sdk#10947](https://github.com/paritytech/polkadot-sdk/pull/10947) , both contained in frame-omni-bencher `v0.17.1`. These fixes on WAH (SDK repo) brought `force_apply_commission extrinsic` time down from ms to µs. The cure doesn't seem to work as well on PAH/KAH.
> 
> Further tests on my local machine:
> 
> * on WAH removing the changes to [whitelist the dummy key and do a fake r/w](https://github.com/paritytech/polkadot-sdk/blob/stable2512/substrate/frame/benchmarking/src/utils.rs#L298-L306) makes the 2r-1w force_apply_min_commision back from 55 µs to few ms (proving the - at least - partial validity of SDK 10974)
> * on PAH using the above "patched" frame-omni-bencher increase the result (from 3265µs to 5030µs)
> * on PAH using vanilla frame-omni-bencher (v0.17.1) adding `--genesis-builder none` brings result down from ms to µs (!)
> * on WAH using vanilla frame-omni-bencher (v0.17.1) adding  `--genesis-builder none` brings result down from 55 us to 9.9µs
> 
> Note : on SDK, CI uses for WAH the equivalent of `--genesis-builder runtime --genesis-builder-preset development` which also has > 26k dev stakers
> 
> cc @kianenigma , @Ank4n

I have investigated the issue - the regression in benchmarks comes with the update of trie-db from 0.30.0 to 0.31.0 - if we use a frame-omni-bencher still pointing at 0.30.0 then we are back to `µs`.
I've created an issue https://github.com/paritytech/trie/issues/230 to tackle this.
A temporary pragmatic fix for this PR might be to just point to an hard-coded version of frame-omni-bencher like v0.16.0 which points to trie-db 0.30.0. Hopefully that CI can build /install it properly, locally I had to do some massage since I have too new clang/gcc toolchain so I had to use stuff like  `CXXFLAGS="-include cstdint" cargo build -p frame-omni-bencher --release`.
What do you think @kianenigma , @ggwpez , @Ank4n ? Does it make sense to go with the pragmatic approach and benchmark staking-async with this ad-hoc frame-omni-bencher?

CHANGELOG.md
sigurpol: Done that, considering also @acatangiu  and @dhirajs0  extra PR (plus potentially bump crates for a revive PR from @yrong )

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
sigurpol: @lexnv  can you help with this one?

relay/kusama/constants/src/lib.rs
```diff
@@ -186,10 +186,6 @@ pub mod proxy {
 		Spokesperson,
 		#[codec(index = 10)]
 		ParaRegistration,
-		/// Operator proxy for validators. Can only manage session keys.
-		/// Cannot do staking operations (those are on Asset Hub).
-		#[codec(index = 11)]
-		StakingOperator,
```
ggwpez, line 192: Needed? I think there is a chance that this makes the storage undecodable if someone (for whatever reason) has such a variant. We can do the cleanup (including migration) at another time.

relay/kusama/src/lib.rs
```diff
@@ -685,7 +679,7 @@ impl pallet_bags_list::Config<VoterBagsListInstance> for Runtime {
 	type BagThresholds = BagThresholds;
 	type Score = sp_npos_elections::VoteWeight;
 	#[cfg(feature = "runtime-benchmarks")]
-	type MaxAutoRebagPerBlock = ConstU32<5>;
+	type MaxAutoRebagPerBlock = ConstU32<10>;
```
ggwpez, line 682: To make it faster?

relay/kusama/constants/src/lib.rs
ggwpez: PS just read the other discussion :)

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
ggwpez: Maybe we can add it to kusama/constants crate and have a test in the kusama runtime to check that they are correct?  
But maybe that would blow up this diff here too much.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
ggwpez, line 359:
```suggestion
	type ReserveData = (); // TODO FIXME BEFORE 2.1
```
? we can filter for that in the final 2.1 release MR

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
```diff
@@ -395,6 +396,7 @@ impl pallet_assets::Config<PoolAssetsInstance> for Runtime {
 	type Extra = ();
 	type WeightInfo = weights::pallet_assets_pool::WeightInfo<Runtime>;
 	type CallbackHandle = ();
+	type ReserveData = ();
```
ggwpez, line 401: ditto

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
ggwpez: This wont break offline signers, right?

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -130,19 +131,24 @@ impl pallet_bags_list::Config<VoterBagsListInstance> for Runtime {
 parameter_types! {
 	pub const DelegatedStakingPalletId: PalletId = PalletId(*b"py/dlstk");
 	pub const SlashRewardFraction: Perbill = Perbill::from_percent(1);
+	pub const DapPalletId: PalletId = PalletId(*b"dap/buff");
```
ggwpez, line 134: I think we normally make them start with `py/` but admittingly i dont know why 😆

system-parachains/asset-hubs/asset-hub-polkadot/src/genesis_config_presets.rs
```diff
@@ -49,13 +54,16 @@ fn asset_hub_polkadot_genesis(
 	foreign_assets: Vec<(Location, AccountId, Balance)>,
 	foreign_assets_endowed_accounts: Vec<(Location, AccountId, Balance)>,
 ) -> serde_json::Value {
+	let mut balances: Vec<(AccountId, Balance)> = endowed_accounts
+		.iter()
+		.cloned()
+		.map(|k| (k, ASSET_HUB_POLKADOT_ED * 4096 * 4096))
+		.collect();
+	balances.push((dap_buffer_account(), ASSET_HUB_POLKADOT_ED));
```
ggwpez, line 62: Do we have a migration that ensures this upon launch?

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
```diff
@@ -1555,6 +1564,9 @@ construct_runtime!(
 		// Contracts
 		Revive: pallet_revive = 90,
 
+		// DAP (Delegated Account for Penalties)
+		Dap: pallet_dap = 100,
```
ggwpez: Could go to "Monetary stuff" section?

CHANGELOG.md
```diff
@@ -10,9 +10,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Add `pallet-assets` on Polkadot People to support HOLLAR (and more assets in the future) ([polkadot-fellows/runtimes/pull/962](https://github.com/polkadot-fellows/runtimes/pull/962))
 
+
 ### Changed
```
ggwpez, line 13: I think 2.0.6 was eaten by a merge 😆

relay/polkadot/src/lib.rs
serban300: Yes, looks good ! All the implementations of `execute_block()` and `check_inherents()` should use `<Block as BlockT>::LazyBlock` instead of `Block` now.

relay/kusama/src/lib.rs
serban300: LGTM !

CHANGELOG.md
sigurpol: we didn't backport https://github.com/polkadot-fellows/runtimes/pull/1069 to main as far as I can tell :sweat_smile:

kianenigma: We are now doing this same again for 2.0.7. 

https://github.com/polkadot-fellows/runtimes/pull/1074#issuecomment-3908534608

sigurpol: one 2.0.7 is out,  I will sanitize the CHANGELOG in this PR :smile:

relay/kusama/src/lib.rs
Lederstrumpf: @sigurpol sorry, I'd missed your first tag on this.
The implementation declaration changes also lgtm (both here & [the polkadot change](https://github.com/polkadot-fellows/runtimes/pull/1065/changes#diff-e7afb81f3cef77f8a1686085f1adaa992122790270bae77eb2e640ad40b12cf4R2641-R2647)).

~~But I believe what's missing is a bump in the `api_version` of both `mmr::MmrApi` (to account for the `generate_ancestry_proof` method's addition there) & `beefy_primitives::BeefyApi` (to account for the method's disappearance there).~~

relay/kusama/src/lib.rs
```diff
@@ -2746,16 +2726,6 @@ sp_api::impl_runtime_apis! {
...
-			best_known_block_number: Option<BlockNumber>,
-		) -> Option<sp_runtime::OpaqueValue> {
-			Mmr::generate_ancestry_proof(prev_block_number, best_known_block_number)
-				.map(|p| p.encode())
-				.map(OpaqueKeyOwnershipProof::new)
-				.ok()
-		}
 	}
 
 	impl mmr::MmrApi<Block, Hash, BlockNumber> for Runtime {
```
Lederstrumpf, line 2731:
```suggestion
	#[api_version(3)]
	impl mmr::MmrApi<Block, Hash, BlockNumber> for Runtime {
```

relay/kusama/src/lib.rs
Lederstrumpf, line 2677:
```suggestion
	#[api_version(6)]
	impl beefy_primitives::BeefyApi<Block, BeefyId> for Runtime {
```

relay/polkadot/src/lib.rs
Lederstrumpf, line 2569:
```suggestion
	#[api_version(6)]
	impl beefy_primitives::BeefyApi<Block, BeefyId> for Runtime {
```

relay/polkadot/src/lib.rs
```diff
@@ -2629,16 +2610,6 @@ sp_api::impl_runtime_apis! {
...
-			best_known_block_number: Option<BlockNumber>,
-		) -> Option<sp_runtime::OpaqueValue> {
-			Mmr::generate_ancestry_proof(prev_block_number, best_known_block_number)
-				.map(|p| p.encode())
-				.map(OpaqueKeyOwnershipProof::new)
-				.ok()
-		}
 	}
 
 	impl mmr::MmrApi<Block, Hash, BlockNumber> for Runtime {
```
Lederstrumpf, line 2622:
```suggestion
	#[api_version(3)]
	impl mmr::MmrApi<Block, Hash, BlockNumber> for Runtime {
```

relay/kusama/src/lib.rs
bkchr: Not needed, it picks the latest automatically.

relay/kusama/src/lib.rs
bkchr: Not needed, it picks the latest automatically.

relay/kusama/src/lib.rs
Lederstrumpf: TIL, thanks!

relay/kusama/src/lib.rs
sigurpol: to make it work 😄  - see the other equivalent thread under relay/polkadot

relay/kusama/src/lib.rs
sigurpol: TIL as well - thanks to both of you!

relay/kusama/src/lib.rs
sigurpol: thank you 🙇 !

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
sigurpol: done

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
sigurpol: in this PR, I have introduced  a test both on Polkadot and Kusama `session_keys_are_compatible_between_ah_and_rc` which verifies
 key_ids() match. 
@kianenigma , for beefy we are indeed using `ecdsa_crypto` already on AH. I have double-checked again key by key and AH and RC are aligned and have identical RelayChainSessionKeys structure.
so they both rely on RC pallets' BoundToRuntimeAppPublic impls in polkadot-sdk (ed25519 for grandpa, sr25519 for
 babe/initializer/session_info/authority_discovery, ecdsa for beefy) match 1:1 AH type defined below

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
sigurpol: Good point - created https://github.com/paritytech/polkadot-sdk/issues/11083, we will do post 2.1.0

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
sigurpol: absolutely, great catch. I have done it for Westend and not here 🫠  Done now. Will need a round of `/cmd bench --runtime asset-hub-polkadot --pallet pallet_staking_async_rc_client ` (and on kusama asset-hub ofc)

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
sigurpol: Well, here we rely on RC pallet-session benchmark and we multiply by x3 to be extra-conservative. I understand your point. I think though "16KB for writing 200 bytes" is a bit too harsh, it's 8 separate trie accesses (1 for Ledger, 1 for NextKeys, 6 of KeyOwner lookups ), and the proof size is the cumulative Merkle proof for all of them so maybe more reasonable that it sounds...

.github/workflows/runtimes-matrix.json
```diff
@@ -109,6 +112,9 @@
     "benchmarks_templates": {
       "pallet_xcm_benchmarks::generic": "templates/xcm-bench-template.hbs",
       "pallet_xcm_benchmarks::fungible": "templates/xcm-bench-template.hbs"
+    },
+    "benchmarks_exclude_extrinsics": {
+      "pallet_staking_async": ["prune_era_total_stake"]
```
sigurpol, line 117: NOTE for the reviewers: the issue with stale prune_era_total_stake has been fixed upstream in https://github.com/paritytech/polkadot-sdk/pull/11050 but we haven't backported the PR to 2512 assuming that assuming the stale extrinsic is good enough. If more changes to staking SDK pallet are needed, then we could backport 1050 as well and remove the temporary hack here

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
sigurpol: good point, added to pre-relase TODO list!

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
sigurpol: added a TODO linking to your gorgeous https://github.com/sigurpol/runtimes/pull/5

sigurpol: done

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
sigurpol: done

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
sigurpol: never deployed

system-parachains/asset-hubs/asset-hub-kusama/tests/tests.rs
sigurpol: see  https://github.com/sigurpol/runtimes/pull/5

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
sigurpol: I don't know either... but it's not always the case e.g. we have also PotStake , fraction, ... without the py/ stuff. If there are no major objection, I would keep it as it is now (it's also what we have in WAH today already)

kianenigma:
/cmd bench --runtime asset-hub-polkadot --pallet pallet_staking_async pallet_staking_async_rc_client

system-parachains/asset-hubs/asset-hub-polkadot/src/genesis_config_presets.rs
sigurpol: manual transfer of ED to the dap buffer account has been done already (see https://assethub-polkadot.subscan.io/account/13UVJyLkAxdQn6zM3GrLGvrFgN9jWneRX2NGvf3u92iunZq1)

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
sigurpol: Moved

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
georgepisaltu: This extension is needed to make the `authorize` macro work. This was introduced by @gui1117 to deprecate unsigned transactions and instead make them pass through the normal TxExt pipeline. If any of the code in this runtime uses `authorize` for transactions, you need it.

About the extension itself, it is just an empty extension with a `PhantomData` inside of it, so it should encode like a `()`. Obviously if you add it to the `TxExtension` pipeline it will end up in the metadata, but I don't think this should break anything. I don't know what offline signers are doing but if PJS works with it, I assume they do too since they may be using the same API lib behind the scenes. Today, I think that any extension that can encode like a `()` for the default value should work out of the box with the tooling we have.

gui1117: yes this transaction extension doesn't break anything except it shows up in the metadata.

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
lexnv: Yep, this needs to be `type RelayParentOffset = ConstU32<RELAY_PARENT_OFFSET>;`. Similar to other chains where we have enabled it

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
sigurpol: thank you all - I've added it to Polkadot AH to be consistent with what we have here

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
sigurpol: Reset to RELAY_PARENT_OFFSET for both PAH and KAH (initially incorrectly set to 0 as in WAH) - thanks guys

.github/workflows/test.yml
```diff
@@ -475,7 +475,9 @@ jobs:
       - name: Checkout polkadot-ecosystem-tests
         uses: actions/checkout@v4
         with:
-          repository: open-web3-stack/polkadot-ecosystem-tests
+          # TODO: revert to open-web3-stack/polkadot-ecosystem-tests once PR #524 merges
```
rockbmb: I think we want to clarify further: https://github.com/open-web3-stack/polkadot-ecosystem-tests/pull/524 can only be merged when this (#1065) PR itself lands on live production networks.

It doesn't have to be both Kusama & Polkadot; after Kusama, there's the option of gating the changes in PET behind a network P vs K check.

.github/workflows/debug-jemalloc.yml
```diff
@@ -0,0 +1,127 @@
+name: Debug Jemalloc
```
ggwpez: Still needed? I think you concluded rust version being the issue.

relay/kusama/src/lib.rs
```diff
@@ -188,7 +182,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: alloc::borrow::Cow::Borrowed("kusama"),
 	impl_name: alloc::borrow::Cow::Borrowed("parity-kusama"),
 	authoring_version: 2,
-	spec_version: 2_000_005,
+	spec_version: 2_000_007,
```
ggwpez: We do that in the release MR.

relay/polkadot/src/lib.rs
```diff
@@ -177,7 +172,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: alloc::borrow::Cow::Borrowed("polkadot"),
 	impl_name: alloc::borrow::Cow::Borrowed("parity-polkadot"),
 	authoring_version: 0,
-	spec_version: 2_000_005,
+	spec_version: 2_000_007,
```
ggwpez: Not needed now

.github/workflows/debug-jemalloc.yml
kianenigma: Is there a chance this will happen again, and we need a CI job for it? 

Though short term I am happy to have it to keep our weights consistent.

kianenigma: (Too many issues with weight lately!)

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
```diff
@@ -199,7 +199,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_name: Cow::Borrowed("statemint"),
 	spec_name: Cow::Borrowed("statemint"),
 	authoring_version: 1,
-	spec_version: 2_000_005,
+	spec_version: 2_000_007,
```
ggwpez: Undo please

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -180,7 +180,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("bridge-hub-kusama"),
 	impl_name: Cow::Borrowed("bridge-hub-kusama"),
 	authoring_version: 1,
-	spec_version: 2_000_005,
+	spec_version: 2_000_007,
```
ggwpez:
```suggestion
	spec_version: 2_000_005,
```

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
```diff
@@ -190,7 +190,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_name: Cow::Borrowed("bridge-hub-polkadot"),
 	impl_name: Cow::Borrowed("bridge-hub-polkadot"),
 	authoring_version: 1,
-	spec_version: 2_000_005,
+	spec_version: 2_000_007,
```
ggwpez:
```suggestion
	spec_version: 2_000_005,
```

CHANGELOG.md
```diff
@@ -42,7 +52,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
-- Add StakingOperator proxy type and the ability for Staking proxy to manage StakingOperator proxies ([#1033](https://github.com/polkadot-fellows/runtimes/pull/1033))
```
ggwpez, line 49: Not sure if we can remove stuff form old releases.

CHANGELOG.md
```diff
@@ -10,16 +10,26 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+- Fix check-migration(asset-hub-polkadot) CI job [#1057](https://github.com/polkadot-fellows/runtimes/pull/1057)
+- Update to polkadot-sdk stable2512 (tag 2512-2), introduce DAP pallet, refine StakingOperator and add Session keys handling on AssetHub  [#1065](https://github.com/polkadot-fellows/runtimes/pull/1065)
 
 ## [2.0.7] 16.02.2026
 
 ### Added
 
 - Add hyperbridge to parachains header root ([#1073](https://github.com/polkadot-fellows/runtimes/pull/1073))
 
+## [2.0.6] 04.02.2026
```
ggwpez: Hm this adds 2.0.6?

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
kianenigma: Thanks @gui1117 @georgepisaltu. if no additional encoded data, it is only as much of a breaking change for the signers, as updating runtime in any form-factor is.

system-parachains/people/people-polkadot/src/assets.rs
```diff
@@ -48,6 +48,7 @@ impl pallet_assets::Config for Runtime {
 	type WeightInfo = weights::pallet_assets::WeightInfo<Runtime>;
 	type CallbackHandle = ();
 	type AssetAccountDeposit = AssetAccountDeposit;
+	type ReserveData = ();
```
kianenigma, line 52: Does this require a TODO as well like other instances? 

Purely pattern matching here, haven't looked at the semantic here 🙈

ggwpez: Yes will add

.github/workflows/debug-jemalloc.yml
ggwpez: removed for now, we can do weights later

ggwpez: Removed

.github/workflows/test.yml
ggwpez: Reverted

.github/workflows/debug-jemalloc.yml
sigurpol: Good to remove - we run the test once with last commit and we have seen that finally weights are back to normal so we can remove .

CHANGELOG.md
sigurpol: I think it was never released so deserves to stay in 2.1.0 section once we release - 2.0.6 and 2.0.7 don’t contain it

PR #1068, opened 2026-02-02
Refactor PET in `test.yml` CI workflow

### LINKED ISSUE #1010: Rework PET job in `test.yml` workflow

rockbmb:
Closes #1010 .

This PR will rework the PET job in `.github/workflows/test.yml`.

- [x] use a test matrix with individual retry/timeout configurations rather than just `yarn test` to avoid restarting the whole suite on a single suite's failure
- [x] To also avoid RPC timeouts due to throttling/stalling by test network's production RPC endpoints, consider using https://github.com/AcalaNetwork/subway or Chopsticks (as in https://github.com/open-web3-stack/polkadot-ecosystem-tests/pull/504)
- [x] Filter end-to-end test suites from PET to be used in finalized workflow
    -  This repository is only concerned with the runtimes that are under the Polkadot Fellowship's purview, while PET contains end-to-end tests for other ecosystme chains; as such, it makes sense to filter what is included in CI to further reduce load.
- [x] Re-enable parts of `test.yml` that were disabled to test PET changes in isolation

# Review notes

Main changes:
* new folder in `.github/subway-configs`; it contains [Subway](https://github.com/AcalaNetwork/subway) configs for the RPC proxies that will be created during CI: one for each Polkadot/Kusama relay + system parachain .
* PET reenabled in `.github/workflows/test.yml`, and completely refactored
    - it uses a matrix strategy with two jobs, one for each network: Polkadot and Kusama

## Details

Each of the above P/K jobs:
- Fetches Rust toolchain + Subway, builds its binary, and starts a Subway proxy for each of that network's relay/SP
       - Polkadot: 1 proxy for relay + 5 proxies for SPs
       - Kusama: 1 proxy for relay + 4 proxies for SPs
- Downloads WASM artifacts from a previous step of the CI pipeline
- Overrides PET's `.env` with said runtimes + proxies' endpoints
    - That means PET will run using runtimes built against the code of whichever PR triggered CI, and cache RPC responses, which are shared between all chains' tests
- Gets `yarn`, checks out PET repo, builds it
- Runs PET's known block number command from within it, to use the latest state in combination with built WASM runtimes
- Runs all end-to-end test suites for that network's chains
    - only runs tests related to relay + SPs; ecosystem chains are skipped, as this repo's purview is only for Fellowship-maintained runtimes
- Cleans up Subway proxies at the end

In particular, the critical change:

```bash
          yarn test -u $TEST_FILES --pool=threads --maxWorkers=3 --retry=3
          # `$TEST_FILES` contains all E2E modules for the network: relay + SPs included
          # 1. `-u` updates snapshots so that they don't fail spuriously e.g. proof size/ref time changes
          # 2. `--pool=threads` runs test files from `$TEST_FILES` in parallel using worker threads. Default
          #    is `--pool=forks`, which uses child processes - heavier.
          # 3. `--maxWorkers=3` combined with the above means at most 3 test files running concurrently, each in its
          #    own thread
          # 4. `--retry=3` attempts a failing test 3 times. This operates at the level of individual tests, not modules,
          #    or the whole test suite; one failure will *not* require a full rerun.
```

---

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -497,8 +497,15 @@ jobs:
         working-directory: ecosystem-tests
         run: |
           echo "Finding Fellowship runtime test files in ${{ matrix.network.test_dir }}..."
-          # Filter: must contain 'e2e' AND the network name ("polkadot/kusama", case insensitive)
-          TEST_FILES=$(find ${{ matrix.network.test_dir }} -name "*.test.ts" | grep -i "e2e" | grep -i "${{ matrix.network.name }}" | tr '\n' ' ')
+          # Only run tests for chains that have a local Subway proxy (i.e. a *_ENDPOINT env var set).
+          # External parachains (bifrost, acala, hydration, ...) never have endpoint vars set, so
+          # they are excluded automatically — no hard-coded allowlist/denylist needed.
+          TEST_FILES=$(find ${{ matrix.network.test_dir }} -name "*.e2e.test.ts" | \
+            while IFS= read -r f; do
+              chain=$(basename "$f" | cut -d. -f1)
+              envvar=$(echo "$chain" | tr '[:lower:]' '[:upper:]')_ENDPOINT
+              [ -n "${!envvar}" ] && echo "$f"
+            done | tr '\n' ' ')
 
           if [ -z "$TEST_FILES" ]; then
             echo "No test files found in ${{ matrix.network.test_dir }}"


.github/workflows/test.yml
```diff
@@ -363,27 +385,89 @@ jobs:
...
         with:
-          repository: open-web3-stack/polkadot-ecosystem-tests
-          path: ecosystem-tests
+          targets: "wasm32v1-none"
+          components: "rust-src"
+          toolchain: "${{env.RUST_STABLE_VERSION}}"
+
+      - name: Build Subway
+        run: |
+          echo "Building subway..."
```
xlc: `cargo install --git https://github.com/AcalaNetwork/subway`

.github/workflows/test.yml
```diff
@@ -392,16 +476,85 @@ jobs:
...
+      - name: Checkout polkadot-ecosystem-tests
+        uses: actions/checkout@v4
+        with:
+          repository: open-web3-stack/polkadot-ecosystem-tests
+          path: ecosystem-tests
+
+      - name: Install test dependencies
         working-directory: ecosystem-tests
-        run: yarn install
+        run: yarn install --immutable
```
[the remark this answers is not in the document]
xlc: should replace the previous `npm install -g yarn` with `corepack`

.github/workflows/test.yml
```diff
@@ -392,16 +476,85 @@ jobs:
...
+          echo "Running tests with limited parallelism..."
+          # Notes on flags below:
+          # 1. `-u` updates snapshots so that they don't fail spuriously e.g. proof size/ref time changes
+          # 2. `--pool=threads` runs test files from `$TEST_FILES` in parallel using worker threads. Default
+          #    is `--pool=forks`, which uses child processes - heavier.
+          # 3. `--maxWorkers=3` combined with the above means at most 3 test files running concurrently, each in its
+          #    own thread
+          # 4. `--retry=3` attempts a failing test 3 times. This operates at the level of individual tests, not modules,
+          #    or the whole test suite; one failure will *not* require a full rerun.
+          yarn test -u $TEST_FILES --pool=threads --maxWorkers=3 --retry=3
```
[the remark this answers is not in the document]
rockbmb: I disagree. In most cases, on-chain events are not only snapshot, but destructured and asserted on during tests. Examples: [1](https://github.com/open-web3-stack/polkadot-ecosystem-tests/blob/5e342b1378e8e2d439b7d98eee7cd4529708da8e/packages/shared/src/accounts.ts#L647), [2](https://github.com/open-web3-stack/polkadot-ecosystem-tests/blob/5e342b1378e8e2d439b7d98eee7cd4529708da8e/packages/shared/src/scheduler.ts#L191), [3](https://github.com/open-web3-stack/polkadot-ecosystem-tests/blob/5e342b1378e8e2d439b7d98eee7cd4529708da8e/packages/shared/src/staking.ts#L135).

Without this flag, someone who submits a PR that fails due to e.g. greatly differing `proofSize/refTime` values in some of the built runtimes - which PET will have run against, not the ones currently live - must wait for those runtimes to hit production, for PET's CI to then update snapshots, and finally for this workflow to use the updated snapshots.

This PR's scope never included such drastic changes to runtime development workflow.

.github/workflows/test.yml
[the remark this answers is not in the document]
rockbmb: The step fails when changed to this; I created a [PR](https://github.com/AcalaNetwork/subway/pull/200) to Subway to bump dependencies.

xlc (approved):

xlc (approved):

.github/subway-configs/people-kusama.yml
bkchr, line 1: Sorry for being that late to the part, but I don't get why we have literally the same file multiple times. Just different endpoints.

bkchr: https://github.com/polkadot-fellows/runtimes/pull/1077

rockbmb: >but I don't get why we have literally the same file multiple times.

We don't! You're right.

This was an omission on my part; I planned to improve that point, and then forgot. Thanks for doing it.

PR #1072, opened 2026-02-13
Allow fellowship's Architects track (ranks 4+) to manage fellowship treasury and salary on Asset Hub

franciscoaguirre:
Changed the XCM configuration from Asset Hub and Collectives to allow:
- Architects track to use the scheduler
- Architects track to dispatch calls as treasurer = `[Plurality { body: BodyId::Technical, part: BodyPart::Voice }, GeneralIndex(4)]`
  - This location semantically means "The 4th rank inside of the Technical Fellowship. I expect we can use this same pattern for other ranks.
- Architects origin to alias (with `AliasOrigin`) to the fellowship's treasury and salary pallets on Asset Hub

This allows better internal management of the fellowship treasury, without needing to go to OpenGov's Root track.

### DIFF
--- a/integration-tests/emulated/tests/assets/asset-hub-polkadot/Cargo.toml
+++ b/integration-tests/emulated/tests/assets/asset-hub-polkadot/Cargo.toml
@@ -63,6 +63,7 @@ runtime-benchmarks = [
 	"pallet-xcm/runtime-benchmarks",
 	"parachains-common/runtime-benchmarks",
 	"polkadot-runtime-common/runtime-benchmarks",
+	"polkadot-runtime-constants/runtime-benchmarks",
 	"polkadot-runtime/runtime-benchmarks",
 	"polkadot-system-emulated-network/runtime-benchmarks",
 	"sp-runtime/runtime-benchmarks",
@@ -70,5 +71,4 @@ runtime-benchmarks = [
 	"xcm-executor/runtime-benchmarks",
 	"xcm-runtime-apis/runtime-benchmarks",
 	"xcm/runtime-benchmarks",
-	"polkadot-runtime-constants/runtime-benchmarks"
 ]
--- a/relay/polkadot/constants/Cargo.toml
+++ b/relay/polkadot/constants/Cargo.toml
@@ -48,5 +48,5 @@ runtime-benchmarks = [
 	"sp-runtime/runtime-benchmarks",
 	"sp-trie",
 	"xcm-builder/runtime-benchmarks",
-	"xcm/runtime-benchmarks"
+	"xcm/runtime-benchmarks",
 ]


system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -331,6 +337,48 @@ impl Contains<Location> for SecretaryEntities {
 	}
 }
 
+/// Allows the Fellowship Treasurer plurality from Collectives to alias into
+/// the Fellowship Treasury or Fellowship Salary pallet locations.
+///
+/// This allows the Architects track (rank 4+ Fellowship members) on the Collectives chain to
+/// manage the fellowship treasury and salary here on Asset Hub. The Architects origin is converted
+/// to a `Plurality { id: BodyId::Treasury, part: BodyPart::Voice }` via
```
bkchr: Why are we using the treasury body id?

franciscoaguirre: Because it makes more sense semantically, "the treasury plurality of the collectives chain" vs "collectives -> pallet instance 65". We can also add as many aliases as we want for this location thus limiting its privileges as much as we want. The Technical body was already taken for track Fellows.

pandres95: Probably a dumb question, but does this mean that the Fellowship treasury would be the only treasury under Collectives? Asking because IIUC, the ambassadors' collective is still on (dumb question, precisely because I'm not certain about the current state of that collective).

system-parachains/collectives/collectives-polkadot/src/lib.rs
```diff
@@ -678,7 +678,7 @@ impl pallet_scheduler::Config for Runtime {
 	type PalletsOrigin = OriginCaller;
 	type RuntimeCall = RuntimeCall;
 	type MaximumWeight = MaximumSchedulerWeight;
-	type ScheduleOrigin = EnsureRoot<AccountId>;
+	type ScheduleOrigin = EitherOfDiverse<EnsureRoot<AccountId>, Architects>;
```
acatangiu, line 681: what is this required for?

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
acatangiu: Even though it is semantically descriptive, the documentation of this `BodyId` disagrees:
```rust
pub enum BodyId {
	/// The unambiguous treasury body (for Polkadot, an opinion on the topic given via a public
	/// referendum on the `treasurer` track).
	Treasury,
}
```

but maybe we just change the doc to discriminate `BodyId` meaning based on the originating Chain too

system-parachains/collectives/collectives-polkadot/src/xcm_config.rs
```diff
@@ -288,10 +288,15 @@ parameter_types! {
 /// Type to convert the Fellows origin to a Plurality `Location` value.
 pub type FellowsToPlurality = OriginToPluralityVoice<RuntimeOrigin, Fellows, FellowsBodyId>;
 
+/// Type to convert the Architects origin to a Treasurer Plurality `Location` value.
+pub type ArchitectsToTreasurerPlurality =
+	OriginToPluralityVoice<RuntimeOrigin, Architects, TreasurerBodyId>;
```
acatangiu: I don't think this is a good idea - it means that ANY and ALL `Architects` messages will be mapped on egress as `TreasurerBodyId` origin.

system-parachains/collectives/collectives-polkadot/src/lib.rs
franciscoaguirre: For the Architects origin to be able to shcedule calls. This is needed e.g. for it to be able to schedule a DCA with just one proposal instead of many.

system-parachains/collectives/collectives-polkadot/src/xcm_config.rs
franciscoaguirre: True, `Architects` might have other things they want to do that are not represented in a good way by `Treasury`

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
franciscoaguirre: That description is indeed quite restrictive. We could change the description though

system-parachains/collectives/collectives-polkadot/src/xcm_config.rs
franciscoaguirre: I could use a [`Moniker`](https://github.com/paritytech/polkadot-sdk/blob/fb71b25620c5ce246a1151d5eca21491bac012b1/polkadot/xcm/src/v3/junction.rs#L128) like `Arch` or `Dan4` or an [`Index`](https://github.com/paritytech/polkadot-sdk/blob/fb71b25620c5ce246a1151d5eca21491bac012b1/polkadot/xcm/src/v3/junction.rs#L130) of 4. This can then be aliased to the corresponding accounts on asset hub.

pandres95: I agree. Something like `{ id: Technical, part: Index(4) }` sounds way better to represent the treasury of the `Technical` plurality.

franciscoaguirre: `Index` is not a `BodyPart` but a `BodyId`. We could use one junction for Technical Plurality and an inner junction with `Index(4)`.

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -331,6 +337,48 @@ impl Contains<Location> for SecretaryEntities {
...
+		matches!(
+			origin.unpack(),
+			(
+				1,
+				[
+					Parachain(system_parachain::COLLECTIVES_ID),
+					Plurality { id: BodyId::Treasury, part: BodyPart::Voice }
+				]
+			)
+		) && matches!(
```
muharem, line 342: OR instead of AND?

system-parachains/collectives/collectives-polkadot/src/xcm_config.rs
muharem: better `Index`, bc we already using them 

https://github.com/polkadot-fellows/runtimes/blob/6c8c94b2132702a3b53d28563d261a43902b6789/relay/polkadot/constants/src/lib.rs#L109

integration-tests/emulated/tests/assets/asset-hub-polkadot/src/tests/aliases.rs
```diff
@@ -357,3 +357,144 @@ fn asset_hub_kusama_root_does_not_alias_into_asset_hub_polkadot_origins() {
 		assert!(!<XcmConfig as xcm_executor::Config>::Aliasers::contains(&origin, &target));
```
muharem, line 358: may be a test where we command something as arch from Collectives to AH, and get the origin converted via alias into desired one?

acatangiu: mandatory :D

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
franciscoaguirre: AND is correct. We want to assert that the `origin` is what we expect can alias AND the `target` what we expect the origin can alias to

muharem: ah, I see

system-parachains/collectives/collectives-polkadot/constants/src/lib.rs
```diff
@@ -22,6 +22,9 @@ pub const FELLOWSHIP_SALARY_PALLET_INDEX: u8 = 64;
 /// Polkadot Fellowship Treasury pallet instance.
 pub const FELLOWSHIP_TREASURY_PALLET_INDEX: u8 = 65;
 
+/// Fellowship Architects rank (rank 4). Used with `GeneralIndex` in XCM locations.
+pub const ARCHITECTS_RANK: u128 = 4;
```
muharem: may be some namespacing? `xcm::body::ARCHITECTS_INDEX` for example?
sounds line architects rank number constant.

system-parachains/collectives/collectives-polkadot/src/xcm_config.rs
```diff
@@ -288,10 +290,28 @@ parameter_types! {
...
+/// GeneralIndex(ARCHITECTS_RANK)]`.
+pub struct ArchitectsToLocation;
+impl TryConvert<RuntimeOrigin, Location> for ArchitectsToLocation {
+	fn try_convert(o: RuntimeOrigin) -> Result<Location, RuntimeOrigin> {
+		let _ = Architects::try_origin(o)?;
+		Ok(Location {
+			parents: 0,
+			interior: [
+				Plurality { id: BodyId::Technical, part: BodyPart::Voice },
+				GeneralIndex(collectives_polkadot_runtime_constants::ARCHITECTS_RANK),
```
muharem: this makes it possible to descend to Architects from Fellows, its Fellows is single interior of `Plurality { id: BodyId::Technical, part: BodyPart::Voice }`

muharem: we use something like `Plurality { id: BodyId::Index(xcm::body::ARCHITECTS_INDEX), part: BodyPart::Voice }`

franciscoaguirre: We could do that but then we lose the "Technical" namespace. Collectives already has the ambassador fellowship in there.

What I can do is make sure Fellows is [Plurality { .. }, GeneralIndex(3)]

muharem: but I think `Technical` is not really a name space. its an ID of the body, `Fellows` body in this case. Same we have for the `Treasurer`, we also do not treat it as a name space. and when we do not have a variant from XCM, like `Technical` we use custom index.

acatangiu: > What I can do is make sure Fellows is [Plurality { .. }, GeneralIndex(3)]

You mean change `Fellows` XCM representation:
```diff
- Plurality { id: BodyId::Technical, part: BodyPart::Voice }
+ [ Plurality { id: BodyId::Technical, part: BodyPart::Voice }, GeneralIndex(3) ]
``` 

And add `Architects` representation as:
```rust
[ Plurality { id: BodyId::Technical, part: BodyPart::Voice }, GeneralIndex(4) ]
```

That sounds good as it's more representative, but you need to triple-check and update places where the old Fellows represention is/was used.

relay/polkadot/constants/src/lib.rs
acatangiu: this is due for removal for some time now, maybe quick followup/cleanup PR? :D

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -315,6 +324,49 @@ impl Contains<Location> for AmbassadorEntities {
...
+		matches!(
+			origin.unpack(),
+			(
+				1,
+				[
+					Parachain(system_parachain::COLLECTIVES_ID),
+					Plurality { id: BodyId::Technical, part: BodyPart::Voice },
+					GeneralIndex(ARCHITECTS_RANK)
+				]
+			)
```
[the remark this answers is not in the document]
franciscoaguirre: GeneralKey(4) means rank 4+ since every fellow rank 4 and above can vote on the Architects track

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -264,9 +268,14 @@ impl Contains<Location> for FellowshipEntities {
...
 				[
 					Parachain(system_parachain::COLLECTIVES_ID),
-					Plurality { id: BodyId::Technical, .. }
+					Plurality { id: BodyId::Technical, .. },
+					GeneralIndex(rank)
 				]
-			) | (
+			)
+			if *rank >= FELLOWS_RANK
+		) || matches!(
```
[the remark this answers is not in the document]
franciscoaguirre: Added the previous behavior back, just in case

system-parachains/collectives/collectives-polkadot/constants/src/lib.rs
```diff
@@ -22,6 +22,15 @@ pub const FELLOWSHIP_SALARY_PALLET_INDEX: u8 = 64;
...
 pub const FELLOWSHIP_TREASURY_PALLET_INDEX: u8 = 65;
 
+pub mod xcm {
+	pub mod body {
+		/// Fellowship Fellows rank (rank 3). Used with `GeneralIndex` in XCM locations.
+		pub const FELLOWS_RANK: u128 = 3;
+		/// Fellowship Architects rank (rank 4). Used with `GeneralIndex` in XCM locations.
+		pub const ARCHITECTS_RANK: u128 = 4;
+	}
+}
```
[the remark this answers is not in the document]
franciscoaguirre: Moved them to polkadot-runtime-constants

relay/polkadot/src/xcm_config.rs
```diff
@@ -176,7 +179,15 @@ impl Contains<Location> for Fellows {
...
-			(0, [Parachain(COLLECTIVES_ID), Plurality { id: BodyId::Technical, .. }])
+			(
+				0,
+				[
+					Parachain(COLLECTIVES_ID),
+					Plurality { id: BodyId::Technical, .. },
+					GeneralIndex(rank)
+				]
+			)
+			if *rank >= FELLOWS_RANK
```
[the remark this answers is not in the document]
franciscoaguirre: Added the other pattern back just in case

system-parachains/bridge-hubs/bridge-hub-polkadot/src/xcm_config.rs
```diff
@@ -151,9 +151,11 @@ impl Contains<Location> for FellowsPlurality {
 				1,
 				[
 					Parachain(system_parachain::COLLECTIVES_ID),
-					Plurality { id: BodyId::Technical, .. }
+					Plurality { id: BodyId::Technical, .. },
+					GeneralIndex(rank)
 				]
 			)
+			if *rank >= FELLOWS_RANK
```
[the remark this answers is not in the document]
franciscoaguirre: Same

xlc (approved):

acatangiu:
One other thing to explicitly call out is the requirement for this set of changes to be applied on all system chains **at the same time** to gracefully handle the Fellows Location change across the system.

acatangiu:
Suggested fix here: https://github.com/franciscoaguirre/polkadot-runtimes/pull/4

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -257,57 +261,100 @@ parameter_types! {
...
-	fn contains(location: &Location) -> bool {
+/// Allows the Fellowship Architects origin from Collectives to alias into
+/// the Fellowship Treasury or Fellowship Salary pallet locations.
+///
+/// This allows the Architects track (rank 4+ Fellowship members) on the Collectives chain to
+/// manage the fellowship treasury and salary here on Asset Hub. The Architects origin is converted
+/// to `[Plurality { id: BodyId::Technical, part: BodyPart::Voice }, GeneralIndex(ARCHITECTS_RANK)]`
+/// via `ArchitectsToLocation` before being sent over XCM.
+pub struct FellowshipArchitectsAlias;
+impl ContainsPair<Location, Location> for FellowshipArchitectsAlias {
```
acatangiu: nit: rename

```suggestion
pub struct FellowshipArchitectsAliases;
impl ContainsPair<Location, Location> for FellowshipArchitectsAliases {
```

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -417,6 +464,7 @@ pub type TrustedAliasers = (
 	AliasChildLocation,
 	AuthorizedAliasers<Runtime>,
 	AliasOriginRootUsingFilter<bridging::to_kusama::AssetHubKusama, KusamaGlobalConsensus>,
+	FellowshipArchitectsAlias,
```
acatangiu:
```suggestion
	FellowshipArchitectsAliases,
```

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
acatangiu, line 464:
```suggestion
/// Defines origin aliasing rules for this chain.
///
/// - Allow any origin to alias into a child sub-location (equivalent to DescendOrigin),
/// - Allow origins explicitly authorized by the alias target location.
/// - Allow cousin Kusama Asset Hub to alias into Kusama (bridged) origins.
/// - Allow Technical Fellowship Architects to alias into Fellowship Treasury and Salary pallets.
pub type TrustedAliasers = (
```

relay/polkadot/src/xcm_config.rs
```diff
@@ -171,15 +170,7 @@ pub type TrustedTeleporters = (
...
-pub struct Fellows;
-impl Contains<Location> for Fellows {
-	fn contains(loc: &Location) -> bool {
-		matches!(
-			loc.unpack(),
-			(0, [Parachain(COLLECTIVES_ID), Plurality { id: BodyId::Technical, .. }])
-		)
-	}
-}
+pub type Fellows = IsFellowshipVoice<CollectivesLocation>;
```
acatangiu, line 173: the two are not equivalent!

`IsFellowshipVoice` is only matching on `Voice` of Fellows, whereas the previous `Fellows` filter would match on other body parts

the new filter might be better actually (stricter) - but definitely requires review/thought!

please check if there is any other use than `Voice` anywhere so at least we know we're not braking some existing functionality

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -253,20 +257,14 @@ parameter_types! {
...
+pub struct FellowshipTreasuryPaymasterEntities;
+impl Contains<Location> for FellowshipTreasuryPaymasterEntities {
 	fn contains(location: &Location) -> bool {
 		matches!(
 			location.unpack(),
 			(
-				1,
-				[
-					Parachain(system_parachain::COLLECTIVES_ID),
-					Plurality { id: BodyId::Technical, .. }
```
acatangiu, line 267: same here, we're now "disregarding" other `BodyPart`s - new code is only matching on `BodyPart::Voice`.

flagging for extra eyes on it 👀

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -315,6 +319,49 @@ impl Contains<Location> for AmbassadorEntities {
...
+			origin.unpack(),
+			(
+				1,
+				[
+					Parachain(system_parachain::COLLECTIVES_ID),
+					Plurality { id: BodyId::Technical, part: BodyPart::Voice },
+					GeneralIndex(rank)
+				]
+			) if *rank >= ARCHITECTS_RANK
+		) && matches!(
```
acatangiu, line 342: nit: could replace the second `matches!` here with `FellowshipTreasuryPaymasterEntities::contains(target)`

system-parachains/bridge-hubs/bridge-hub-polkadot/src/xcm_config.rs
```diff
@@ -142,21 +142,7 @@ impl Contains<Location> for ParentOrParentsPlurality {
...
-pub struct FellowsPlurality;
-impl Contains<Location> for FellowsPlurality {
-	fn contains(location: &Location) -> bool {
-		matches!(
-			location.unpack(),
-			(
-				1,
-				[
-					Parachain(system_parachain::COLLECTIVES_ID),
-					Plurality { id: BodyId::Technical, .. }
```
acatangiu, line 154: same here, only `Voice` is now "allowed" (matching the filter)

system-parachains/coretime/coretime-polkadot/src/xcm_config.rs
```diff
@@ -153,21 +153,7 @@ impl Contains<Location> for ParentOrParentsPlurality {
...
-pub struct FellowsPlurality;
-impl Contains<Location> for FellowsPlurality {
-	fn contains(location: &Location) -> bool {
-		matches!(
-			location.unpack(),
-			(
-				1,
-				[
-					Parachain(system_parachain::COLLECTIVES_ID),
-					Plurality { id: BodyId::Technical, .. }
```
acatangiu, line 165: same here, reduction in scope to `Voice`

system-parachains/people/people-polkadot/src/xcm_config.rs
```diff
@@ -185,21 +185,7 @@ impl Contains<Location> for ParentOrParentsPlurality {
...
-pub struct FellowsPlurality;
-impl Contains<Location> for FellowsPlurality {
-	fn contains(location: &Location) -> bool {
-		matches!(
-			location.unpack(),
-			(
-				1,
-				[
-					Parachain(system_parachain::COLLECTIVES_ID),
-					Plurality { id: BodyId::Technical, .. }
```
acatangiu, line 197: same here, reduction in scope to `Voice`

PR #1073, opened 2026-02-15
Add hyperbridge to parachains header root 

Wizdave97:
This PR manually adds hyperbridge to the parachains header root for the following reason.

Hyperbridge transitioned to a para thread recently since our lease period expired and our beefy consensus client on connected networks relies on Hyperbridge being part of the parachain heads root. The client can no longer be updated until this is fixed and it expires in 21 days after which it can no longer be updated permanently freezing the bridge.

You can find a Link to the thread on element chat here https://matrix.to/#/!bMWYTUFeSoYacMOlpS:matrix.parity.io/$heRddjL-qs_cRcJVjl-la19NibMYOQK5pvtGHnjVRmU?via=parity.io&via=matrix.org&via=matrix.parity.io

### DIFF
--- a/relay/polkadot/src/lib.rs
+++ b/relay/polkadot/src/lib.rs
@@ -433,10 +433,10 @@ pub struct ParaHeadsRootProvider;
 impl BeefyDataProvider<H256> for ParaHeadsRootProvider {
 	fn extra_data() -> H256 {
 		let para_heads: BTreeMap<u32, Vec<u8>> = parachains_paras::Parachains::<Runtime>::get()
-			.into_iter()
-			.chain(BEEFY_WHITELISTED_PARATHREADS.into_iter().cloned())
+			.iter()
+			.chain(BEEFY_WHITELISTED_PARATHREADS.iter())
 			.filter_map(|id| {
-				parachains_paras::Heads::<Runtime>::get(id).map(|head| (id.into(), head.0))
+				parachains_paras::Heads::<Runtime>::get(id).map(|head| ((*id).into(), head.0))
 			})
 			.collect();
 


relay/polkadot/src/lib.rs
```diff
@@ -426,7 +426,11 @@ parameter_types! {
 pub struct ParaHeadsRootProvider;
 impl BeefyDataProvider<H256> for ParaHeadsRootProvider {
 	fn extra_data() -> H256 {
-		let mut para_heads: Vec<(u32, Vec<u8>)> = parachains_paras::Parachains::<Runtime>::get()
+		// Manually add hyperbridge to parachains header root
+		let hyperbridge: u32 = 3367;
+		let mut parachains = parachains_paras::Parachains::<Runtime>::get();
+		parachains.push(hyperbridge.into());
```
bkontur: What about the check if contains already (when parathread turns back to para)? Could duplicit entry cause some issues?

seunlanlege: yeah makes sense, we should collect to a hashset

Wizdave97: Done

relay/polkadot/src/lib.rs
```diff
@@ -426,7 +426,13 @@ parameter_types! {
 pub struct ParaHeadsRootProvider;
 impl BeefyDataProvider<H256> for ParaHeadsRootProvider {
 	fn extra_data() -> H256 {
-		let mut para_heads: Vec<(u32, Vec<u8>)> = parachains_paras::Parachains::<Runtime>::get()
+		// Manually add hyperbridge to parachains header root
+		let hyperbridge: u32 = 3367;
```
vgeddes: nit: I would suggest making this more generic by incorporating a constant like the following:

```rust
// Parathreads to include in the BEEFY protocol
const BEEFY_PARATHREADS_WHITELIST = [
    // Hyperbridge
    3367
];
```

Wizdave97: Done

relay/polkadot/src/lib.rs
```diff
@@ -421,12 +421,23 @@ parameter_types! {
 	pub LeafVersion: MmrLeafVersion = MmrLeafVersion::new(0, 0);
 }
 
+/// Parathreads whitelisted to be added to the beefy mmr leaf parachains header root
+const BEEFY_WHITELISTED_PARATHREADS: [ParaId; 1] = [
+	// Hyperbridge
+	ParaId::new(3367),
+];
```
bkchr:
```suggestion
const BEEFY_WHITELISTED_PARATHREADS: &[ParaId] = &[
	// Hyperbridge
	ParaId::new(3367),
];
```

relay/polkadot/src/lib.rs
```diff
@@ -421,12 +421,23 @@ parameter_types! {
...
 pub struct ParaHeadsRootProvider;
 impl BeefyDataProvider<H256> for ParaHeadsRootProvider {
 	fn extra_data() -> H256 {
-		let mut para_heads: Vec<(u32, Vec<u8>)> = parachains_paras::Parachains::<Runtime>::get()
+		let mut parachains = parachains_paras::Parachains::<Runtime>::get()
+			.into_iter()
+			.collect::<alloc::collections::BTreeSet<_>>();
+		// Manually add whitelisted parathreads to parachains header root
+		parachains.extend(BEEFY_WHITELISTED_PARATHREADS);
+		let mut para_heads: Vec<(u32, Vec<u8>)> = parachains
```
bkchr:
```suggestion
		let mut para_heads: Vec<(u32, Vec<u8>)> = parachains_paras::Parachains::<Runtime>::get()
			.into_iter()
			.chain(BEEFY_WHITELISTED_PARATHREADS.into_iter()))
```

bkontur: @bkchr So we don’t need to worry about duplicates, right? For example, if at some point parachains_paras::Parachains contains some of the BEEFY_WHITELISTED_PARATHREADS again, the duplicate para_ids would have the same head data. In that case, the Merkle tree would simply be constructed with duplicate leaves for the same key/value pairs.

```
[
 (para1, head1),
 (para3367, head2),
 (para3367, head2),
]
```

bkchr: Good point @bkontur, I did not thought about it.

relay/polkadot/src/lib.rs
bkchr:
```suggestion
		para_heads.sort_by_key(|k| k.0);
		para_heads.dedup_by_key(|k| k.0);
```

bkchr: But we could also make para_heads directly a BTreeMap to not require these two calls here.

bkchr: @bkontur ^^

seunlanlege: It's already collected to btreeset. Dedupe is unneeded

bkontur: > It's already collected to btreeset. Dedupe is unneeded

@Wizdave97 please, apply that Basti's `.chain` and `.dedup_by_key` suggestions (imo, looks better than `collect::<alloc::collections::BTreeSet`) and we should be good to go.

relay/polkadot/src/lib.rs
```diff
@@ -421,18 +421,26 @@ parameter_types! {
...
+];
+
 /// A BEEFY data provider that merkelizes all the parachain heads at the current block
 /// (sorted by their parachain id).
 pub struct ParaHeadsRootProvider;
 impl BeefyDataProvider<H256> for ParaHeadsRootProvider {
 	fn extra_data() -> H256 {
-		let mut para_heads: Vec<(u32, Vec<u8>)> = parachains_paras::Parachains::<Runtime>::get()
+		let parachains = parachains_paras::Parachains::<Runtime>::get();
+		let para_heads: BTreeMap<u32, Vec<u8>> = parachains
```
bkontur: +fmt
```suggestion
		let para_heads: BTreeMap<u32, Vec<u8>> = parachains_paras::Parachains::<Runtime>::get()
```

relay/polkadot/src/lib.rs
```diff
@@ -421,18 +421,25 @@ parameter_types! {
...
+
 /// A BEEFY data provider that merkelizes all the parachain heads at the current block
 /// (sorted by their parachain id).
 pub struct ParaHeadsRootProvider;
 impl BeefyDataProvider<H256> for ParaHeadsRootProvider {
 	fn extra_data() -> H256 {
-		let mut para_heads: Vec<(u32, Vec<u8>)> = parachains_paras::Parachains::<Runtime>::get()
+		let para_heads: BTreeMap<u32, Vec<u8>> = parachains_paras::Parachains::<Runtime>::get()
 			.into_iter()
+			.chain(BEEFY_WHITELISTED_PARATHREADS.into_iter().cloned())
```
bkontur: clippy?
```suggestion
			.chain(BEEFY_WHITELISTED_PARATHREADS.iter())
```

Wizdave97: The iterators return a reference, I guess it i use iter in both iterators, it should be fine

PR #1085, opened 2026-02-23
Asset Hubs: add reserves info to support non-teleportable Foreign Assets

acatangiu:
Full integration of https://github.com/paritytech/polkadot-sdk/pull/9948

## Description

For assets registered on Asset Hub, only explicitly configured assets' reserve locations will be trusted.
Besides the actual reserve location(s), a `teleportable: bool` flag is also set for each reserve, specifying whether the asset is teleportable or not between the local chain and said reserve location.
Any other chain that should be assumed as a trusted reserve needs to be marked as such in `pallet-assets` by the asset's `Owner`.

By default, on asset creation, no explicit reserve information is configured. This results in foreign assets not being transferable cross-chain by default. To enable cross-chain transfers, one of the following configurations is required on the `Asset` to be done by its `Owner`:

- For AH to accept incoming reserve transfer of asset "A" from remote chain "X", that chain needs to be added as a trusted reserve for asset "A", and `teleportable` flag set to `false`:
```rust
      let asset_id = "A";
      let origin = Signed(OwnerOfA);
      let reserves = vec![
        ForeignAssetReserveData { reserve: "Remote Chain X location", teleport: false }
      ];
      pallet_assets::set_reserves(origin, asset_id, reserves);
```

- For AH to accept teleports of asset "A" from remote chain "X", that chain needs to be added as trusted reserves for asset "A", and `teleportable` flag set to `true`:
```rust
      let asset_id = "A";
      let origin = Signed(OwnerOfA);
      let reserves = vec![
        ForeignAssetReserveData { reserve: "Remote Chain X location", teleport: true }
      ];
      pallet_assets::set_reserves(origin, asset_id, reserves);
```

## Migrations

Since there are already a number of Foreign Assets already registered, and have been teleportable or reserve-transferable since registration based on static rules, this PR also adds a migration that configures the required reserves information for existing foreign assets, so as not to change any existing behaviors.


claravanstaden:
Thanks @acatangiu! The Snworbridge v2 pallet makes provision for this: https://github.com/paritytech/polkadot-sdk/blob/master/bridges/snowbridge/primitives/inbound-queue/src/v2/converter.rs#L308-L313
Snowbridge V1 does not though, so we can't use Snowbridge V1 to register assets anymore, after this change. cc @yrong @alistair-singh

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
```diff
@@ -265,12 +266,31 @@ impl frame_system::Config for Runtime {
...
+	type RuntimeEvent = RuntimeEvent;
+	#[cfg(not(feature = "runtime-benchmarks"))]
+	type Migrations = migrations::MbmMigrations;
+	// Benchmarks need mocked migrations to guarantee that they succeed.
+	#[cfg(feature = "runtime-benchmarks")]
+	type Migrations = pallet_migrations::mock_helpers::MockedMigrations;
+	type CursorMaxLen = ConstU32<65_536>;
+	type IdentifierMaxLen = ConstU32<256>;
+	type MigrationStatusHandler = ();
+	type FailedMigrationHandler = frame_support::migrations::FreezeChainOnFailedMigration;
```
ggwpez: Since we do governance on AH now, this is not an option. If the migration errors, we kind of need to ignore any error to allow Governance intervention. There is a `EnterSafeModeOnFailedMigration` but we dont have the safe-mode pallet deployed yet (TBD #1089).  

So I think something like:  
```rust
/// Ignore any MBM errors and unlock all calls that were locked during migration.
///
/// This implies that any storage invariants that were violated by a faulty MBM could now be
/// exposed to users via calls. It is equivalent to how a faulty single-block-migration would
/// be handled. 
pub struct ForceUnstuckOnFailedMigration;

impl frame_support::migrations::FailedMigrationHandler for ForceUnstuckOnFailedMigration {
	fn failed(_faulty_migration: Option<u32>) -> FailedMigrationHandling {
		FailedMigrationHandling::ForceUnstuck
	}
}
```

Would be reasonable. I will also add that to the SDK for our convenience. PS: https://github.com/paritytech/polkadot-sdk/pull/11194

system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs
ggwpez: Maybe also worth changing it here

system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
acatangiu: good catch! fixed

PR #1088, opened 2026-02-24
Fix upgrade guide for parachains and provide a script for running proposals

bkchr:
This fixes the runtime upgrade guide. It was using an incorrect block number to schedule calls on parachains. On parachains we are using the relay parent block numbers and not the chain local block numbers. Leading to proposals never getting executed in the dry run.

To simplify the entire process, a script is provided that does all the steps automatically.


dry-run/proposal.sh
```diff
@@ -0,0 +1,324 @@
...
+  exit 1
+fi
+
+if [ ! -d "$SCRIPT_DIR/node_modules/@polkadot/api" ]; then
+  echo "Error: @polkadot/api not found. Run 'npm install' from the dry-run/ directory first." >&2
+  exit 1
+fi
+
+cd "$SCRIPT_DIR" && exec node --input-type=module - "$@" << 'SCRIPT'
+import { ApiPromise, WsProvider } from '@polkadot/api';
```
xlc: why writing js in sh?

bkchr: Just to have these extra checks, now have splitted the scripts.

claravanstaden (approved):

PR #1090, opened 2026-02-25
Bump staking and EPMB crates

sigurpol:
- Bump election provider multiblock to include  [SDK PR#11156](https://github.com/paritytech/polkadot-sdk/pull/11156) : reduce overclaimed proof_size during Signed phase
- Bump staking pallets to include [SDK PR#11168](https://github.com/paritytech/polkadot-sdk/pull/11168): add a configurable MinSetKeysBond threshold that rejects set_keys when active bond is insufficient.

### DIFF
--- a/system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
+++ b/system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
@@ -513,7 +513,7 @@ impl pallet_staking_async_rc_client::Config for Runtime {
 	type MaxValidatorSetRetries = ConstU32<64>;
 	type ValidatorSetExportSession = ValidatorSetExportSession;
 	type RelayChainSessionKeys = RelayChainSessionKeys;
-	type MinSetKeysBond = ConstU128<{ 1_000 * UNITS }>;
+	type MinSetKeysBond = ConstU128<{ 10_000 * UNITS }>;
 	type Balance = Balance;
 	// | Key                 | Crypto  | Public Key | Signature |
 	// |---------------------|---------|------------|-----------|


system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -513,6 +513,7 @@ impl pallet_staking_async_rc_client::Config for Runtime {
 	type MaxValidatorSetRetries = ConstU32<64>;
 	type ValidatorSetExportSession = ValidatorSetExportSession;
 	type RelayChainSessionKeys = RelayChainSessionKeys;
+	type MinSetKeysBond = ConstU128<{ 10_000 * UNITS }>;
```
sigurpol: Aligned with min validator bond we are going to enforce post 2.1.0

system-parachains/asset-hubs/asset-hub-kusama/src/staking/mod.rs
```diff
@@ -452,6 +452,7 @@ impl pallet_staking_async_rc_client::Config for Runtime {
 	type ValidatorSetExportSession = ValidatorSetExportSession;
 	type RelayChainSessionKeys = RelayChainSessionKeys;
 	type Balance = Balance;
+	type MinSetKeysBond = ConstU128<{ 250 * UNITS }>;
```
sigurpol: up to debate 😄

Ank4n: I would suggest keeping it to bare minimum. Something like 1 or 10 KSM, just enough to prevent spam.
(But I guess there is no right answer)

sigurpol: yeah, probably 1 KSM is already sufficient with current value and something easier to communicate to Kusama folks - something we should not forget to do since we have talked in the forum about Polkadot and setting eventually min validator bond etc whereas this constraint on set_keys comes out of the blue for Kusama.

sigurpol: I went for 1 KSM

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
bkchr: Isn't this number already configured somewhere else? So, we use everywhere the same?

bkchr: >  with min validator bond we are going to enforce post 2.1.0

When is this gonna happen exactly? Not directly with this upgrade or?

sigurpol: Min validator bond  of 10k will be enabled via open gov call  in a separate referendum  few weeks after the runtime upgrade - once a significant part of validators will be compliant with the requirement.  It is not enabled in the 2.1.0 release directly on purpose. 2.1.0 introduces support for StakingOperator which will ease the transition for some of the validators.

bkchr: Until this has not happened, I would not say that we should require 10K Dots for key registration. We can still increase this later on. Generally I don't know why this value is not just automatically following the required self bond, would have made this much easier.

Let's set this to 1k for now. We can still increase this later on.

sigurpol: I am fine with 1K as short-term spam protection. Again, this is a temporary stop-gap solution, it is not meant to stay.  We can't really enforce a min self stake validator on day 1 considering the current landscape where a significant part of validators do not meet the requirement. 10k was a way to create an incentive to have all new validators be compliant, that's why I am/was proposing that.

sigurpol: I've lowered now to 1k DOT -  which is more-than-enough as anti-spam protection (but helps less to incentivate new validators to be compliant - also true that we need primarily to convince the current existing validators first and most importantly so I think any value would work here)

sigurpol: and after further discussion with Basti + W3F friends back to 10k DOT :rofl:

ggwpez, line 516: Not sure if I understand: we require 10K dot as a storage deposit? Seems excessive, no?

ggwpez: resolved in dm

sigurpol: (discussed via DM - for preventing spam yes - 250 DOT or so would be more than fine, to enforce alignment towards the upcoming min validator bond requirement,  the consensus has been to go for 10k DOT

PR #1096, opened 2026-02-27
EPMB: Ensure that only root, admin and manager can manage

ggwpez:
Changes:
- Add sanity origin check for the `EPMB::manage` call


system-parachains/asset-hubs/asset-hub-polkadot/src/lib.rs
```diff
@@ -3023,4 +3027,41 @@ mod tests {
...
+
+			// try_successful_origin is feature gated
+			#[cfg(feature = "runtime-benchmarks")]
+			{
+				// manager origin could call
+				let manager = <Runtime as pallet_election_provider_multi_block::Config>::ManagerOrigin::try_successful_origin().unwrap();
+				hypothetically_ok!(call.clone().dispatch(manager));
+
+				// admin origin could call
+				let admin = <Runtime as pallet_election_provider_multi_block::Config>::AdminOrigin::try_successful_origin().unwrap();
```
bkchr: We know them here and can just hard code them. So, we don't need the feature to be activated.

ggwpez: They are actually both set to root so we already tested 🙃 I added an assert now that they are root.

PR #1103, opened 2026-03-09
Revert cap enactment block number

ggwpez:
Changes:
- Revert "Adjust hard cap enactment block number (#1076)"

We should revert this since the original block is going to be reached in 4 days and we cannot get the Polkadot ref in before that.  
Without reverting, it would change the enactment number after it was already enacted and could mess with the issuance calculation.

I dont think it would be a big issue, but also dont want to find out :D

### DIFF
--- a/.github/workflows/ai-review.yml
+++ b/.github/workflows/ai-review.yml
@@ -43,7 +43,7 @@ jobs:
           config: .config/review.polkacodes.yml
         env:
           POLKA_API_PROVIDER: openrouter
-          POLKA_MODEL: openai/gpt-5.2
+          POLKA_MODEL: openai/gpt-5.4
           GITHUB_TOKEN: ${{ github.token }}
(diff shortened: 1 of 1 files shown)


system-parachains/asset-hubs/asset-hub-polkadot/src/staking/mod.rs
```diff
@@ -885,7 +885,7 @@ mod tests {
 		}));
 	}
 
-	const MARCH_14_2026: RC_BlockNumber = 30_354_008;
+	const MARCH_14_2026: RC_BlockNumber = 30_349_908;
```
bkchr, line 888: Why do we have these numbers duplicated?

ggwpez: One is just for the test

PR #1111, opened 2026-03-16
Do not skip PET when running triggered test workflow

rockbmb:
PET is currently set to run against opened PRs, but not when `.github/workflows/test.yml` is manually triggered; see https://github.com/polkadot-fellows/runtimes/actions/runs/23109436806/job/67125996164

This PR changes that to allow PET runs on manually triggered workflows.


rockbmb:
Closed as #1145 fixes this.

PR #1116, opened 2026-03-24
people-polkadot: Enable elastic scaling with 2s block times and 3 cores

lexnv:
This PR enables elastic scaling for the people-polkadot chain.

Part of:
- https://github.com/paritytech/polkadot-sdk/issues/11483

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,6 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-
 ## [Unreleased]
 
 ### Added


system-parachains/people/people-polkadot/src/lib.rs
```diff
@@ -572,7 +579,7 @@ impl pallet_proxy::Config for Runtime {
 	type CallHasher = BlakeTwo256;
 	type AnnouncementDepositBase = AnnouncementDepositBase;
 	type AnnouncementDepositFactor = AnnouncementDepositFactor;
-	type BlockNumberProvider = System;
+	type BlockNumberProvider = RelaychainDataProvider<Runtime>;
```
lexnv, line 579: cc @gui1117 Please take another look over this change 🙏

gui1117: For existing proxy with announced calls before the runtime upgrade, it will make the required delay before dispatching the announced call being 0.

The doc say it shouldn't be changed for already deployed pallets.

I am looking now if there is any proxy deployed with non-zero delay on people chain. If yes, then the proper way is to have a migration to change all their announced calls from the system chain block number to the relay chain block number. Or accept the risk that the delay is ignored for announcement before the runtime upgrade.

gui1117: Apparently there is only 17 proxy accounts, and all have 0 delay configured. So maybe we can accept the risk of ignoring the delay for the all the proxy accounts that will be created before the runtime upgrade.

gui1117:
> We need confirmation that the `BlockNumberProvider` change doesn't break anything.
> 
> Otherwise, the sooner we get PC on 2s, the more time we have to identify and fix any issues on the node side.

It breaks a bit:
All proxy that announce a call before the runtime upgrade, will be able to execute the announced call and ignore the required delay after the runtime upgrade (for the calls announced before the runtime upgrade).
The delay for such announcement becomes 0, due to the change of the block provider.

That said there is only 17 proxy account on polkadot people chain. and all have a configured delay of 0.

So it sounds ok to accept the risk. Or we need a migration.

sandreim:
> It breaks a bit: All proxy that announce a call before the runtime upgrade, will be able to execute the announced call and ignore the required delay after the runtime upgrade (for the calls announced before the runtime upgrade). The delay for such announcement becomes 0, due to the change of the block provider.
> 
> That said there is only 17 proxy account on polkadot people chain. and all have a configured delay of 0.
> 
> So it sounds ok to accept the risk. Or we need a migration.

Also, sounds acceptable from my side, but can we get more opinions on this ?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+
```
lexnv:
```suggestion
```

PR #1143, opened 2026-04-10
Add `AuthorizeCall` transaction extension to all runtimes

gui1117:
This transaction extension is barely used for now, but we may use it in the future we should support it, better than expecting it and having it missed, and there is no downside to have it.


franciscoaguirre (approved):

bkontur (approved):
I think we will also need this when we remove ValidateUnsigned stuff

bkchr (approved):

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - [[#1114](https://github.com/polkadot-fellows/runtimes/pull/1114)] Asset Hub Kusama & Polkadot: EVM ERC-20 precompiles for foreign assets.
 - [[#1114](https://github.com/polkadot-fellows/runtimes/pull/1114)] Polkadot & Kusama relay: Set `max_relay_parent_session_age` to `0` in genesis presets (to be configured via governance).
 - Add Bulletin Polkadot system parachain runtime ([#1120](https://github.com/polkadot-fellows/runtimes/pull/1120)).
+- [[#1143](https://github.com/polkadot-fellows/runtimes/pull/1143)] All runtimes now support the transaction extension `AuthorizeCall`.
```
bkontur:
```suggestion
```

gui1117:
> I think we will also need this when we remove ValidateUnsigned stuff

Yes lots of ValidateUnsigned will be replaced by AuthorizeCall, but I imagine we write them in a backward compatible way that doesn't remove the ValidateUnsigned, only allows the 2 paths.

PR #1145, opened 2026-04-11
Disable some CI checks on non-release

ggwpez:
Speed up CI.  

Changes:
- **Only run PET tests on release MR or Release branch push**
- **microoptimize fmt**
- **forbid warnings**
- **quiet benchmark logging**
- **Only run zombienet tests on releases as well**
- **skip slow checks for drafts**


.github/workflows/test.yml
```diff
@@ -404,10 +439,10 @@ jobs:
 
   ecosystem-tests:
     name: "PET (${{ matrix.network.name }})"
-    needs: [build-runtimes]
+    needs: [build-runtimes, check-release-pr]
     runs-on: self-hosted
     timeout-minutes: 120
-    if: github.event_name == 'pull_request'
+    if: needs.check-release-pr.outputs.is-release-pr == '1'
```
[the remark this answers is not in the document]
ggwpez: Yes it now also runs on release branches.

.github/workflows/test.yml
```diff
@@ -126,8 +158,8 @@ jobs:
           echo "Running benchmarking for RUNTIME_BLOB_PATH=$RUNTIME_BLOB_PATH $EXCLUDE_EXTRINSICS"
           ./frame-omni-bencher v1 benchmark pallet --runtime $RUNTIME_BLOB_PATH --all --steps 2 --repeat 1 $EXCLUDE_EXTRINSICS --heap-pages 4096 --min-duration 0 --quiet
         env:
-          RUSTFLAGS: "-C debug-assertions -A warnings" # FAIL-CI AHM
-          RUNTIME_LOG: "error"
+          RUSTFLAGS: "-C debug-assertions"
+          RUNTIME_LOG: "off"
```
bkchr:
```suggestion
          RUNTIME_LOG: "error"
```

We should not deactivate it.

bkchr (approved):

.github/workflows/test.yml
ggwpez: Yea okay, but we mostly never check the output 😆

.github/workflows/test.yml
```diff
@@ -15,8 +17,37 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  check-release-pr:
```
bkontur, line 20: What about extracting this and doing also "check-migrations" just for the release branches?
I think Fellows can run it manually if needed,

.github/workflows/test.yml
bkontur, line 447: Could we somehow fix/if this, so we still have possibility to run manual workflows that will also run `ecosystem-tests` or `zombienet-smoke`? Just in case.

bkontur (approved):

.github/workflows/test.yml
ggwpez: Yes changed.

.github/workflows/test.yml
ggwpez: We could slim it down further, but for now the worst offenders are PET. The migration checks at least dont produce false-positives.

rockbmb: Makes sense to do this; I forgot to consider this when introducing the check. It really only makes sense to run against `main`.

joepetrowski (approved):

Issue #1119, opened 2026-03-27
Add the Bulletin Polkadot Chain runtime


bkontur:
## Summary

Onboard Bulletin Chain as a new system parachain (paraId 1010) in the Fellowship runtimes repository. Bulletin Chain provides content-addressed ephemeral storage for the Polkadot ecosystem.

For full context and rationale, see the [onboarding proposal](https://hackmd.io/@bkontur/SJWyQEdq-e).

### Stage 0 — Empty shell runtime
- [x] Fellowship [PR](https://github.com/polkadot-fellows/runtimes/pull/1120) with shell runtime (`system-parachains/bulletin/bulletin-polkadot/`, minimal scaffolding, explicitly **excludes** `pallet-bulletin-transaction-storage`)
- [x] Genesis config and collator setup (@franciscoaguirre) https://github.com/polkadot-fellows/runtimes/pull/1146
- [x] OpenGov referendum: ParaId 1010 reservation + registration, coretime allocation (@karolk91) https://github.com/polkadot-fellows/runtimes/issues/1153
- [x] Part of the Release <del>2.2.0</del>2.2.1 or 2.2.2
- [x] [Post-onboarding steps](https://github.com/paritytech/polkadot-bulletin-chain/issues/343)

### Stage 1 — Storage pallet activation (target: before Web3 Summit Berlin, June 18–19, 2026)
- [x] Fellowship PR: integrate `pallet-bulletin-transaction-storage`
- [ ] HRMP channels to People Chain and Asset Hub
- [x] Initial storage authorization (OpenGov root + PeopleChain root)
- [ ] Part of the Release 2.3.0

## References
- [Onboarding proposal](https://hackmd.io/@bkontur/SJWyQEdq-e)
- [Bulletin Chain repo](https://github.com/paritytech/polkadot-bulletin-chain)
- Already live on Westend (paraId 2487) and Paseo (paraId 5118)

bkontur:
closing, already live

PR #1146, opened 2026-04-14
feat: add bulletin chain spec to generator

franciscoaguirre:
Part of: https://github.com/polkadot-fellows/runtimes/issues/1119

I added the Bulletin Chain live genesis preset and used it in the chain spec generator

### DIFF
--- a/chain-spec-generator/src/system_parachains_specs.rs
+++ b/chain-spec-generator/src/system_parachains_specs.rs
@@ -376,7 +376,6 @@ pub fn bulletin_polkadot_config() -> Result<Box<dyn sc_chain_spec::ChainSpec>, S
 	properties.insert("tokenSymbol".into(), "DOT".into());
 	properties.insert("tokenDecimals".into(), 10.into());
 
-	// TODO: Add post-launch.
 	let boot_nodes: [&str; 0] = [];
 
 	Ok(Box::new(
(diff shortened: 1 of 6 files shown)


bkchr (approved):

chain-spec-generator/src/system_parachains_specs.rs
```diff
@@ -369,6 +369,40 @@ pub fn people_polkadot_local_testnet_config() -> Result<Box<dyn sc_chain_spec::C
...
 }
 
+#[cfg(feature = "bulletin-polkadot")]
+pub fn bulletin_polkadot_config() -> Result<Box<dyn sc_chain_spec::ChainSpec>, String> {
+	let mut properties = sc_chain_spec::Properties::new();
+	properties.insert("ss58Format".into(), 0.into());
+	properties.insert("tokenSymbol".into(), "DOT".into());
+	properties.insert("tokenDecimals".into(), 10.into());
+
+	// TODO: Add post-launch.
```
bkontur: I think that we don't need to update this after release. We should better remove this `fn bulletin_polkadot_config()` (we keep just genesis live preset (should never change)), because the next commit after the release commit, there will be a different `bulletin_polkadot_runtime::WASM_BINARY`, which means different genesis hash, so changing bootNodes here does not help that much.

@seadanda should we also remove similar `fn coretime_polkadot_config()` https://github.com/polkadot-fellows/runtimes/blob/93b5bcc82ea9927829083221f2df1c59f27cf4e0/chain-spec-generator/src/system_parachains_specs.rs#L290 or is it needed?

```suggestion
```

bkontur (approved):

chain-spec-generator/src/system_parachains_specs.rs
seadanda: After discussion in DM, yea we should remove both after launch. The concrete genesis chainspecs are added to the sdk so they are tracked somewhere, then afterwards they shouldn't change

PR #1149, opened 2026-04-15
bulletin-chain: switch to 24s aura slots

sandreim:
We missed this during initial review. Longer AURA slots improve elastic scaling tput and censorship resistance.

### DIFF
--- a/system-parachains/bulletin/bulletin-polkadot/src/lib.rs
+++ b/system-parachains/bulletin/bulletin-polkadot/src/lib.rs
@@ -69,7 +69,7 @@ use sp_version::NativeVersion;
 use sp_version::RuntimeVersion;
 use system_parachains_constants::{
 	async_backing::{
-		AVERAGE_ON_INITIALIZE_RATIO, HOURS, MAXIMUM_BLOCK_WEIGHT, MILLISECS_PER_BLOCK,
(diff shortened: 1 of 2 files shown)


sandreim:
This needs to go with 2.2.1 or 2.2.2 release, depending on which will be used for onboarding. CC @ggwpez

system-parachains/bulletin/bulletin-polkadot/src/lib.rs
```diff
@@ -91,8 +91,8 @@ use xcm_runtime_apis::{
 	fees::Error as XcmPaymentApiError,
 };
 
-/// Bulletin uses 6s slot duration (same as block time).
-pub const SLOT_DURATION: u64 = MILLISECS_PER_BLOCK;
+/// Bulletin uses 24s slot duration.
+pub const SLOT_DURATION: u64 = 4 * MILLISECS_PER_BLOCK;
```
skunert: Cant we just put 24s here directly? `MILLISECS_PER_BLOCK ` is a bit misleading because we might switch to 2s blocks at some point, but then this will not be 8s. Its just a nit though.

skunert (approved):

bkontur (approved):

ggwpez (approved):

PR #1152, opened 2026-04-20
Update Polkadot runtime construction and aggregation implementation

knzeng-e:
This PR implements the new runtime construction format.


shawntabrizi (approved):

muharem (commented):
how about updating all the chains?

PR #1155, opened 2026-04-23
Add elastic scaling zombienet tests for Polkadot AH and PC

sandreim:
Add some basic ES tests to ensure we produce the expected amount of blocks with 1 and 3 cores.
To ensure the test throughput is stable, we are now waiting on all validators to have compiled the PVF. Without this change, the PVF compilation would have happened during the measurement period making the test flaky. 

I've also fixed the existing smoke test.


integration-tests/zombienet/tests/elastic_scaling_asset_hub.rs
```diff
@@ -0,0 +1,74 @@
+//! Elastic scaling integration test for **Asset Hub Polkadot** (para_id 1000).
```
alindima: could we deduplicate these tests into one and just use rstest for parameterisation?

Ideally also using test helpers we already defined in polkadot-sdk?

sandreim: I'll also add people chain in the same test, but first I want to make the test stable.

integration-tests/zombienet/src/helpers.rs
```diff
@@ -0,0 +1,85 @@
+//! Throughput helper used by the elastic scaling integration test.
+//!
+//! Mirrors `cumulus-zombienet-sdk-helpers::assert_para_throughput` from polkadot-sdk;
```
pepoviola, line 3: Maybe make sense to move this logic to zombienet-sdk itself as helper module. wdyt?

sandreim: Yes a good idea, but also look at Polkadot SDK, there is a lot of stuff there that I can't pull in here from the helper. crate, because of conflicts of crate depedency versions.

ggwpez (approved):

ggwpez (approved):

integration-tests/zombienet/tests/elastic_scaling.rs
```diff
@@ -0,0 +1,94 @@
...
+	let first_collator = network.get_node(collators[0])?;
+	assert!(
+		first_collator.wait_until_is_up(120u64).await.is_ok(),
+		"collator {} failed to come up",
+		collators[0]
+	);
+
+	wait_for_pvf_prepared(&network, ELASTIC_VALIDATORS, 1, 300).await?;
+
+	let measurement_result = assert_para_throughput(
```
alindima, line 53: would be nice to also assert the finality lag, as we do in polkadot-sdk tests

alindima (approved):
LGTM

kianenigma (approved):

PR #1164, opened 2026-05-05
feat: deploy pallet-safe-mode and pallet-tx-pause to Asset Hub Polkadot & Kusama

### LINKED ISSUE #1089: Deploy `tx-pause`/`safe-mode`
We should deploy one or both of the pallets on all system chains + relay chain. They can be used to lock down certain functionality, for example in the case of a security issue.

We will need to support normal transactions + XCM (which may requires some work to make the pallets compatible).

Kanasjnr:
Closes #1089

## Summary
Deploys `pallet-safe-mode` and `pallet-tx-pause` to Asset Hub Polkadot (PAH) and Asset Hub Kusama (KAH). 

**Major fixes in the latest update (per xlc review):**
*   **Activated Call Filters**: Wired both pallets into `BaseCallFilter` in `frame_system::Config`. Previously they were inert.
*   **Fixed TxPause**: Changed `WhitelistedCalls` from `Everything` to `Nothing` so that pausing actually functions (in this pallet, "whitelisted" means "cannot be paused").
*   **Governance Safety**: Expanded the `SafeModeWhitelist` to include `ConvictionVoting`, `Preimage`, and `TxPause`. This ensures OpenGov stays functional during a lockdown and allows governance to transition from a general halt to targeted pauses.
*   **Origin Precision**: Switched forced origins from local Asset Hub tracks to the official Technical Fellowship XCM voice (`EnsureXcm<IsFellowshipVoice>`).

### SafeMode (both PAH & KAH)
| Parameter | Value |
|---|---|
| Enter deposit | 100,000 UNITS (permissionless) |
| Extend deposit | 100,000 UNITS |
| Enter duration | 1 day |
| Extend duration | 1 day |
| **Release delay** | **60 days** |
| **Force origins** | `EnsureRoot` \| `EnsureXcm<TechnicalFellowship>` |
| **Whitelisted calls** | `System`, `Timestamp`, `ParachainSystem`, `Referenda`, `ConvictionVoting`, `Preimage`, `TxPause` |

*Note: Whitelist includes minimal governance and TxPause to ensure the chain remains governable during a safe-mode activation.*

### TxPause (both PAH & KAH)
| Parameter | Value |
|---|---|
| Pause/Unpause origin | `EnsureRoot` \| `EnsureXcm<TechnicalFellowship>` |
| **Whitelisted calls** | **Nothing** (no call is permanently exempt from pausing) |

## Weights
Using each pallet's built-in `SubstrateWeight<Runtime>` as a conservative placeholder. Runtime-specific benchmarks will be added in a follow-up once `frame-omni-bencher` is run against the production WASM.

## To-do / Discussion
*   **XCM compatibility**: To be discussed during review per the original issue.


Kanasjnr:
@bkchr ready for review

system-parachains/asset-hubs/asset-hub-polkadot/src/safe_mode.rs
```diff
@@ -0,0 +1,62 @@
...
+use super::*;
+use frame_support::traits::{Contains, EitherOf, MapSuccess};
+use sp_runtime::traits::Replace;
+
+parameter_types! {
+	pub const EnterDepositAmount: Option<Balance> = Some(100_000 * UNITS);
+	pub const ExtendDepositAmount: Option<Balance> = Some(10_000 * UNITS);
+	pub const EnterDuration: BlockNumber = 72 * RC_HOURS; // 3 days
+	pub const ExtendDuration: BlockNumber = 24 * RC_HOURS; // 1 day
+	pub const ReleaseDelay: Option<BlockNumber> = Some(RC_DAYS); // 1 day
```
ggwpez:
```suggestion
	pub const EnterDepositAmount: Option<Balance> = Some(100_000 * UNITS);
	pub const ExtendDepositAmount: Option<Balance> = Some(100_000 * UNITS);
	pub const EnterDuration: BlockNumber = 24 * RC_HOURS; // 1 day
	pub const ExtendDuration: BlockNumber = 24 * RC_HOURS; // 1 day
	pub const ReleaseDelay: Option<BlockNumber> = Some(30 * RC_DAYS); // 30 days
```

Release delay needs to be long enough for governance to slash malicious enter calls. so it has 30-1=29 days to do that.

system-parachains/asset-hubs/asset-hub-polkadot/src/safe_mode.rs
```diff
@@ -0,0 +1,62 @@
...
+}
+
+pub struct SafeModeWhitelist;
+impl Contains<RuntimeCall> for SafeModeWhitelist {
+	fn contains(call: &RuntimeCall) -> bool {
+		matches!(
+			call,
+			RuntimeCall::System(_) | RuntimeCall::Timestamp(_) | RuntimeCall::ParachainSystem(_)
```
ggwpez: Need to think about adding minimal governance. Since otherwise we cannot propose a relieve ref while safe-mode is enabled.

system-parachains/asset-hubs/asset-hub-polkadot/src/safe_mode.rs
```diff
@@ -0,0 +1,62 @@
...
+	// ForceEnterOrigin success value must be BlockNumber (duration); use Replace to map () ->
+	// EnterDuration
+	type ForceEnterOrigin =
+		MapSuccess<EitherOf<EnsureRoot<AccountId>, FellowshipAdmin>, Replace<EnterDuration>>;
+	// ForceExtendOrigin success value must be BlockNumber (duration); use Replace to map () ->
+	// ExtendDuration
+	type ForceExtendOrigin =
+		MapSuccess<EitherOf<EnsureRoot<AccountId>, FellowshipAdmin>, Replace<ExtendDuration>>;
+	type ForceExitOrigin = EitherOf<EnsureRoot<AccountId>, FellowshipAdmin>;
+	type ForceDepositOrigin = EitherOf<EnsureRoot<AccountId>, FellowshipAdmin>;
```
ggwpez: Not sure about the fellowship admin here. Lets keep for now until I asked back but we will probably remove it

system-parachains/asset-hubs/asset-hub-kusama/src/safe_mode.rs
```diff
@@ -0,0 +1,68 @@
...
+use frame_support::traits::{Contains, EitherOf, MapSuccess};
+use kusama_runtime_constants::currency::UNITS;
+use sp_runtime::traits::Replace;
+
+parameter_types! {
+	pub const EnterDepositAmount: Option<Balance> = Some(100_000 * UNITS);
+	pub const ExtendDepositAmount: Option<Balance> = Some(100_000 * UNITS);
+	pub const EnterDuration: BlockNumber = 24 * RC_HOURS; // 1 day
+	pub const ExtendDuration: BlockNumber = 24 * RC_HOURS; // 1 day
+	pub const ReleaseDelay: Option<BlockNumber> = Some(30 * RC_DAYS); // 30 days
```
bkchr: IMO we should increase this to 48 days (double of the root proposal time). Mainly to be on the safe side, if there are any issues.

xlc (changes requested):
I asked codex to review and here are the findings (and I checked and believe they are valid)

----

P0: The pallets are currently inert. The PR adds SafeMode/TxPause to construct_runtime!, but does not compose them into frame_system::Config::BaseCallFilter. PAH still uses AhMigrator and KAH still uses Everything at asset-hub-polkadot/src/lib.rs (line 243) and asset-hub-kusama/src/lib.rs (line 193). Both SDK pallets only affect dispatch when used as call filters. SafeMode::enter and TxPause::pause would store state/events, but calls would keep executing.

P0: TxPause is configured to pause nothing. The PR sets type WhitelistedCalls = Everything in [tx_pause.rs](https://github.com/polkadot-fellows/runtimes/blob/2ae5b0e009ad37a95b174c4f90b10b506c4e4b3e/system-parachains/asset-hubs/asset-hub-polkadot/src/tx_pause.rs#L25-L29). In the SDK, whitelisted calls are “calls that cannot be paused”, and is_paused immediately returns false for them at tx-pause/src/lib.rs (line 124) and tx-pause/src/lib.rs (line 233). Use Nothing or a narrow custom whitelist instead.

P1: Safe mode does not preserve the governance recovery flow. The whitelist only allows System, Timestamp, ParachainSystem, and Referenda in [safe_mode.rs](https://github.com/polkadot-fellows/runtimes/blob/2ae5b0e009ad37a95b174c4f90b10b506c4e4b3e/system-parachains/asset-hubs/asset-hub-polkadot/src/safe_mode.rs#L28-L39). But voting is ConvictionVoting, proposal material often needs Preimage, and the intended “broad halt -> targeted tx pause -> exit safe mode” flow needs TxPause callable while safe mode is active. Without those, a permissionless 100k DOT/KSM activation can block most emergency handling until timeout.

P1: Be precise about FellowshipAdmin. The configured FellowshipAdmin is the Asset Hub local OpenGov custom origin, defined in origins.rs (line 52), not the Technical Fellowship XCM voice. Other runtime config uses EnsureXcm<IsFellowshipVoice<...>> when it means the collectives fellowship, for example governance/mod.rs (line 65). If the intended actor is Technical Fellowship, this origin is wrong; if the intended actor is the local AH FellowshipAdmin track, document that it can force-enter, force-extend, force-exit, and slash/release deposits.

P2: Once the filters are wired, weights need a broader refresh. The PR only uses the new pallets’ built-in SubstrateWeight, but the filters add state reads to dispatch. Current Asset Hub weights do not include SafeMode::EnteredUntil or TxPause::PausedCalls; compare the SDK filter behavior at safe-mode/src/lib.rs (line 593). This should be covered by runtime-specific benchmarking or an explicit weight accounting decision.

-----

We need some integration tests, which should hopefully catch most of the configuration issue listed above.
I will say no need to worry about weight in this PR.

100% need some full e2e tests with https://github.com/open-web3-stack/polkadot-ecosystem-tests

monsieurbulb:
**Security review — permissionless trigger analysis**

First, thanks for the quick turnaround on xlc's review. The `BaseCallFilter` wiring, `WhitelistedCalls = Nothing` fix, governance whitelist expansion, and `EnsureXcm<IsFellowshipVoice>` origins all look right now. 

A few things I think are worth flagging before this ships:

### 1. Permissionless trigger economics differ significantly between PAH and KAH

The permissionless entry cost is 100,000 UNITS, locked for the duration of safe mode (1 day) plus a **60-day release delay** — so ~61 days of capital lockup. The deterrent strength varies enormously by chain:

| Chain | 100k at ~spot | Approx capital locked |
|-------|--------------|----------------------|
| PAH (DOT) | ~$4–10 | $400k–$1M |
| KAH (KSM) | significantly lower | proportionally lower |

On PAH this is a strong deterrent. On KAH, depending on spot price, 100k KSM might be cheap enough that a well-capitalised attacker can absorb the lockup cost. Worth considering whether the deposit should be parameterised differently across the two chains.

### 2. The 60-day release delay works against legitimate triggers too

The permissionless trigger argument — "a single alert person can save the chain at 3am" — assumes someone is willing to lock 100k UNITS for ~2 months. A genuine good samaritan with 100k DOT might think twice about locking that capital for 61 days. The delay protects against deposit-griefing exit scams but also raises the bar for the exact use case the permissionless design is meant to enable. Could a graduated release (e.g., partial release after exit, full after delay) or a lower deposit + shorter delay pair be worth exploring?

### 3. Extension griefing: can this stack indefinitely?

The `ForceExtendOrigin` is gated behind Root/Fellowship, so the concern there is minimal. But the permissionless extend path (`pallet_safe_mode::Call::extend`) — can it be called repeatedly by different accounts, each paying 100k UNITS to stack another day? If so, a well-funded attacker could chain-extend safe mode (and TxPause could be similarly played if pauses can be restacked). Could you clarify whether there's a cap on total extension length?

### 4. XCM behaviour during safe-mode / tx-pause

Still flagged as TBD in the PR description, and I think it's the most important thing to nail down before merging. Specifically:
- **Inbound XCM during safe mode**: Are incoming reserve transfers, teleports, and `Transact` calls filtered? If not, an attacker who triggers safe mode on AH can exploit price dislocation while freely moving assets onto the frozen chain.
- **Inbound XCM during tx-pause**: If a call is paused locally but reachable via XCM `Transact`, you get an asymmetry where native users are blocked but XCM callers aren't.
- **Outbound XCM**: If outbound XCM is paused but users can still interact with the chain locally, you may trap assets. Worth documenting the intended behaviour and ensuring it's consistent across both pallets.

### 5. TxPause + SafeMode interaction edge case

With `WhitelistedCalls = Nothing` and no permanent exemptions, it's technically possible (though governance-gated) to pause the `TxPause::unpause` call itself if SafeMode is already active. That would leave the chain in safe mode with no way to transition to targeted pauses. Since pause/unpause requires Root/Fellowship, this is a governance-compromise scenario rather than a permissionless attack vector, but worth a note in the docs for operators about the intended safe-mode→tx-pause→exit workflow.

### 6. Integration tests

Echoing xlc — e2e tests covering at minimum: permissionless enter, governance force-exit, tx-pause during safe mode, and XCM behaviour across all three states (normal / safe-mode / paused) would catch most of the configuration foot-guns.

---

None of the above should be blockers if the XCM question gets a clear answer during review, but points 1–4 are worth addressing in the PR description or a follow-up issue before this hits production governance.

oshakarishvili:
First off, this is genuinely needed, so thanks for pushing this forward. That said, coming at this from a governance rather than a technical angle, I have some concerns about the mechanism design that I think are worth working through before this goes to a production vote.

1. 100,000 DOT locked for 61 days is $120K–$500K+ in capital tied up by a single address with no pooling mechanism and no distinction between a good-faith trigger and a malicious one. The "permissionless" framing implies any community member can pull this, but realistically, only a small set of wealthy actors can afford to. A genuinely good participant acting at 3 am during an exploit would have their capital locked for two months regardless of outcome, so there's no path and no reward for acting correctly. 

2. The Fellowship fast-track still requires a community vote, meaning even the fastest recovery path takes several days minimum. Safe Mode only lasts 24 hours per deposit. To cover a full fix cycle, you'd need 4+ consecutive extensions, that's 400,000 + DOT locked across ~62 days of capital commitment per depositor. That's maybe not a realistic ask.

3. If I understood correctly, TxPause, as configured, gives the Fellowship unilateral power to pause any call on Asset Hub without a community vote. That's executive control, which sits outside what the Fellowship manifesto defines as their mandate: they're a technical standards body, not an ops committee. Pausing Assets::transfer for users without community approval is a meaningful power grant that I don't think has been explicitly debated or ratified anywhere. This should be a deliberate community decision, not a config detail in a runtime PR.

One thing that might be worth exploring as an alternative or complement is Lido's Dual Governance mechanism that solves a structurally similar problem in a way that could translate well here. Rather than one address posting a large fixed deposit, any token holder contributes any amount to a shared on-chain escrow. Nothing happens until the pool crosses a meaningful collective threshold, and the duration scales proportionally with how much is locked. No single actor carries the full risk or cost.

Adapted to Polkadot's actual numbers (1.66B DOT issuance, 5% support required for whitelisted proposals = ~83M DOT as the meaningful participation anchor):

- **5% of issuance (~83M DOT) in escrow** → Safe Mode activates. This matches the existing whitelisted proposal support threshold, hence the community has already decided this is a meaningful signal. Duration scales from 5 days upward, proportional to the amount locked.
- **10% (~166M DOT)** → Extended freeze, maximum duration.
- **15% (~250M DOT)** → Full rage quit, all pending governance execution halted until escrow clears.

The key design principles this preserves are that it's harder to freeze the network than to approve a fix (thresholds are at or above the existing governance participation bar), no single actor can trigger it alone, depositors get their capital back when the crisis resolves with no 60-day lockup, a governance referendum determines whether the trigger was legitimate and can slash the pool if it was malicious. The honest caveat is that any meaningful threshold still requires whale coordination, given current participation levels, but the pooling and adjudication mechanics are fairer than a single-address deposit, regardless. 

This is just an idea, and I'd be happy to explore it further if there's interest. But either way, the mechanism parameters and the Fellowship authority question feel like they warrant a broader governance conversation before this hits production.

system-parachains/asset-hubs/asset-hub-kusama/src/tx_pause.rs
```diff
@@ -0,0 +1,35 @@
...
+
+parameter_types! {
+	pub const MaxNameLen: u32 = 256;
+}
+
+impl pallet_tx_pause::Config for Runtime {
+	type RuntimeEvent = RuntimeEvent;
+	type RuntimeCall = RuntimeCall;
+	type PauseOrigin = EitherOfDiverse<EnsureRoot<AccountId>, EnsureXcm<IsFellowshipVoice>>;
+	type UnpauseOrigin = EitherOfDiverse<EnsureRoot<AccountId>, EnsureXcm<IsFellowshipVoice>>;
```
ggwpez:
```suggestion
	type PauseOrigin = EnsureRoot<AccountId>;
	type UnpauseOrigin = EnsureRoot<AccountId>;
```

ggwpez:
> The "permissionless" framing implies any community member can pull this, but realistically, only a small set of wealthy actors can afford to.

Yes. All of this is intended as improvement over the status quo. The general idea is that it is better to have 100K DOT locked up than to loose it due to not having save-mode available at all.

But agreed that it is not ideal. It requires a large chunk of capital to be bound and trust in OpenGov to not slash it.

>A genuinely good participant acting at 3 am during an exploit would have their capital locked for two months regardless of outcome, so there's no path and no reward for acting correctly.

Correct, but I would expect OpenGov to pay some bug bounty in that case.

>The Fellowship fast-track still requires a community vote, meaning even the fastest recovery path takes several days minimum. Safe Mode only lasts 24 hours per deposit. To cover a full fix cycle, you'd need 4+ consecutive extensions, that's 400,000 + DOT locked across ~62 days of capital commitment per depositor. That's maybe not a realistic ask.

Yes. Do you know alternatives? This is the most simple approach, even though quite capital inefficient.

>If I understood correctly, TxPause, as configured, gives the Fellowship unilateral power to pause any call on Asset Hub

This is indeed a misconfiguration. Requested it to be changed now.

>Nothing happens until the pool crosses a meaningful collective threshold, and the duration scales proportionally with how much is locked. No single actor carries the full risk or cost.

This can partially be done with a Smart Contract, but contracts can only access transferrible balance.  
So they would have to be transferred to a contract account that then posts it to the safe-mode.  

If we want also locked balance to be usable (like for voting in general) then we would have to change the pallet, since it uses reserves that do not overlap with staking or referenda locks.

Kanasjnr:
@ggwpez @xlc @bkchr pushed a follow-up added Whitelist + Utility to the safe-mode whitelist, release delay is now 30d on PAH / 48d on KAH.  ready for another look when you get a chance

PolkadotDom:
Have spoken on this a bit on element but will bring it here-  My issue with this setup is that we have both permissionless bridging (hyperbridge / snowbridge) and permissionless contracts. Meaning it's not just the Polkadot community who can pause our economy. This is a huge national security risk. Do we really feel economically powerful adversaries would care about 100k a day to shut us down?

I would prefer (at a minimum) we modify the pallets to use PoP for both the extrinsic and the crowdsourcing, then also have a blacklist the extrinsics check against. I don't feel the current setup would be worth it as an interim state.

Kanasjnr:
> Have spoken on this a bit on element but will bring it here- My issue with this setup is that we have both permissionless bridging (hyperbridge / snowbridge) and permissionless contracts. Meaning it's not just the Polkadot community who can pause our economy. This is a huge national security risk. Do we really feel economically powerful adversaries would care about 100k a day to shut us down?
> 
> I would prefer (at a minimum) we modify the pallets to use PoP for both the extrinsic and the crowdsourcing, then also have a blacklist the extrinsics check against. I don't feel the current setup would be worth it as an interim state.

@PolkadotDom I don't have a good answer yet 🙁, with permissionless bridging and permissionless contracts live, the threat model isn't "griefer wastes 100k," it's "well-funded actor extracts more than 100k/day elsewhere while Asset Hub is frozen." A flat deposit doesn't scale with that.

Before pushing this further I'd like to either get a rough bound on realistic extraction value during a freeze to see if the deposit is even in the right order of magnitude, or treat PoP-gating as a real follow-up rather than dismiss it, given People chain already exists in-ecosystem. Would rather get this right than ship an interim state that's cheaper to attack than it looks. 
 @ggwpez @bkchr what do you think?

monsieurbulb:
Agreed on gating the crowdsourcing rather than only the extrinsic. A personhood gate changes who may pull the lever though, not the economics underneath it. Entry cost stays fixed while the payoff still scales with the attacker's position elsewhere, so a capital attack becomes a rented-identity attack, and identity is a good deal cheaper to rent than 100k DOT. Necessary, not sufficient.

Two constraints I would size a personhood gate against rather than assume away:

1. Verified population. A distinct-persons threshold only means something once enough people hold a proof for N to be reachable at 3am, and getting to that number means paying for adoption in some form. Whoever subsidises it sets the price of the marginal alias, which is what an attacker pays to rent one. So the threshold wants sizing against verified population and the cost of acquiring it, and the gate is cheapest to defeat while that population is small.
2. Attribution. Aliases are unlinkable across contexts, so a malicious trigger cannot be traced back, the remedy stops at the slashed deposit, and the same human re-enters under a fresh alias elsewhere. That is the part that makes @PolkadotDom's blacklist hard, since there is nothing stable to list.

On @ggwpez's June question, the alternative I would suggest: keep the deposit as a spam floor, require a threshold of **distinct persons** to enter, and scale duration with pool size rather than a flat day per deposit. No single actor can buy a freeze at any price, and the honest 3am case gets cheaper per participant rather than more expensive. Pallet work either way, per @ggwpez on reserves not overlapping the staking and referenda locks.

Two things for this PR:

- Ship both pallets with permissionless entry disabled and only the force origins wired. The Fellowship gets the tool now, the lever stays shut until there is something better than a deposit behind it, and enabling it later is a parameter change.
- On bounding the extraction value, the quantity is what reprices while AH cannot settle. Bridged assets in flight, contract-held collateral, anything depending on an AH oracle or venue. Estimable from on-chain data, unlike a round 100k.

Kanasjnr:
Been sitting on this and I think the deposit-size framing (100k, PoP-gated or not) is solving the wrong layer of the problem.

The mismatch is structural, not a calibration problem. A fixed bond is only a valid deterrent when the harm it gates is bounded and known at config time. `SafeMode::enter's `payoff to an attacker is the value of freezing Asset Hub and once permissionless bridges and permissionless contracts sit on top of AH, that value is a claim on external state we don't control and can't bound: bridged assets in flight, contract-held collateral, anything pricing off an AH oracle. 

There's no D such that D > attacker payoff, because payoff isn't a property of this pallet it's a property of everything built on AH, including things that don't exist yet. Raising the deposit, or gating it behind PoP, moves the crossover point; it doesn't remove it, and we have no way to know which side of it we're on.

It's also worse than "one attacker sizes up against a bond." I went and checked `pallet_safe_mode::extend` directly: it's open to any signed account, not just the original enterer, and uncapped which means a sustained freeze doesn't require one well-capitalized actor holding a growing position  it can be a rotating cast of independent, uncoordinated accounts, each paying for one day, none bearing the cumulative cost or cumulative risk. 

Given that the directions that have been mentioned 

First distinct-persons quorum (@monsieurbulb's proposal) prices coordinated humans instead of capital, and capital is something a well-funded adversary simply has. But it's only as strong as verified People-chain population, and that number is outside this PR's control shipping this now means betting safety on an adoption curve we don't own.

also root/Fellowship-only, permissionless entry disabled removes the mismatch entirely by removing the economic game. But we should be honest about what it is: Root and Fellowship-XCM can already do this today; this configuration gives a faster invocation path (one extrinsic vs. a runtime upgrade to add an emergency filter), not new authority. That's real value, but it's a governance-speed tool, not the permissionless circuit breaker #1089 asked for, and i dont think we should frame it that way.

another angle we have not discussed is to reduce the payoff instead of only raising the cost. Since cost can't be bounded against an unbounded payoff, the other lever is scoping the freeze so triggering it doesn't help an extraction attacker 

for instance: blocking arbitrary Transact / asset movement while leaving user-protective exits open, instead of a blanket halt that freezes whatever the attacker wants frozen too.

Issue #1071, opened 2026-02-11
update multi-asset bounties pallet account derivation (mbt/mcb) so runtimes can drop local types


dhirajs0:
## Context

We use `pallet_multi_asset_bounties` on Asset Hub (Kusama & Polkadot) with the same treasury `PalletId` as the legacy single-asset bounties (`pallet_bounties` / `pallet_child_bounties`). The pallet’s built-in account derivation uses fixed sub-account prefixes:

- **Bounty account:** `PalletId + "bt" + bounty_id`
- **Child-bounty account:** `PalletId + "cb" + parent_id + child_id`

Those match the legacy bounties pallets. So for the same `PalletId` and same numeric ids, legacy bounty index 5 and multi-asset bounty index 5 would resolve to the **same** account.

## Current Solution

We added local types in `system-parachains/common` that implement the same `TryConvert` behaviour but use distinct prefixes so account IDs do not collide: [PR #1070](https://github.com/polkadot-fellows/runtimes/pull/1070)

- **Bounty:** `PalletId + "mbt" + bounty_id` (instead of `"bt"`)
- **Child-bounty:** `PalletId + "mcb" + parent_id + child_id` (instead of `"cb"`)

- **Location:** `system-parachains/common/src/multi_asset_bounty_sources.rs`
- **Types:** `MultiAssetBountySourceFromPalletId`, `MultiAssetChildBountySourceFromPalletId`
- **Used in:** Asset Hub Kusama & Polkadot treasury config (`BountySource` / `ChildBountySource`)

Once [Issue#11052](https://github.com/paritytech/polkadot-sdk/pull/11052) is merged and the new version of crate is published, we will remove the local types and use the types from `pallet_multi_asset_bounties`.

## Acceptance criteria (for runtimes)
- [ ] Pallet provides (or supports) bounty/child-bounty source types that derive accounts with distinct prefixes (e.g. `"mbt"` / `"mcb"`) so they do not collide with legacy `"bt"` / `"cb"`.
- [ ] Runtimes can switch Asset Hub treasury config to use the pallet’s types for `BountySource` and `ChildBountySource`.
- [ ] Remove local account-derivation types and the `multi-asset-bounties` feature from `system-parachains/common` (delete `multi_asset_bounty_sources.rs`, revert related `Cargo.toml` / `lib.rs` and treasury config changes).

## References

- Local implementation: `system-parachains/common/src/multi_asset_bounty_sources.rs`
[...]

PR #1168, opened 2026-05-06
refactor(asset-hubs): use upstream multi-asset bounty source types

dhirajs0:
Closes issue #1071 

## Description

Asset Hub runtimes were using local wrapper source types from `system-parachains-common`
for `multi-asset bounties` and `child-bounties`. Those wrappers were already marked for
removal and duplicated logic that now exists upstream in `pallet_multi_asset_bounties`.

This PR removes the duplicate local module, wires both Asset Hub runtimes to the upstream
source types directly, and migrates the funds of any existing multi-asset bounty/child-bounty
to the new account derivation introduced upstream in
[paritytech/polkadot-sdk#11052](https://github.com/paritytech/polkadot-sdk/pull/11052).

### Source-type consolidation
- Switch `BountySource`/`ChildBountySource` in both runtimes to:
  - `pallet_multi_asset_bounties::BountySourceFromPalletId`
  - `pallet_multi_asset_bounties::ChildBountySourceFromPalletId`
- Use upstream prefix constants:
  - `pallet_multi_asset_bounties::BountyAccountPrefix` (`*b"mbt"`)
  - `pallet_multi_asset_bounties::ChildBountyAccountPrefix` (`*b"mcb"`)
- Remove local implementation module: `system-parachains/common/src/multi_asset_bounty_sources.rs`
- Remove the `multi-asset-bounties` optional feature from `system-parachains/common/Cargo.toml`
  and from both Asset Hub `Cargo.toml`s.
- Regenerate lockfile entries affected by the dependency cleanup.

### Encoding shift (why a migration is needed)
The wrapper's prefix was `&'static str` (`"mbt"` / `"mcb"`); the upstream prefix is `[u8; 3]`
(`*b"mbt"` / `*b"mcb"`). SCALE-encodes differently: `&str` carries a 1-byte compact-length
tag, `[u8; 3]` does not — so the sub-account seeds (and therefore the derived AccountIds) are
not byte-identical. Upstream's `[u8; 3]` is deliberate (predictable seed size, see SDK PR);
Asset-Hub-Westend already runs on this encoding. We converge KAH/PAH to it via a one-shot
balance migration.

### Migration: `MigrateBountyAccountAssets`
For each entry in `pallet_multi_asset_bounties::Bounties` and `ChildBounties`:
- Compute the OLD AccountId (legacy `&str` derivation) and the NEW AccountId (upstream
  `[u8; 3]` derivation).
- Use the runtime's existing `pallet_bounties::Config::TransferAllAssets`
  (`TransferAllFungibles<AccountId, NativeAndAssets, BountyRelevantAssets>`) to sweep every
  fungible the pot might hold from OLD → NEW.
- On KAH the migration uses a *local* `BountyMigrationAssets` parameter that extends
  `BountyRelevantAssets` with DOT, since multi-asset bounties on KAH are funded in DOT (a
  foreign asset the legacy bounties pallet does not need to handle). KAH's
  `BountyRelevantAssets` is left untouched.
- Native account collisions with the legacy `pallet-bounties` are not possible — legacy uses
  `"bt"`/`"cb"` prefixes, multi-asset uses `"mbt"`/`"mcb"`, so the derived sub-accounts are
  disjoint. Sweeping native is safe.

### Tests
Two end-to-end migration tests (one per runtime) seed a `Bounties` entry, credit the OLD
pot account with each asset class, run `on_runtime_upgrade()`, and assert the funds land at
the NEW account:
- PAH: native DOT + USDT + USDC.
- KAH: native KSM + foreign DOT + USDT.
Each test also credits an unrelated asset to confirm it is left at the OLD account.

## Scope / Behavior

Behavior of the multi-asset bounties pallet itself is unchanged. Account *derivation* shifts
to the upstream byte-for-byte form; the migration moves funds to match.


dhirajs0:
@muharem @ggwpez It is now ready for review with the migration.

system-parachains/asset-hubs/asset-hub-kusama/src/migrations.rs
```diff
@@ -29,13 +30,106 @@ pub type SingleBlockMigrations = (Unreleased, Permanent);
...
+	/// the legacy bounties pallet does not need to handle.
+	///
+	/// Kept local to the migration so the legacy `BountyRelevantAssets`
+	/// stays focused on the legacy bounties pallet's needs.
+	pub BountyMigrationAssets: alloc::vec::Vec<xcm::latest::Location> = {
+		let mut assets = crate::treasury::BountyRelevantAssets::get();
+		assets.push(crate::xcm_config::bridging::to_polkadot::DotLocation::get());
+		assets
+	};
 }
```
dhirajs0, line 48: Extended the BountyMigrationAssets with DotLocation for DOT based bounties on KAH

system-parachains/asset-hubs/asset-hub-kusama/src/migrations.rs
```diff
@@ -29,13 +30,106 @@ pub type SingleBlockMigrations = (Unreleased, Permanent);
...
+/// the derived accounts are disjoint.
+pub struct MigrateBountyAccountAssets;
+impl frame_support::traits::OnRuntimeUpgrade for MigrateBountyAccountAssets {
+	fn on_runtime_upgrade() -> frame_support::weights::Weight {
+		use frame_support::traits::Get;
+		use pallet_bounties::TransferAllAssets;
+		use sp_runtime::traits::AccountIdConversion;
+
+		let pallet_id = <crate::Runtime as pallet_treasury::Config>::PalletId::get();
+		let assets_per_bounty = BountyMigrationAssets::get().len() as u64;
```
dhirajs0, line 87: Though a bounty can have only one asset type, it's usage has minimal cost as there will be only few bounties at max by the time of runtime upgrade. So, went with it for simplicity.

ggwpez (approved):

muharem (approved):

system-parachains/asset-hubs/asset-hub-kusama/src/migrations.rs
```diff
@@ -19,6 +19,7 @@
 pub type Unreleased = (
 	RemoveAhMigratorPallet,
 	cumulus_pallet_xcmp_queue::migration::v6::MigrateV5ToV6<crate::Runtime>,
+	MigrateBountyAccountAssets,
```
acatangiu, line 22: should this be MBM? iiuc attacker could create many bounties before runtime upgrade enacts and make the upgrade be overweight - in RC case, that used to mean long/delayed block, but for AH it can mean overweight PoV and rejected block, bricked chain or at least failed upgrade

dhirajs0: The input set is governance-bounded, not attacker-bounded: `fund_bounty` requires `SpendOrigin` (TreasurySpender), and `fund_child_bounty` is capped at `MaxActiveChildBountyCount = 100` per parent (with the curator spending real balance — not free spam).

At enactment we expect ~5 parent bounties(at max, at present 1 bounty on PAH and 3 on KAH will be created before the enactment). Per-bounty work is `2r + 2w` per asset over `{KSM, USDT, DOT}`, so the upper bound is:

```
5 × (1 + 100) × 3 × (2r + 2w) ≈ 1,500 reads + 1,500 writes
```

— well below KAH/PAH per-block PoV budget ( assuming each curator creates 100 child-bounties which is unlikely ). And it's a one-shot migration, so future growth isn't the relevant question.

Happy to convert to MBM if you'd prefer belt-and-suspenders, but I think the single-block version is defensible here. WDYT?

acatangiu: this is fine if but let's also keep an eye on the `TreasurySpender` track when this goes to referenda - there is no upper limit to the number of `Small/Medium/BigSpender` refs afaik - but I agree in practice it is safe and just keeping an eye on it is enough

muharem (approved):

acatangiu (approved):
let's merge this to `main` after `oty-prepare-release-2604` is merged, rather than piling on top of [oty-prepare-release-2604](https://github.com/polkadot-fellows/runtimes/tree/oty-prepare-release-2604)

ggwpez (approved):

PR #1170, opened 2026-05-06
Add Bulletin storage pallets and business (Stage 1)

bkontur:
Relates: https://github.com/polkadot-fellows/runtimes/issues/1119

`bulletin-polkadot` has been running as a plain system parachain. This adds the chain's actual business — three pallets at indices 40/41/42:

- `pallet-bulletin-transaction-storage` — authorization-gated data storage
- `pallet-bulletin-data-renewal` — extends a stored blob's lifetime past the retention period, bounded by the account's allowance and the chain-wide `MaxPermanentStorageSize`
- `pallet-bulletin-hop-promotion` — HOP promotions

The pallets are developed in [`paritytech/polkadot-bulletin-chain`](https://github.com/paritytech/polkadot-bulletin-chain) and consumed as published crates. They already run on the Bulletin Westend and Bulletin Paseo testnets maintained there

### What this unlocks

- Authorized accounts can store data. Authorizations are granted by Root/governance, or by the People Chain over XCM.
- Stored data can be renewed past the retention period — one-shot or auto-renewed per cycle — within the account's allowance and the chain-wide `MaxPermanentStorageSize` cap.
- Runtime APIs for clients: query an account's authorization, pre-check `can_store` / `can_renew`, build HOP promotion extrinsics.
- Calls that commit data are not dispatchable over XCM.

### Testing

- This PR: runtime tests (origins, feeless calls, per-block capacity, wrapper rejection, `SafeCallFilter`, HOP promotion) and emulated XCM tests (teleports, aliases, governance from the relay and Asset Hub, storage authorization).
- Upstream: pallet unit tests in `polkadot-bulletin-chain`, whose CI also builds this runtime out-of-tree and runs its
  zombienet integration tests and try-runtime checks against it, alongside the Bulletin Westend/Paseo runtimes it hosts.
- Live-chain state: https://github.com/open-web3-stack/polkadot-ecosystem-tests/pull/682 adds Bulletin Polkadot to PET's system, upgrade and XCM suites. They pass against the WASM built from this PR via PET's runtime override, and this repo's `ecosystem-tests` CI job runs them once that PR lands.

### Reviewer decision needed

- [ ] **Keep or drop the temporary Fellowship storage grant** (`AuthorizeFellowshipMembers`): on enactment, 58 hard-coded  rank-1+ Fellowship accounts get 100 transactions / 200 MiB each for a short window to sanity-check the chain. Self-expires after 14 days and is deletable in one line, but renewals made during the window commit permanent bytes (up to ~11.6 GiB total), so the 200 MiB sizing needs confirming too.

### TODO

  - [x] Add more testing
  - [x] Release and bump final Bulletin crates
  - [x] Update to bulletin 0.11.0 crates (stable2606-1) lands after #1223 merges
  - [x] Make sure testing genesis accounts
  - - @karolk91: not sure what the item ask for but  dev/local presets aligned with the upstream Bulletin testnet presets and
  live seeds nothing
  - [x] Bump `transaction_version` — `TxExtension` gains two extensions
  - [x] ~Set initial storage versions for the new pallets (added by upgrade, so genesis does not run)~
  - - @karolk91: Seems like nothing to do, manually tested with PET and it can be seen:
   ```
      🐥 New pallet "TransactionStorage" detected … Initializing the on-chain storage version … StorageVersion(5)
      🐥 New pallet "HopPromotion" detected … no defined storage version … StorageVersion(0)
      🐥 New pallet "DataRenewal" detected … StorageVersion(1)
  ```

  - [x] Re-run `/cmd bench` for the three pallets (with the 0.11.0 crates, after #1223)
  - [x] Verify integration testing pass at latest commit of this PR at https://github.com/paritytech/polkadot-bulletin-chain/pull/746


bkontur:
/cmd bench --runtime bulletin-polkadot --pallet pallet_bulletin_transaction_storage pallet_bulletin_data_renewal pallet_bulletin_hop_promotion

integration-tests/emulated/tests/bulletin/bulletin-polkadot/src/tests/storage.rs
```diff
@@ -0,0 +1,106 @@
...
+
+	// Verify the Bulletin chain processed the message and created the authorization.
+	BulletinPolkadot::execute_with(|| {
+		type RuntimeEvent = <BulletinPolkadot as Chain>::RuntimeEvent;
+
+		assert_expected_events!(
+			BulletinPolkadot,
+			vec![
+				RuntimeEvent::TransactionStorage(
+					pallet_bulletin_transaction_storage::Event::AccountAuthorized { .. }
```
franciscoaguirre: We could assert the `who` here

system-parachains/bulletin/bulletin-polkadot/src/lib.rs
```diff
@@ -507,6 +555,14 @@ mod runtime {
...
 	pub type AuraExt = cumulus_pallet_aura_ext;
 
+	// The main business of the Bulletin chain.
+	#[runtime::pallet_index(40)]
+	pub type TransactionStorage = pallet_bulletin_transaction_storage;
+	#[runtime::pallet_index(41)]
+	pub type HopPromotion = pallet_bulletin_hop_promotion;
+	#[runtime::pallet_index(42)]
+	pub type DataRenewal = pallet_bulletin_data_renewal;
+
```
franciscoaguirre: nit: we could move after the XCM stuff to keep the numbers in the right order

system-parachains/bulletin/bulletin-polkadot/src/migrations.rs
```diff
@@ -28,3 +50,433 @@ pub type SingleBlockMigrations = (Unreleased, Permanent);
...
+	/// Taken from `FellowshipCollective::Members` on Collectives Polkadot at block `9,496,200`
+	/// (hash `0x23aee9a890b2800f95e74871617401805604ebcc5bb15aefd9552a935eed7c99`), keeping
+	/// every member whose `MemberRecord::rank` is `>= 1` — i.e. all inducted Fellows, excluding
+	/// rank-0 candidates. Each entry carries a Subsquare link so the rank can be verified by
+	/// hand; [`remote_tests`] re-checks the whole list against live Collectives state.
+	///
+	/// This is a *snapshot*: the Fellowship changes over time, and the Bulletin chain cannot
+	/// read Collectives state, so the list has to be hard-coded. Refresh it (and the block
+	/// reference above) whenever the migration is re-armed.
+	pub const FELLOWSHIP_RANK1_MEMBERS: [[u8; 32]; 58] = [
```
franciscoaguirre:
```suggestion
	pub const FELLOWSHIP_RANK1_PLUS_MEMBERS: [[u8; 32]; 58] = [
```

system-parachains/bulletin/bulletin-polkadot/src/storage.rs
```diff
@@ -0,0 +1,214 @@
...
+const CLEANUP_PRIORITY: TransactionPriority = TransactionPriority::MAX;
+// Base priority for `store`. Picked well below `TransactionPriority::MAX` so
+// `AllowanceBasedPriority` can add its boost without saturating `u64`, while still
+// leaving plenty of headroom above generic transactions.
+const STORE_PRIORITY: TransactionPriority = TransactionPriority::MAX / 4;
+const TX_LONGEVITY: TransactionLongevity = DAYS as TransactionLongevity;
+
+parameter_types! {
+	// TODO: @bkontur @franciscoaguirre @karolk91 confirm
+	pub const AuthorizationPeriod: crate::BlockNumber = 14 * DAYS;
```
franciscoaguirre, line 51: I agree with this value

franciscoaguirre (approved):

integration-tests/emulated/tests/bulletin/bulletin-polkadot/src/tests/storage.rs
karolk91: addressed in https://github.com/bkontur/runtimes/pull/33

system-parachains/bulletin/bulletin-polkadot/src/lib.rs
karolk91: addressed in https://github.com/bkontur/runtimes/pull/33

system-parachains/bulletin/bulletin-polkadot/src/storage.rs
karolk91: removed the TODO in https://github.com/bkontur/runtimes/pull/33

Cargo.toml
```diff
@@ -41,8 +41,10 @@ bp-relayers = { version = "0.28.0", default-features = false }
 bp-runtime = { version = "0.27.1", default-features = false }
 bp-xcm-bridge-hub = { version = "0.13.0", default-features = false }
 bp-xcm-bridge-hub-router = { version = "0.24.0", default-features = false }
+bulletin-pallets-common = { version = "0.10.0-draft", default-features = false }
 bulletin-polkadot-emulated-chain = { path = "integration-tests/emulated/chains/parachains/bulletin/bulletin-polkadot" }
 bulletin-polkadot-runtime = { path = "system-parachains/bulletin/bulletin-polkadot" }
+bulletin-transaction-storage-primitives = { version = "0.10.0-draft", default-features = false }
```
karolk91: note for reviewers: `-draft` versions will be soon exchanged for non draft versions before merging this PR

acatangiu (approved):

.github/workflows/runtimes-matrix.json
```diff
@@ -306,8 +306,12 @@
     "package": "bulletin-polkadot-runtime",
     "path": "system-parachains/bulletin/bulletin-polkadot",
     "max_code_size": 3145728,
+    "uris": [
+      "wss://bulletin-rpc.polkadot.io",
+      "wss://rpc-bulletin.luckyfriday.io"
+    ],
     "is_relay": false,
-    "blocktime": 6000,
+    "blocktime": 24000,
```
eskimor: Why? Slot time yes, but blocktime?

eskimor:
Claude found a few things which at least partly seem solid, can you please check:

# Bulletin Polkadot — Stage 1 review findings

Review of the PR adding the storage pallets and business logic to
`bulletin-polkadot-runtime`.

**Key context:** `bulletin-polkadot` is a **live, block-producing chain** (Stage 0
shell, spec `2002001`), not a genesis launch. This is a runtime upgrade to a
deployed chain, which drives the severity of the compatibility items below.

Each item has a checkbox. `[ ]` = open, `[x]` = done, `[~]` = rejected/invalid
(please add a one-line reason).

---

## Blocking

- [ ] **B1 — Live-state migration test runs in CI on every PR and will flake.**
  `authorize_fellowship::remote_tests::fellowship_rank1_snapshot_is_current` in
  `migrations.rs` is gated `#[cfg(all(test, feature = "try-runtime"))]`, and the
  test job runs `cargo test -p bulletin-polkadot-runtime --all-features`, which
  enables `try-runtime`. The test queries live Collectives Polkadot and asserts the
  hard-coded 58-member snapshot equals the current rank-1+ set in both directions.
  It will fail the build whenever anyone is inducted to / drops below rank 1
  (routine, and on a chain this PR does not touch), and on any RPC outage or pruning
  of the pinned block. **Fix:** gate it behind a dedicated feature not implied by
  `--all-features`, or move it to a scheduled / non-blocking job.

---

## Should fix

- [ ] **B2 — `AuthorizeFellowshipMembers::post_upgrade` can fail the
  `check-migrations` job on a member who already holds a smaller grant.**
  `on_runtime_upgrade` skips any member with an unexpired authorization
  (`migrations.rs:290`, guard `account_has_active_authorization`, which checks only
  existence + non-expiry, not size). But `post_upgrade` asserts
  `bytes_allowance >= FELLOWSHIP_BYTES` for **every** member (`migrations.rs:359`),
  and `bytes_allowance` is the granted total, not the remaining. If a listed Fellow
  already holds a live grant smaller than 100 tx / 200 MiB — e.g. one issued by the
  People chain or an `AllowedAuthorizers` account in the window between the snapshot
  and enactment — they are skipped by the upgrade and then trip the `post_upgrade`
  `ensure!`.
  Scope: `post_upgrade` is `#[cfg(feature = "try-runtime")]`, so this does **not**
  affect live enactment — it fails the try-runtime `check-migrations` verification
  (a false-positive migration failure). (Note: a prior *migration* run cannot cause
  this — the additive re-grant path only ever raises `bytes_allowance` — so the
  trigger is specifically a pre-existing grant from a different authorizer.)
  **Fix:** either have the skip-guard also require the existing allowance to meet
  the floor (top up otherwise), or assert the floor in `post_upgrade` only for
  members the migration actually granted to (thread the granted set through
  `pre_upgrade` state).

- [ ] **B3 — Bump `transaction_version`.**
  `system-parachains/bulletin/bulletin-polkadot/src/lib.rs:161` still has
  `transaction_version: 1`. This PR appends two extensions to `TxExtension`
  (lib.rs:127-128: `ValidateBulletinCalls`, `StoragePriorityBoost`), changing the
  runtime's transaction-extension set. Convention is to bump `transaction_version`
  whenever that set changes so metadata-driven signing tools resync.
  Note the practical impact here is limited: both appended extensions are unit
  structs with `Implicit = ()` that encode to **zero bytes** and are appended
  **after** the existing extensions, so the signed-payload bytes are unchanged and
  transactions signed against old metadata still verify — this is not a
  signing-breakage, and `CheckMetadataHash` is configured disabled (`new(false)`).
  It is a correctness/convention item: the extension set changed without the
  version signal that tooling and future review rely on. **Fix:** bump to `2`.

- [ ] **S1 — Capacity comment disagreeing with impl (off by 4×).**
  `tests.rs:1125` and `:1137` say "512 transactions × 8 MiB = 4 GiB per block" and
  "maximum-sized transaction (8 MiB)". `DEFAULT_MAX_TRANSACTION_SIZE` is **2 MiB**,
  so the real figure is 512 × 2 MiB = 1 GiB. The assertion under the comment
  (`assert_eq!(max_size, DEFAULT_MAX_TRANSACTION_SIZE)`) is self-referential and
  passes regardless, so the comment is the only statement of intent and it is
  wrong. **Fix:** correct the comment; or, if 8 MiB was intended, override
  `MaxTransactionSize` instead of taking the pallet default.

- [ ] **S2 — `pallet_bulletin_hop_promotion` ships with zero test coverage.**
  Live at `lib.rs:572` (pallet index 41). No unit tests in the runtime test file
  and nothing in the emulated tests touch it (only its `PromoteTxParams` constant
  is referenced). In particular `create_promotion_extrinsic` — a runtime-API
  composition that only exists in `impl_runtime_apis!` — is untested.
  **Fix:** add coverage for `promote` / `is_promoted_on_chain` /
  `create_promotion_extrinsic` (equivalents exist in the upstream
  `polkadot-bulletin-chain` `bulletin-paseo` test suite and can be ported).

- [ ] **S3 — `transaction_storage_max_throughput_per_block` no longer tests
  throughput.**
  `tests.rs:1123`. The body asserts two constants, then does a single
  `TransactionStorage::store(RuntimeOrigin::root(), …)` direct pallet call and
  checks `max_size + 1` is refused — which duplicates
  `transaction_storage_runtime_sizes` (`tests.rs:1050`). There is no per-block
  accumulation, no block-full (`ExhaustsResources`) rejection, and no
  allowance-consumption check; the test would pass with `MaxBlockTransactions = 1`.
  **Fix:** restore the multi-store-in-one-block + Nth-rejection + consumption
  assertions.

- [ ] **S4 — No negative-origin test for `add_authorizer`.**
  Every `add_authorizer` call in the test file is `assert_ok!` from Root. There is
  no test that a non-privileged signed origin is rejected with `BadOrigin`. This is
  the call that mints authorizers, so the `AuthorizerRegistrarOrigin` boundary is
  worth an explicit test. **Fix:** add a `RuntimeOrigin::signed(...)` →
  `BadOrigin` case.

---

## Decision required (not a code defect)

- [ ] **D1 — `authorize_fellowship` grants standing write capability; confirm it is
  wanted.**
  `migrations.rs`. On the next upgrade this grants 58 hard-coded rank-1+ Fellowship
  accounts 100 transactions / 200 MiB each (~11.6 GiB total) of storage allowance.
  It is the one piece of behaviour with no counterpart in the upstream reference
  runtimes, so it warrants an explicit accept/drop decision rather than riding
  along in a storage-logic PR. Safeguards verified in code: self-expiring after
  `AuthorizationPeriod` (14 days) with nothing renewing it; idempotent skip-guard;
  per-entry failures logged, not fatal; duplicate-guarded in `pre_upgrade`; the
  author flags it as a single line to delete. Rationale (a bring-up window to
  sanity-check the chain before Proof-of-Personhood becomes the only authorizer) is
  coherent. **Action:** decide keep vs drop, and record it. See also B1 and B2,
  which are defects in this same module regardless of the decision.

---

## Minor / polish

- [ ] **M1 — `dev` genesis preset seeds no authorizer while `local` seeds Eve.**
  `genesis_config_presets.rs:106` (local seeds `//Eve`) vs `:128` (dev passes an
  empty list). If dev is meant to be usable single-node, give it the same seed;
  otherwise document why it differs.

- [ ] **M2 — Harden the local-preset authorizer seed against misuse.**
  `genesis_config_presets.rs:109` bakes the well-known `//Eve` key in as an
  authorizer. It only affects a chain built from the `local` preset (the `live` and
  `dev` presets pass empty lists) and genesis presets do not apply on a runtime
  upgrade, so it cannot reach the deployed chain. Residual risk is purely
  operational: a new production chain accidentally launched from `local` would bake
  in a public key as authorizer. **Optional fix:** `#[cfg(feature = "…")]`-gate the
  `local` arm, or drop the seed and have zombienet authorize via a Root/XCM call at
  runtime.

- [ ] **M3 — Module doc narrower than file contents.**
  `integration-tests/emulated/tests/bulletin/bulletin-polkadot/src/tests/storage.rs:15`
  says "storage authorization via XCM from the People chain"; the file also covers
  BridgeHub barrier rejection, Asset Hub origin rejection, and SafeCallFilter
  blocking. Widen the doc.

- [ ] **M4 — `create_extension` hand-duplicates the `TxExtension` tuple.**
  `lib.rs:487-503` repeats the extension list by hand. A future reordering of
  `TxExtension` that is not mirrored here compiles but breaks at runtime (arity
  changes are caught; reordering of same-shaped extensions is not). Consider a
  shared constructor or a compile-time assertion tying the two together.

- [ ] **M5 — Other weakened test assertions.**
  - `allowance_based_priority_works` (`tests.rs:1183`) drops the over-budget →
    no-boost transition and the eager-compute demotion cases. The over-budget case
    is the core purpose of an allowance-based priority strategy.
  - `authorized_storage_transactions_are_for_free` (`tests.rs:1149`) tests only
    `store`; the feeless-ness of `force_renew` / `enable_auto_renew` is no longer
    covered (`enable_auto_renew` feeless is untested anywhere).
  - `store_with_cid_config_works` (`tests.rs:1002`) uses direct root calls,
    dropping the signed-extrinsic path and the authorization-extent assertions.
  - `force_renew` inside a `Utility` wrapper is no longer tested (the surviving
    wrapper-rejection matrix covers only `renew`).
  - Loss of the "after removal, ex-authorizer's `authorize_account` is rejected"
    assertion; removal is checked only via the provider refcount.

- [ ] **M6 — Undocumented divergences from the upstream reference runtime.**
  - `TxExtension` places `ValidateBulletinCalls` / `StoragePriorityBoost` *after*
    `CheckMetadataHash`; the upstream reference places them *before*. Internally
    consistent (matches `create_extension`), so it works, but it is an unexplained
    layout change (related to B1). Match upstream or add a comment.
  - `SetRetentionPeriodIfZero` is placed in `Unreleased` (runs once) rather than
    `Permanent` (runs every upgrade, as upstream). Fine for a chain past genesis,
    but worth a one-line note.

---

## Checked and considered fine (no action expected)

- [ ] Matrix `blocktime` `6000 → 24000`: consistent with the repo convention (the
  field tracks Aura *slot* duration; other 24s-slot chains use 24000). Feeds only
  try-runtime's `--blocktime`. The "blocks authored every 6s" doc comment is also
  correct (`BLOCK_PROCESSING_VELOCITY = 1` against 6s relay slots; time constants
  come from `async_backing`).
- [ ] `STORE_PRIORITY = MAX/4` + `ALLOWANCE_PRIORITY_BOOST` (1e9): no `u64`
  overflow.
- [ ] `is_renewal_committing_call` (net-new in `storage.rs`): correct — closes a
  gap the upstream filter leaves open (upstream blocks only storage-pallet mutating
  calls, but here Root reaches `force_renew` over XCM via `LocationAsSuperuser`).
  Deny-by-default allowlist; the two exemptions (`disable_auto_renew`,
  `process_pending_renewals`) are justified; wrapper-depth fail-safe matches the
  storage walk. Worth reporting upstream.
- [ ] `AuthorizeFellowshipMembers` weight accounting: matches worst case (174
  reads / 116 writes); idempotence guard holds; snapshot is 58 entries as declared.
- [ ] Weight files for the three new pallets: genuinely regenerated against the
  Polkadot runtime WASM, and structurally identical to upstream (same fn sets, same
  storage-access annotations) — not copy-pasted values.
- [ ] `Authorizer` narrowed from "any parachain in `AllowedParachainIds`" to
  `Equals<PeopleLocation>`: correct production hardening. (Consequence: adding a
  second authorizer chain later needs a runtime upgrade, not `set_storage`.)
- [ ] `SafeCallFilter = EverythingBut<StorageCallInspector>`: correctly blocks
  `store*` / `renew*` (committing calls) over XCM while leaving management calls
  (e.g. `authorize_account`) dispatchable.
- [ ] Cross-chain blast radius: this PR adds no mechanism by which a Bulletin flaw
  reaches other chains. Outbound `send` is locked
  (`SendXcmOrigin = EnsureXcmOrigin<_, ()>`); local `execute` is scoped to the
  caller's own account; the new storage pallets emit no XCM. The only cross-chain
  trust Bulletin has is the standard system-sibling set (same-account alias, DOT
  teleport), granted by the other chains' configs and bounded to
  act-as-your-own-account — unchanged here.
- [ ] `apis.rs`, `benchmarks.rs`, pallet indices, offchain impls: faithful ports of
  the upstream reference.
- [ ] HOP `promote` v1 payload not being genesis-bound: inherited from upstream and
  forced by `sp-hop 0.2.0` exposing only the v1 constructor (the pallet's
  `promote_v2` binds genesis + recipients to prevent cross-chain replay but is not
  reachable via the runtime API). Mitigated by `TransactionSource::External`
  rejection and `propagate: false`. Consider tracking an `sp-hop` bump so `v2`
  becomes reachable, since Paseo/Westend Bulletin run the same pallet. (Not a defect
  of this PR.)

karolk91:
Thanks @eskimor, I will be addressing one by one

> B1 — Live-state migration test runs in CI on every PR and will flake.

Indeed I was also aware but I think the initial intention was to remove this migration along with any tests altogether once deployed. I believe its fair to just remove this testing now - the list of fellowship members will just stay as it is up-to-date today and no live checks will be further performed

Commit that removes the live check: https://github.com/polkadot-fellows/runtimes/pull/1170/commits/152b9297b8b171d422f7f9b5f75a2beb8fe8fa79

karolk91:
> * B2 — `AuthorizeFellowshipMembers::post_upgrade` can fail the
>       `check-migrations` job on a member who already holds a smaller grant.

I don't think this is a real case - its not possible for any account to hold authorization prior to this migration as this runtime upgrade brings bulletin pallets for the first time. Additionally it should never be the case in the future that any of this account gets authorization via other mechanism

karolk91:
> * B3 — Bump `transaction_version`.

I see no harm in bumping the tx version so bumped in the commit https://github.com/polkadot-fellows/runtimes/pull/1170/commits/6a9a636833c0225784ec9950b859155a83f7ce07

karolk91:
> * **S1 — Capacity comment disagreeing with impl (off by 4×).**

Comment corrected in https://github.com/polkadot-fellows/runtimes/pull/1170/commits/4fca88b6e0610e324dc195cfb1bf377eb813461a

karolk91:
> * S2 — `pallet_bulletin_hop_promotion` ships with zero test coverage.

Indeed, while we have tests for the upstream runtimes, these were missed here. Added in https://github.com/polkadot-fellows/runtimes/pull/1170/commits/9c524e022844891421199e31305df6e19d392338

karolk91:
> * S3 — `transaction_storage_max_throughput_per_block` no longer tests
>       throughput.

Improved testing in https://github.com/polkadot-fellows/runtimes/pull/1170/commits/1922105fdbbf72f852b1f3996b9e216d9ff94ccd

karolk91:
> * S4 — No negative-origin test for `add_authorizer`.

Tests added in https://github.com/polkadot-fellows/runtimes/pull/1170/commits/f750ee6c7d6534c7f6d4593d9125111e14d7e564

karolk91:
> * M1 — `dev` genesis preset seeds no authorizer while `local` seeds Eve.

`dev` and `local` presets were aligned in: 4ef89c960

> * M2 — Harden the local-preset authorizer seed against misuse.

I assume nothing to do here. If production chain is launched from local/dev presets then I think there are more serious problems than the authorizer (eg Alice account with 1M+ DOT balance)

> * M3 — Module doc narrower than file contents.

Improved docs in 6f9e9a15b

> * M4 — `create_extension` hand-duplicates the `TxExtension` tuple.

Looks like the problem doesn't apply here because (same-type reorder) has zero instances in this tuple (There is no same-typed pair to swap)

> * M5 — Other weakened test assertions.

Tests improved in 18ce23e6c

> * M6 — Undocumented divergences from the upstream reference runtime.

Now changed to match upstream reference for both cases in 1bb27a8b6 and 7cf75eddf

karolk91:
@eskimor, I believe everything is addressed now from the list - the only thing left that is a decision:
> D1 — authorize_fellowship grants standing write capability; confirm it is
wanted.

is now part of this PR description for everyone to see it clearly

@acatangiu , @franciscoaguirre - I pushed few commits since your approvals, so you may want to review again

mudigal (commented):
LGTM - with the completion of TODOs of 0.11.0 bump + fresh benchmarks.

On the decision of "Keep or drop the temporary Fellowship storage grant" - I lean towards dropping them to keep it clean, as I do not see much of the value add for time being. Individuality runtime is also getting updating and if there is no sufficient gap between these 2 there is not much testing that can happen.

mudigal (approved):

system-parachains/bulletin/bulletin-polkadot/src/genesis_config_presets.rs
```diff
@@ -87,9 +91,33 @@ fn bulletin_polkadot_genesis(
...
+		},
 	})
 }
 
+/// Authorizer seeded into the `dev` and `local_testnet` presets: `//Eve`, the account the
+/// bulletin-chain integration tests authorize with. The runtime has no sudo and reaches Root
+/// only over XCM from the relay chain or the Asset Hub, so a chain built from these presets
+/// would otherwise start with no way to authorize storage at all. The `live` preset seeds
+/// nothing, and presets never apply on a runtime upgrade, so this cannot reach the deployed
+/// chain.
```
ggwpez: slopus

system-parachains/bulletin/bulletin-polkadot/src/lib.rs
```diff
@@ -153,7 +158,7 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	spec_version: 2_003_002,
 	impl_version: 1,
 	apis: RUNTIME_API_VERSIONS,
-	transaction_version: 1,
+	transaction_version: 2,
```
ggwpez, line 161: Why bumped?

system-parachains/bulletin/bulletin-polkadot/src/migrations.rs
```diff
@@ -18,13 +18,384 @@
...
+	//
+	// Once Individuality lands (https://github.com/polkadot-fellows/runtimes/pull/1233), Proof of
+	// Personhood on the People Chain is the only authorizer, so there is no way to store data here
+	// until that logic is switched on live. This grant opens a short window to sanity-check the
+	// chain itself first — Bitswap retrieval, collator p2p setup, and the like — before PoP/DIM
+	// go live.
+	//
+	// NOTE: this is entirely optional. If we decide we do not want this testing window, drop this
+	// one line before release — nothing else in the runtime depends on it, and the rest of
+	// `authorize_fellowship` (snapshot, tests) goes with it.
```
ggwpez, line 38: Do we not have the PET tests? And what if we do this but it does not work? It is just a diagnostic, not a fix, right?

system-parachains/bulletin/bulletin-polkadot/src/storage.rs
```diff
@@ -0,0 +1,213 @@
...
+	DEFAULT_MAX_BLOCK_TRANSACTIONS, DEFAULT_MAX_TRANSACTION_SIZE, MAX_WRAPPER_DEPTH,
+};
+use pallet_xcm::EnsureXcm;
+use sp_runtime::transaction_validity::{TransactionLongevity, TransactionPriority};
+
+parameter_types! {
+	/// Cap on the total bytes committed to permanent storage (via `renew`) across all
+	/// authorizations on this chain. Seeded at 1.7 TiB; storage-backed so governance
+	/// (root) can raise/lower it via `system.set_storage` without a runtime upgrade.
+	pub storage MaxPermanentStorageSize: u64 = 17 * 1024 * 1024 * 1024 * 1024 / 10;
```
ggwpez, line 38: No please use pallet parameters it it should be adjustable. No set_storage.

system-parachains/bulletin/bulletin-polkadot/src/storage.rs
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// Permissionless cleanup sits at the top so it always runs before stores compete for
+// blockspace.
+const CLEANUP_PRIORITY: TransactionPriority = TransactionPriority::MAX;
```
ggwpez, line 43: Should this cleanup TX not just have weight class Operational? Then it is first as well.

system-parachains/bulletin/bulletin-polkadot/src/storage.rs
```diff
@@ -0,0 +1,213 @@
...
+// Base priority for `store`. Picked well below `TransactionPriority::MAX` so
+// `AllowanceBasedPriority` can add its boost without saturating `u64`, while still
+// leaving plenty of headroom above generic transactions.
+const STORE_PRIORITY: TransactionPriority = TransactionPriority::MAX / 4;
```
ggwpez, line 47: This value sounds quite random. What priority do normal TX have?

system-parachains/bulletin/bulletin-polkadot/src/storage.rs
```diff
@@ -0,0 +1,213 @@
...
+	/// Maximum allowable skew between the user's submit timestamp and the on-chain
+	/// time when validating a HOP promotion: 48 hours, in milliseconds.
+	pub const SubmitTimestampTolerance: u64 = 48 * 60 * 60 * 1000;
+	// Lowest priority: promotion only fills blockspace stores would not have used.
+	pub const PromoteTxParams: ValidTransactionParams =
+		ValidTransactionParams::new("HopPromotion", 0, 5);
+}
+
+impl pallet_bulletin_hop_promotion::Config for Runtime {
+	type SubmitTimestampTolerance = SubmitTimestampTolerance;
```
ggwpez, line 210: Note for the future: names like this should always either include the unit or use a wrapper type. It is not clear to me if its MS, S or something else. And if it ever changes then this will silently fail.

system-parachains/bulletin/bulletin-polkadot/src/storage.rs
```diff
@@ -0,0 +1,213 @@
...
+				// Releases a registration rather than committing one; Root needs it for cleanup.
+				RenewalCall::disable_auto_renew { .. } |
+					// Mandatory inherent — `ensure_none` rejects any `Transact` origin anyway.
+					RenewalCall::process_pending_renewals { .. }
+			);
+		}
+		<Self as CallInspector<Runtime>>::inspect_wrapper(call).is_some_and(|inner_calls| {
+			// Same fail-safe as the storage-call walk: a wrapper too deep to inspect counts as
+			// committing.
+			depth >= MAX_WRAPPER_DEPTH ||
```
ggwpez, line 127: This returns false when it is too deep? so the XCM dispatch filter lets it pass???

system-parachains/bulletin/bulletin-polkadot/src/lib.rs
karolk91: related to
> B3 — Bump transaction_version

from the comment https://github.com/polkadot-fellows/runtimes/pull/1170#issuecomment-5342525266

tldr: new tx extensions added and `CheckMetadataHash` changed its index in the tuple: https://github.com/polkadot-fellows/runtimes/pull/1170/files#diff-db6292f32d0eb9f1f8b239199a0c4c2cb9f116326af6bb9412be63456a2a5fa5R128

system-parachains/bulletin/bulletin-polkadot/src/migrations.rs
karolk91: I only have a local PET test to test this migration and confirm some example account can be used - I never intented to push it to upstream as its just one-time action for the initial launch and didn't make sense to me to include this as part of PET suite that run continuously every day

system-parachains/bulletin/bulletin-polkadot/src/storage.rs
karolk91: As far as I'm checking the Operational would not raise the priority as its based on fees, and cleanup calls are feeless (no fee == no additional priority). As DOS protection I think it makes sense that these also stay `Normal` instead of `Operational`

system-parachains/bulletin/bulletin-polkadot/src/storage.rs
karolk91: Its not super strict explanation and I'm no expert in this area but "normal tx" priorities seem based on formula that takes into account the tip and weight - if my research is correct if you would like `transfer_keep_alive` tx to have equal priority to `TransactionPriority::MAX / 4` you would need to pay 150k DOT of fee

Bulletin is mostly based on feeless transactions and `Normal` class, and being a storage solution, these are higher priority than some regular non-storage pallets related transactions. This value may seem random but it make sense to easily compare bulletin calls to each other - eg `MAX / 2` will be higher priority than `MAX / 4` and the exact number doesn't matter. It can be also compared to `apply_authorized_upgrade` that uses just `MAX` priority and `Operational` class

system-parachains/bulletin/bulletin-polkadot/src/storage.rs
karolk91: Thanks for the advice

system-parachains/bulletin/bulletin-polkadot/src/storage.rs
karolk91: `depth >= MAX_WRAPPER_DEPTH` returns `true` when wrapper is too deep - so it blocks the call

system-parachains/bulletin/bulletin-polkadot/src/genesis_config_presets.rs
karolk91: tried to unslope here https://github.com/polkadot-fellows/runtimes/pull/1170/commits/42f7cffcd7b9d8eda4400bca801db9be902c82f5

system-parachains/bulletin/bulletin-polkadot/src/storage.rs
karolk91: We have a PR for that: https://github.com/paritytech/polkadot-bulletin-chain/pull/576 but unfortunately is not yet merged/tested etc. I'm thinking if this could be brought by future runtime upgrade

karolk91 (approved):

acatangiu (approved):

system-parachains/bulletin/bulletin-polkadot/src/storage.rs
ggwpez: Ah yea i read the `EverythingBut<StorageCallInspector>` in the opposite way.

ggwpez (approved):

franciscoaguirre (approved):

PR #1159, opened 2026-05-01
Integrate SDK 2604

ggwpez:
Integrates SDK release 2604.

## Contracts

- [x] @0xRVE — [#11460](https://github.com/paritytech/polkadot-sdk/pull/11460) `CallbackHandle` check
- [x] @0xRVE — [#11590](https://github.com/paritytech/polkadot-sdk/pull/11590) Asset-conversion precompile · [4759eb9e4](https://github.com/polkadot-fellows/runtimes/commit/4759eb9e402742fd3a4d3b303e7a755331a99e2c)
- [x] @pgherveou — [#11416](https://github.com/paritytech/polkadot-sdk/pull/11416) Auto-mapping (`AutoMap`) · [6dd3ddc6b](https://github.com/polkadot-fellows/runtimes/commit/6dd3ddc6be113f32585e6b11dee4be98dd806bdf) https://github.com/polkadot-fellows/runtimes/pull/1165
- [x] @mokita-j — [#11507](https://github.com/paritytech/polkadot-sdk/pull/11507) Revive debug enable

## Bridges

- [x] @acatangiu — [#11694](https://github.com/paritytech/polkadot-sdk/pull/11694) `pallet_xcm_bridge_hub_router` `UnpaidExport` flag · [635ac29a8](https://github.com/polkadot-fellows/runtimes/commit/635ac29a86572729df8832d2526b6f15ddf5c2b8)
- [x] @yrong — [#11354](https://github.com/paritytech/polkadot-sdk/pull/11354) Snowbridge inbound nonce API · [c6fa9f236](https://github.com/polkadot-fellows/runtimes/commit/c6fa9f23692995a9e9640466dfe2320080ced35c)

## Async backing / collator

- [x] @sandreim — [#11778](https://github.com/paritytech/polkadot-sdk/pull/11778) Relay-parent offset

## Account recovery

- [x] @ggwpez — [#10482](https://github.com/paritytech/polkadot-sdk/pull/10482) pallet-recovery modernization (HoldConsideration follow-up + re-bench) · [7401ea900](https://github.com/polkadot-fellows/runtimes/commit/7401ea90008b841f98ab037c8ba1675185efbe4b)

## Penpal / integration tests

[...]

PR #1151, opened 2026-04-20
[DRAFT / DO NOT MERGE UNTIL WE BUMP TO SDK2604] Polkadot: configure DAP satellite, route XCM fees and slashes to DAP/DAP satellite and drain legacy old treasury local chain accounts

sigurpol:
[ still DRAFT]

:information_source:   **UPDATE: we have decided to review and eventually merge  only after the DAP satellite pallet is replaced by a generic `accumulate-and-forward` pallet (PR in the SDK  ~yet to come~ https://github.com/paritytech/polkadot-sdk/pull/11881 - currently in review). This might or not be ready by SDK 2604 cut-off (foreseen on April 30th)**

Polkadot only:
- Configured DAP satellite on Polkadot RC and all system chains except AssetHub
- XCM delivery fees and slashed routed to DAP / DAP satellite
- Sealed every residual inflow to the legacy Polkadot RC `py/trsry` account and drain historical balances on next runtime .

## TODO

- Changelog
- Migration of funds from  AH legacy RC treasury account to DAP (see https://github.com/paritytech/polkadot-sdk/issues/11705#issuecomment-4222920754)
- Once we include crates coming from SDK 2604, look for `SDK2604` and all related TODO and address them as explained in the comment: e.g update dap / dap satellite runtime config to mimic Westend wiring, replace shim with real dap / dap_satellite methods etc. **UPDATE**: actually, for dap-satellite, we will need to use instead the new accumulate and forward pallet. Note that `dap/satl` has been already prefunded with >=ED on RC and all system chains !=AH (e.g. see https://polkadot.subscan.io/account/13UVJyLkAxdQnd3UUQvJkLycM8NSybBUTECBRecdyiJ9Bhqp) since it was done before taking the decision to introduce the generic pallet. 

## NOTES

- We can't merge until we bump crates to `SDK 2604` - once we do so, we can remove the temporary DAP and DAP satellite shims and just use DAP / DAP satellite methods directly (see TODO in the code) or the new generic pallet replacing DAP satellite. This was done to already implement migration from legacy Polkadot RC `py/trsy` to DAP satellite (https://github.com/paritytech/polkadot-sdk/pull/11763) and to have support for NegativeImbalance (https://github.com/paritytech/polkadot-sdk/pull/11716)
- **We can't merge until  PRs introducing DAP satellite are properly audited** 
- This PR is a prereq for pallet_treasury  + OpenGov pallets removal from the Polkadot relay chain post-AHM (#1102)

## Related issues

- #1137 
[...]

Issue #1144, opened 2026-04-10
Remove RC and local treasury accounts


sigurpol:
Similar to SDK https://github.com/paritytech/polkadot-sdk/issues/11705, and related to #1137, we should get rid of all non AH local treasury accounts on **Polkadot** chains != AH, and do a migration to transfer funds on DAP satellites or AH treasury account on Polkadot chains (probably the former? To be discussed!).

## Polkadot 

We can actually get rid and drain local / RC treasury accounts after the following has happened:

  - Transaction fees → DAP satellite (DAP Satellite (80%) + block author (20%) on Polkadot, #1137 )                 
  - XCM delivery fees → DAP Satellite (#1137 )                             
  - Slashing → DAP Satellite (#1137)                                                   
  - Dust removal → DAP Satellite (#1137) 
  - collectives (fellowship, ambassador, alliance) and people identity slash to DAP satellite : this requires the SDK issue https://github.com/paritytech/polkadot-sdk/issues/11704 to be fixed and a SDK bump (related SDK PR: https://github.com/paritytech/polkadot-sdk/pull/11716)     
  - Treasury extrinsics  on RC→ blocked by PostAhmFilter (already done post AHM)                              
  - Bounties on RC → blocked by PostAhmFilter   (already done post AHM)                                         

Similar to SDK 11705, in all system chains different from AH we should look for references to `RelayTreasuryLocation` , `pub TreasuryAccount: AccountId = TREASURY_PALLET_ID.into_account_truncating();` and similar and be sure to remove what is not meant to be used anymore and to ensure the migration of funds on the accounts we are removing.

Similar to [SDK#11763](https://github.com/paritytech/polkadot-sdk/pull/11763), we should aim to remove the treasury pallet on RC completely.
Look at  issue #1102 for the whole OpenGov removal from RC (similar to Westend's SDK#11775](https://github.com/paritytech/polkadot-sdk/issues/11775)).

**NOTE1**:  currently post AHM - RC treasury account has still a balance !=0 and growing as you can see in https://polkadot.subscan.io/account/13UVJyLnbVp9RBZYFwFGyDvVd1y27Tt8tkntv6Q7JVPhFsTB?tab=xcm_transfer&chart_tab=time - which is unexpected and needs to be understood before removing the related account - but we still have e.g. XCM and tx fees redirected to local RC treasury until #1137 is fixed.

## Kusama 

[...]

PR #1173, opened 2026-05-07
[WIP] Redirect burns and fees to DAP

Ank4n:
UPDATE (18 May): Not to be released with 2.3.0.

Rework of #1151 by @sigurpol, rebased on #1159 (sdk2604). 

## TODO 

- [x] Polkadot RC
- [x] Polkadot AH
- [x] Polkadot Bridge Hub
- [x] Polkadot Collectives
- [x] Polkadot Coretime
- [x] Polkadot People
- [x] integration-tests
- [ ] bench 
  - [ ] accumlate-and-forward pallet
  - [ ] pallet-dap
  - [ ] pallet-broker


relay/polkadot/src/lib.rs
```diff
@@ -1594,13 +1611,27 @@ impl pallet_delegated_staking::Config for Runtime {
...
 	type Currency = Balances;
-	// slashes are sent to the treasury.
-	type OnSlash = ResolveTo<TreasuryAccountId<Self>, Balances>;
+	type OnSlash = AccumulateForward;
 	type SlashRewardFraction = SlashRewardFraction;
 	type RuntimeHoldReason = RuntimeHoldReason;
 	type CoreStaking = Staking;
 }
 
+impl pallet_accumulate_and_forward::Config for Runtime {
```
Ank4n, line 1622: Not super sure if we should add this in RC given we want to remove balances altogether at some point.

relay/polkadot/constants/src/lib.rs
```diff
@@ -205,6 +205,38 @@ pub mod system_parachain {
...
+		.into();
+	}
+}
+
+/// Accumulate-and-forward constants.
+pub mod accumulate_forward {
+	use frame_support::{parameter_types, PalletId};
+	use polkadot_primitives::{Balance, BlockNumber};
+
+	parameter_types! {
```
Ank4n: should this be in constants?

relay/polkadot/src/lib.rs
```diff
@@ -489,11 +488,21 @@ parameter_types! {
 	/// This value increases the priority of `Operational` transactions by adding
 	/// a "virtual tip" that's equal to the `OperationalFeeMultiplier * final_fee`.
 	pub const OperationalFeeMultiplier: u8 = 5;
+	/// Percentage of fees that go to the accumulation account.
+	/// The remainder goes to block author. Tips always go 100% to author.
+	pub const AccumulateForwardFeePercent: Percent = Percent::from_percent(100);
```
Ank4n: should be 80

relay/polkadot/src/lib.rs
```diff
@@ -1594,13 +1605,36 @@ impl pallet_delegated_staking::Config for Runtime {
...
 	type SlashRewardFraction = SlashRewardFraction;
 	type RuntimeHoldReason = RuntimeHoldReason;
 	type CoreStaking = Staking;
 }
 
+parameter_types! {
+	/// The pallet ID used to derive the accumulation account.
+	pub const AccumulateForwardPalletId: PalletId = PalletId(*b"acf/dott");
+	/// How often the accumulation account forwards to the destination.
+	pub const ForwardPeriod: BlockNumber = HOURS;
```
Ank4n, line 1617: what should be this value?

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
```diff
@@ -160,7 +161,13 @@ pub mod migrations {
 	use super::*;
 
 	/// Unreleased migrations. Add new ones here:
-	pub type Unreleased = (cumulus_pallet_xcmp_queue::migration::v6::MigrateV5ToV6<Runtime>,);
+	pub type Unreleased = (
+		cumulus_pallet_xcmp_queue::migration::v6::MigrateV5ToV6<Runtime>,
+		// Drain residual legacy `py/trsry` balance into the accumulation account. Idempotent.
+		pallet_accumulate_and_forward::migrations::DrainLegacyTreasuryToAccumulationAccount<
```
Ank4n, line 167: probably should have passed Treasury account/pallet-id here.

system-parachains/bridge-hubs/bridge-hub-polkadot/src/xcm_config.rs
```diff
@@ -199,7 +191,6 @@ pub type Barrier = TrailingSetTopicAsId<
 pub type WaivedLocations = (
 	Equals<RootLocation>,
 	RelayOrOtherSystemParachains<AllSiblingSystemParachains, Runtime>,
-	Equals<RelayTreasuryLocation>,
```
Ank4n, line 202: should replace with accumulation account

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
```diff
@@ -660,6 +669,30 @@ impl cumulus_pallet_weight_reclaim::Config for Runtime {
 	type WeightInfo = weights::cumulus_pallet_weight_reclaim::WeightInfo<Runtime>;
 }
 
+parameter_types! {
+	/// The pallet ID used to derive the accumulation account.
+	pub const AccumulateForwardPalletId: PalletId = PalletId(*b"acf/dott");
+	/// How often the accumulation account forwards to the destination, in relay blocks.
+	pub const ForwardPeriod: BlockNumber = polkadot_runtime_constants::time::HOURS;
```
Ank4n, line 676: what should be this value?

Ank4n:
@sigurpol @andreitrand
need your reviews. Touches quite a few files.

relay/polkadot/src/lib.rs
sigurpol: to be double-checked with @andreitrand  (can't ping him directly here) , no strong opinion, can be also daily or so

sigurpol:
Generic question: do we have a task to to prefund "acf/dott" (`13UVJyLjaBdLEGEpgE7dexnp5fiAG5Zu6D1SWSt6j2Q49w52`) with ED on RC + all system chains +  dap staging account (`13UVJyLkAxdQn6zM3Gz49SmCLi8SZW3bdtm7DTY29ScavqW2`) too on AH @Ank4n , @andreitrand ?  Please double check the addresses too :) 

note: dap/buff was already prefunded (13UVJyLkAxdQn6zM3GrLGvrFgN9jWneRX2NGvf3u92iunZq1) on AH, and so were the deprecated dap/satl (not an issue, just some dead accounts with ED before we decided to move to the acf pallet, we can live with that).

relay/polkadot/src/lib.rs
sigurpol: I think it's ok to keep it for now, maybe with a TODO pointing to #1144 .
We have tackled in the SDK (https://github.com/paritytech/polkadot-sdk/issues/11705) and this can be a followup task post 2.3.0 imho

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
sigurpol: I agree,  this seems a bug in the SDK here:

https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/accumulate-and-forward/src/migrations.rs#L36-L36

This works only on RC and matches pre-AHM RC local treasury but matches nothing on the other system chains (IIUC) so we would need a fix on upstream SDK to fix it properly. 
Before doing that, we should check the current amount on each `RelayTreasuryPalletAccount`

```rust
  pub RelayTreasuryLocation: Location = (Parent, PalletInstance(TREASURY_PALLET_ID)).into();
  pub RelayTreasuryPalletAccount: AccountId =
      LocationToAccountId::convert_location(&RelayTreasuryLocation::get())
          .unwrap_or(TreasuryAccount::get());
```

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
sigurpol: same as before - no strong opinion, let's 👂 from @andreitrand too

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
sigurpol: actually doublechecking on subscan vs `14xmwinmCEz6oRrFdczHKqHgWNMiCysE2KrA4jXXAAM1Eogk` (assuming it's the correct RelayTreasuryPalletAccount 😄 ), it's zero on all system chains so there is nothing to migrate and we probably don't need to rush any fix in the SDK @Ank4n , @andreitrand. 
(To be noted : on AH see https://assethub-polkadot.subscan.io/account/14xmwinmCEz6oRrFdczHKqHgWNMiCysE2KrA4jXXAAM1Eogk, it has 0 DOT but  still USDT and USDC)

relay/polkadot/src/lib.rs
andreitrand: We can leave it to 1 hour, this number was agreed upon with Adrian as well.

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
andreitrand: The current 1 hour is a good compromise. It can remain as-is.

relay/polkadot/src/lib.rs
```diff
@@ -489,11 +488,21 @@ parameter_types! {
 	/// This value increases the priority of `Operational` transactions by adding
 	/// a "virtual tip" that's equal to the `OperationalFeeMultiplier * final_fee`.
 	pub const OperationalFeeMultiplier: u8 = 5;
+	/// 80% of tx fees go to the accumulation account, 20% to block author. Tips always go 100% to
+	/// author.
+	pub const AccumulateForwardFeePercent: Percent = Percent::from_percent(80);
```
Ank4n, line 493: Double check the old split

system-parachains/bridge-hubs/bridge-hub-polkadot/src/bridge_to_ethereum_config.rs
```diff
@@ -278,7 +278,7 @@ impl snowbridge_pallet_system::Config for Runtime {
 	type OutboundQueue = EthereumOutboundQueue;
 	type SiblingOrigin = EnsureXcm<AllowSiblingsOnly>;
 	type AgentIdOf = snowbridge_core::AgentIdOf;
-	type TreasuryAccount = RelayTreasuryPalletAccount;
+	type TreasuryAccount = AccumulateAccount;
```
Ank4n, line 281: Recheck this

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
Ank4n: so lets just remove this migration?

sigurpol: I would take two actions:
1. remove the migration for every chain != RC (needed for the release)
2. create an issue for the SDK to make account-forward pallet more generic. Actually, I am not a fan of accumulate-and-forward pallet having embedded DrainLegacyTreasuryToAccumulationAccount at all in it since it defeats its scope to be a generic pallet, I think we shouldn't have put it in the pallet in the 1st place. Post RU, maybe we should just nuke the code. Wdyt @Ank4n , @andreitrand , @kianenigma ?

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
```diff
@@ -312,7 +319,7 @@ parameter_types! {
 impl pallet_balances::Config for Runtime {
 	/// The type for recording an account's balance.
 	type Balance = Balance;
-	type DustRemoval = ();
+	type DustRemoval = AccumulateForward;
```
Ank4n, line 322: Ensure this does not keep looping

Ank4n: We can keep this PR minimal where we only redirect high value burns/slashes to accumulation

system-parachains/bridge-hubs/bridge-hub-polkadot/src/lib.rs
sigurpol: for the upstream SDK change:  tentative PR -> https://github.com/paritytech/polkadot-sdk/pull/12092

PR #1174, opened 2026-05-07
Use 24s AURA slots for Kusama/Polkadot AH and PC

sandreim:
Migration from 6 to 24s has been tested on Westend AH and also live on Polkadot Bulletin Chain.

Longer Aura slots improve elastic scaling throughput and censorship resistance, while keeping the parachain block time unchanged. During it's own slot a collator can produce multiple blocks, depending on the number of cores assigned. For a single core, 24s AURA slot means, that 6 consecutive blocks will be produced by the same author.


bkchr (approved):
Looks like try-runtime doesn't like this change. We need to fix it before merging this.

sandreim:
> Looks like try-runtime doesn't like this change. We need to fix it before merging this.

fixed, should pass now

ggwpez (approved):

acatangiu:
can you add details about what this change does in practice and WHY we're doing it?

sandreim:
> can you add details about what this change does in practice and WHY we're doing it?

Added in change log and description.

CHANGELOG.md
```diff
@@ -10,6 +10,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Asset Hub Polkadot & Kusama: add `PrevalidateAttests` transaction extension to the `TxExtension` ([#1156](https://github.com/polkadot-fellows/runtimes/pull/1156))
 - Zombienet SDK integration tests for elastic scaling on Asset Hub Polkadot and People Polkadot.
+
+### Changed
+
+- Asset Hub Polkadot & Kusama, People Polkadot & Kusama: switch to 24s Aura slot duration. Longer Aura slots improve elastic scaling throughput and censorship resistance, while keeping the parachain block time unchanged.
```
bkchr: Missing link to this pr.

acatangiu (approved):

sandreim:
coretime integration tests should be fixed by xcm-emulator 0.28 which comes with SDK 2604. Will pass CI after https://github.com/polkadot-fellows/runtimes/pull/1159 is merged

ggwpez:
Screenshot 2026-06-02 at 18 22 39

The KAH migration job seems broken, there is no way that since yesterday, a compressed KAH snapshot lost 111 MB.  
So will merge anyway and investigate.

PR #1177, opened 2026-05-11
release: verify runtime WASM size against thresholds

### LINKED ISSUE #940: improve runtime release automation to verify wasm size is not too big
as part of the release automation, it should verify wasm size is not greater than the hard limit
should probably make some warnings when it is greater than some warning limit so we are less likely need to make last minute investigation

Nathy-bajo:
Adds a release-workflow gate that warns at 80% and blocks publication at 95% of each runtime's configured compressed-WASM size limit, opening a tracking issue summarizing any breaches; closes #940 


Nathy-bajo:
@bkchr @xlc Ready for review

xlc:
Looks fine. Did you test it? Maybe setup a fork, update the params so that a normal build triggeres warning / error and see if it creates the issue correctly.

Nathy-bajo:
@xlc Tested end-to-end on my fork with a temporary workflow that exercises the gate scripts under real Actions and asserts the expected outputs (evidence preserved via the run and issue URLs below).

The [ok](https://github.com/Nathy-bajo/runtimes/actions/runs/25742312326) run produced no violation and skipped issue creation; the [warning](https://github.com/Nathy-bajo/runtimes/actions/runs/25742316720), [error](https://github.com/Nathy-bajo/runtimes/actions/runs/25742321204), and [missing](https://github.com/Nathy-bajo/runtimes/actions/runs/25742326275) runs each crossed the threshold and opened the corresponding issue ([#1](https://github.com/Nathy-bajo/runtimes/issues/1), [#2](https://github.com/Nathy-bajo/runtimes/issues/2), [#3](https://github.com/Nathy-bajo/runtimes/issues/3)) with the rendered markdown table, threshold explanation, Release label, and a link back to the workflow run. Every run's Assert expected outcome step is green, confirming the gate produced the right verdict in each scenario.

xlc:
I think we should fail the release CI on the error case to stop generate the new release. Otherwise all looking good.

Nathy-bajo:
@xlc  Good catch. The failure path was already wired via the default `needs` behavior, but the gating wasn't visible in the YAML.  Pushed an update that makes it explicit by adding `if: needs.check-runtime-sizes.result == 'success'` to `publish-release` and tightens the failing step's message to say it's blocking the release.

xlc (approved):

Nathy-bajo:
@bkchr All done, your approval needed for second blocker

xlc (approved):

Nathy-bajo:
@ggwpez Second approval needed to merge this

ggwpez:
Dont we already have this check from Joseph that runs it for all runtimes? I dont remember, need to check

Nathy-bajo:
> Dont we already have this check from Joseph that runs it for all runtimes? I dont remember, need to check

There's no existing size check on `main`, only an informational size line in the release notes; the enforcing check (warn 80% / error 95%, blocks `publish-release`, covers all 14 runtimes) is new in this PR.

sigurpol:
> Dont we already have this check from Joseph that runs it for all runtimes? I dont remember, need to check

@ggwpez , we have a workflow running https://github.com/polkadot-api/check-runtime - which AFAIK doesn't cover this use case - are you suggesting to extend that to cover also for it instead of adding an ad-hoc workflow?

.github/scripts/create-size-issue.sh
ggwpez, line 1: I think creating an issue is overkill. I personally also think that the scripts here could be much shorter, just iterating over the json result and directly failing when a WASM is too big.

We dont expect this to routinely happen and a hard CI stop is the easiest way to debug. Less script code is also easier to maintain.

ggwpez:
> we have a workflow running https://github.com/polkadot-api/check-runtime - which AFAIK doesn't cover this use case - are you suggesting to extend that to cover also for it instead of adding an ad-hoc workflow?

Ah okay I thought that already checked size, but does not. Then its fine to add it here I guess.

.github/scripts/create-size-issue.sh
Nathy-bajo: Done. dropped the issue creation and the separate aggregate job; the per-runtime build step now just reads the srtool JSON and hard-fails CI at 95% of the configured limit (warning at 80%), which blocks `publish-release` directly with far less script to maintain.

ggwpez (approved):
Thanks!

bkchr (approved):

PR #1180, opened 2026-05-19
Shard PET ecosystem tests; switch to vitest forks pool

rockbmb:
Brings the PET ecosystem-tests job in this repo's CI into line with the perf and reliability work landed in [open-web3-stack/polkadot-ecosystem-tests](https://github.com/open-web3-stack/polkadot-ecosystem-tests) over the last few weeks.

## Test execution

- Shard the matrix 3 ways per network via Vitest's `--shard`, so the two ecosystem-test jobs (polkadot, kusama) become six running in parallel on separate self-hosted runners.
- Switch from `--pool=threads --maxWorkers=3` to `--pool=forks --maxWorkers=8`. The shared-client refactor ([open-web3-stack/polkadot-ecosystem-tests#609](https://github.com/open-web3-stack/polkadot-ecosystem-tests/pull/609)) showed retry-induced state leaks under threads; forks isolate each test file in its own child process and tolerate higher `maxWorkers` without contention.
- `--retry=3` → `--retry=2`; endpoint flakiness was cut by the ranking work in [open-web3-stack/polkadot-ecosystem-tests#616](https://github.com/open-web3-stack/polkadot-ecosystem-tests/pull/616).
- `timeout-minutes` 120 → 60 to match the new envelope.

## Subway

- Install Subway by extracting the binary from `acala/subway:v0.1.1` instead of rebuilding from source (`cargo install --git`, ~4 min per job). `docker cp` into a user-owned tmp path then `sudo install` works around root not having access to the docker socket on some IBP runners.
- Fetch both `pet-chain-endpoints.json` and `subway-template.yml` from PET's `master` at run time, in place of `runtimes-matrix.json.uris` and a local template snapshot. PET's endpoint list is actively probed and ranked by an endpoint-health workflow; PET's template honors that ranking (`shuffle_endpoints: false`) and sets `request_timeout_seconds: 90` for heavy upstream queries that exceed Subway's default 30s per-upstream timeout. Keeping local copies of either was a recipe for drift.


.github/workflows/test.yml
```diff
@@ -493,36 +494,52 @@ jobs:
...
+          # `docker cp` writes to a user-owned tmp path because the daemon
+          # socket isn't accessible to root on these self-hosted runners
+          # (so `sudo docker cp` straight to /usr/local/bin/ fails). Then
+          # `install` moves the binary into place with the executable bit
+          # set, under sudo for the privileged write only.
+          docker cp subway-extract:/usr/local/bin/subway /tmp/subway
+          docker rm subway-extract
+          sudo install -m 0755 /tmp/subway /usr/local/bin/subway
+          rm /tmp/subway
+          subway --version
```
ggwpez, line 554: Why are you extracting the subway binary from a docker image?

rockbmb: Workaround.

`acala/subway` v0.1.1's GitHub Release has no `.tar.gz` asset attached because `cargo build --release --locked` failed against a stale `Cargo.lock` ([run](https://github.com/AcalaNetwork/subway/actions/runs/26141877372)); the Docker image at the same tag built fine and is currently the only consumption path for that version.

v0.1.1 is the first tag containing [AcalaNetwork/subway#203](https://github.com/AcalaNetwork/subway/pull/203), which exposes `request_timeout_seconds`. Subway's default 30s per-upstream timeout is too short for the heavy storage queries this CI runs.

[AcalaNetwork/subway#204](https://github.com/AcalaNetwork/subway/pull/204) asks for a re-cut; once the artifact is up, this can revert to `curl | tar -xz`.

rockbmb: Fixed, so Docker is gone.

ggwpez (approved):

xlc (approved):

ggwpez (approved):

PR #1186, opened 2026-05-26
Configurable salary parameters for Fellowship and Secretary

muharem:
Integrates `pallet-parameters` to the Collectives runtime and exposes the Fellowship and Secretary salary configurations as dynamic parameters that can be updated on-chain.

Supports: https://collectives.subsquare.io/fellowship/referenda/526

Dynamic parameters:
- `fellowship_salary::SalaryConfig` -`{ asset, budget }`
- `secretary_salary::SalaryConfig` -`{ asset, budget, salary_rank1 }`

`asset` is a `VersionedLocatableAsset` defaulting to USDT on Asset Hub. `budget` and `salary_rank1` are raw asset units (decimals included).

Access: Root, the FellowshipAdmin origin, or a vote among Fellows.

Notes:
- since budgets and salaries are stored as raw units, changing the asset (different decimals) requires updating those values together with it.
- for the Fellowship, the per-rank salaries stored in `pallet-core-fellowship` are also raw asset units and must be updated alongside the asset change.
- asset/value updates should be scheduled at the end of the current salary payout period to avoid mid-period mismatches between budgets and outstanding claims.


system-parachains/collectives/collectives-polkadot/src/secretary/mod.rs
```diff
@@ -81,22 +80,51 @@ pub struct SalaryForRank;
...
+pub type SecretarySalaryPaymaster = PayOverXcm<
+	crate::fellowship::FellowshipSalaryInteriorLocation,
+	crate::xcm_config::XcmConfig,
+	crate::PolkadotXcm,
+	ConstU32<{ 6 * HOURS }>,
+	AccountId,
+	(),
+	SalaryAssetId,
+	AliasesIntoAccountId32<(), AccountId>,
+>;
```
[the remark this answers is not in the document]
muharem: this is intended

system-parachains/collectives/collectives-polkadot/src/parameters.rs
```diff
@@ -0,0 +1,164 @@
...
+
+	fn try_origin(
+		origin: RuntimeOrigin,
+		key: &RuntimeParametersKey,
+	) -> Result<Self::Success, RuntimeOrigin> {
+		match key {
+			// Fellowship salary parameters can be set by Root, the FellowshipAdmin
+			// origin (i.e. token holder referendum), or by a vote among all Fellows.
+			RuntimeParametersKey::FellowshipSalary(_) |
+			RuntimeParametersKey::SecretarySalary(_) => EitherOfDiverse::<
```
ggwpez, line 135: I think for secretary it makes sense for the fellowship to be able to change it, but for the fellowship salary itself it should probably be only FellowshipAdmin | Root.

muharem: I was thinking about this as well. But the funds are already in the Fellowship salary account, and the configuration does not change that account.

ggwpez: Hm will defer it to the next reviewer.

ggwpez (approved):

muharem:
@bkchr can you have a look please?

Krayt78:
LGTM - only caveat :

Since `set_parameter` now lets governance swap the salary `asset` on-chain, i think we need to limit the risks of mistakes. 
So a value like `salary_rank1 = 6_666 * 1_000_000` only means "6,666" because USDT has 6 decimals. If a future referendum changes `asset` to one with a different number of decimals and forgets to rescale the amounts, payouts silently come out off by the decimals ratio.

so only questions i have are :

1. Is the expectation that any asset change will, in practice, only ever be a same-decimals swap (e.g. USDT↔USDC, both 6 decimals)?
2. Would it be worth adding a short warning comment on the Fellowship `asset` field pointing at the `pallet-core-fellowship` rank-salary dependency, so whoever drafts a future referendum doesn't have to rediscover the coupling?

system-parachains/collectives/collectives-polkadot/src/parameters.rs
Krayt78: So right now fellows can already change their per rank-salaries i think - so it wouldnt really change anything on that front. 

I understand the concern but imo its fine to let the fellows be handle to self govern on that because in the end the fellowship only has a limited amount of funds given by the treasury so the worst we can do is use all of the allocated funds.

muharem:
@Krayt78 updated the docs

bkchr (approved):

system-parachains/collectives/collectives-polkadot/src/parameters.rs
```diff
@@ -0,0 +1,170 @@
...
+		///
+		/// Defaults to USDT on Asset Hub (`PalletInstance(50)/GeneralIndex(1984)`) asset with 6
+		/// decimals and budget value of 250,000 USDT (i.e., `250_000_000_000` means
+		/// 250,000.000000 USDT).
+		#[codec(index = 0)]
+		pub static SalaryConfig: crate::parameters::FellowshipSalaryConfig =
+			crate::parameters::FellowshipSalaryConfig {
+				asset: Box::new(VersionedLocatableAsset::V5 {
+					location: AssetHubLocation::get(),
+					asset_id: AssetId(Location::new(0, [PalletInstance(50), GeneralIndex(1984)])),
```
bkontur: what about reusing the existing: https://github.com/polkadot-fellows/runtimes/blob/5e18de5f89d5bf05cf9cd77be8d25b66315e2805/system-parachains/collectives/collectives-polkadot/src/xcm_config.rs#L81

```suggestion
					location: crate::xcm_config::AssetHubUsdt::get().location,
					asset_id: crate::xcm_config::AssetHubUsdt::get().asset_id,
```

btw. can we better have some direct conversion between `LocatableAssetId` -> `VersionedLocatableAsset`?

system-parachains/collectives/collectives-polkadot/src/parameters.rs
```diff
@@ -0,0 +1,170 @@
...
+		/// Defaults to USDT on Asset Hub (`PalletInstance(50)/GeneralIndex(1984)`) asset with 6
+		/// decimals and budget value of 13,332 USDT (i.e., `13_332_000_000` means
+		/// 13,332.000000 USDT) and salary for rank 1 of 6,666 USDT (i.e., `6666_000000` means
+		/// 6,666.000000 USDT).
+		#[codec(index = 0)]
+		pub static SalaryConfig: crate::parameters::SecretarySalaryConfig =
+			crate::parameters::SecretarySalaryConfig {
+				asset: Box::new(VersionedLocatableAsset::V5 {
+					location: AssetHubLocation::get(),
+					asset_id: AssetId(Location::new(0, [PalletInstance(50), GeneralIndex(1984)])),
```
bkontur:
```suggestion
					location: crate::xcm_config::AssetHubUsdt::get().location,
					asset_id: crate::xcm_config::AssetHubUsdt::get().asset_id,
```

system-parachains/collectives/collectives-polkadot/src/parameters.rs
bkontur: maybe we could at least refactor and reuse `impl TryConvert<(), LocatableAssetId> for SalaryAssetId {` ?

system-parachains/collectives/collectives-polkadot/src/secretary/mod.rs
```diff
@@ -81,22 +80,51 @@ pub struct SalaryForRank;
...
 		}
 	}
 }
 
+/// Reads the Fellowship salary asset from the runtime parameters pallet and
+/// resolves it to a [`LocatableAssetId`].
+pub struct SalaryAssetId;
+impl TryConvert<(), LocatableAssetId> for SalaryAssetId {
+	fn try_convert(_: ()) -> Result<LocatableAssetId, ()> {
+		let versioned_asset = *crate::dynamic_params::secretary_salary::SalaryConfig::get().asset;
```
bkontur: @muharem nit - why not deduplicate `SalaryAssetId`? My proposal: https://github.com/polkadot-fellows/runtimes/pull/1196

system-parachains/collectives/collectives-polkadot/src/parameters.rs
franciscoaguirre: [Architects (rank 4+)](https://github.com/polkadot-fellows/runtimes/pull/1072) can already manage the fellowship salary account. I would change this from `Fellows` to `Architects` to make it consistent with that.
I agree the fellowship should be able to manage their own resources and worst case OpenGov can just create a ref to demote bad actors or take away their funds.

franciscoaguirre (approved):
Sweet

Just missing benchmarks and I'd also add the cleanup from @bkontur

Issue #603, opened 2025-02-24
CI: Code blob size check


kianenigma:
As we have learned in https://github.com/paritytech/polkadot-sdk/issues/7664#issuecomment-2678983053, the size of a code blob, when un-compressed, should not be more than a threshold that is hardcoded in the Polkadot node client. 

If this is confirmed, there should be a CI check that verifies the wasm code sizes before triggering a new release.

PR #1191, opened 2026-05-28
ci: enforce uncompressed runtime wasm size limit before release

Nathy-bajo:
closing #603 

Adds a CI check that fails the workflow when a runtime's uncompressed wasm exceeds the node-side decompression bomb limit — 50 MiB for relay chains (`sp_maybe_compressed_blob::CODE_BLOB_BOMB_LIMIT`) and 30 MiB for parachains (`validation_code_bomb_limit = max_code_size * 10`) 


Nathy-bajo:
@kianenigma Ready for review

PR #1193, opened 2026-05-28
Flexible minimum validator set size on RCs 

kianenigma:
And, this PR lowers the default limits to: 

* 100 for Kusama
* 250 for Polkadot

Reasoning: 

* Context: `MinimumValidatorSetSize` on the RCs is not a parameter that controls the validator set-size in any way (that's `staking.ValidatorCount` in AHs). It is only a safety mechanism that ensures no matter what AH does, RC ignores a set smaller than this size. 
* The `StakingAsync` on AHs already has strict checks where any set smaller than `ValidatorCount` are rejected on the AH side.
* Why keeping `MinimumValidatorSetSize` flexible via a storage item vs. `const`? Because if not, any reduction of validator set requires a runtime upgrade on the RC first, slowing down the process. 
* Why `pallet-parameters` for the flexibility? Unlike `parameter_types { pub storage }` the origin can be flexible, and the proposal to upgrade it is more transparent than a `system.setStorage`.
* Why lower defaults? 
  * As mentioned above, these lower limits won't have any meaningful impact, until `Staking.ValidatorCount` on AH is changed.  
  * Since reduction of sets [is already envisioned](https://forum.polkadot.network/t/polkadot-staking-changes-progress-timeline/17436) as a part of the DAP proposal, this flexibility + lower defaults speeds up that effort. 
* Why 100 and 250? They follow the the core count and usage in each respective network.


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 - Zombienet SDK integration tests for elastic scaling on Asset Hub Polkadot and People Polkadot.
+- Flexible validator set size limit on rc ([#PR1193](https://github.com/polkadot-fellows/runtimes/pull/1193))
```
sigurpol: Dumb-or-philosophical question: this PR doesn't touch `ValidatorCount` or the active set, but on enactment it significantly lowers the RC's safety floor: Kusama 1000 → 100 , Polkadot 600 → 250  i.e. it widens the range of validator-set sizes the RC will accept from AH without rejecting them. 
AFAIK no prior referendum agreed to these specific floors (I might be wrong). Since the runtime-upgrade referendum is then the approval, I think a change like this must be called out explicitly here (old → new values + rationale, "safety floor, not the active count" etc), so the community can knowingly vote it up or down.

relay/kusama/src/lib.rs
```diff
@@ -796,6 +796,15 @@ pub mod dynamic_params {
...
 		pub static BurnDestination: BurnDestinationAccount = Default::default();
 	}
+
+	/// Parameters used by `pallet-staking-async-ah-client`.
+	#[dynamic_pallet_params]
+	#[codec(index = 2)]
+	pub mod ah_client {
+		/// Minimum size of the validator set that the relay chain will accept from Asset Hub.
+		#[codec(index = 0)]
+		pub static MinimumValidatorSetSize: u32 = 100;
```
muharem, line 805: from current `1000` to `100`?

sigurpol: more context: https://forum.polkadot.network/t/proposal-discussion-adapting-kusama-to-demand-and-its-jam-future/17141/8 (and  https://hackmd.io/@Lederstrumpf/referendum-kusama-validator-set-reduction-to-700) where a first step is the reduction from 1000 to 700 (considered a no-brainer) - aiming then to a final figure in the ballpark of 100 or so (to be discussed / agreed / decided). This PR doesn't touch validator set size but make more ergonomic to play with it in the future (if I read it right, @kianenigma  :smile: ).

CHANGELOG.md
kianenigma: My thinking was that this is only lowering the floor on a safety check that was too conservative/strict to begin with, and as you said it has no effect on the network, until OpenGov touches `ValidatorCount` on AH.

If you prefer, we can keep the values on the old values, and let them be upgraded in sync. 

For Kusama, the old pattern would have worked, but it was un-ergonomic. For Polkadot, the `const` had us locked in, and had to be changed. 

Main motivation here is to lower the effort and time-delay of "when OpenGov eventually votes to lower the real `ValidatorCount`". Making both `dynamic-parameter` + speculatively reducing them helps, but the latter is optional indeed.

Alternatively, if you think I should only call out the change better in the PR, that's pretty reasonable to do.

sigurpol: My personal preference: 
> Alternatively, if you think I should only call out the change better in the PR, that's pretty reasonable to do.


but I am also perfectly fine with 
> If you prefer, we can keep the values on the old values, and let them be upgraded in sync.

kianenigma:
@muharem @ggwpez PR description updated for more transparency, ready for review.

CHANGELOG.md
kianenigma: PR description updated.

CHANGELOG.md
```diff
@@ -27,6 +27,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - PAH: filter `staking.reap_stash` from both `BaseCallFilter` and the XCM `SafeCallFilter` while the validator self-stake transition to higher bond. See Ref: [#1890](https://polkadot.subsquare.io/referenda/1890) ([#1159](https://github.com/polkadot-fellows/runtimes/pull/1159))
 - Collectives Polkadot: Increase the Secretary Collective salary budget ([#1172](https://github.com/polkadot-fellows/runtimes/pull/1172))
 - Align all runtimes to use the same values for e.g. `TransactionByteFee` and some other clean ups. ([#1175](https://github.com/polkadot-fellows/runtimes/pull/1175))
+- Flexible validator set size limit on rc ([#PR1193](https://github.com/polkadot-fellows/runtimes/pull/1193))
```
kianenigma:
```suggestion
- Flexible minimum validator set size on rc ([#PR1193](https://github.com/polkadot-fellows/runtimes/pull/1193))
```

CHANGELOG.md
sigurpol: thanks for updating the description :bow:  Do we want to do the final extra mile and mention also in the CHANGELOG that we are lowering default limits ? I am just assuming that the average community member before voting will mostly read the CHANGELOG without reading any single PR description.

muharem (approved):

ggwpez (approved):

PR #1201, opened 2026-06-11
AHP: pin nomination-pools unbonding bound across the fast-unbond flip

sigurpol:
`SubPools::with_era` is bounded by `TotalUnbondingPools = bonding_duration() + PostUnbondingPoolsWindow`. Setting `AreNominatorsSlashable = false` drops the nominator bonding duration 28 -> 2, shrinking the bound 32 -> 6: oversized historical maps then fail to decode and the next `unbond` overwrites them with an empty default, destroying per-era unbonding accounting.

Make `PostUnbondingPoolsWindow` a `Get` returning `MAX - nominator_bonding_duration()`, keeping the bound pinned at 32 regardless of the flag (subtracting the same value the pallet adds). Post-flip the effective merge window widens to 30 eras, covering deferred pre-flip slashes (SlashDeferDuration = 27).
AHP only, AHK is unaffected.

Interim runtime fix ahead of the in-pallet decoupling in polkadot-sdk.

### DIFF
--- a/system-parachains/asset-hubs/asset-hub-polkadot/src/staking/nom_pools.rs
+++ b/system-parachains/asset-hubs/asset-hub-polkadot/src/staking/nom_pools.rs
@@ -16,14 +16,49 @@
 
 //! Config for the nomination pools.
 
-use crate::{staking::StakingAdmin, *};
-use frame_support::traits::Nothing;
+use crate::{
+	staking::{BondingDuration, StakingAdmin},
+	*,
+};
+use frame_support::traits::{Get, Nothing};
 use sp_runtime::FixedU128;
 
 parameter_types! {
 	pub const PoolsPalletId: PalletId = PalletId(*b"py/nopls");
 	// Allow pools that got slashed up to 90% to remain operational.
 	pub const MaxPointsToBalance: u8 = 10;
+}
+
+/// Bound for `pallet_nomination_pools::SubPools::with_era`, frozen at its historical maximum so it
+/// does not shrink when the nominator bonding duration drops.
+///
+/// The pallet derives its `TotalUnbondingPools` bound as `StakeAdapter::bonding_duration() +
+/// PostUnbondingPoolsWindow`, and `StakeAdapter::bonding_duration()` resolves to
+/// `StakingInterface::nominator_bonding_duration()`, which drops from `BondingDuration` (28) to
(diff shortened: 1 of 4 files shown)


sigurpol:
@Ank4n , @cirko33, @kianenigma , @ggwpez  PTAL :bow:

system-parachains/asset-hubs/asset-hub-polkadot/src/staking/nom_pools.rs
```diff
@@ -26,6 +29,38 @@ parameter_types! {
...
+/// destroying per-era unbonding accounting (bugbounty report #90).
+///
+/// We pin the bound at `BondingDuration + 4 = 32` by computing the window as `MAX -
+/// nominator_bonding_duration()`. Subtracting the same value the pallet adds keeps
+/// `TotalUnbondingPools == MAX` by construction, regardless of the flag. Post-flip this also widens
+/// the effective merge window to `32 - 2 = 30` eras, keeping each per-era pool on its correct
+/// points-to-balance ratio long enough for deferred pre-flip slashes (`SlashDeferDuration = 27`) to
+/// still be applied.
+///
+/// TODO: this is an interim runtime fix, meant to be superseded by the in-pallet decoupling in
```
ggwpez, line 50: Okay so we also have to ensure that we dont increase it more now than the permanent fix would set it to?  
Otherwise we have the same problem again if someone fills up the limit now.

sigurpol: not really, this fix makes it so that TotalUnbondingPools = 32 no matter what (assuming BondingDuration doesn't change ofc), basically `TotalUnbondingPools = nominator_bd + (BondingDuration + 4 − nominator_bd) = BondingDuration + 4 = 32` so nominator_bd doesn't count in the result.

On top of this, this is meant to be removed with next SDK upgrade when the proper fix lands in-pallet.

ggwpez (approved):
CI migration check should also catch if there is a decoding error.

kianenigma (approved):

Ank4n (approved):

sigurpol:
@kianenigma , @ggwpez  can you retrigger https://github.com/polkadot-fellows/runtimes/actions/runs/27360466093/job/80859887486?pr=1201 ? IT failed due to a timeout :bow:

kianenigma (approved):

Issue #460, opened 2024-09-25
Enable async backing for system parachains


s0me0ne-unkn0wn:
# Overview

We had tried enabling 6-second block times for all the system parachains in #266, as it seemed to be easy, but it was discovered that this is blocked by paritytech/polkadot-sdk#3268. Since, considering the current system parachains transaction traffic, such an enablement doesn't provide any real benefit for the network, that work has been put on hold.

This issue serves as a tracking issue, should anyone want to perform a transition for a single system parachain or a group of them.

# Current status
## Kusama
- [ ] AssetHub
- [ ] BridgeHub
- [ ] Coretime
- [ ] Encointer
- [x] People

## Polkadot
- [ ] AssetHub
- [ ] BridgeHub
- [ ] Collectives
- [ ] Coretime
- [ ] People

clangenb:
Encointer is also on 6s in v1.3.3, since https://github.com/polkadot-fellows/runtimes/pull/462

eskimor:
With the new slot based collator - please keep slot times to 12s (single collator will produce two blocks in a row).

clangenb:
Can you please elaborate that? Does this affect encointer that is already at 6 seconds too?

eskimor:
There is just no need to. When writing above message I also thought that shorter slot times affects censorship resistance, but @bkchr rightly just pointed out that these points become moot with elastic scaling anyways. Nevertheless, I would recommend for chains to keep slot time at 12s, if there is no need to reduce it, as increasing is harder than decreasing. So if in doubt, better to leave it as is. (No harm done for chains which already changed it, 6s are fine too.)

muharem:
I checked the pallets of Asset Hub, in 2503 sdk release, all AH pallets using block numbers, have `BlockNumberProvider` type within it's config. We can setup it to use Relay Chain blocks.

muharem:
Async Backing for Kusama Asset Hub 
https://github.com/polkadot-fellows/runtimes/pull/659

skunert:
Just an update, @lexnv will keep  track of bringing async-backing to polkadot-assethub.

skunert:
@lexnv While the averag block time seems to be okay, I am seeing a ton of blocks produced by the same collator every 12s. This is a bit weird, in the current configuration the unincluded segment should get filled relatively fast and then each collator should just build one block every 6s. This should be looked into before we continue with enablement on other chains.

PR #1202, opened 2026-06-12
Enable async backing for BridgeHub, Coretime and Collectives system parachains

Stephenlawrence00:
Enables async backing (6s block time, 24s Aura slot duration) for the remaining system parachains tracked in #460: Bridge Hub Polkadot/Kusama, Coretime Polkadot/Kusama, and Collectives Polkadot.

### DIFF
--- a/.github/workflows/runtimes-matrix.json
+++ b/.github/workflows/runtimes-matrix.json
@@ -139,7 +139,7 @@
       "wss://ksm-rpc.stakeworld.io/bridgehub"
     ],
     "is_relay": false,
-    "blocktime": 12000,
+    "blocktime": 24000,
     "extra_args": "--disable-mbm-checks",
     "benchmarks_templates": {
       "pallet_xcm_benchmarks::generic": "templates/xcm-bench-template.hbs",
@@ -163,7 +163,7 @@
       "wss://dot-rpc.stakeworld.io/bridgehub"
     ],
     "is_relay": false,
-    "blocktime": 12000,
+    "blocktime": 24000,
     "extra_args": "--disable-mbm-checks",
     "benchmarks_templates": {
       "pallet_xcm_benchmarks::generic": "templates/xcm-bench-template.hbs",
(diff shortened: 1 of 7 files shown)


acatangiu:
What is the reasoning behind this change?

This improves the latency and effective bandwidth/blockspace for these chains at the cost of extra Polkadot RC load (cpu, storage, networking, etc). But afaict we don’t really need either better latency or bandwidth for any of these chains. So, unless there’s another reason to do it, I’m against it.

Stephenlawrence00:
> What is the reasoning behind this change?

The overall main goal is to finish #460. These are the last system parachains still on the legacy 12s non-async-backing setup, so this brings them in line with the rest (AH, People, etc. already moved). Beyond consistency, async backing also gives Bridge Hub lower message relay latency and headroom for traffic spikes rather than steady-state bandwidth. If the extra RC load is the concern, happy to hold off on the ones where there's clearly no benefit (Collectives and Coretime) and keep just Bridge Hub.

acatangiu:
@bkchr @sandreim please weigh in here. Does it help with code maintenance or otherwise to finish #460? From the Polkadot network point of view I only see minor downsides, but could be worth it if you think applying this everywhere simplifies the system or codebase.

sandreim:
> @bkchr @sandreim please weigh in here. Does it help with code maintenance or otherwise to finish #460? From the Polkadot network point of view I only see minor downsides, but could be worth it if you think applying this everywhere simplifies the system or codebase.

I don't see how it simplifies the codebase. The only benefit is that these chains will have shorter block times and 4x more ref CPU per block.

acatangiu:
> I don't see how it simplifies the codebase. The only benefit is that these chains will have shorter block times and 4x more ref CPU per block.

Yes, but it comes at a network+cpu cost to the operators ergo to the Polkadot network. The 3 chains here are operating with mostly empty blocks, and also have no latency-sensitive use cases. So the benefit is only theoretical while the cost is real. That’s why I am leaning against doing it.

PR #1204, opened 2026-06-12
Polkadot: correct TotalIssuance over-count from block 313222 killStorage dusting

### LINKED ISSUE #147: Block `313222` messed up `TotalIssuance` by 20K DOT
https://polkadot.subscan.io/extrinsic/313222-3 killed account `11gh7WpYe5CKN82opEkrZUJsHeaFZa3pdEkEZFSMkV3k1Xj` that was holding 20K DOT at that time without updating TI.  
Discovered in https://substrate.stackexchange.com/questions/10842

I have no historical context on this. Currently the TI is off by ~25K DOT. Not sure if there are other reasons for this.

Stephenlawrence00:
fixes #147

Adds a one-off, idempotent runtime migration that decreases Polkadot's `Balances::TotalIssuance` by exactly 19,999.998 DOT via `force_adjust_total_issuance(Decrease, ..)`, realigning the ledger total after block 313222's `sudo`→`system.killStorage` deleted an account without decrementing total issuance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


bkchr (changes requested):
Ty! 

But this should just be done as an individual root/whitelisted referendum calling the function to correct this.

Stephenlawrence00:
> Ty!
> 
> But this should just be done as an individual root/whitelisted referendum calling the function to correct this.

Ah, I see. That makes more sense 😅

PR #1207, opened 2026-06-14
Fix undecodable legacy Proxy.Proxies entries on Polkadot & Kusama

### LINKED ISSUE #453: An entry in the Proxy.proxies storage map cannot be decoded
Hello!

While working on decoding historic storage entries, I ran into an issue whereby the storage entry in `Proxy.Proxies` keyed by account ID (twox64_concated) `0x0E6DE68B13B82479FBE988AB9ECB16BAD446B67B993CDD9198CD41C7C6259C49` and with value `0x0898c8a3d01da9877b7b30877c717ae8f2a7e726cefa176c5dfcdcebc9b6a122230098c8a3d01da9877b7b30877c717ae8f2a7e726cefa176c5dfcdcebc9b6a1222304005109bd2e0000000000000000000000` fails to decode. it decodes OK in spec version 18, and starts failing to decode as of spec version 23 and until the present day by the looks of it.

The value type for this map in spec version 18 is `(Vec<(T::AccountId, T::ProxyType)>, BalanceOf<T>)`.

The value type as of spec version 23 is `(Vec<ProxyDefinition<T::AccountId, T::ProxyType, T::BlockNumber>>, BalanceOf<T>)`, where `ProxyDefinition` is:

```rust
#[derive(Encode, Decode, Clone, Copy, Eq, PartialEq, Ord, PartialOrd, RuntimeDebug)]
pub struct ProxyDefinition<AccountId, ProxyType, BlockNumber> {
	/// The account which may act on behalf of another.
	delegate: AccountId,
	/// A value defining the subset of calls that it is allowed to make.
	proxy_type: ProxyType,
	/// The number of blocks that an announcement must be in place for before the corresponding call
	/// may be dispatched. If zero, then no announcement is needed.
	delay: BlockNumber,
}
```

In other words, the size of the value type grew by 4 bytes when this `delay: BlockNumber` field was added (and then the tuple turned to a struct).

In my tests, the value of this field never changed when the type did, and so the value is unable to be decoded to this day.

Should a storage migration be applied to remove this entry from the map?

Stephenlawrence00:
closes  #453 

Migration to repair `Proxy.Proxies` entries undecodable since spec 23 (added `delay` field + removed `SudoBalances` type); keeps valid proxies, refunds deposits when none survive.


Stephenlawrence00:
@bkchr @kianenigma Can you please take a look when/if you have time? 🙇

PR #1211, opened 2026-06-22
Migrate off deprecated parachain runtime APIs on both relays

### LINKED ISSUE #1199: Migrate off deprecated parachain runtime APIs suppressed with #[allow(deprecated)] on both relays
## Context

Three collation/block-production-critical parachain runtime APIs are implemented with deprecation warnings silenced rather than migrated.

## Affected sites

| API | Polkadot | Kusama |
|---|---|---|
| `candidate_pending_availability` | [lib.rs#L2469](https://github.com/polkadot-fellows/runtimes/blob/main/relay/polkadot/src/lib.rs#L2469) | [lib.rs#L2538](https://github.com/polkadot-fellows/runtimes/blob/main/relay/kusama/src/lib.rs#L2538) |
| `backing_state` (via `para_backing_state`) | [lib.rs#L2566](https://github.com/polkadot-fellows/runtimes/blob/main/relay/polkadot/src/lib.rs#L2566) | [lib.rs#L2635](https://github.com/polkadot-fellows/runtimes/blob/main/relay/kusama/src/lib.rs#L2635) |
| `async_backing_params` | [lib.rs#L2571](https://github.com/polkadot-fellows/runtimes/blob/main/relay/polkadot/src/lib.rs#L2571) | [lib.rs#L2640](https://github.com/polkadot-fellows/runtimes/blob/main/relay/kusama/src/lib.rs#L2640) |

## Risk

These are deprecated upstream. When the SDK drops the deprecated impls, the runtimes break. The successor path partly exists already (e.g. `claim_queue` at [relay/polkadot/src/lib.rs#L2588](https://github.com/polkadot-fellows/runtimes/blob/main/relay/polkadot/src/lib.rs#L2588)).

## Tasks

- [ ] Map each deprecated API to its non-deprecated successor
- [ ] Migrate impls and remove `#[allow(deprecated)]`
- [ ] Verify collators/clients consuming these APIs are unaffected

Nathy-bajo:
closes #1199.

Reimplements the `candidate_pending_availability` and `async_backing_params` `ParachainHost` methods on the Polkadot and Kusama relays using their non-deprecated successors and drops the corresponding `#[allow(deprecated)]`, while documenting why `para_backing_state` must keep delegating to the deprecated helper.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+### Changed
+
+- Polkadot & Kusama relay: Migrate the `ParachainHost` runtime API off the deprecated `candidate_pending_availability` and `async_backing_params` helpers (now implemented via `candidates_pending_availability` and a direct configuration read respectively) and drop the corresponding `#[allow(deprecated)]`. `para_backing_state` still delegates to the deprecated `backing_state` helper, as the full `BackingState` cannot be reconstructed from public runtime APIs; clients should use the already-exposed `backing_constraints` and `candidates_pending_availability` instead ([#1199](https://github.com/polkadot-fellows/runtimes/issues/1199)).
+
 ## [2.3.1] 12.06.2026
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


alindima:
this is not really useful. it's the API that is considered deprecated. not the actual API implementation.

the point was that these APIs were superseded by some others, and we wanted to mark them as deprecated so that nodes stop using them and we could eventually remove them.

Nathy-bajo:
> this is not really useful. it's the API that is considered deprecated. not the actual API implementation.
> 
> the point was that these APIs were superseded by some others, and we wanted to mark them as deprecated so that nodes stop using them and we could eventually remove them.

You're right; the deprecation targets node-side callers, and since the runtime must implement these trait methods anyway, this just swaps internals and silences the warning without moving anyone off the deprecated API, so I'll close it.

PR #1212, opened 2026-06-22
Disable relay session set_keys/purge_keys post-AHM

### LINKED ISSUE #1200: type KeyDeposit = () on both relays post-AHM session keys can be registered with no deposit (spam vector)
## Summary

`pallet_session` on both relays sets `type KeyDeposit = ();`, so `session::set_keys` reserves nothing and `SessionKeys` entries can be created for free, a post-AHM storage-spam vector.

## Expected

Registering session keys is no longer free. Post-AHM the session-key lifecycle lives on Asset Hub (which already has the full deposit mechanism), and those keys are propagated to the relay via `ah_client::set_keys_from_ah`, so the direct relay `set_keys`/`purge_keys` extrinsics should be disabled rather than kept open.

## Actual

The relay extrinsics are still callable and take no deposit. The `KeyDeposit` value is defined but left unwired, and the relay path was only kept enabled temporarily to give validators time to migrate to Asset Hub:

```rust
pub KeyDeposit: Balance = deposit(1, 5 * 32 + 33);

// TODO: we will set this post-AHM
type KeyDeposit = ();
```

## Evidence

- [relay/polkadot/src/lib.rs#L552](https://github.com/polkadot-fellows/runtimes/blob/main/relay/polkadot/src/lib.rs#L552)
- [relay/kusama/src/lib.rs#L561](https://github.com/polkadot-fellows/runtimes/blob/main/relay/kusama/src/lib.rs#L561)

## Fix

Disable `set_keys` and `purge_keys` on both Polkadot and Kusama relays by extending `PostAhmFilter` with `Session(set_keys | purge_keys) => false`, leaving `KeyDeposit` as `()` and removing the now obsolete deposit scaffolding (the unused `KeyDeposit` parameter and the `validators_who_cannot_afford_session_key_deposit` remote test). Add tests asserting both calls are rejected with `CallFiltered`. Disabling the relay extrinsics does not affect the Asset Hub to relay key propagation through `ah_client::set_keys_from_ah`. Flipping this on is partly a timing call (the validator grace period needs to be considered over), so it needs sign off from the staking leads (cc @Ank4n @kianenigma).

Nathy-bajo:
closes #1200 

Disables the `session.set_keys` and `session.purge_keys` extrinsics on the Polkadot and Kusama relays via `PostAhmFilter`, since post-AHM session keys are managed on Asset Hub and forwarded to the relay through `ah_client::set_keys_from_ah`.


sigurpol:
As per my comment https://github.com/polkadot-fellows/runtimes/issues/1200#issuecomment-4772207030, I believe the best way to fix #1200  would be to disallow setting / purging keys via RC and allowing to do it only via AssetHub via `stakingRcClient`.

sigurpol (approved):
LGTM, I'll check with validators as well if we are ready for the full deprecation of setting/purging keys via RC in the next runtime upgrade. @kianenigma , @Ank4n , @ggwpez  please have a look too.

cirko33 (approved):
LGTM! Agreeing with @sigurpol solution proposal.

Ank4n (approved):
Looks good

Ank4n:
> Looks good

Maybe one comment: should we do Kusama first and Polkadot later, in case some old validators are still using RC extrinsics? Lets us break things gradually.

Given Polkadot RC's ED is 1 DOT, bloating storage is cheaper but not free.

sigurpol:
> > Looks good
> 
> Maybe one comment: should we do Kusama first and Polkadot later, in case some old validators are still using RC extrinsics? Lets us break things gradually.
> 
> Given Polkadot RC's ED is 1 DOT, bloating storage is cheaper but not free.

We are having some offline communications with validators to timely inform them (and they were already aware that RC path  was going to be deprecated, as clearly stated in the documentation). Considering that, I would be honestly fine in keeping the usual strategy to 1. deploy on  Kusama first 2. deploy  on Polkadot after with a reasonably short delta (weeks, not months). Then it really depends on what the content of the next runtime upgrade will be - but I wouldn't let this PR alone justify a big delay between Polkadot and Kusama deployment.

sigurpol:
As a note: I was thinking that this PR could have as well set `ProxyType::Staking => false` but then I think it's out of scope, it should be done once we cleanup RC similarly to what we are doing in SDK 11940.

sigurpol:
@kianenigma , @ggwpez  PTAL so we can move on with this PR :pray:

bkontur (approved):

ggwpez (approved):

sigurpol:
Thanks @Nathy-bajo for the great work :rocket:

Nathy-bajo:
> Thanks @Nathy-bajo for the great work 🚀

Thanks for the amazing reviews and follow-ups also 🙌

PR #1213, opened 2026-06-24
Remove redundant relay-common crate, use polkadot-runtime-common

clangenb:
The relay-common crate held a copy-pasted `EraPayoutParams/relay_era_payout` from the SDK. The current `polkadot-runtime-common` dependency already exports these at `polkadot_runtime_common::impls`, so the local copy is now obsolete.

So now there is no need for maintaining the current relay-common for the time being, and hence I removed it altogether.


relay/kusama/src/lib.rs
```diff
@@ -884,8 +884,7 @@ impl pallet_staking::EraPayout<Balance> for EraPayout {
 				None
 			},
 		};
-		log::debug!(target: "runtime::kusama", "params: {params:?}");
-		relay_common::relay_era_payout(params)
```
clangenb, line 905: Upstream does no longer implement debug. But we do not log this in the other runtimes. So I thought it is fine to remove it.

Edit: Prepared an upstream PR to derive stuff in case we ever need it: https://github.com/paritytech/polkadot-sdk/pull/12449

acatangiu (approved):

bkchr (approved):

carlosala (approved):

PR #1217, opened 2026-06-24
Use ForceUnstuckOnFailedMigration from upstream frame-support

clangenb:
Pending todo that has been resolved with this.


acatangiu (approved):

ggwpez (approved):
It would also make sense to add a blanket 1-day max-duration for the configured MBMs.

Currently the `max_blocks` of the aggregated MBM is just the sum of all the individual tuple elements. But if one migration does not set any, then there is no fallback.  

So as a follow up if you are interested I would propose to add a 1-day `max_blocks` as fallback to ensure that no MBM on Polkadot Hub ever takes more than 1 day. Currently it would just keep going indefinitely if any MBM has a bug where it does not make progress.

PR #1219, opened 2026-07-08
Restrict `inherent digests` to at most ~100KiB

bkchr:
This restricts the size of the header at initialization to ~100KiB.


acatangiu (approved):

relay/kusama/src/lib.rs
```diff
@@ -255,11 +255,25 @@ impl Contains<RuntimeCall> for PostAhmFilter {
...
+		})
+		.max_header_size(100 * 1024)
+		.build();
+}
+
 impl frame_system::Config for Runtime {
 	type RuntimeEvent = RuntimeEvent;
 	type BaseCallFilter = PostAhmFilter;
 	type BlockWeights = BlockWeights;
-	type BlockLength = BlockLength;
```
bkontur, line 262: I was just curious why `BlockLength` is not removed from imports and it looks like `BlockLength` is still used by `OffchainSolutionLengthLimit` https://github.com/paritytech/polkadot-sdk/blob/bbbb695f4564f0bb2435924d1608d2d483c93608/polkadot/runtime/common/src/elections.rs#L41 over `impl_elections_weights` from [`impl_runtime_weights`](https://github.com/polkadot-fellows/runtimes/blob/31ba937fdbc2bb307979a99005ef6af636e824ed/relay/kusama/src/lib.rs#L169).

Maybe we should change/fix `impl_elections_weights` and `OffchainSolutionWeightLimit` / `OffchainSolutionLengthLimit` to get the values from `frame_system::Config::BlockLength` / `frame_system::Config::BlockWeights` instead of using `BlockWeights` generated by `impl_runtime_weights` and macro expecting `BlockLength` value. Just to avoid possible future inconsistencies between `polkadot_runtime_common::BlockLength` used by `OffchainSolutionLengthLimit` vs runtime-provided `RuntimeBlockLength` (potentially also `BlockWeights` vs `RuntimeBlockWeights`).

Also, it looks like that we could just remove `polkadot_runtime_common::BlockLength` import and rename `RuntimeBlockLength` -> `BlockLength` and the macro `impl_runtime_weights` will work.


Also, another possibility is to remove `impl_runtime_weights!` and define `BlockWeights` directly here.

---

cc: @sigurpol Do we still need this `impl_elections_weights` for the relaychain?

bkontur (approved):
Fix for failing tests for bridge hubs: https://github.com/polkadot-fellows/runtimes/pull/1220 (just integrity test, not affecting runtime setup, it is more for offchain relayer setup + follow-up)

To my [comment](https://github.com/polkadot-fellows/runtimes/pull/1219#discussion_r3551363369), the actual values should be ok for now, but worth to follow-up

PR #1229, opened 2026-07-22
[Testing] Asset Hub local/dev genesis: minimal dev_stakers, add opt-in large-staker-set preset

bkontur:
## Problem / Context

The Asset Hub Polkadot & Kusama `dev`/`local_testnet` genesis presets seed a large `dev_stakers` set — `(2000, 25000)` for Polkadot and `(4000, 15000)` for Kusama. These stakers are generated programmatically in `pallet-staking-async`'s `build_state`, which makes the **raw** local chain spec huge and slow to build and dominates local/zombienet network spawn time — even though a local network authors blocks via collators (`collatorSelection` + `session`), not staking.

- **Discovered while** doing some Polkadot/Kusama bridges testing — paritytech/parity-bridges-common#3299.
- **Originally introduced** as part of the Asset Hub Migration (AHM) in #856 ([this line](https://github.com/polkadot-fellows/runtimes/pull/856/changes#diff-4181a5e90d735e214ba911625514bded1f1528341bf7c25ff4b6ab14b52e9cf6R72)).
- **Purpose of the large set:** runtime **benchmarking**. The multi-block election benchmarks need a full election snapshot page, and the numbers are deliberately over-provisioned to over-estimate election weights (per review discussion).

## Proposed solution

- **`dev`** keeps the large `dev_stakers` set — `frame-omni-bencher` uses the `development` preset by default, so benchmarking works unchanged (no bot/CI changes).
- **`local_testnet`** is shrunk to a tiny `(10, 15)` set — all the `-local` zombienet/bridge chains need (fast spawn).
- The large set is also available as an **opt-in** `local_testnet_large_staker_set` preset (chain-spec-generator: `asset-hub-{polkadot,kusama}-local-large-staker-set`) for spinning up a large-staker local network.

Production genesis is **unchanged**.

## Numbers

Raw genesis via the `chain-spec-generator` (`--no-default-features --features fast-runtime,asset-hub-polkadot,asset-hub-kusama`):

```sh
time ./target/release/chain-spec-generator <chain> --raw > raw.json
jq -r '.genesis.raw.top | keys | length' raw.json   # keys
```

| preset | `chain-spec-generator <chain>` | `dev_stakers` | raw keys | size | `--raw` gen time |
|---|---|---|--:|--:|--:|
| `local_testnet` | `asset-hub-polkadot-local` | (10, 15) | **~400** | ~5.2 MiB | **~1.7s** |
| `local_testnet` | `asset-hub-kusama-local` | (10, 15) | **~400** | ~5.3 MiB | **~1.7s** |
| `local_testnet_large_staker_set` | `asset-hub-polkadot-local-large-staker-set` | (2000, 25000) | 216,227 | **81 MiB** | **114.2s** |
| `local_testnet_large_staker_set` | `asset-hub-kusama-local-large-staker-set` | (4000, 15000) | 152,206 | **64 MiB** | **82.6s** |

`dev` uses the same large set as `local_testnet_large_staker_set` (it's what benchmarking runs against); it isn't exposed as a chain-spec-generator chain.

`local_testnet` still authors blocks (`polkadot-omni-node --chain <raw>.json --dev-block-time 1000 --tmp`), and the `dev` (large) preset passes the `pallet_election_provider_multi_block` benchmarks that the tiny set fails.


bkontur:
> ## TODO / open questions
> 
>   - [ ] What is the purpose of this large dev_stakers set — staking benchmarking, plain local testing, or the Asset Hub Migration (AHM)?
>   - [ ] Do we still need it this big?
>   - [ ] If it must stay the default for dev/local_testnet, I can flip the approach instead: keep the large set as the default and add a dedicated local_minimal_stakers preset for the lean/fast case.

@ggwpez @sigurpol @Ank4n guys, please, I could use some help with the reasoning here :point_up_2:, could you take a look?

Ank4n:
The large set is for [benchmarking the election provider pallet](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/election-provider-multi-block/src/benchmarking.rs#L26). The bench cli uses the default present ([DEV_RUNTIME_PRESET](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/utils/frame/benchmarking-cli/src/pallet/mod.rs#L221C50-L221C68)). We can either preserve that and reduce only the local testnet preset, or pass the large staker genesis preset to [bench invocation](https://github.com/polkadot-fellows/runtimes/blob/main/.github/scripts/cmd/cmd.py#L164). 

We can run a bench update in this PR and check the weights (for epmb*) look sane - then we should be good.

bkontur:
> The large set is for [benchmarking the election provider pallet](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/election-provider-multi-block/src/benchmarking.rs#L26). The bench cli uses the default present ([DEV_RUNTIME_PRESET](https://github.com/paritytech/polkadot-sdk/blob/master/substrate/utils/frame/benchmarking-cli/src/pallet/mod.rs#L221C50-L221C68)). We can either preserve that and reduce only the local testnet preset, or pass the large staker genesis preset to [bench invocation](https://github.com/polkadot-fellows/runtimes/blob/main/.github/scripts/cmd/cmd.py#L164).
> 
> We can run a bench update in this PR and check the weights (for epmb*) look sane - then we should be good.

Thank you, meantime we also found out that it is required to pass benchmarking and DEV_RUNTIME_PRESET preset:
https://github.com/polkadot-fellows/runtimes/pull/1229/changes/784991145e9d55a0d07d221db54bb83faeb42e76

Basically, this PR adjusts:
- for DEV_RUNTIME_PRESET uses the minimal working big set to pass benchmarks `(2_500, 1_600)`
- local uses very low number (`(10, 20)`)
- new local preset for the very big (`(4_000, 15_000)`)

@Ank4n do you suggest for DEV_RUNTIME_PRESET to lower them to [minimum that passes (`(2_500, 1_600)`)](https://github.com/polkadot-fellows/runtimes/pull/1229/changes/784991145e9d55a0d07d221db54bb83faeb42e76) or keep those big random numbers (`4_000, 15_000)`)?

Ank4n:
> @Ank4n do you suggest for DEV_RUNTIME_PRESET to lower them to [minimum that passes (`(2_500, 1_600)`)](https://github.com/polkadot-fellows/runtimes/pull/1229/changes/784991145e9d55a0d07d221db54bb83faeb42e76) or keep those big random numbers (`4_000, 15_000)`)?

Ideally, this number only needs to be as big as the actual on-chain numbers (PAH is up to 2k validators and 22.5k nominators). @kianenigma can confirm later (he's OOO), but we deliberately put in bigger numbers to overestimate the weight in case our benchmark isn't super accurate for worst case scenario (underestimating could stall AH).

Ank4n:
> > @Ank4n do you suggest for DEV_RUNTIME_PRESET to lower them to [minimum that passes (`(2_500, 1_600)`)](https://github.com/polkadot-fellows/runtimes/pull/1229/changes/784991145e9d55a0d07d221db54bb83faeb42e76) or keep those big random numbers (`4_000, 15_000)`)?
> 
> Ideally, this number only needs to be as big as the actual on-chain numbers (PAH is up to 2k validators and 22.5k nominators). @kianenigma can confirm later (he's OOO), but we deliberately put in bigger numbers to overestimate the weight in case our benchmark isn't super accurate for worst case scenario (underestimating could stall AH).

I think the best option would be to keep the existing (large) numbers for benchmarking, but move them into a new dedicated large-staker preset that the benchmark uses. Wdyt?

bkontur:
> > > @Ank4n do you suggest for DEV_RUNTIME_PRESET to lower them to [minimum that passes (`(2_500, 1_600)`)](https://github.com/polkadot-fellows/runtimes/pull/1229/changes/784991145e9d55a0d07d221db54bb83faeb42e76) or keep those big random numbers (`4_000, 15_000)`)?
> > 
> > 
> > Ideally, this number only needs to be as big as the actual on-chain numbers (PAH is up to 2k validators and 22.5k nominators). @kianenigma can confirm later (he's OOO), but we deliberately put in bigger numbers to overestimate the weight in case our benchmark isn't super accurate for worst case scenario (underestimating could stall AH).
> 
> I think the best option would be to keep the existing (large) numbers for benchmarking, but move them into a new dedicated large-staker preset that the benchmark uses. Wdyt?

@Ank4n actually, I implemented this, but it was to much changes around cmd.py and matrix passing another profile, so I decided to keep it simple:
- keep `development` preset with the big overestimated dev_stakers, which is used for benchmarking by default
- change the `local_testnet` for very small dev_stakers, so the local runs are by default quicker and no need to generated ~80MiB chain-spec by zombienet 
- added new preset `local_testnet_large_staker_set`, just in case `development` is not enough and somebody wants to run locally with big dev_stakers (actually this is kind of unnecessary, you can run `development`)

Ank4n:
> @Ank4n actually, I implemented this, but it was to much changes around cmd.py and matrix passing another profile, so I decided to keep it simple:
> 
> * keep `development` preset with the big overestimated dev_stakers, which is used for benchmarking by default
> * change the `local_testnet` for very small dev_stakers, so the local runs are by default quicker and no need to generated ~80MiB chain-spec by zombienet

Sounds good. 

> * added new preset `local_testnet_large_staker_set`, just in case `development` is not enough and somebody wants to run locally with big dev_stakers (actually this is kind of unnecessary, you can run `development`)

Yeah doesn't make sense to add this.

kianenigma:
Conclusion sounds good to me. What Ankan said [above](https://github.com/polkadot-fellows/runtimes/pull/1229#issuecomment-5047581337) is the best theoretical option but if that's too much of an overhead for the `cmd.py`'s logic, what Branislav suggested is also good. 

All in all, as long as for benchmarking we use the large presets, it is all fine. 

It would be nice to make sure that there are some checks in place, so that in the future, no one accidentally reduces `dev_stakers` of the preset used in benchmarking. I think the runtime already fails on running benchmarks though. 

All in all, sorry about this requirement for large `dev_stakers`. It is the only (easy) way to keep good benchmarks for these pallets, and it is slow as hell. We can hopefully deprecate these pallets by EOY.

sigurpol:
>We can hopefully deprecate these pallets by EOY.

which year are you talking about :rofl: ?

I am fine with @bkontur 's pragmatic approach if what @Ank4n suggests is too much extra work. As Kian said, eventually this election machinery is meant to go away

PR #1234, opened 2026-07-30
Fix remote-proxy dropping newest relay root when multiple parachain blocks share a relay parent

### LINKED ISSUE #1230: `RemoteProxyRelayChain.BlockToRoot` drops new relay roots when full
## Situation

`on_validation_data` prunes by relay-block age, then pushes one entry per parachain block:

https://github.com/polkadot-fellows/runtimes/blob/31ba937fdbc2bb307979a99005ef6af636e824ed/pallets/remote-proxy/src/lib.rs#L214-L226

The pruning cutoff is `relay_parent - MaxStorageRootsToKeep`, but the hook runs once per *parachain* block, and every parachain block on the same relay parent pushes a duplicate entry.

On KAH, `MaxStorageRootsToKeep` is `RC_MINUTES`, so the vec capacity is 10:

https://github.com/polkadot-fellows/runtimes/blob/31ba937fdbc2bb307979a99005ef6af636e824ed/system-parachains/asset-hubs/asset-hub-kusama/src/lib.rs#L1218-L1220

Since [#1174](https://github.com/polkadot-fellows/runtimes/pull/1174) switched AHK to 24s Aura slots, several parachain blocks are produced per relay block. `BlockToRoot` fills with duplicates that are too young to prune, so `try_push` fails and the newest relay root is dropped. The failure is a `debug_assert!`, so it is silent in production.

## Evidence

At AHK block `19295078`:

- `BlockToRoot` keys: `34471287, 34471287, 34471287, 34471288, 34471288, 34471288, 34471289, 34471289, 34471290, 34471290`
- `parachainSystem.lastRelayChainBlockNumber`: `34471296`

So the last stored key lags the actual relay parent by 6 blocks.

The same pattern appears at the other failing pins from the scheduled `Update Known Good Block Numbers` workflow:

| Run | AHK block | last stored key | relay parent | lag |
|---|---:|---:|---:|---:|
| [29833774560](https://github.com/open-web3-stack/polkadot-ecosystem-tests/actions/runs/29833774560) | `19295078` | `34471290` | `34471296` | `6` |
| [29884062319](https://github.com/open-web3-stack/polkadot-ecosystem-tests/actions/runs/29884062319) | `19315869` | `34478562` | `34478568` | `6` |
| [29902845087](https://github.com/open-web3-stack/polkadot-ecosystem-tests/actions/runs/29902845087) | `19326433` | `34482255` | `34482261` | `6` |

Control pins from passing runs have lag `0`:

| AHK block | last stored key | relay parent | lag |
|---|---:|---:|---:|
| `19286333` | `34468289` | `34468289` | `0` |
| `19304697` | `34474672` | `34474672` | `0` |

The runtime boundary lines up with the slot change:

- AHK block `17836000`: spec `2002000`, Aura slot `12000`
- AHK block `17836700`: spec `2003000`, Aura slot `24000`

At failing pin `19295078`, the root for relay block `34471290` matches the real Kusama relay header root exactly, while relay block `34471296` has no entry in `BlockToRoot`:

- relay `34471290` header root = stored root
- relay `34471296` header root = present on relay, missing from `BlockToRoot`

## Impact

A valid proof anchored at a dropped relay block fails with `UnknownProofAnchorBlock`.

Verified with a real `Proxy.Proxies` read proof from the Kusama relay chain at relay block `34471296`:

- with live AHK state, where `34471296` is absent from `BlockToRoot`, `remote_proxy` fails with `Module { index: 43, error: 0x01 }`
- with the same proof and the real root for `34471296` inserted into `BlockToRoot`, the call succeeds and emits `proxy.ProxyExecuted`

This is also the cause of the PET failures worked around in [open-web3-stack/polkadot-ecosystem-tests#662](https://github.com/open-web3-stack/polkadot-ecosystem-tests/pull/662).

This behavior contradicts the pallet docs, which describe `MaxStorageRootsToKeep` as the effective proof lifetime window:

https://github.com/polkadot-fellows/runtimes/blob/31ba937fdbc2bb307979a99005ef6af636e824ed/pallets/remote-proxy/src/lib.rs#L187-L194

## Related

#792

## Fix

In `on_validation_data`:

1. do not push when `roots.last()` already has the current relay block number
2. if the vec is still full after pruning, evict the oldest entry instead of dropping the newest one

A regression test should call the hook multiple times per relay block and assert that the newest relay root remains retrievable.

Stephenlawrence00:
Fixes `pallet-remote-proxy` silently dropping the newest relay storage root by skipping duplicate `BlockToRoot` entries for the same relay parent and evicting the oldest entry instead of the newest when the bounded vector is full, so remote-proxy proofs anchored at recent relay blocks no longer fail with `UnknownProofAnchorBlock` closes #1230.


pallets/remote-proxy/src/lib.rs
```diff
@@ -212,15 +212,25 @@ pub mod pallet {
...
 					roots.remove(0);
 				}
 
-				// We always remove all the old items before, thus there should always be space in
-				// the vector.
+				// Safety net: if age-based pruning didn't free a slot, evict the oldest root
+				// rather than silently dropping the newest one.
+				if roots.is_full() {
+					roots.remove(0);
+				}
```
bkchr: There can not be more than `MaxStorageRootsToKeep` in `roots`. So, this is not required. 

It was broken before, because it did not checks for the same relay parents, but this is now fixed by your check above.

bkchr (commented):
One nit

Stephenlawrence00:
> One nit

Thanks for the review! Removed the `is_full` branch; with the duplicate check in place there can be at most `MaxStorageRootsToKeep` entries, so `try_push` can never fail.

bkchr (approved):

pallets/remote-proxy/src/lib.rs
```diff
@@ -212,15 +212,21 @@ pub mod pallet {
...
 					block.clone().saturating_sub(T::MaxStorageRootsToKeep::get().into());
 
 				while roots.first().is_some_and(|f| f.0 <= delete_up_to) {
 					roots.remove(0);
 				}
 
-				// We always remove all the old items before, thus there should always be space in
-				// the vector.
+				// The duplicate check above ensures we never store the same relay block twice, so
+				// there can be at most `MaxStorageRootsToKeep` entries and there is always space.
```
rockbmb:
```suggestion
				// The pruning loop above, combined with the duplicate check before it, ensure that
				// 1. all stored relay blocks are unique
				// 2. there can be at most `MaxStorageRootsToKeep - 1` entries (corollary: there is always room in the `roots` vec)
```

rockbmb (approved):

bkchr (approved):

rockbmb (approved):

acatangiu (approved):

rockbmb (approved):

rockbmb (approved):

PR #1236, opened 2026-08-06
[AHP] Add `technical_maintenance` and `prosperity_emergency` OpenGov tracks

sigurpol:
## Why

Two independent problems to solve:

1. The upcoming integrations needed to bring App support on mainnet (Individuality, dotUSD, PSM, Coinage, Bulletin, statement store, ...) bring many operational values that might/will need changing repeatedly and quickly once live: statement store slots, per person storage quotas, onboarding batch sizes, durations, anti-spam fees and more. These are all changes with no financial impact meaning they  move no funds, set no monetary or market policy and do not change who controls what, and should be executed fast as part of normal technical maintenance. 
`Root` is the wrong instrument for these: it is slow, serialised (`max_deciding: 1`) and unbounded.
`whitelisted_caller` dispatches with `Root` power and needs one Fellowship whitelist per call instance, which does not fit repeated tuning and quick technical maintenance. Also its decay is slower than needed because of its unbounded scope. Existing admin tracks carry other domains and their curves were priced for other powers. `technical_maintenance` track is the solution to this problem.

2. The monetary mechanisms behind `Prosperity` additionally needs a protective lane: a way to react fast when
something goes wrong, holding only **very limited** powers that reduce risk. `prosperity_emergency` track is the solution to this problem.

## The solution

Two new tracks with dedicated custom origins are added. Both keep the same hard 5% support floor, reached along different decays.

- `technical_maintenance`: operational settings such as quotas, allowances, limits and anti-spam fees. Support: about 8% of active issuance at day 1, 6% at day 3, easing to the hard 5% floor by day 28. Fast only when uncontested.  `max_deciding` 100, `deposit` 10,000 DOT.
- `prosperity_emergency`: guard for the `Prosperity` monetary mechanisms (PSM, potentially Coinage in the future), intended for the protective tier like operating the circuit breakers (disable minting while redemption stays open, halt swaps, restore via `set_asset_status`). Wider administrative powers (e.g. `set_minting_fee`, `set_redemption_fee`, `set_max_debt`, `set_asset_ceiling_weight`, `add_external_asset`, `remove_external_asset`, `remove_psm` or Coinage's `create_sufficient_instance`, `make_instance_sufficient`, `make_instance_sponsored`) stay on stricter paths until deliberately wired. Support and approval curves are `whitelisted_caller`'s: support 20% of active issuance at day 1, 7.4% at day 7, 5% floor; approval near unanimous in the first days.  `max_deciding` 10, `deposit` **50,000 DOT**.

Both origins get an XCM plurality voice so other chains can accept them from Asset Hub.

`technical_maintenance` is meant to be a cross-domain bounded `catch_all` track whereas `prosperity_emergency` is domain-specific, tailored around `Prosperity`'s needs.

This PR adds the governance plumbing only. Wiring lands with the integrations that consume them.

### Curve rationale

`technical_maintenance` never requires less than the  absolute 5% minimum as `whitelisted_caller`, but is a deliberately more aggressive curve, requiring less support than `whitelisted_caller` in the earlier parts of the decision period.

Its decay is much faster because the origin is bounded: unlike `whitelisted_caller` it cannot dispatch as `Root`, only reach the config positions the runtime explicitly names it in.

On `technical_maintenance` the approval curve keeps the `whitelisted_caller` property that an uncontested proposal passes quickly while a small share of `Nay` votes blocks the fast window.

On `prosperity_emergency` the curves match `whitelisted_caller` exactly: passing is never cheaper than on the proven track at any point in time, and there is no Fellowship gate. A genuine incident is expected to mobilise far more support than routine maintenance does, landing a well backed protective action within one to two weeks.
`prosperity_emergency`'s deposit is `50,000 DOT`, 5× technical_maintenance's deposit. It matches `referendum_killer`. Occupying the guard's deciding slots costs `500k DOT` locked, at slash risk if killed

## Future integrations

Per single parameter, through `pallet-parameters`. Its `AdminOrigin` is an `EnsureOriginWithArg` receiving the key, so individual keys route to the track while neighbouring keys can stay on Root:

```rust
RuntimeParametersKey::SomePallet(some_pallet::ParametersKey::SomeAllowance(_)) =>
    EitherOfDiverse::<EnsureRoot<AccountId>, TechnicalMaintenance>::ensure_origin(origin.clone())
        .map(|_| ())
        .map_err(|_| origin),
```

Per pallet `Config` position, for calls that are actions rather than settings:

```rust
type SomeAdminOrigin = EitherOfDiverse<EnsureRoot<AccountId>, TechnicalMaintenance>;
```

On other chains, as the plurality voice of the body from Asset Hub, following the existing
`StakingAdmin` pattern on People and Bulletin:

```rust
EnsureXcm<IsVoiceOfBody<AssetHubLocation, TechnicalMaintenanceBodyId>>
```

For `prosperity_emergency` specifically, pallets that store their admin origins in state (the PSM stores `PalletsOrigin` values with distinct `full` and `emergency` tiers) can be given this origin as the stored emergency admin, so the protective actions happen through referenda on this track with no further runtime wiring, while full administrative powers stay on stricter paths until deliberately assigned.


sigurpol:
@seadanda , @kianenigma  this is the initial draft version for two new tracks aiming to ease the integrations needed to bring support for App on mainnet. I believe it's easier to discuss / comment on the PR directly - I've put some very opinionated and debatable values for both and explained the idea behind in the PR description.
PTAL :bow:

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/tracks.rs
```diff
@@ -200,6 +207,34 @@ const TRACKS_DATA: [pallet_referenda::Track<u16, Balance, BlockNumber>; 16] = [
...
+			min_approval: APP_TECHNICAL_MAINTENANCE,
+			min_support: SUP_TECHNICAL_MAINTENANCE,
+		},
+	},
+	pallet_referenda::Track {
+		id: 17,
+		info: pallet_referenda::TrackInfo {
+			name: s("prosperity_admin"),
+			max_deciding: 10,
+			decision_deposit: 5 * GRAND,
```
kianenigma: I would presume prosperity needs higher deposit than technical, no? it is more critical

sigurpol: the initial number here comes simply from `technical_maintenance` mirroring `whitelisted_caller`'s economics (decision: 10k, max_deciding 100), `prosperity_admin` mirroring `staking_admin`'s (decision: 5k, max_deciding 10). I can surely bump it - but matches the other "admin" family, and it's not a security parameter by itself but it's true that the lower, the easier to clog the track with junk / spam referenda (again, the same applies currently to other admin tracks). I can bump to 10k and call it a day though :smile:

sigurpol:
> For prosperity_admin specifically, pallets that store their admin origins in state (the PSM stores PalletsOrigin values with distinct full and emergency tiers) can be given this origin as the stored emergency admin, so the protective actions happen through referenda on this track with no further runtime wiring, while full administrative powers stay on stricter paths until deliberately assigned.

@seadanda , here in particular I would like to hear your opinion. What I am suggesting is that PSM `emergency_admin = ProsperityAdmin` vs `full_admin != ProsperityAdmin` (so there we would go for the usual root/whitelisted path)

@kianenigma  to make it clearer I have updated the description and comment as well:

> prosperity_admin: guard for the Prosperity monetary mechanisms (PSM, coinage), intended for the protective tier like operating the circuit breakers (disable minting while redemption stays open, halt swaps, restore via set_asset_status). Wider administrative powers (e.g. `set_minting_fee`, `set_redemption_fee`, `set_max_debt`, `set_asset_ceiling_weight`, `add_external_asset`, `remove_external_asset`, `remove_psm` or Coinage's `set_underlying_asset_id`) stay on stricter paths until deliberately wired. 

So initially the way I see prosperity_admin integrated is just to link it to `set_asset_status` and no more than that (no enabling/ disabling of PSM or Coinage or anything else).

Now that I think more about that, `maybe` a better name would be `prosperity_emergency_admin` then.

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/origins.rs
```diff
@@ -66,6 +66,20 @@ pub mod pallet_custom_origins {
...
+		/// Origin for guarding the monetary mechanisms around Prosperity: PSM, coinage and related
+		/// financial parameters.
+		///
+		/// Intended to be designated as the protective tier of those mechanisms once they are
+		/// deployed: operating their circuit breakers, for example disabling the minting of a
+		/// stable asset while keeping redemptions open, halting its swaps entirely, or restoring
+		/// normal operation, so that a fast referendum on this track can react when something
+		/// goes wrong. Wider administrative powers, such as fees, debt ceilings and asset
+		/// composition, stay on stricter paths until deliberately wired here.
+		ProsperityAdmin,
```
sigurpol: `ProsperityEmergencyAdmin` is probably a better name? It would also make the couple with `emergency_admin` origin in PSM code pretty obvious

sigurpol:
with Kian OOO this week, can I get some more :eyes:  on this PR?  @ggwpez , @seadanda , @joepetrowski  can you have a look :pray: ?

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/tracks.rs
```diff
@@ -200,6 +207,34 @@ const TRACKS_DATA: [pallet_referenda::Track<u16, Balance, BlockNumber>; 16] = [
...
+			min_enactment_period: 10 * RC_MINUTES,
+			min_approval: APP_TECHNICAL_MAINTENANCE,
+			min_support: SUP_TECHNICAL_MAINTENANCE,
+		},
+	},
+	pallet_referenda::Track {
+		id: 17,
+		info: pallet_referenda::TrackInfo {
+			name: s("prosperity_admin"),
+			max_deciding: 10,
```
seadanda, line 229: This combination gives the track a 50,000 DOT DOS cost: ten referenda with decision deposits can occupy all deciding slots for the full 28-day period with support unless killed. The deposits are even refundable unless killed. IMO the decision deposit should be much higher on this fast track

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/origins.rs
```diff
@@ -66,6 +66,20 @@ pub mod pallet_custom_origins {
 		WhitelistedCaller,
 		/// Origin for signaling that the network wishes for some change.
 		WishForChange,
+		/// Origin for technical maintenance of operational values across system pallets: quotas,
+		/// allowances, limits, durations, fees and similar settings whose change moves no funds
```
seadanda: `fees` in general do not fit the in with what I'd infer from the name of the track. we should specify exactly what is meant here, as a change in fees can be a material economic change vs just an anti spam mechanism or whatever to protect the system

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/origins.rs
```diff
@@ -66,6 +66,20 @@ pub mod pallet_custom_origins {
...
+		/// and does not change who controls what.
+		TechnicalMaintenance,
+		/// Origin for guarding the monetary mechanisms around Prosperity: PSM, coinage and related
+		/// financial parameters.
+		///
+		/// Intended to be designated as the protective tier of those mechanisms once they are
+		/// deployed: operating their circuit breakers, for example disabling the minting of a
+		/// stable asset while keeping redemptions open, halting its swaps entirely, or restoring
+		/// normal operation, so that a fast referendum on this track can react when something
```
seadanda, line 80: Is it worth enforcing a pallet side restriction so this origin can only move toward a safer state? `ProsperityEmergencyAdmin` would also be clearer if we intend to only go that way, otherwise it really is the ProsperityAdmin and has bidirectional authority to pause and resume

sigurpol:
> with Kian OOO this week, can I get some more 👀 on this PR? @ggwpez , @seadanda , @joepetrowski can you have a look 🙏 ?

Kian, Oliver, Dónal: I've added you as collaborators in case you want to push more stuff on his PR while I am away next week

seadanda (commented):
> The 5% support floor on both tracks matches whitelisted_caller, so passing a referendum here is never cheaper than on the proven track, whatever the timing.

`TechnicalMaintenance` is "cheaper" to force something to pass everywhere along the curve except the end. taking some values along the period:
day1: 8.1% vs 20%
day3: 6% vs 11%
day28: 5 vs 5

This sentence would be clearer as:
> It never requires less than the same absolute 5% minimum, but is a deliberately more aggressive curve, requiring less support than `whitelisted_caller` in the earlier parts of the decision period.

sigurpol:
> > The 5% support floor on both tracks matches whitelisted_caller, so passing a referendum here is never cheaper than on the proven track, whatever the timing.
> 
> `TechnicalMaintenance` is "cheaper" to force something to pass everywhere along the curve except the end. taking some values along the period: day1: 8.1% vs 20% day3: 6% vs 11% day28: 5 vs 5
> 
> This sentence would be clearer as:
> 
> > It never requires less than the same absolute 5% minimum, but is a deliberately more aggressive curve, requiring less support than `whitelisted_caller` in the earlier parts of the decision period.

:100:  agree - updated

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/origins.rs
sigurpol: Agree with you, I have rephrased (`fees` -> `anti-spam fees`. What I was hinting at here was more specific and was about `set_alias_fee` that IIUC is about  "anti-spam" `fees` (so needed to protect the system)  as opposed to `fees` that have an impact on economy. Happy to further improve to make it clearer

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/tracks.rs
sigurpol: makes sense - raised deposit to 50,000 DOT (matching `referendum_killer` example which is another `emergency` track) so we move from 50k to 500k DOT to occupy all 10 slots

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/origins.rs
sigurpol: That's a very good question, and what we decide in the upstream PSM pallet dictates the right name here.

My mental model so far: this track maps 1:1 to  PSM's `EmergencyAdmin` tier . As of today (AFAIK), that tier is bidirectional: `set_asset_status` writes whatever `CircuitBreakerLevel` it is given, identically for Full and Emergency tiers, so the emergency admin can pause and resume.

If we give `CircuitBreakerLevel` an ordering (`AllEnabled < MintingDisabled < AllDisabled`) and make `set_asset_status` reject, for Emergency callers only, any move toward a less safe state:

1. Speed becomes **asymmetric** by design. Toward safety we stay fast (this track, no whitelisting needed). Back to normal goes through `full_admin`, i.e. Root or the whitelisted path today. I think that is the right shape for a circuit breaker  but it is a real cost and we must  agree on explicitly. It can also be softened later without lowering any bar: `full_admin` could be a separate `root-strength` track as you have suggested already, which shortens the resume path while keeping the same support requirements as Root. 
2. (stating the obvious...) The one-way property is only  a contract kept upstream in `frame/psm`, not something this PR can enforce, and the name `ProsperityEmergency` or similar would sell a promise the runtime cannot check. Slightly more annoying when it contains the `Emergency` word inside...

About naming,  if the one-way restriction lands upstream (and so we go in the direction that  emergency can only lower ceilings or pause and never raise ceilings or resume), an emergency name is right. If the tier stays bidirectional, the current name is probably the right one.

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/tracks.rs
sigurpol: done - following https://github.com/polkadot-fellows/runtimes/pull/1236#discussion_r3750978967

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/origins.rs
sigurpol: closing  - we are discussing about it here https://github.com/polkadot-fellows/runtimes/pull/1236#discussion_r3750978983

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/origins.rs
sigurpol: If we go for the one-way approach, I've a  draft PR with PSM upstream changes here -> https://github.com/paritytech/polkadot-sdk/pull/12886

bkontur:
@karolk91 please, track or find out if we need/want to align Bulletin's cfg with these new tracks (AllowedAuthorizers or anything else)

sigurpol:
@seadanda , @kianenigma , following some recent discussions , I have modified name and curves of the prosperity track - rationale in the PR description.

sigurpol:
About being a blocker: @seadanda , @kianenigma, @ggwpez , @re-gius , @Zebedeusz  correct me if I am wrong, but we are ok in launching `2.5` without the technical maintenance fast track,  if we assume `dotNS` doesn't need it vs whitelisting / registering short names. For statement store, bulletin-chain  it's mostly a nice to have to speed up **future** maintenance, but definitely not a blocker. 

That said, I think that 
- `technical maintenance` fast track is pretty not controversial as  it is now - it could ease #1233 integration and I don't see any strong reason to keep it on hold, it could be merged rn if there are no objections
- `prosperity emergency` is not going to be used from day 1 so we have time to tune / refine its related curve later on, it doesn't harm to merge it in the current form but I can also drop it


So guys, please review :pray:

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/origins.rs
```diff
@@ -66,6 +66,21 @@ pub mod pallet_custom_origins {
 		WhitelistedCaller,
 		/// Origin for signaling that the network wishes for some change.
 		WishForChange,
+		/// Origin for technical maintenance of operational values across system pallets: quotas,
+		/// allowances, limits, durations and anti-spam fees, and similar settings whose change
+		/// moves no funds, does not price the system's economics and does not change who controls
+		/// what.
+		TechnicalMaintenance,
```
gui1117, line 73: by changing the anti-spam fees it does "price the system's economics", no?
saying "does not price the system's economics" seems wrong.

What it must not price is the consensus system economics.

gui1117 (commented):
`TechnicalMaintenance` seems ok for the usage we want to make in coinage and other individuality pallets.

I don't know about `ProsperityEmergency` because I don't know the consequences in PSM. It also depends on how many DOT holders are responsive and in how much time. And I don't know those numbers.

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/origins.rs
gui1117: But it is fine as a general rule.

gui1117:
> prosperity emergency is not going to be used from day 1 so we have time to tune / refine its related curve later on, it doesn't harm to merge it in the current form but I can also drop it

I would not merge it if it is not completely approved, or we should make it explicit in the type name. People will forget to review it again later and nobody will be accountable.

EDIT: yes it is also fair, when using it, the reviewer of the PR must make sure it is appropriate. all is good to me then

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/origins.rs
sigurpol: Fair point, the wording can def be improved.  I can reword in something like "whose change moves no funds, sets no monetary or market policy, and does not change who controls what". So if changing a value redistributes value or influences market policy, it does not belong on this track. In that sense, something like alias-account anti-spam fee can belong here since its purpose is purely and only abuse deterrence.

sigurpol:
> I don't know about `ProsperityEmergency` because I don't know the consequences in PSM. 

On the PSM consequences: the intent is that this origin only ever holds the emergency tier, which so far  in the PSM pallet is exactly one call, `set_asset_status` so the circuit breaker (normal / minting disabled with redemptions still open / all halted).  Everything else like fees, ceilings, asset composition,  admin reassignment etc require the `Full tier` and stay on the usual stricter tracks. Also  nothing is wired here in this PR.

> It also depends on how many DOT holders are responsive and in how much time. And I don't know those numbers.

If we check last whitelisted ref,  we can see that support ranged between 6 and 12% of active issuance, mostly concentrated in a limited number of large holders. We can't and do not want assume that they will necessarily vote for this emergency tier. But  the underlying (**debatable**) assumption here is that an emergency will mobilize more support than "ordinary routine maintenance" so that's why the track copies whitelisted_caller's curves:  same 5% floor, same fast decay - just very bounded unilke whitelisted's one. Ofc this is a debatable assumption, we can make the curve more aggressive but starting from a well known path like whitelisted one seems a good starting point. All in all, a call mostly for @seadanda  / PSM friends

> I would not merge it if it is not completely approved, or we should make it explicit in the type name. People will forget to review it again later and nobody will be accountable.

I am fine with both keeping or discarding it (I was suggesting this as an open point) - again, mostly a call for @seadanda  / PSM folks vs how we plan to handle an emergency requiring to stop minting or so.

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/origins.rs
sigurpol: (improved the comment)

gui1117 (approved):
The curves are sensible to me, taking from the curve from whitelisted caller makes sense for the emergency, and the other one is only about potential spam and non-critical parameters.

PR that make use of those curve should ensure that they are fit for their new usage.

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/origins.rs
```diff
@@ -66,6 +66,21 @@ pub mod pallet_custom_origins {
 		WhitelistedCaller,
 		/// Origin for signaling that the network wishes for some change.
 		WishForChange,
+		/// Origin for technical maintenance of operational values across system pallets: quotas,
+		/// allowances, limits, durations and anti-spam fees, and similar settings whose change
+		/// moves no funds, sets no monetary or market policy and does not change who controls
+		/// what.
+		TechnicalMaintenance,
```
kianenigma, line 73: I learned while reviewing #1233 that we have, at times, given this right to `Fellows` directly. For example, `ControllerOrigin` of XCMP queue.

I think this track should not replace such use-cases. Fellowship's boundary is cleaner and it does one thing: Whitelist or supervise runtime upgrades.

kianenigma (approved):
I am happy to merge this and we will try to refine it more as we start using it. As long as not used, it is harmless.

system-parachains/asset-hubs/asset-hub-polkadot/src/governance/origins.rs
sigurpol: Totally agree - and it's not meant to.  Suspending XCMP queue or similar is not normal ordinary technical maintenance and needs to keep been handling by a separate track. 
The technical maintenance track is meant for routine maintenance of some settings like  adjusting per-person statement storage quotas, how many new people can be onboarded to the People system per batch and this kind of stuff - so  normal operations, no incidents, no funds to be moved, etc.

I would also argue that maybe  in the past we have used of the Fellowship for routine parameters only for lack of better origin and this track helps to solve this problem , offering an alternative where Fellowship is not involved at all. 

Specifically about ControllerOrigin, it's maybe debatable if Root + Whitelist should be better than having also `Fellows` but it's also out of the scope of this PR.

sigurpol:
> I am happy to merge this and we will try to refine it more as we start using it. As long as not used, it is harmless.

On the other hand, I am happy to merge now if it's useful. Having it as harmless and not wired anywhere is maybe not ideal.
Let's hear also from @seadanda

seadanda (approved):
Some naming and docs will probably not age well since they assume some usecases, depending on how these tracks are actually linked up. Not a reason not to merge though and it at least motivates why we're adding them a bit better

PR #1238, opened 2026-08-07
Cap individual MBM migrations to one day

clangenb:
In theory, misconfigured MBM migrations could run indefinitely. As per https://github.com/polkadot-fellows/runtimes/pull/1217#pullrequestreview-4592477489, this PR introduces a fallback safety mechanism to cap them to one day. After hitting the timeout, the migration will be considered a failure and the `ForceUnstuckOnFailedMigration` will the triggered.

I have also drafted an upstream PR to integrate this into the migrations-pallet: https://github.com/paritytech/polkadot-sdk/pull/12841.


bkchr:
@clangenb I would propose we merge the upstream one, if there is no big need for this version?

clangenb:
I am fine with this, we have been rolling like this for quite a while already.

It will take a while I guess until upstream is merged and released, so my thinking was that it makes sense to close this (rather unlikely to happen) security issue.

PR #1240, opened 2026-08-13
[draft] support any asset that has a rate for transaction fee, xcm transaction fee and xcm delivery fee on Polkadot People Chain

gui1117:
On Polkadot People Chain AssetRate has rate set by governance, currently only HOLLAR is usable as transaction fee and xcm transaction fee.

This PR wants to generalize to any asset with a rate, and to also include xcm delivery fee.

That said we may prefer to directly switch to pools like we do on Polkadot Asset Hub. This would make those asset permissionless, and the price adjusted by arbitrers rather than a potentially slow governance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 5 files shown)


gui1117:
closed in favor of this stack of PR: https://github.com/polkadot-fellows/runtimes/pull/1257

PR #1242, opened 2026-08-18
Define canonical burn account

### LINKED ISSUE #1235: Add system Burn account
For https://forum.polkadot.network/t/proposal-introduce-a-permanent-burn-account-for-polkadot/18280/10

ggwpez:
Define the canonical burn account that can be used to burn funds. Burning in this sense means to render them permanently inaccessible - the 2.1B DOT total issuance cap stays unaffected by this.

Calculated from PalletId  `py/cabrn` for canonical burn.
  - Public key: `0x6d6f646c70792f636162726e0000000000000000000000000000000000000000`
  - SS58: `13UVJyLnbVp77Q2CRfD1Uryp655A696vD62DbmzF7woS1Vf8`

Closes #1235

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+
+- Asset Hub Polkadot: add the canonical burn account ([#1242](https://github.com/polkadot-fellows/runtimes/pull/1242)).
+
 ### Fixed
 
 - Asset Hub Polkadot & Kusama: fix `pallet-remote-proxy` dropping the newest relay storage root when multiple parachain blocks share one relay parent. `on_validation_data` now skips storing duplicate relay blocks, which used to fill the bounded `BlockToRoot` vector with copies too young to prune and silently drop the newest root, so remote-proxy proofs anchored at recent relay blocks no longer fail with `UnknownProofAnchorBlock` ([#1230](https://github.com/polkadot-fellows/runtimes/issues/1230)).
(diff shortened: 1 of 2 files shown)


xlc:
do we want to expose it to the metadata somehow?

kianenigma (dismissed):

sigurpol:
my 2c:  I also believe we need more discussion around this PR and more in general around how  referendum #1926 fits in the current DAP model - which was also approved by community so there is some clear conflict / tension here.

I fully respect what community expressed with its vote but I still believe Jonas made a solid point with his comment [here](https://forum.polkadot.network/t/proposal-introduce-a-permanent-burn-account-for-polkadot/18280/12?u=sigurpol). Burning doesn't really play well with the DAP model where the scope of the DAP buffer was to fund future protocol expenses when issuance declines and that's why burns were / are redirected to it right now.
I also believe there are many points in https://polkadot.subsquare.io/referenda/1926 that need a proper analysis, including DAP buffer being a separate concept from treasury,  the scarcity story helping with DOT price which is pretty hard to prove, the assumption that we are hard-coding now the rule at protocol level instead of deferring to a future referendum  which is also not what a burning account fully achieves and so on.  
We need probably a RFC  or some further technical discussion where we can agree on a proper technical solution and a clear future direction including both economic sustainability of the system and DOT value, with broader scope than just "let's burn 100% of DOT revenue from JAMKB sales" so I wouldn't rush anything here.

ggwpez:
I dont see myself in the position to arbitrate OpenGov. The sentence:

>with that DOT burned and taken out of circulation permanently.

Sounds pretty clear to me. Will close this for now until the fellowship has approved an RFC.

PR #1254, opened 2026-08-24
Add PSM remote tests for Asset Hub Polkadot

Szegoo:
Adds remote tests for PSM on Asset Hub Polkadot, covering minting, redemption, fees, debt tracking, and circuit breakers.

The tests use the runtime’s PSM asset adapter and load live USDT asset data from Asset Hub Polkadot.


muharem (approved):

system-parachains/asset-hubs/asset-hub-polkadot/src/remote_tests/psm.rs
```diff
@@ -0,0 +1,284 @@
...
+		external_asset.clone(),
+		Permill::from_percent(100),
+	));
+	assert_ok!(PsmAssets::mint_into(external_asset.clone(), &caller, 2_000 * PSM_UNIT,));
+
+	PsmTestEnv { internal_asset, external_asset, caller, psm_account }
+}
+
+#[tokio::test]
+#[ignore = "requires a live Asset Hub Polkadot RPC"]
```
rockbmb, line 157: The PSM pallet will also be enabled in KAH (right?); should this (and other such) comments be changed to reflect this?

More importantly, though, why are tests ignored?

* because the pallet is currently not live anywhere, or
* because they should be ignored even if PSM were live? E.g. CI performance/resource concerns, they are to be run manually only

In either case, this should be documented with a TODO or a NOTE.

system-parachains/asset-hubs/asset-hub-polkadot/src/remote_tests/psm.rs
```diff
@@ -0,0 +1,284 @@
...
+			.expect("PSM configured")
+			.fee_destination;
+		assert!(
+			PsmAssets::balance(local_asset_location(PSM_INTERNAL_ASSET_ID), &fee_destination) > 0
+		);
+	});
+}
+
+#[tokio::test]
+#[ignore = "requires a live Asset Hub Polkadot RPC"]
```
rockbmb, line 203: ⬆️

rockbmb (approved):

system-parachains/asset-hubs/asset-hub-polkadot/src/remote_tests/psm.rs
Szegoo: > The PSM pallet will also be enabled in KAH 

It will only be on PAH

Szegoo: They’re ignored because they depend on a live RPC and would make CI flaky. This applies even when PSM is live. However, if you think they should run in CI, I can remove the ignore.

rockbmb: No need to remove the `#[ignore]`; I agree with your assessment that introducing them would introduce flakiness (just consider PET in this repo's CI; it's improved, but still happens).

I asked only to clarify the nature of the tests, and that this should be documented (why they are intentionally avoided in CI, and how to run them oneself).

muharem (approved):

rockbmb (approved):

ggwpez (approved):

PR #1257, opened 2026-08-25
PeopleChain: Accept any rated asset for XCM execution fee (as we do for transaction fee)

gui1117:
On Polkadot People Chain, we currently accept any rated asset for transaction fee, but only HOLLAR for XCM execution. This PR makes it on par.

Note that we just take the first asset if it has a rate, not the best rate of any asset.


gui1117:
> Two actionable issues found.
> ### Review Suggestions
> 
> **system-parachains/people/people-polkadot/src/lib.rs:L1074-L1078**
> 
> This runtime API now iterates and allocates a `Vec` for the entire unbounded `ConversionRateToNative` map. As governance adds rates, an ordinary RPC caller can trigger arbitrarily many storage reads and allocations, potentially exceeding runtime-call limits or tying up RPC workers. Maintain a bounded index of fee-eligible assets and enforce its maximum when registering rates, then use that bounded index here.

I think it is ok, governance will not set tons of asset here.

> **system-parachains/people/people-polkadot/src/lib.rs:L1074-L1076**
> 
> `RelayLocation` is inserted unconditionally and can also appear in `ConversionRateToNative`, since `pallet_asset_rate::AssetKind` accepts any `Location`. Registering such a rate makes the API return DOT twice. Filter `RelayLocation` out of the iterated keys, or deduplicate the collected asset IDs before passing them to `query_acceptable_payment_assets`.

fixed

acatangiu (approved):

kianenigma (approved):

gui1117:
> One actionable issue found.
> ### Review Suggestions
> 
> **system-parachains/people/people-polkadot/src/xcm_config.rs:L287-L292**
> 
> `TakeFirstAssetTrader` selects the first fungible asset accepted by `ForeignAssetsMatcher`, but that matcher accepts every concrete foreign asset and does not check `AssetRate`. If holding contains an unrated asset before a rated one, conversion fails for the unrated asset and the later rated asset is never considered, contrary to the stated “first asset with a rate” behavior. Use a rate-aware matcher/trader that skips assets when `AssetRate` has no conversion rate, and add an execution test where an unrated asset sorts before a rated asset in holding.

indeed I updated the PR description, it is the first asset if it has a rate.

PR #1258, opened 2026-08-25
Polkadot People Chain: accept payment for XCM delivery fee in any rated asset.

gui1117:


system-parachains/people/people-polkadot/src/xcm_config.rs
```diff
@@ -328,11 +425,11 @@ impl xcm_executor::Config for XcmConfig {
 	type PalletInstancesInfo = AllPalletsWithSystem;
 	type MaxAssetsIntoHolding = MaxAssetsIntoHolding;
 	type AssetLocker = ();
-	// TODO: Will need for delivery fees.
-	type AssetExchanger = ();
+	// Delivery fees are priced in DOT but can be settled in any asset with a registered rate.
+	type AssetExchanger = FeesAtAssetRate;
```
gui1117, line 429: This is too dangerous, asset exchange happens for both delivery fee but also on the instruction `ExchangeAsset`, and in case it gets enabled, this allows to mint money by using the difference between the rate from governance and the market rate. So the type system doesn't allow us to use asset rate for delivery fee.

PR #1263, opened 2026-08-28
[draft] Polkadot: Make Asset Hub Whitelisted Caller accessible in People Chain.

gui1117:
This is generated by Claude, I barely read it. It is written to show the potential way to resolve it.

TODO:
* add this into the RootOrTechnicalMaintenance for all manager operation
* use it maybe to some root only origin as well.


kianenigma (approved):

franciscoaguirre (commented):
If we just want People Chain to be able to receive an XCM from Asset Hub dispatched from the WhitelistedCaller track then we don't need this.

When a referendum on that track successfully enacts (meaning it was whitelisted by a fellows rank 3+ referendum on Collectives) then it dispatches the inner call [as Root](https://github.com/paritytech/polkadot-sdk/blob/d0393007a3ed1ba82097b5f7cf891a70f0ab90ee/substrate/frame/whitelist/src/lib.rs#L372). Because of that we can just have the inner call be an XCM send to People with origin_kind: Superuser and that will work with just EnsureRoot<AccountId>, no need for EnsureRootOrWhitelistedCaller or anything like that

kianenigma:
Thanks for clarifying @franciscoaguirre!

PR #1266, opened 2026-09-01
CI: Prevent constant cache churn

ggwpez:
We have 10GB cache in the github free repo. Currently we upload 50+GB per run.  
This causes try-runtime caches to be evicted which makes the checks painfully slow. I am addressing the obvious low-hanging fruit.

Changes:
- Remove all caching besides try-runtime (already uses 8.8GB peak)
- Explicitly delete old try-runtime snapshots after taking new ones to free up space

Next step is to carefully re-enable what we can without causing churn; possibly Clippy cache.


.github/workflows/check-migrations.yml
```diff
@@ -100,27 +106,70 @@ jobs:
...
+          steps.cache-restore.outputs.cache-matched-key == '' &&
+          github.event_name != 'schedule' &&
+          github.event_name != 'workflow_dispatch'
         uses: actions/upload-artifact@v4
         with:
           name: try-runtime-snapshot-${{ matrix.runtime.name }}
           path: snapshot.raw
           retention-days: 1
 
+      - name: Delete previous snapshot for ${{ matrix.runtime.name }}
```
bkchr, line 127: Why do we need this? Because they should be overridden by default?

ggwpez: Otherwise we rely on the LRU logic. I wanted to remove everything here to see how much actual space in the cache we have left.

I think there is only about effectively 1.5GB cache space left, but cleaning up now to confirm.

karolk91 (approved):
Looks good to me to try how that affects builds overall. The other option worth some exploring is also publishing the snapshots as artifacts and other jobs getting the snapshot from articats instead of cache (then leave cache for rust builds) - but of course lets see how it goes with current changes first

sigurpol (approved):
let's see how it goes 🍿

seadanda (approved):

.github/workflows/check-migrations.yml
```diff
@@ -47,6 +50,9 @@ jobs:
   prepare-snapshots:
     needs: [runtime-matrix]
     runs-on: self-hosted
+    permissions:
+      actions: write
+      contents: read
```
franciscoaguirre, line 55: The workfllow file has `permissions: {}`. Could we at least add `persist-credentials: false` on the checkout step?

ggwpez: Will mark as TODO. Since we have cutoff today, any additional CI run that can be avoided, i will happily skip 😄

kianenigma (approved):

PR #1275, opened 2026-09-03
Integrate pallet-assets force_create

ggwpez:
Update crates to pull in:
- https://github.com/paritytech/polkadot-sdk/pull/13067
- https://github.com/paritytech/polkadot-sdk/pull/13072

And configure auto Id increment for pallet assets.

PR #1277, opened 2026-09-03
Benchmark Staking Weights

ggwpez:
Weight refresh after updating the staking pallet here https://github.com/polkadot-fellows/runtimes/pull/1275


sigurpol:
@kianenigma , @ggwpez  I checked why

```
rc_on_session_report	2.42ms	31.34ms	+1196.59
```

It comes from   https://github.com/paritytech/polkadot-sdk/pull/13023 "Scale validator self-stake incentive by era-points performance". The PR rewrote the benchmark from `rc_on_session_report()` to `rc_on_session_report(v: Linear<1, MaxValidatorSet>)`, where v is the number of validators in the relay chain's session report. The payout logic now keeps the weighted-points denominator incrementally: on every session report, `reward_active_era` reads `ErasValidatorIncentiveWeight` for each validator in the report. One storage read per validator.  The old code only accumulated points in memory and wrote one map entry, so its weight was flat.
With 1000 validators like in benchmark case we are talking ~31ms so we should be good

sigurpol (approved):

kianenigma (approved):

sigurpol:
> One access-control defect found.
> 
> ### Review Suggestions
> **system-parachains/people/people-polkadot/src/parameters.rs:L237-L251**
> 
> `TechnicalMaintenance` is granted access to `BulletinChainLocation`, `BulletinTransactionStoragePalletIndex`, and `LitePersonhood::RegistrationFee` by grouping them with operational quotas. These settings were intended to remain Root-only: the first two control where long-term-storage transactions are sent, while the registration fee is an economic parameter. Split these three variants into a branch that calls `frame_system::ensure_root`, leaving only the long-term-storage quotas/cleanup/allowances under `RootOrTechnicalMaintenance`, and update `dynamic_parameter_origin_routes_keys_by_scope` so these keys are tested as Root-only.

@ggwpez , not anymore with #1279  (merged on Kian's branch and not yet on master).
Also fix for the weights for Polkadot ended up in #1280 (again, merged only in Kian's branch - I wanted to minimize CI churn but maybe it would have been better to merge on master instead... the real fix  for weights is in upstream SDK is https://github.com/paritytech/polkadot-sdk/pull/13092)

PR #1281, opened 2026-09-06
Correct Polkadot/Kusama proxy type lattice

rockbmb:
`is_superset` on both relays claims `NonTransfer` is a superset of `Auction`. It is not: the `Auction` filter admits `Registrar::swap`, which `NonTransfer` blocks on purpose. `pallet_proxy` trusts `is_superset` when it gates `add_proxy`, so a `NonTransfer` proxy could add an `Auction` proxy and reach `swap`. One new match arm removes the false claim.

# Tests added

Each relay gets a test that asserts every declared superset edge against the filters. The test covers this regression and every superset-filter mismatch for the call families in the current filters. A filter change that adds a new pallet family also needs a new entry in the test's call list. About 1 ms per relay, at most 1120 filter probes on Polkadot and 1664 on Kusama.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- Polkadot & Kusama: `is_superset` no longer declares `NonTransfer` a superset of `Auction`; the `Auction` filter admits `Registrar::swap`, which `NonTransfer` omits on purpose.
+
 - Asset Hub Polkadot & Kusama: enable `pallet-assets::force_create` by moving asset ID allocation to the new `AssetIdAllocator` config item ([#1275](https://github.com/polkadot-fellows/runtimes/pull/1275)).
 - People Polkadot & Asset Hub Polkadot: wire the `technical_maintenance` track into the operational Individuality settings: quotas, allowances and housekeeping on People, the Individuality parameters on Asset Hub ([#1269](https://github.com/polkadot-fellows/runtimes/pull/1269)).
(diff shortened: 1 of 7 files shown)


bkchr (approved):

xlc (approved):

rockbmb:
> **relay/kusama/src/lib.rs:L1471-L1472**
> 
> This only blocks a `NonTransfer` proxy from creating new `Auction` proxies. Any `Auction` proxy already added through the previous superset rule remains in proxy storage and can continue dispatching `Registrar::swap`, so the vulnerability persists for previously affected accounts after the runtime upgrade. Add a bounded migration that invalidates or quarantines existing `Auction` definitions—requiring owners to recreate legitimate ones—or introduce a legacy encoded proxy type whose filter excludes `swap` so existing definitions cannot retain the escalated permission.
> 
> **relay/polkadot/src/lib.rs:L1273-L1274**
> 
> This only blocks a `NonTransfer` proxy from creating new `Auction` proxies. Any `Auction` proxy already added through the previous superset rule remains in proxy storage and can continue dispatching `Registrar::swap`, so the vulnerability persists for previously affected accounts after the runtime upgrade. Add a bounded migration that invalidates or quarantines existing `Auction` definitions—requiring owners to recreate legitimate ones—or introduce a legacy encoded proxy type whose filter excludes `swap` so existing definitions cannot retain the escalated permission.

>Add a bounded migration that invalidates or quarantines existing `Auction` definitions—requiring owners to recreate legitimate ones—or introduce a legacy encoded proxy type whose filter excludes `swap` so existing definitions cannot retain the escalated permission.

This is not worth it - there remain 0 `Auction`-type proxies on both Polkadot/Kusama relays.

PR #1282, opened 2026-09-07
Coretime: forward bulk revenue to  DAP on Polkadot, burn on AssetHub on Kusama

sigurpol:
All the revenues reaching the broker's `OnRevenue` hook (bulk sales, renewals and the system share of on-demand revenue), were teleported to the relay chain and burnt there, a path `AHMv2` blocks. This obsoletes [RFC-010](https://github.com/polkadot-fellows/RFCs/blob/main/text/0010-burn-coretime-revenue.md) on Polkadot, coherently with DAP model introduced as per  [ref#1827](https://polkadot.subsquare.io/referenda/1827).

Coretime Polkadot adds `pallet-accumulate-and-forward`. `OnRevenue` and `DustRemoval` resolve into its account, which forwards  to the DAP staging account on Asset Hub, where `pallet-dap` deactivates the funds (similarly to what Westend Coretime already does).

On Kusama, we burn revenue on Asset Hub instead of the relay chain, since KSM total issuance is tracked on AssetHub post AHM.

On Coretime Polkadot only, we  retire the `py/ctbrn` holding account with an idempotent migration that sweeps any residual, to the accumulation account, and reaps the account.

Related SDK issue: https://github.com/paritytech/polkadot-sdk/issues/13063

**NOTE**: Asset Hub Polkadot must be upgraded before Coretime Polkadot, otherwise the forwards are rejected and the teleported DOT is lost.

**TODO:**
- [x] fund `13UVJyLjaBdLEGEpgE7dexnp5fiAG5Zu6D1SWSt6j2Q49w52` (`acf/dott`) on Coretime with `ED` (similarly to what we did for DAP staging and main account)
- [ ]  followup task: switch to `RelaychainDataProvider` once  https://github.com/paritytech/polkadot-sdk/issues/13149 lands. Not a blocker for the release imho since accumulate-and-forward pallet is doing merely low prio housekeeping activity, we don't need strictly need a proper clock there
- [ ] followup task not in scope of the PR: re-run benchmarks after upstream SDK fixes https://github.com/paritytech/polkadot-sdk/issues/13129 

**Out of scope**:  moving the whole on-demand coretime ordering from the Relay Chain to Core Time Chain is not part of this PR, it's tracked in https://github.com/paritytech/polkadot-sdk/issues/12715. Related PR  https://github.com/paritytech/polkadot-sdk/pull/13009. See also   RFC https://github.com/polkadot-fellows/RFCs/pull/175


sigurpol:
/cmd bench --runtime coretime-polkadot --pallet pallet_accumulate_and_forward pallet_broker pallet_balances

sigurpol:
/cmd bench --runtime coretime-polkadot --pallet pallet_accumulate_and_forward pallet_broker pallet_balances

system-parachains/constants/src/polkadot.rs
```diff
@@ -38,6 +38,9 @@ pub mod account {
 	pub const FELLOWSHIP_TREASURY_PALLET_ID: PalletId = PalletId(*b"py/feltr");
 	/// Ambassador treasury pallet ID
 	pub const AMBASSADOR_TREASURY_PALLET_ID: PalletId = PalletId(*b"py/ambtr");
+	/// Accumulate-and-forward pallet ID. Derives the account on the Coretime chain that gathers
+	/// coretime revenue before it is forwarded to the DAP on Asset Hub.
+	pub const ACCUMULATE_FORWARD_PALLET_ID: PalletId = PalletId(*b"acf/dott");
```
sigurpol, line 43: already funded  above ED see [here](https://coretime-polkadot.subscan.io/account/13UVJyLjaBdLEGEpgE7dexnp5fiAG5Zu6D1SWSt6j2Q49w52)

system-parachains/constants/src/polkadot.rs
```diff
@@ -38,6 +38,9 @@ pub mod account {
 	pub const FELLOWSHIP_TREASURY_PALLET_ID: PalletId = PalletId(*b"py/feltr");
 	/// Ambassador treasury pallet ID
 	pub const AMBASSADOR_TREASURY_PALLET_ID: PalletId = PalletId(*b"py/ambtr");
+	/// Accumulate-and-forward pallet ID. Derives the account on the Coretime chain that gathers
```
sigurpol, line 41: [debatable] This is meant to be used by any system chain and not only coretime but since coretime is the 1st wiring it, I have just mentioned that in the comment

sigurpol:
@Ank4n, @cirko33, @Szegoo  ( + @andreitrand  for the ACF/DAP integration), PTAL :pray:

CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ Changelog for the runtimes governed by the Polkadot Fellowship.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+### Changed
+
+- Coretime Polkadot & Kusama, Asset Hub Polkadot: coretime bulk revenue is no longer teleported to the relay chain to be burnt. Polkadot forwards revenue to the DAP staging account on Asset Hub through `pallet-accumulate-and-forward`, and Asset Hub grants that forward free execution. Kusama teleports it to Asset Hub and burns it there, since KSM issuance is tracked on Asset Hub. A migration retires the `py/ctbrn` holding account on Coretime Polkadot. ([#1282](https://github.com/polkadot-fellows/runtimes/pull/1282)).
```
Szegoo: Worth mentioning that the dust on Polkadot is now also being forwarded to the DAP.

Szegoo (approved):
The changes look good to me.

CHANGELOG.md
sigurpol: done [here](56dc0250e25c373f74c93ad05ec4f3f388627d27)

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -394,6 +408,9 @@ pub type Barrier = TrailingSetTopicAsId<
 							Equals<bridging::SiblingBridgeHub>,
 							AmbassadorEntities,
 							IsSiblingSystemParachain<ParaId, parachain_info::Pallet<Runtime>>,
+							// Forwards coretime revenue to the DAP staging account without paying
+							// fees.
+							CoretimeAccumulationAccount,
```
acatangiu: we already know all system chains will redirect burns to AH.

even if this PR doesn't do that for all of them, it can set the stage correctly for the pieces it touches.

use [SystemChainAccumulationAccounts](https://github.com/paritytech/polkadot-sdk/blob/2124c778df3a31dd3638684608951fa2ded15280/cumulus/parachains/runtimes/assets/asset-hub-westend/src/xcm_config.rs#L311) here instead.
```suggestion
							SystemChainAccumulationAccounts,
```

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -886,3 +903,18 @@ fn foreign_pallet_has_correct_local_account() {
 	let address = Ss58Codec::to_ss58check_with_version(&account, polkadot);
 	assert_eq!(address, "13w7NdvSR1Af8xsQTArDtZmVvjE8XhWNdL4yed3iFHrUNCnS");
 }
+
+#[test]
+fn coretime_accumulation_account_matches_only_the_forwarder_origin() {
```
acatangiu: this test will also need to cover all sys chains

acatangiu (approved):
Looks good. When is the plan to use DAP on the other system chains?

sigurpol:
> Looks good. When is the plan to use DAP on the other system chains?

Part of https://github.com/polkadot-fellows/runtimes/issues/1283 - hopefully soon / ASAP, depending on other priorities though

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
sigurpol: fair enough - I was mentioning something similar in my own comment for reviewers [here](https://github.com/polkadot-fellows/runtimes/pull/1282/changes#r3950229720) 😄

system-parachains/constants/src/polkadot.rs
sigurpol: done here e0969efd7267fba3a9f196f334de7f962332c23d

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
sigurpol: done here e0969efd7267fba3a9f196f334de7f962332c23d

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
sigurpol: done here e0969efd7267fba3a9f196f334de7f962332c23d

system-parachains/coretime/coretime-polkadot/src/lib.rs
```diff
@@ -309,6 +313,28 @@ impl pallet_balances::Config for Runtime {
...
+	type PalletId = AccumulateForwardPalletId;
+	type Forwarder = xcm_builder::TeleportForwarderForAccountId32<
+		xcm_config::XcmConfig,
+		AssetHubLocation,
+		DotRelayLocation,
+		DapStagingLocation,
+	>;
+	type TransferPeriod = ForwardPeriod;
+	type MinTransferAmount = MinForwardAmount;
+	type BlockNumberProvider = System;
```
seadanda, line 338: parachain block numbers are not a good clock, and get worse if the chain is moved away from regular block production (which may well happen on coretime)
```suggestion
	type BlockNumberProvider = RelaychainDataProvider<Runtime>;
```

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
```diff
@@ -394,6 +417,9 @@ pub type Barrier = TrailingSetTopicAsId<
 							Equals<bridging::SiblingBridgeHub>,
 							AmbassadorEntities,
 							IsSiblingSystemParachain<ParaId, parachain_info::Pallet<Runtime>>,
+							// System chains forward their burns to the DAP staging account without
+							// paying fees.
+							SystemChainAccumulationAccounts,
```
seadanda, line 422: Before this passes coretime burns would be charged fees? or fail? It basically means AH should be upgraded first either way

system-parachains/coretime/coretime-polkadot/src/lib.rs
seadanda: Checked the accumulate and forward pallet - annoyingly it can miss an edge since it checks for strict equality of the modulus, but IMO we should fix this in the pallet and accept that we miss some edges til it's fixed rather than using local block number on a parachain, which we've tried to move away from completely.

seadanda: Or make a note to change this when the pallet is fixed.

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
sigurpol: Good catch! Fail - and teleported DOT would be lost (AH rejects the message before execution while CT already burnt it).   
So yes, AH should be upgraded first. @Ank4n , @cirko33 is this problematic vs MinRelay deployment plan? 
I'll make sure to add it in the CHANGELOG (+ PR body).

system-parachains/coretime/coretime-polkadot/src/lib.rs
sigurpol: Agree that main fix should land upstream - I will create an SDK issue + add the TODO here to switch to RelayChainDataProvider .  
NOTE: I don't think it's critical  since we are talking about housekeeping activity so an irregular block production in practice has no relevant effect here.

sigurpol: SDK issue: https://github.com/paritytech/polkadot-sdk/issues/13149 + TODO added here c479b7de0638cdbc05cf362983e4a51d4cc296f9
I'll prepare the upstream fix in the SDK.

system-parachains/asset-hubs/asset-hub-polkadot/src/xcm_config.rs
sigurpol: Changelog updated here c479b7de0638cdbc05cf362983e4a51d4cc296f9 (and PR body updated too)

Ank4n: This is independent of the AHMv2 deployment, right? Either way, the AHMv2 code ships before the actual migration kicks off. So this can go in whatever release comes next, and if it happens to be the AHMv2 one, we just deploy AH first.

sigurpol: correct, this can go in in the next RU without waiting for AHMv2 (with or without the upstream SDK fix in pallet-accumulate-and-forward)

system-parachains/coretime/coretime-polkadot/src/coretime.rs
```diff
@@ -301,3 +220,76 @@ impl pallet_broker::Config for Runtime {
...
+}
+
+impl OnRuntimeUpgrade for RetireCoretimeBurnAccount {
+	fn on_runtime_upgrade() -> Weight {
+		let burn_account = Self::burn_account();
+		let residual =
+			Balances::reducible_balance(&burn_account, Preservation::Expendable, Fortitude::Polite);
+		if residual > 0 {
+			let accumulation_account = AccumulateForward::accumulation_account();
+			if let Err(e) = Balances::transfer(
```
Ank4n, line 245: There is a `Balances::transfer_all` function as well (reduces few lines but fine as it is).

system-parachains/coretime/coretime-polkadot/src/coretime.rs
```diff
@@ -301,3 +220,76 @@ impl pallet_broker::Config for Runtime {
...
+		PalletId(*b"py/ctbrn").into_account_truncating()
+	}
+}
+
+impl OnRuntimeUpgrade for RetireCoretimeBurnAccount {
+	fn on_runtime_upgrade() -> Weight {
+		let burn_account = Self::burn_account();
+		let residual =
+			Balances::reducible_balance(&burn_account, Preservation::Expendable, Fortitude::Polite);
+		if residual > 0 {
```
Ank4n, line 243: Nit: could exit early in the else arm of this (avoids a few reads if this upgrade code is left behind in the next upgrade).

Ank4n (approved):

system-parachains/coretime/coretime-polkadot/src/coretime.rs
sigurpol: that would be an extrinsic though instead of a fungible trait so doable but maybe not worth it?

system-parachains/coretime/coretime-polkadot/src/coretime.rs
sigurpol: Done in d741b04236a29f32cfcfa93e4e2321a63906e6a3 - actually on account existance rather than on the residual so even if balance 0 with a provider left, we reap the account (current live state...).

system-parachains/coretime/coretime-polkadot/src/lib.rs
seadanda: LGTM, fine to approve. Thanks

seadanda (approved):

PR #1285, opened 2026-09-10
[AHMv2] Migrator state machine + snapshot test harness

Ank4n:
## Context

AHM v2 moves the relay chain's remaining state (balances with their reserves, registrar, HRMP, proxies) to the Coretime chain and Asset Hub. 

This PR scaffolds all the stages (scheduling, warm up, data stages, cool off), wires them into four runtimes behind a feature (so it is no-op for prod builds), and adds a harness that tests against try-runtime chain snapshots. 

I will create separate PRs for the migration stages (rough [checklist](https://github.com/paritytech/polkadot-sdk/issues/12719#issue-4942712626)). This is split out of the reference PoC: https://github.com/polkadot-fellows/runtimes/pull/1284

I have written the full migration implementation (high level) logic here: https://hackmd.io/@ak0n/B16n8fltMx

### rc2-migrator
- **Scheduling** by timestamp (not block number). The migration starts at the first block after the timestamp has passed. This avoids block time drift (assuming schedule is set weeks ahead).
- **Handshake** with CT, a warm up window and then a cool-off window before finishing. 
-  Full stage list (accounts → proxies → registrar → HRMP → sweep → dust → TI correction) is in the enum but nothing implements it yet: each data stage advances in one block and sends nothing, so current run is `Scheduled → WaitingForCt → WarmUp → (data stages) → CoolOff → MigrationDone`.
- Migration is driven by a `Manager` account (as in v1). Also callable by AdminOrigin (and a future manager multisig).

### ct-migrator
No stage machine of its own; the relay chain drives every transition. Stores the stage so the chain can check locally whether the migration is over. 

### Runtime
- All four runtime wiring (RC + CT * 2 networks) is  gated on `all(feature = "ahm-v2", not(feature = "on-chain-release-build"))`. 
- Not benchmarked. Weights are just placeholders for now.

## Testing

Snapshot suite in `integration-tests/ahmv2`: RC and CT loaded from try-runtime snapshots of live state.

```bash
cd integration-tests/ahmv2
just test                 # Polkadot
NETWORK=kusama just test  # Kusama
```

`just test` creates any missing snapshot and runs the suite; it needs the [try-runtime CLI](https://github.com/paritytech/try-runtime-cli). Snapshots are saved in `snapshots/<network>/` (gitignored) and reused until deleted. Override the RPC endpoints with `RC_URI` / `CT_URI`, or the snapshot location with `SNAP_DIR`.

## CI

Optional (non-blocking) `ahm-v2-test` job. Runs unit test and integration test on cached network snapshot (per network).

## Also see 
- #1285 (this)
- #1305

All PRs can be reviewed independently. I've left `TODO(ahm-v2)` markers that we should go through before wiring everything up.

The reason for splitting is simply to make reviewing and iterating on each change easier.

---

---


pallets/ct-migrator/src/lib.rs
```diff
@@ -72,8 +115,98 @@ pub mod pallet {
...
 	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// The relay chain asks whether this chain can receive migrated state.
+		///
+		/// The CT response unblocks the relay chain's first data stage. A relay chain rewound to
+		/// `Scheduled` re-runs the handshake, which is accepted here without a matching rewind.
+		#[pallet::call_index(0)]
+		#[pallet::weight(T::DbWeight::get().reads_writes(3, 2))]
```
cirko33, line 144: Do we need here actual weights and should they be against `CtMigrationStage`? 
If it intended for the future that's okay.

Ank4n: In this case its pretty straightforward so hard coding is probably fine, but yes we should look back and proper benchmark these pallets once all flows are in.

relay/polkadot/src/lib.rs
```diff
@@ -1948,6 +1952,10 @@ construct_runtime! {
 		// The pallet must be located below `MessageQueue` to get the XCM message acknowledgements
 		// from Asset Hub before we get the `RcMigrator` `on_initialize` executed.
 		RcMigrator: pallet_rc_migrator = 255,
+		// AHM v2 migrator. Below `MessageQueue` for the same reason as `RcMigrator`: its
+		// `on_initialize` has to see the block's inbound messages.
+		#[cfg(all(feature = "ahm-v2", not(feature = "on-chain-release-build")))]
+		Rc2Migrator: pallet_rc2_migrator = 254,
```
cirko33, line 1958: nit: would move it up before RcMigrator, but it's okay here as well
applies to all changed construct runtimes

cirko33 (approved):
Overall look good, only nits from my side
CI should be looked at by someone who knows more about it

cirko33 (approved):
Overall looks good to me, only nits. 
Just for CIs, should be looked at by someone more familiar with fellowship CI.

pallets/rc2-migrator/src/lib.rs
```diff
@@ -85,8 +173,131 @@ pub mod pallet {
...
+			Ok(())
+		}
+
+		/// The Coretime chain confirms that it can receive migrated state.
+		///
+		/// Sent by `pallet-ct-migrator` in response to [`CtMigratorCall::StartMigration`]. Nothing
+		/// is drained before it arrives.
+		#[pallet::call_index(2)]
+		#[pallet::weight(T::DbWeight::get().reads_writes(2, 1))]
+		pub fn ct_ready(origin: OriginFor<T>) -> DispatchResult {
```
sigurpol, line 314: in the hackmd, you were mentioning a warm-up / handshake phase:
```
RC tells CT to lock down and waits for an ack before any data moves. Allows time to drain any XCM queue, on-demand coretime calls, etc.
```
Are we dropping it  / am I missing it ?

pallets/rc2-migrator/src/lib.rs
```diff
@@ -85,8 +173,131 @@ pub mod pallet {
...
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Schedule the migration to begin at `start`.
+		///
+		/// The only way out of [`MigrationStage::Pending`] other than `force_set_stage`.
+		#[pallet::call_index(0)]
+		#[pallet::weight(T::DbWeight::get().reads_writes(2, 1))]
+		pub fn schedule_migration(origin: OriginFor<T>, start: MomentOf<T>) -> DispatchResult {
+			ensure_root(origin)?;
```
sigurpol: is it intentional to have ensure_root instead of an AdminOrigin (here and in other places)? The hackmd note was hinting at governance + manager driving so probably better to settle on ensure_root vs AdminOrigin now rather than later

pallets/ct-migrator/src/lib.rs
```diff
@@ -72,8 +115,98 @@ pub mod pallet {
...
+		pub fn start_migration(origin: OriginFor<T>) -> DispatchResult {
+			// relay chain origin converts to root.
+			ensure_root(origin)?;
```
sigurpol, line 147: same  as the other comment around ensure_origin vs AdminOrigin (that seemed the original plan in the hackmd)

pallets/rc2-migrator/src/lib.rs
```diff
@@ -85,8 +173,131 @@ pub mod pallet {
...
+			ensure!(
```
sigurpol, line 282: how do I cancel and reschedule (see hackmd)?

sigurpol (commented):
Asking some preliminary DQ vs hackmd plan before proceeding further 😄

pallets/rc2-migrator/src/lib.rs
Ank4n: By handshake I meant: RC::WaitingForCt → CT start_migration → RC::ct_ready. We don't start migration (moving data) until CT responds.

What's missing is the lockdown logic: call filters based on the migration stage, then waiting for the XCM queues to drain. It will be in a follow-up PR; also adding to the followup TODO list in the description so its tracked.

pallets/rc2-migrator/src/lib.rs
Ank4n: I wanted to add the local multisig-manager + governance-appointed manager origin in a separate PR, keeping almost exactly the same logic as AHM v1 (see [old rc-migrator](https://github.com/Ank4n/runtimes/blob/985df25829b3385730ff66acc50161ac57f0692c/pallets/rc-migrator/src/lib.rs#L1005)). Then AdminOrigin, the manager multisig (which resolves to manager origin), or the manager account can all drive it.

I think I should add AdminOrigin here already (or the full thing if you'd prefer). Just trying to break this into reviewable pieces and not sure where to draw the line 🤦

pallets/rc2-migrator/src/lib.rs
Ank4n: There will be cancel migration call. Makese sense to add here. 
(Adding)

pallets/rc2-migrator/src/lib.rs
sigurpol: Agree it's hard to draw the line :sweat_smile:  my 2c:  add type `AdminOrigin` to both pallets in this PR and replace all `ensure_root` sites. And defer the manager account, the multisig and the ensure_admin_or_manager helper for a self-contained PR later on. But it's all very debatable. I am also fine if you leave things as they are now and add a  TODO(ahmv2) or similar  around so it's easy to track in followup PRs.

pallets/rc2-migrator/src/lib.rs
sigurpol: thanks for the clarification - resolving (and yes, checklist in PR body will help as well)

pallets/ct-migrator/src/lib.rs
sigurpol: discussed here already https://github.com/polkadot-fellows/runtimes/pull/1285#discussion_r4016666276 - closing here

relay/kusama/src/ahm_v2.rs
```diff
@@ -0,0 +1,59 @@
...
+use crate::{
+	xcm_config::{Broker, XcmRouter},
+	BlockNumber, BrokerId, Runtime, RuntimeEvent,
+};
+use frame_support::{parameter_types, traits::Equals};
+use kusama_runtime_constants::time::MINUTES;
+use pallet_xcm::EnsureXcm;
+
+parameter_types! {
+	/// Manual verification window between the last data stage and finishing.
```
sigurpol: in the same spirit of other comments vs something  to be implemented yet,  should  we add a `TODO(ahmv2)` or similar for anything that will need a final pass before landing ?

integration-tests/ahmv2/src/tests.rs
```diff
@@ -0,0 +1,266 @@
...
+		assert_ok!(pallet_ct_migrator::Pallet::<network::ct::Runtime>::start_migration(
+			network::ct::RuntimeOrigin::root(),
+		));
+		take_ump::<CoretimePara>()
+	});
+	assert!(!ump.is_empty(), "the Coretime chain queued no answer to copy");
+
+	// WHEN another parachain sends that same message.
+	rc.execute_with(|| {
+		enqueue_ump(IMPOSTOR_PARA.into(), ump);
```
sigurpol: Since 4242 is not a system para, IIUC the RC barrier (`IsChildSystemParachain`) refuses this message before `Transact` runs.  Meaning that the test tests the barrier, but  not the `CtOrigin = EnsureXcm<Equals<CoretimeLocation>>` wiring of `ct_ready`.  Should we test also vs a system para id too (or only)?

integration-tests/ahmv2/src/mock.rs
```diff
@@ -0,0 +1,403 @@
...
+	let name = chain.name();
+	let now = frame_system::Pallet::<T>::block_number() + One::one();
+	log::debug!(target: chain.log_target(), "Executing block: {now:?}");
+	frame_system::Pallet::<T>::set_block_number(now);
+	frame_system::Pallet::<T>::reset_events();
+	let weight = hooks(now);
+
+	// A message the executor refused outright, such as one the barrier turned away, is discarded
+	// as `ProcessingFailed`; one that started executing and then errored is `Processed` with
+	// `success: false`.
```
sigurpol: Linked to https://github.com/polkadot-fellows/runtimes/pull/1285/changes#r4017769736: are we returning `Processed { success: false }` also for a `Transact` whose call returns `Err` or only for XCM-level errors?  To be checked by some XCM wizards (@acatangiu ? @franciscoaguirre ?) but a `Transact` whose call returns an `Err`, might not be an instruction error in this executor so the instruction would return `Ok`, and the error would be stored in `transact_status`. So we might end up receiving `Processed { success: true }` . If this holds true, we would miss a migrator call that errored on the receiving chain. One way to verify is probably to run the impostor test with a system parachain and check.

pallets/rc2-migrator/src/lib.rs
Ank4n: so I added bunch of TODO(ahm-v2) as well as just admin origin and manager account (w/o the multisig).

pallets/rc2-migrator/src/lib.rs
Ank4n: Added

relay/kusama/src/ahm_v2.rs
Ank4n: Yes, added more TODO(ahm-v2) markers so we can make sure they're all handled before wiring up the runtimes.

relay/polkadot/src/lib.rs
Ank4n: index wise makes sense but logically rc2 should be below rc1, no? :)

integration-tests/ahmv2/src/tests.rs
Ank4n: Right. 4242 gets rejected before Transact.
Rewritten to cover both: 4242 for the barrier path, and AH for the origin path, where the message is admitted and CtOrigin is the thing refusing the call.

integration-tests/ahmv2/src/tests.rs
```diff
@@ -0,0 +1,289 @@
...
+		enqueue_ump(OUTSIDER_PARA.into(), ump.clone());
+		next_block_rc_expecting_rejection();
+		assert_eq!(rc_stage(), RcStage::WaitingForCt);
+	});
+
+	// WHEN a system para sends it. THEN the barrier admits the message and `Transact` runs, so
+	// this is `CtOrigin` refusing the call rather than the barrier refusing the message.
+	rc.execute_with(|| {
+		enqueue_ump(SYSTEM_IMPOSTOR_PARA.into(), ump);
+		next_block_rc();
```
sigurpol: the snapshot might have in relay queue already other UMP messages for this paraid ahead of this one in theory . Can we somehow also assert that the message was processed or that a Processed event for the related  origin has been emitted or something like that?

pallets/rc2-migrator/src/lib.rs
```diff
@@ -85,8 +193,217 @@ pub mod pallet {
...
+
+		/// Send a `pallet-ct-migrator` call to the Coretime chain.
+		fn send_to_ct(call: CtMigratorCall) -> Result<(), Error<T>> {
+			let call = CtRuntimeCall::CtMigrator(call);
+			// `Superuser` converts to Root on the Coretime chain; the receiving calls check for
+			// Root.
+			let message = Xcm(vec![
+				UnpaidExecution { weight_limit: WeightLimit::Unlimited, check_origin: None },
+				Transact {
+					origin_kind: OriginKind::Superuser,
```
sigurpol, line 465: related to https://github.com/polkadot-fellows/runtimes/pull/1285#discussion_r4017840507

IIUC, a failed dispatch inside `Transact` completes the XCM and the queue reports `Processed { success: true }`, so a migrator call that errors on the other chain is invisible on-chain or to the harness.  Should we consider adding `ExpectTransactStatus(MaybeErrorCode::Success)` after the Transact? This way, if I get it right, a failed dispatch inside Transact turns into a  `Processed { success: false }` and we should be good.

pallets/rc2-migrator/src/lib.rs
```diff
@@ -85,8 +193,217 @@ pub mod pallet {
...
+
+			WarmUpPeriod::<T>::put(warm_up);
+			CoolOffPeriod::<T>::put(cool_off);
+			Self::transition(MigrationStage::Scheduled { start });
+			Ok(())
+		}
+
+		/// Set the migration stage directly.
+		///
+		/// Root-only escape hatch for a lost message or a stage that needs re-running.
```
sigurpol: it's not Root-only anymore 😄

pallets/rc2-migrator/src/lib.rs
```diff
@@ -85,8 +193,217 @@ pub mod pallet {
...
+				// TODO(ahm-v2): the data stages run from here, once they exist.
+				MigrationStage::WarmUp { end_at } if now >= end_at => {
+					let end_at = now + CoolOffPeriod::<T>::get();
+					Self::transition(MigrationStage::CoolOff { end_at });
+					T::DbWeight::get().reads_writes(2, 2)
+				},
+				// wait cool off period before finishing migration
+				MigrationStage::CoolOff { end_at } if now >= end_at => {
+					if Self::send_to_ct(CtMigratorCall::EndLockdown).is_ok() {
+						Self::transition(MigrationStage::MigrationDone);
```
sigurpol, line 443: DQ: are we fine with RC entering MigrationDone independently from what CT says? (E.g. CT not in `DataMigrationOngoing` would refuse it and recover would be to `force_set_stage` on CT or similar). Probably yes - just asking for a friend 😅

pallets/rc2-migrator/src/tests.rs
```diff
@@ -0,0 +1,619 @@
...
+
+		// THEN the machine has not moved, nothing was sent, and no event was emitted.
+		assert_stage(Stage::Pending);
+		assert_eq!(sent().len(), 0);
+		assert_eq!(transitions(), vec![]);
+	});
+}
+
+#[test]
+fn only_root_can_schedule_and_only_into_the_future() {
```
sigurpol: not only root now

pallets/rc2-migrator/src/lib.rs
Ank4n: Great suggestion. Added.

integration-tests/ahmv2/src/mock.rs
Ank4n: You were right ofc.

Fixed by your suggestion to add a `ExpectTransactStatus(MaybeErrorCode::Success)` instruction after `Transact`.

integration-tests/ahmv2/src/tests.rs
Ank4n: I thought I was draining the queue but apparently not.

Fixed now: the test first drains the queue, and ump_outcome(para) runs blocks until the mq reports on that para's origin

pallets/rc2-migrator/src/lib.rs
Ank4n: Yeah I think its fine here. RC is the authority/driver anyway. If CT rejects, it means something arrived out of order, and `force_set_stage` is the way to go.

For data stages actually, we should have a `ReportTransactStatus` so if something fails, migration pauses.

pallets/rc2-migrator/src/lib.rs
Ank4n: fixed

pallets/rc2-migrator/src/tests.rs
Ank4n: fixed

pallets/rc2-migrator/src/lib.rs
```diff
@@ -85,8 +194,227 @@ pub mod pallet {
...
+			Self::transition(MigrationStage::Scheduled { start });
+			Ok(())
+		}
+
+		/// Set the migration stage directly.
+		///
+		/// Escape hatch for a lost message or a stage that needs re-running.
+		#[pallet::call_index(1)]
+		#[pallet::weight(T::DbWeight::get().reads_writes(1, 1))]
+		pub fn force_set_stage(origin: OriginFor<T>, stage: MigrationStageOf<T>) -> DispatchResult {
```
Kanasjnr, line 300: `force_set_stage `doesn't validate stage at all, and `progress_migration `only knows how to move Scheduled → WarmUp → CoolOff forward, everything else (the _ arm) just no-ops:
 https://github.com/polkadot-fellows/runtimes/blob/9f4440fe79a52c56e07aea5f4f2cb4d08675fb7c/pallets/rc2-migrator/src/lib.rs#L357-L386

So if this ever gets pointed at `AccountsInit`..`TiCorrection`, it silently sits there forever, no error, nothing. Worth a guard, or at least a comment ? 😅

pallets/rc2-migrator/src/lib.rs
```diff
@@ -85,8 +194,227 @@ pub mod pallet {
...
+	pub type WarmUpPeriod<T: Config> = StorageValue<_, BlockNumberFor<T>, ValueQuery>;
+
+	/// How long [`MigrationStage::CoolOff`] holds, as the scheduled migration set it.
+	#[pallet::storage]
+	pub type CoolOffPeriod<T: Config> = StorageValue<_, BlockNumberFor<T>, ValueQuery>;
+
+	/// An account that may drive the migration alongside [`Config::AdminOrigin`], except that it
+	/// cannot appoint a manager itself.
+	#[pallet::storage]
+	pub type Manager<T: Config> = StorageValue<_, T::AccountId, OptionQuery>;
```
Kanasjnr, line 206: this is just an `AccountId` for now, not the multisig the TODO mentions 
https://github.com/polkadot-fellows/runtimes/blob/9f4440fe79a52c56e07aea5f4f2cb4d08675fb7c/pallets/rc2-migrator/src/lib.rs#L345

so `Manager` already has full root-equivalent power today, incl. skipping warm-up/cool-off via `force_set_stage`. 

Assuming that's tracked before this ever ships for real, just double-checking 🙂

integration-tests/ahmv2/src/tests.rs
```diff
@@ -0,0 +1,355 @@
...
+		next_block_para::<CoretimePara>();
+		assert_eq!(ct_stage(), pallet_ct_migrator::MigrationStage::MigrationDone);
+
+		// Nothing was migrated, so nothing was minted here.
+		assert_eq!(
+			pallet_balances::Pallet::<network::ct::Runtime>::total_issuance(),
+			ct_issuance_before,
+			"a migration with no data stages must not change Coretime issuance"
+		);
+	});
```
Kanasjnr, line 248: i just want to ask to understand the scope,this only drives `MessageQueue` + `Rc2Migrator` hooks, nothing else (staking/session/xcmp/etc). 

Totally fine for the handshake here, i was just wondering if this grows once the real data stages land, or if there's a beefier test planned for that.

Kanasjnr (commented):
Really solid work ,i just a few questions 🙂

pallets/rc2-migrator/src/lib.rs
Ank4n: Thanks for reviewing. The data stages are to be filled in follow up PRs, but now I have added all stages in the progress migration as no-op steps, so `force_set_stage` won't take it into a stranded stage.

pallets/rc2-migrator/src/lib.rs
Ank4n: Yeah, I will have multisig in a separate PR. Before this ships, we should resolve all `TODO(ahm-v2) markers`.

integration-tests/ahmv2/src/tests.rs
Ank4n: As we add data stages, we should extend/refactor the tests. But for the hooks itself (mq + rc2migrator) , they should be enough as staking / session / xcmp are untouched by this migration.

sigurpol (approved):
LGTM!
I have some minor nits / open points/ DQ but not relevant enough to block this PR, just mentioning here so you can see if it makes sense to consider or ignore them.

1.  A pause doesn't extend the window so end_at stays fixed when a given stage is entered. E.g if I decide to Pause during CoolOff and resume later on, Done fires at the same block as if I didn't pause (IIUC). Might be fine, but maybe worth documenting somehow or adding a comment around resume_migration saying that  we need to rearm with force_set_stage (e.g. `force_set_stage(CoolOff {new end_at})` or something like that)
2. `Paused` can outlive the migration if while in paused, I force to Pending or MigrationDone and I forget to resume. Should we refuse a `schedule_migration` if paused?  Or should we clear Pause when we call force_set_stage?

Ank4n:
> A pause doesn't extend the window so end_at stays fixed when a given stage is entered. E.g if I decide to Pause during CoolOff and resume later on, Done fires at the same block as if I didn't pause (IIUC). Might be fine, but maybe worth documenting somehow or adding a comment around resume_migration saying that we need to rearm with force_set_stage (e.g. force_set_stage(CoolOff {new end_at}) or something like that)

Good point. Documented it.

> Paused can outlive the migration if while in paused, I force to Pending or MigrationDone and I forget to resume. Should we refuse a schedule_migration if paused? Or should we clear Pause when we call force_set_stage?

changed. So pause only persists while data stages are ongoing. If we set stage to `Pending`, `Scheduled` or `MigrationDone`, it is cleared.

diff: https://github.com/polkadot-fellows/runtimes/pull/1285/changes/ce690492a1f85bd28edbe90fb4fb2661bba8febe

cc: @sigurpol

sigurpol (approved):

pallets/rc2-migrator/src/lib.rs
```diff
@@ -76,17 +194,425 @@ pub mod pallet {
...
+		///   transition to the cool-off stage.
+		///
+		/// Read [`MigrationStage::Scheduled`] documentation for more details.
+		#[pallet::call_index(0)]
+		#[pallet::weight(T::DbWeight::get().reads_writes(2, 3))]
+		pub fn schedule_migration(
+			origin: OriginFor<T>,
+			start: MomentOf<T>,
+			warm_up: BlockNumberFor<T>,
+			cool_off: BlockNumberFor<T>,
```
muharem: should not be `warm_up` and `cool_off` relative to the `start`? can it be accurately relative if they use different measures?

pallets/rc2-migrator/src/lib.rs
```diff
@@ -76,17 +194,425 @@ pub mod pallet {
...
+					T::DbWeight::get().reads_writes(2, 2)
+				},
+				// wait cool off period before finishing migration
+				MigrationStage::CoolOff { end_at } if now >= end_at => {
+					if Self::send_to_ct(CtMigratorCall::EndLockdown).is_ok() {
+						Self::transition(MigrationStage::MigrationDone);
+					}
+					T::DbWeight::get().reads_writes(3, 3)
+				},
+				_ => T::DbWeight::get().reads(1),
```
muharem, line 582: explicit handle for Pending and Done would be a bit safer

pallets/ct-migrator/src/lib.rs
```diff
@@ -64,16 +121,187 @@ pub mod pallet {
...
+
+		/// Set the manager account id.
+		///
+		/// The manager has the similar to [`Config::AdminOrigin`] privileges except that it
+		/// can not set the manager account id via `set_manager` call. Each chain's manager is set
+		/// on that chain; the Relay Chain does not carry one over.
+		#[pallet::call_index(3)]
+		#[pallet::weight(T::DbWeight::get().reads_writes(1, 1))]
+		pub fn set_manager(origin: OriginFor<T>, new: Option<T::AccountId>) -> DispatchResult {
+			T::AdminOrigin::ensure_origin(origin)?;
```
muharem, line 247: can be root too at this level, like for `start_migration` above to not forget

system-parachains/coretime/coretime-kusama/src/lib.rs
```diff
@@ -716,6 +719,10 @@ construct_runtime!(
 
 		// The main stage.
 		Broker: pallet_broker = 50,
+
+		// AHM v2 migrator.
+		#[cfg(all(feature = "ahm-v2", not(feature = "on-chain-release-build")))]
+		CtMigrator: pallet_ct_migrator = 100,
```
muharem, line 725: can be something like 255

.github/workflows/test.yml
```diff
@@ -229,6 +229,98 @@ jobs:
...
+          RUSTFLAGS: "-C debug-assertions"
+          # The test binary runs with its package as the working directory, so these must be
+          # absolute.
+          SNAP_RC: ${{ github.workspace }}/snap_rc.snap
+          SNAP_CT: ${{ github.workspace }}/snap_ct.snap
+          NETWORK: ${{ matrix.network }}
+
+      - name: Report a missing snapshot
+        if: steps.snap-rc.outputs.cache-matched-key == '' || steps.snap-ct.outputs.cache-matched-key == ''
+        run: echo "::notice::No cached ${{ matrix.network }} snapshot; skipped the AHM v2 snapshot suite"
```
muharem, line 322: just a heads-up: CI is green even if the snapshot is missing.

relay/kusama/Cargo.toml
```diff
@@ -12,6 +12,8 @@ version.workspace = true
 
 # Asset Hub Migration concerning deps
 pallet-rc-migrator = { workspace = true }
+# Behind the `ahm-v2` feature; see `src/ahm_v2.rs`.
+pallet-rc2-migrator = { workspace = true }
```
muharem, line 16: The migrator pallets are non-optional deps of the four runtimes, so they are compiled into every build even when ahm-v2 is off (not included into build, but every time compiled). you can make them optional.

integration-tests/ahmv2/Cargo.toml
```diff
@@ -0,0 +1,53 @@
...
+sp-io = { workspace = true, default-features = true }
+sp-runtime = { workspace = true, default-features = true }
+sp-storage = { workspace = true, default-features = true }
+sp-tracing = { workspace = true, default-features = true }
+tokio = { features = ["macros", "rt-multi-thread", "sync"], workspace = true }
+xcm = { workspace = true, default-features = true }
+
+# Runtimes for one network are compiled per build; see the `polkadot` and `kusama` features.
+coretime-polkadot-runtime = { workspace = true, optional = true, features = ["ahm-v2"] }
+polkadot-runtime = { workspace = true, optional = true, features = ["ahm-v2"] }
```
muharem, line 34: heads up: bare cargo build at root, or --workspace, will build runtimes builds with pallet included.

muharem (approved):

pallets/rc2-migrator/src/lib.rs
Ank4n: done here https://github.com/polkadot-fellows/runtimes/pull/1285/commits/f9d57e78ff37d6b292d1e9c053c486d8c8b8ced3

pallets/ct-migrator/src/lib.rs
Ank4n: made it consistent https://github.com/polkadot-fellows/runtimes/pull/1285/changes/bf9b54b8f1a6870c736451997515e7937b2cdae9

system-parachains/coretime/coretime-kusama/src/lib.rs
Ank4n: done: https://github.com/polkadot-fellows/runtimes/pull/1285/commits/fd77911854b44d80e4b6ac547c96dcac87af1471

.github/workflows/test.yml
Ank4n: Made it fail on missing snap: https://github.com/polkadot-fellows/runtimes/pull/1285/changes/62d40f22853ccec478db4568fd8f1d527aaafd11

integration-tests/ahmv2/Cargo.toml
Ank4n: dropped it from default

Ank4n: https://github.com/polkadot-fellows/runtimes/pull/1285/changes/19d2e704fda3ded41d9ae7efb0eec1103a55dd5d

cirko33 (approved):

sigurpol (approved):

relay/kusama/Cargo.toml
Ank4n: I tried this but it doesn't work. With `optional = true` and `ahm-v2 = ["pallet-ct-migrator"]` (and `pallet-ct-migrator?/std` etc.), `cargo check -p coretime-polkadot-runtime` (w/o `ahm-v2`) fails,

```
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `pallet_ct_migrator`
    |         CtMigrator: pallet_ct_migrator = 255,
```

so a `#[cfg]`-gated entry inside `construct_runtime!` still needs the crate to resolve. Same for the relay side. Afaict v1 didn't make `pallet-rc-migrator` optional either.

seadanda (approved):

Issue #1283, opened 2026-09-08
Mint only on AssetHub and redirect all system-chain burns to AH  (burn on Kusama, forward to DAP on Polkadot)


sigurpol:
**Target:  Asset Hub's `TotalIssuance` is the network total and AH's XCM checking account must equal the sum of the other chains' `TotalIssuance`**

This is not meant to be a new requirement but aims to consolidate what AHM and XCM checking account aimed to do already.

A teleport out of Asset Hub does not lower its `TotalIssuance`: the amount moves to the `py/xcmch` checking account (an ordinary keyless account on AH whose balance is already inside AH's TI), which records what lives on other chains. [^1]
**The checking account tracks the overall tokens in RC + systemchains != AH <=>   only AH mints and burns.**

A burn on any other system chain lowers that chain's total only. Asset Hub's total and checking account never move, so the AH's TI  is overstated. This is currently the case (e.g. Coretime's bulk revenue burns on RC before #1282 ). [^2]  
This has also some impact on staking:  Asset Hub Kusama's `EraPayout` takes AH's TI   as `total_stakable` in its issuance curve - which doesn't match the real TI if other chains burn.

Reserve transfers do not affect this. DOT reserve-transferred from Asset Hub stays on Asset Hub in the destination parachain's sovereign account. A parachain that trusts another system chain as its DOT reserve leaves the DOT in its sovereign account on that chain. In both cases the DOT stays in some chain's `TotalIssuance` and remains covered by the checking account. Only native mints and burns on chains other than Asset Hub move the figures apart.

## Rules 
In order to meet the target, we enforce the following rules:

-  No chain other than Asset Hub mints.
- No chain other than Asset Hub burns. 
  - Polkadot: every sink is redirected to the DAP staging account through `pallet-accumulate-and-forward` (aligned with [DAP's WFC](https://polkadot.subsquare.io/referenda/1827))
  - Kusama: every sink is teleported to Asset Hub and burnt there.

Once the rules are enforced, **a reconciliation step is needed** since both on Polkadot and Kusama we have burnt / are burning on chains != AH. [^3] so today the real supply is Asset Hub's `TotalIssuance` minus whatever other chains burnt without Asset Hub knowing plus whatever other chains minted w/o AH knowing.

## Tasks

We have created in the past  few issues spread in runtime repo and SDK. **TODO**: link / aggregate all of them under this umbrella task

[...]

PR #1287, opened 2026-09-11
Kusama: teleport dust to Asset Hub and burn it where issuance is tracked

Stephenlawrence00:
Bridge Hub, Coretime, People and Encointer Kusama route `DustRemoval` into a
`pallet-accumulate-and-forward` account and teleport it to Asset Hub to be burned there, so a
burn on a system chain moves the network `TotalIssuance` that Asset Hub tracks (#1283).

## Checklist
- [x] Accounts funded (`F3opxRYLmNnYbWQsbYMryaRxM7PeLKRhKEorQVeDeF15KKy`, one per chain)
- [x] Benchmarks re-run
- [x] Coretime Kusama follow-up tracked in #1290, linked to #1283
- [x] Created a followup task https://github.com/paritytech/polkadot-sdk/issues/13238  Once SDK, the copies in polkadot-fellows/runtimes can be deleted and the Kusama relay and system chains use the same type as Polkadot's.


sigurpol:
Thanks for the PR :bow: ! Will review on Monday. Couple of notes from a very quick initial look:
1. migrations jobs are failing and it's expected until we fund acf/ksmt (once we agree on the approach, we can do manually outside the PR as we did for DAP buffer and staging account or  the coretime acf account)
2. coretime kusama  ends up with two teleport-and-burn paths. I would retire py/ctbrn and use AccumulateAndForward as only way (routing OnRevenue to AccumulateAndForward). Can also be a follow-up PR but must be tracked under #1283 in case
3.  we need to re-run benchmarks - maybe to be added in a TODO section in PR body so we don't forget

sigurpol:
cc @Ank4n , @andreitrand , @cirko33

Stephenlawrence00:
> Thanks for the PR 🙇 ! Will review on Monday. Couple of notes from a very quick initial look:

Thanks! Addressed 1 and 3; 2 will be a follow-up.

> 1. migrations jobs are failing and it's expected until we fund acf/ksmt (once we agree on the approach, we can do manually outside the PR as we did for DAP buffer and staging account or  the coretime acf account)

Happy to fund them unless you'd rather own it. Same address on all four chains, `F3opxRYLmNnYbWQsbYMryaRxM7PeLKRhKEorQVeDeF15KKy`, each needing the system-parachain ED of 33,333,333 planck; simplest route is four teleports from Asset Hub with that account as beneficiary. (Derivation sanity-checked by reproducing #1282's published `acf/dott` address.) Note #1289 reuses the same pallet ID, so the relay's account is the same address at the relay ED.

sigurpol:
> > Thanks for the PR 🙇 ! Will review on Monday. Couple of notes from a very quick initial look:
> 
> Thanks! Addressed 1 and 3; 2 will be a follow-up.
> 
> > 1. migrations jobs are failing and it's expected until we fund acf/ksmt (once we agree on the approach, we can do manually outside the PR as we did for DAP buffer and staging account or  the coretime acf account)
> 
> Happy to fund them unless you'd rather own it. Same address on all four chains, `F3opxRYLmNnYbWQsbYMryaRxM7PeLKRhKEorQVeDeF15KKy`, each needing the system-parachain ED of 33,333,333 planck; simplest route is four teleports from Asset Hub with that account as beneficiary. (Derivation sanity-checked by reproducing #1282's published `acf/dott` address.) Note #1289 reuses the same pallet ID, so the relay's account is the same address at the relay ED.

Great, feel free to go ahead and fund them (I would suggest slightly more than ED to account for  destination execution fee, and avoid that a transfer of exactly ED lands below ED and the deposit fails). You can probably ask these KSM to be refunded, although I am not sure of the best procedure (other fellows might help here, @kianenigma  maybe?).
Will review #1289 as well

sigurpol:
> Bridge Hub, Coretime, People and Encointer Kusama accumulate the imbalances they used to burn locally and teleport them to Asset Hub to be burned there, so that a burn on a system chain actually moves the network TotalIssuance that Asset Hub tracks (https://github.com/polkadot-fellows/runtimes/issues/1283).

I think we can improve PR body in couple of ways to ease the review / track the progress:
1. [niptpick] Say just DustRemoval rather than "the imbalances they used to burn locally"
2. Add a checklist:  
- [ ] account funded
- [ ] benchmarks re-run
- [ ]  Coretime Kusama follow-up tracked in <issue to be created and linked to #1283 >

I will run benchmarks  rn

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- Bridge Hub, Coretime, People and Encointer Kusama: the sinks that used to burn locally now accumulate and are teleported to Asset Hub to be burned there ([#1287](https://github.com/polkadot-fellows/runtimes/pull/1287)).
```
sigurpol:
```suggestion
- Bridge Hub, Coretime, People and Encointer Kusama:  dust from reaped accounts now accumulates and is teleported to Asset Hub to be burned there ([#1287](https://github.com/polkadot-fellows/runtimes/pull/1287)).
```

sigurpol:
/cmd bench --runtime bridge-hub-kusama --pallet pallet_accumulate_and_forward

sigurpol:
/cmd bench --runtime coretime-kusama --pallet pallet_accumulate_and_forward

sigurpol:
/cmd bench --runtime people-kusama --pallet pallet_accumulate_and_forward

sigurpol:
/cmd bench --runtime encointer-kusama --pallet pallet_accumulate_and_forward

integration-tests/emulated/tests/people/people-kusama/src/tests/burn.rs
```diff
@@ -0,0 +1,102 @@
...
+
+use crate::*;
+use frame_support::traits::{
+	fungible::{Inspect as FungibleInspect, Mutate as FungibleMutate},
+	Hooks,
+};
+
+/// Accumulated KSM is teleported to Asset Hub and burned there, where the network's
+/// `TotalIssuance` is tracked.
+#[test]
```
sigurpol: we are doing e2e of People only. We should e2e test also the other chains. Even if we use the same forwarder, each runtime does its own wiring (`Dest`,  asset location, `XcmConfig`...) so UTs wouldn't catch a wrong wiring. Try to avoid code duplication as much as possible -maybe an helper in the emulated common crate used by all chains would do, mirroring what People tests here are doing. Wdyt?

system-parachains/common/src/accumulate_and_forward.rs
```diff
@@ -0,0 +1,147 @@
...
+		frame_support::ensure!(
+			Self::is_funded(),
+			"accumulation account is not funded with the existential deposit"
+		);
+		Ok(())
+	}
+}
+
+/// [`pallet_accumulate_and_forward::Forwarder`] teleporting to Asset Hub to burn there, since
+/// Kusama tracks `TotalIssuance` on Asset Hub. A failed send rolls back; once queued, a failure
```
sigurpol: Can you check if it's always true that a failure on arrival traps the assets?  It is if `BurnAsset` fails, but is it the case for ReceiveTeleportedAsset failure or barrier rejection? Or wouldn't we end up with KSM gone locally and AH checking account unchanged?

integration-tests/emulated/tests/people/people-kusama/src/tests/burn.rs
Stephenlawrence00: Agreed, done. Added `test_accumulated_funds_are_burnt_on_asset_hub!` to `integration-tests/emulated/helpers`, invoked from all four chains. Each call passes only its own chain, Asset Hub and the two EDs, so `Dest`, the asset location and `XcmConfig` are exercised per runtime. A macro rather than a generic fn since the emulated types are concrete per chain.

Stephenlawrence00:
> > Bridge Hub, Coretime, People and Encointer Kusama accumulate the imbalances they used to burn locally and teleport them to Asset Hub to be burned there, so that a burn on a system chain actually moves the network TotalIssuance that Asset Hub tracks (#1283).
> 
> I think we can improve PR body in couple of ways to ease the review / track the progress:
> 
> 1. [niptpick] Say just DustRemoval rather than "the imbalances they used to burn locally"
> 2. Add a checklist:
> 
> * [ ]  account funded
> * [ ]  benchmarks re-run
> * [ ]   Coretime Kusama follow-up tracked in <issue to be created and linked to [Mint only on AssetHub and redirect all system-chain burns to AH  (burn on Kusama, forward to DAP on Polkadot) #1283](https://github.com/polkadot-fellows/runtimes/issues/1283) >
> 
> I will run benchmarks rn

Both done. The description now says DustRemoval and carries the checklist. Thanks for running the benchmarks!

system-parachains/common/src/accumulate_and_forward.rs
Stephenlawrence00: You're right, the comment was wrong. Corrected.

`post_process` only traps when holding is non-empty, so `BurnAsset` failing traps but a barrier rejection or failed `ReceiveTeleportedAsset` traps nothing. With the checking account set to `()`, `withdraw_asset` just burns locally, so in those cases the KSM is gone here with Asset Hub untouched, exactly as you describe.

Both need Asset Hub to stop trusting the chain as a teleporter, which the new per chain e2e tests pin down. Same exposure as `burn_at_asset_hub` in #1282, so not new, but worth stating accurately.

integration-tests/emulated/helpers/Cargo.toml
```diff
@@ -12,6 +12,9 @@ paste = { workspace = true }
 hex-literal = { workspace = true }
 
 # Substrate
+frame-support = { workspace = true, default-features = true }
```
sigurpol, line 15: to make zepter happy, you need to propagate also

```
"frame-support/runtime-benchmarks",
"frame-system/runtime-benchmarks",
"pallet-accumulate-and-forward/runtime-benchmarks"
```

in the `runtime-benchmarks` list in this file

integration-tests/emulated/helpers/src/common.rs
```diff
@@ -30,3 +30,96 @@ pub mod snowbridge {
 		pub EthLocation: Location =  Location::new(2, [GlobalConsensus(EthereumNetwork::get())]);
 	}
 }
+
+/// Asserts that a chain's accumulated funds are teleported to Asset Hub and burned there.
```
sigurpol: This describes Kusama's forwarder, not a property of accumulated funds. Polkadot chains forward to the DAP staging account instead, and Coretime Polkadot covers already that with the SDK's `dap_helpers::test_accumulate_forward_transfers_to_asset_hub`. Since this crate is shared by both networks, the doc  should say at least that this is meant for Kusama-like chains imo. Name of the macro can be kept generic. The alternative would be an upstream SDK change so a burn sibling of `dap_helpers` in the SDK's emulated common crate, then this repo carries neither the forwarder nor the test helper.  
My 2c is that this rework (the change in the SDK) could/should be a followup and not a change to ask in this PR

integration-tests/emulated/helpers/Cargo.toml
Stephenlawrence00: Added all three to the runtime-benchmarks list, thanks

integration-tests/emulated/tests/people/people-kusama/Cargo.toml
```diff
@@ -13,9 +13,11 @@ codec = { workspace = true, default-features = true }
 # Substrate
 sp-runtime = { workspace = true, default-features = true }
 frame-support = { workspace = true, default-features = true }
+frame-system = { workspace = true, default-features = true }
```
sigurpol: do we need changes in this `Cargo.toml` at all now that test moved into shared macro?

Stephenlawrence00: Good catch, they were leftovers from the bespoke test. Removed both, plus their `runtime-benchmarks` entries. The macro reaches everything through `$crate::`, which the other three chains already demonstrate since they run it without those deps.

sigurpol:
@Stephenlawrence00 , I've funded F3opxRYLmNnYbWQsbYMryaRxM7PeLKRhKEorQVeDeF15KKy (acf/ksmt) on all system chains !=AH above ED - will retrigger the check migration job :popcorn:

sigurpol (approved):

system-parachains/common/src/accumulate_and_forward.rs
```diff
@@ -0,0 +1,148 @@
...
+
+/// Fails `try-runtime` if the accumulation account lacks the ED, without which dust is burned
+/// rather than accumulated. Fund it out of band, as for `acf/dott` in #1282.
+pub struct EnsureAccumulationAccountFunded<T>(PhantomData<T>);
+
+impl<T: pallet_accumulate_and_forward::Config> EnsureAccumulationAccountFunded<T> {
+	fn is_funded() -> bool {
+		use frame_support::traits::fungible::Inspect;
+
+		let account = pallet_accumulate_and_forward::Pallet::<T>::accumulation_account();
```
kianenigma, line 40: very JAM compatible naming :D

system-parachains/common/src/accumulate_and_forward.rs
```diff
@@ -0,0 +1,148 @@
...
+		<T as pallet_accumulate_and_forward::Config>::Currency::balance(&account) >=
+			<T as pallet_accumulate_and_forward::Config>::Currency::minimum_balance()
+	}
+}
+
+impl<T: pallet_accumulate_and_forward::Config> OnRuntimeUpgrade
+	for EnsureAccumulationAccountFunded<T>
+{
+	fn on_runtime_upgrade() -> Weight {
+		if !Self::is_funded() {
```
kianenigma, line 50: So before upgrade, some champion is going to fund the account? nice. The migration test should fail and at that point you will be reminded.

integration-tests/emulated/helpers/src/common.rs
```diff
@@ -30,3 +30,99 @@ pub mod snowbridge {
...
+
+			let accumulation_account = $chain::execute_with(|| {
+				pallet_accumulate_and_forward::Pallet::<ChainRuntime>::accumulation_account()
+			});
+			let amount = 2 * $chain::execute_with(|| {
+				<ChainRuntime as pallet_accumulate_and_forward::Config>::MinTransferAmount::get()
+			});
+			$chain::fund_accounts(vec![(accumulation_account.clone(), amount)]);
+
+			// The emulated chain minted that KSM itself, so top up the checking account.
```
kianenigma: Isn't it better to do happens in reality: 

* teleport from `$asset_hub` to `$chain`
* burn in `$chain`
* observe what happens next + what is the state of all checking accounts and TIs before + during + after each step

This also helps verify `TeleportAndBurnForwarder` which is the only meaty part.

kianenigma (commented):
suggest revisiting https://github.com/polkadot-fellows/runtimes/pull/1287/changes#r4014090960 otherwise LGTM

system-parachains/common/src/accumulate_and_forward.rs
sigurpol: I just did so now migration test should start succeeding (last famous words...)

sigurpol: or will do once we clear the cache (manually or wait for tomorrow's schedule to refresh the caches), snapshots otherwise we get this from stale snapshot:
```
2026-09-15T09:18:25Z ERROR runtime] panicked at /_work/ibp-ng/runtimes/runtimes/system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs:1481:65:
      called `Result::unwrap()` on an `Err` value: Other("accumulation account is not funded with the existential deposit")

  https://github.com/polkadot-fellows/runtimes/actions/runs/34840104206/job/104318963549?pr=1287
  ```

File: system-parachains/common/src/accumulate_and_forward.rs [hunk 10]
```diff
@@ -0,0 +1,148 @@
...
+			Junction::AccountId32 { network: None, id: source }.into_location();
+		let context = XcmContext { origin: None, message_id: [0; 32], topic: None };
+
+		let withdrawn = Transactor::<XcmConfig>::withdraw_asset(&asset, &source_location, None)?;
+
+		// Credit the checking account before the message goes out.
+		// TODO https://github.com/polkadot-fellows/runtimes/issues/404
+		Transactor::<XcmConfig>::can_check_out(&dest, &asset, &context)?;
+
+		let assets = withdrawn.reanchored_assets(
```
kianenigma, line 100: Not sure if relevant but if the inner `reanchor` fails here, this will return an empty set of assets. Probably you want to check the assets returned, and log a warning if it is empty. 

Probably some test can ensure the inner `reanchor` always works. I presume since chain locations don't change, if it works now, it will always work.

system-parachains/common/src/accumulate_and_forward.rs
sigurpol: cleared cache manually and retriggered CI :popcorn:

system-parachains/common/src/accumulate_and_forward.rs
```diff
@@ -0,0 +1,148 @@
...
+		);
+		Ok(())
+	}
+}
+
+/// [`pallet_accumulate_and_forward::Forwarder`] teleporting to Asset Hub to burn there, since
+/// Kusama tracks `TotalIssuance` on Asset Hub. A failed send rolls back. Once queued, only a
+/// `BurnAsset` failure traps the assets; a barrier rejection or a failed `ReceiveTeleportedAsset`
+/// leaves the KSM burned here with Asset Hub untouched.
+pub struct TeleportAndBurnForwarder<XcmConfig, Dest, NativeAsset>(
```
kianenigma, line 75: All in all what we are doing here is quite similar to `TeleportForwarderForAccountId32`, right? 

Except: 

* `TeleportForwarderForAccountId32` is doing a slightly better job with using higher level functions like `prepare_and_execute`
* Has a rollback if send fails 
* if we make its `remote_xcm` part configurable, won't it be possible to reuse it here?

system-parachains/common/src/accumulate_and_forward.rs
```diff
@@ -0,0 +1,148 @@
...
+			Self::is_funded(),
+			"accumulation account is not funded with the existential deposit"
+		);
+		Ok(())
+	}
+}
+
+/// [`pallet_accumulate_and_forward::Forwarder`] teleporting to Asset Hub to burn there, since
+/// Kusama tracks `TotalIssuance` on Asset Hub. A failed send rolls back. Once queued, only a
+/// `BurnAsset` failure traps the assets; a barrier rejection or a failed `ReceiveTeleportedAsset`
```
kianenigma: Elaborate bit more here please. `BurnAsset` as far as I can see is executed without any failures. So what could cause a trap?

system-parachains/common/src/accumulate_and_forward.rs
sigurpol: (slightly related discussion in the twin PR: https://github.com/polkadot-fellows/runtimes/pull/1289#discussion_r4003497788) vs long-term plan

Stephenlawrence00: Yes, making `remote_xcm` configurable would let us delete this, same conclusion as @sigurpol's `xcm-builder` point, though `BurnAsset` needs concrete assets so the reanchor guard would still be needed.

system-parachains/common/src/accumulate_and_forward.rs
Stephenlawrence00: Nothing can: `BurnAsset` is `saturating_take` and cannot fail, and a failed `ReceiveTeleportedAsset` rolls back leaving holding empty, so the comment was wrong and is fixed.

system-parachains/common/src/accumulate_and_forward.rs
Stephenlawrence00: Ha, `acf/ksmt` has to fit in eight bytes either way.

Stephenlawrence00:
> suggest revisiting https://github.com/polkadot-fellows/runtimes/pull/1287/changes#r4014090960 otherwise LGTM

Thanks for the review! I have addressed the fix

system-parachains/common/src/accumulate_and_forward.rs
sigurpol: (to be fair, this is also copy-paste of pre-existing code, not really something this PR added but it's a good point)

system-parachains/common/src/accumulate_and_forward.rs
sigurpol: Agreed on the direction: one forwarder in xcm-builder. And agree on Steven's comment about BurnAsset wanting concrete assets. 
My proposal:  @Stephenlawrence00  can you create an issue in the SDK + add a TODO here pointing at the issue ? Once we bump SDK, we can then drop this code and reuse SDK's variant. @kianenigma , is that ok for you?

sigurpol: ok, I've created https://github.com/paritytech/polkadot-sdk/issues/13238 - so you can reference it here

Stephenlawrence00: Added, pointing at paritytech/polkadot-sdk#13238. Thanks for opening it

sigurpol (approved):
Great work as usual 🙇 ! Thanks @Stephenlawrence00

kianenigma (approved):

system-parachains/common/src/accumulate_and_forward.rs
kianenigma: Is there anything that could potentially cause a trap in this pipeline though?

kianenigma: I've only heard traps being possible in very specific edge cases, and it is when we already sent the message (no rollback on sender chain), but it became unrecoverable on destination.

Stephenlawrence00: No, not here. `BurnAsset` names the exact `Assets` the receive put in holding and uses `saturating_take`, so holding is always empty at the end. It would take a mismatch between what Asset Hub mints and what we burn, and KSM comes back identical through the `FungibleAdapter`.

You're right about where the risk is: the receive failing after we've sent. It rolls back, so holding is empty and nothing is claimable, and the KSM is burned here with Asset Hub untouched.

sigurpol: I've double-checked this pipeline and seems to me that  @Stephenlawrence00 's comment about ReceiveTeleportedAsset and BurnAsset holds so unless I am missing something, we should be good trap-wise.
What could happen would be a rejection on AH that leaves the KSM burnt on the sender with checking account `py/xcmch` untouched.  I guess one way this can happen is if the checking account doesn't cover the amount -  which is not realistic today (https://assethub-kusama.subscan.io/account/F3opxRbN5ZcGCL9nNgBRwNtmbuy3Z7fxADRPaVNrriP5Man -> ~29,809KSM) but could be an issue in the moment we do the reconciliation step in #1283 

> Once the rules are enforced, a reconciliation step is needed since both on Polkadot and Kusama we have burnt / are burning on chains != AH.


On KSM, today checking account is around ~940KSM less than the sum of relay + System chains (the difference is what they have burnt w/o AH post AHM). 
`ReceiveTeleportedAsset` on Asset Hub checks that the checking account can cover the incoming amount. Today not an issue - but if we reconcile as plain snapshot difference with some forwards in flight, the next teleport-and-burn would arrive at a checking account too small to cover it. It would be rejected, the KSM would already be burnt on the sender, and the account would end up overstated again :melting_face:  Probably easiest to reconcile giving a small margin above the exact diff.

system-parachains/bridge-hubs/bridge-hub-kusama/src/lib.rs
```diff
@@ -318,6 +322,29 @@ impl pallet_balances::Config for Runtime {
 	type DoneSlashHandler = ();
 }
 
+parameter_types! {
+	pub const AccumulateForwardPalletId: PalletId = ACCUMULATE_FORWARD_PALLET_ID;
+	pub const ForwardPeriod: BlockNumber = HOURS;
+	pub const MinForwardAmount: Balance = UNITS / 10;
```
acatangiu: units is granular enough for teleporting + burning fees

```suggestion
	pub const MinForwardAmount: Balance = UNITS;
```

system-parachains/common/src/accumulate_and_forward.rs
```diff
@@ -0,0 +1,159 @@
...
+};
+use sp_runtime::DispatchError;
+use xcm::latest::prelude::*;
+use xcm_executor::traits::TransactAsset;
+
+const LOG_TARGET: &str = "xcm::accumulate-forward";
+
+/// Fails `try-runtime` if the accumulation account lacks the ED, without which dust is burned
+/// rather than accumulated. Fund it out of band, as for `acf/dott` in #1282.
+pub struct EnsureAccumulationAccountFunded<T>(PhantomData<T>);
```
acatangiu: remove this if the accounts are already funded at this point - minimize dead code / technical debt

system-parachains/common/src/accumulate_and_forward.rs
```diff
@@ -0,0 +1,159 @@
...
+	PhantomData<(XcmConfig, Dest, NativeAsset)>,
+);
+
+impl<XcmConfig, Dest, NativeAsset> TeleportAndBurnForwarder<XcmConfig, Dest, NativeAsset>
+where
+	XcmConfig: xcm_executor::Config,
+	Dest: Get<Location>,
+	NativeAsset: Get<Location>,
+{
+	fn teleport_and_burn(source: [u8; 32], amount: u128) -> Result<(), XcmError> {
```
acatangiu: don't do all these error-prone manual steps here.

just copy/paste what `TeleportForwarderForAccountId32` is doing and replace the `AccountId32` part with a `Burn`.

specifically, like Kian said, make sure to use `prepare_and_execute()` and have a rollback if it fails

system-parachains/coretime/coretime-kusama/src/lib.rs
```diff
@@ -301,6 +305,29 @@ impl pallet_balances::Config for Runtime {
 	type DoneSlashHandler = ();
 }
 
+parameter_types! {
+	pub const AccumulateForwardPalletId: PalletId = ACCUMULATE_FORWARD_PALLET_ID;
+	pub const ForwardPeriod: BlockNumber = HOURS;
+	pub const MinForwardAmount: Balance = UNITS / 10;
```
acatangiu:
```suggestion
	pub const MinForwardAmount: Balance = UNITS;
```

system-parachains/encointer/src/lib.rs
```diff
@@ -364,6 +367,29 @@ impl pallet_balances::Config for Runtime {
 	type DoneSlashHandler = ();
 }
 
+parameter_types! {
+	pub const AccumulateForwardPalletId: PalletId = ACCUMULATE_FORWARD_PALLET_ID;
+	pub const ForwardPeriod: BlockNumber = HOURS;
+	pub const MinForwardAmount: Balance = UNITS / 10;
```
acatangiu:
```suggestion
	pub const MinForwardAmount: Balance = UNITS;
```

system-parachains/people/people-kusama/src/lib.rs
```diff
@@ -275,6 +277,29 @@ impl pallet_balances::Config for Runtime {
 	type DoneSlashHandler = ();
 }
 
+parameter_types! {
+	pub const AccumulateForwardPalletId: PalletId = ACCUMULATE_FORWARD_PALLET_ID;
+	pub const ForwardPeriod: BlockNumber = HOURS;
+	pub const MinForwardAmount: Balance = UNITS / 10;
```
acatangiu:
```suggestion
	pub const MinForwardAmount: Balance = UNITS;
```

system-parachains/common/src/accumulate_and_forward.rs
Stephenlawrence00: Done, now mirrors `TeleportForwarderForAccountId32` with `prepare_and_execute` and rollback; it also needs the accumulation account in `WaivedLocations` and, like #1282 did for Asset Hub Polkadot, in Asset Hub Kusama's `AllowExplicitUnpaidExecutionFrom`.

system-parachains/common/src/accumulate_and_forward.rs
Stephenlawrence00: Removed, the accounts are funded.

system-parachains/common/src/accumulate_and_forward.rs
```diff
@@ -139,18 +53,52 @@ where
...
+
+		// The XCM flow: `ReceiveTeleportedAsset → UnpaidExecution → BurnAsset`.
+		let xcm: Xcm<XcmConfig::RuntimeCall> = Xcm(vec![
+			UnpaidExecution { weight_limit: WeightLimit::Unlimited, check_origin: None },
+			DescendOrigin(Junction::AccountId32 { network: None, id: source.into() }.into()),
+			WithdrawAsset(asset.into()),
+			InitiateTransfer {
+				destination: dest,
+				remote_fees: None,
+				preserve_origin: false,
```
sigurpol, line 76: this is problematic , I believe we  end up dropping the burn message at the barrier here (I would expect CI to fail too now).
One way we could make it work while keeping the (good) @acatangiu's suggestion:  mirror the SDK fully so `preserve_origin: true`, and add an accumulation-accounts arm to Asset Hub Kusama's `AllowExplicitUnpaidExecutionFrom`, copying SystemChainAccumulationAccounts from Asset Hub Polkadot .

Note that SDK work because `preserve_origin: true` turns the origin into `Parachain(id)/AccountId32(acf)` via AliasOrigin, and Asset Hub Polkadot lists those origins in its unpaid barrier thanks to #1282

Stephenlawrence00: Thanks, that was exactly it. Done: `preserve_origin`: `true` plus `SystemChainAccumulationAccounts` copied from 1282, with its unit test. All four chains pass e2e.

The relay arm is included so 1289 won't need to touch Asset Hub, and Asset Hub Kusama now has to be upgraded before the four system chains, same ordering as 1282.

acatangiu: it should've worked even with `preserve_origin: false`, the real origin is a system chain that is allowed unpaid execution, then origin would've been cleared before running `BurnAssets` which doesn't need any specific origin as it just burns.

The cleaner setup for Kusama is actually one with `preserve_origin: false` and no fake accounts and special `UnpaidBarrier` managing.

Stephenlawrence00: Right idea, though not quite that reason: `AllowExplicitUnpaidExecutionFrom` rejects `ClearOrigin` outright, before execution reaches `BurnAsset`:

```
ClearOrigin => {
    // We don't support the `ClearOrigin` instruction since we always need
    // to know the origin to know if it's allowed unpaid execution.
    return Err(ProcessMessageError::Unsupported);
},
```

But the paid barrier tolerates it and accepts `PayFees`, so using `remote_fees` gives `preserve_origin: false` with no Asset Hub change. Switched to that and dropped `SystemChainAccumulationAccounts`; all four chains pass e2e.

Only trade-off: Asset Hub keeps its execution fee rather than burning it, ~0.0004% of the forward. It stays in Asset Hub's issuance, so the checking account still reconciles exactly.

acatangiu: > Only trade-off: Asset Hub keeps its execution fee rather than burning it, ~0.0004% of the forward. It stays in Asset Hub's issuance, so the checking account still reconciles exactly.

The asset-hub execution fees also eventually get burned further down the line, no?

either way, it's fine

sigurpol:
A note for future us: once we settle on the final version of the forwarder (see [this thread](https://github.com/polkadot-fellows/runtimes/pull/1287#discussion_r4025677331), we will have to run the benchmark once again ( or remember to do it when we release 2.5.1 or whatever the next version is)

system-parachains/common/src/accumulate_and_forward.rs
sigurpol: @acatangiu , can you please check this new approach :bow:  ?

Stephenlawrence00: > The asset-hub execution fees also eventually get burned further down the line, no?
> 
> either way, it's fine

Not burned, no: Asset Hub Kusama's `Trader` resolves execution fees to `StakingPot`, which is `CollatorSelection::account_id()`, so that sliver pays collators instead. It stays in Asset Hub's issuance either way, which is what keeps the checking account exact. Thanks!

Stephenlawrence00:
> A note for future us: once we settle on the final version of the forwarder (see [this thread](https://github.com/polkadot-fellows/runtimes/pull/1287#discussion_r4025677331), we will have to run the benchmark once again ( or remember to do it when we release 2.5.1 or whatever the next version is)

Yes, please re-run benchmarks :)

system-parachains/common/src/accumulate_and_forward.rs
```diff
@@ -0,0 +1,105 @@
...
+		// The XCM flow: `ReceiveTeleportedAsset → ClearOrigin → PayFees → RefundSurplus →
+		// BurnAsset`, which Asset Hub's existing paid barrier accepts unchanged.
+		let remote_xcm = Xcm(vec![RefundSurplus, BurnAsset(remote_asset.clone().into())]);
+		let xcm: Xcm<XcmConfig::RuntimeCall> = Xcm(vec![
+			UnpaidExecution { weight_limit: WeightLimit::Unlimited, check_origin: None },
+			DescendOrigin(Junction::AccountId32 { network: None, id: source.into() }.into()),
+			WithdrawAsset(asset.clone().into()),
+			InitiateTransfer {
+				destination: dest,
+				remote_fees: Some(AssetTransferFilter::Teleport(remote_asset.into())),
```
sigurpol: IIUC `remote_fees` is matched against local  holding,  so why are we using `remote_asset` here?

system-parachains/common/src/accumulate_and_forward.rs
```diff
@@ -0,0 +1,105 @@
...
+		let asset = Asset { id: AssetId(NativeAsset::get()), fun: Fungible(amount.into()) };
+
+		// `BurnAsset` takes concrete assets, so the amount has to be named as Asset Hub sees it.
+		let remote_asset = asset
+			.clone()
+			.reanchored(&dest, &XcmConfig::UniversalLocation::get())
+			.map_err(|asset| {
+				log::error!(target: LOG_TARGET, "🚨 could not reanchor {asset:?} for {dest:?}");
+			})?;
+		// The XCM flow: `ReceiveTeleportedAsset → ClearOrigin → PayFees → RefundSurplus →
```
sigurpol: aren't we emitting PaysFee before ClearOrigin?

integration-tests/emulated/helpers/src/common.rs
```diff
@@ -30,3 +30,124 @@ pub mod snowbridge {
...
+					$asset_hub,
+					vec![AssetHubEvent::MessageQueue(
+						$crate::pallet_message_queue::Event::Processed { success: true, .. }
+					) => {},]
+				);
+				// Everything that arrives is burned except Asset Hub's execution fee, which stays
+				// in its issuance and so keeps the checking account correct.
+				let burned = asset_hub_issuance_before - AssetHubBalances::total_issuance();
+				assert!(
+					burned > 0 && burned <= forwarded,
```
sigurpol: can we have an exact assertion here again (we know where the fee ends and we could check the collator pot before and after  in theory)?

sigurpol (commented):
I'll let Adrian comment on the new approach in the forwarder since he is the XCM undisputed  🫅  - added few nits if we keep current approach there

system-parachains/common/src/accumulate_and_forward.rs
sigurpol: For DAP runtimes we have https://github.com/paritytech/polkadot-sdk/issues/11409, it's a bit unclear to me if for Kusama we want to keep things as they are or burned but yeah, out of scope of this PR most likely

sigurpol:
> > A note for future us: once we settle on the final version of the forwarder (see [this thread](https://github.com/polkadot-fellows/runtimes/pull/1287#discussion_r4025677331), we will have to run the benchmark once again ( or remember to do it when we release 2.5.1 or whatever the next version is)
> 
> Yes, please re-run benchmarks :)

Sure thing, I was waiting for the forwarder impl to settle first, then I will re-run all of them :popcorn:

system-parachains/common/src/accumulate_and_forward.rs
sigurpol: and one extra note not for this PR but maybe worth raising now that we are discussing about it: this new paid pattern  would also let the Polkadot forwarder drop the `SystemChainAccumulationAccounts` arm on Asset Hub Polkadot and the whole aliasing dance - and once we do SDK#11409 anything would flow in the DAP anyway.

system-parachains/common/src/accumulate_and_forward.rs
acatangiu: @sigurpol is right, this should be the local asset - it happens to work as is because reanchoring `Parent` between parachains results in the same relative location (`Parent`), but the is technically incorrect and should be corrected

acatangiu (commented):
looks good, will approve once the remaining nits are also addressed

sigurpol:
/cmd bench --runtime bridge-hub-kusama --pallet pallet_accumulate_and_forward

sigurpol:
/cmd bench --runtime coretime-kusama --pallet pallet_accumulate_and_forward

sigurpol:
/cmd bench --runtime people-kusama --pallet pallet_accumulate_and_forward

sigurpol:
/cmd bench --runtime encointer-kusama --pallet pallet_accumulate_and_forward

integration-tests/emulated/helpers/src/common.rs
```diff
@@ -30,3 +30,142 @@ pub mod snowbridge {
...
+			// Enough that the forward still clears `MinTransferAmount` after arrival fees.
+			let teleported = 10 * $chain::execute_with(|| {
+				<ChainRuntime as pallet_accumulate_and_forward::Config>::MinTransferAmount::get()
+			});
+
+			let staking_pot =
+				$asset_hub::execute_with($crate::pallet_collator_selection::Pallet::<
+					AssetHubRuntime,
+				>::account_id);
+			let (asset_hub_issuance_before, check_balance_before, pot_before) = $asset_hub::execute_with(|| {
```
sigurpol: [nit] pot_before is never used so dead code - remove it, leftover from previous commits

sigurpol (approved):
Nice! Approving on my side while waiting for new weights to land. Thanks @Stephenlawrence00  (and thanks @acatangiu and @kianenigma  for the reviews 🙇 )

Stephenlawrence00:
> Nice! Approving on my side while waiting for new weights to land. Thanks @Stephenlawrence00 (and thanks @acatangiu and @kianenigma for the reviews 🙇 )

Thank you! All done @acatangiu

acatangiu (approved):

PR #1289, opened 2026-09-12
Kusama relay: stop burning dust locally, route it through Asset Hub

Stephenlawrence00:
The Kusama relay accumulates the dust it used to burn locally and teleports it to Asset Hub to be burned there, so the burn moves the network `TotalIssuance` that Asset Hub tracks, with the rest of the relay's post-AHM sinks inventoried and found to be dormant or non-burning (#1283).

## Checklist
- [x] Account funded (`F3opxRYLmNnYbWQsbYMryaRxM7PeLKRhKEorQVeDeF15KKy`, relay ED)
- [x] Weights regenerated
- [x] Rebase on #1287 once it lands, dropping its duplicate forwarder and pallet id


sigurpol:
cc @Ank4n , @andreitrand , @cirko33

sigurpol:
(note: migration job failing for the same reason mentioned [here](https://github.com/polkadot-fellows/runtimes/pull/1287#issuecomment-5644391702) so once account is funded, we can re-run the job)

sigurpol:
Same as per its twin suggestion: please add a checklist in the PR body so it's easy to track progress and what's missing e.g. 1. fund the account 2. regenerate weights  3. ...

sigurpol:
/cmd bench --runtime kusama --pallet pallet_accumulate_and_forward

relay/kusama/constants/src/accumulate_and_forward.rs
```diff
@@ -0,0 +1,144 @@
...
+
+// Polkadot is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Polkadot. If not, see <http://www.gnu.org/licenses/>.
+
+//! XCM adapter for `pallet-accumulate-and-forward`.
```
sigurpol, line 17: We would probably benefit from some de-duplication with #1287 see [system-parachains/common/src/accumulate_and_forward.rs‎](https://github.com/polkadot-fellows/runtimes/pull/1287/changes#diff-f8995a73f631547934ab65405105a064add498088d2f9c82c95310cb40926cad)

Also,  acf/ksmt is now defined in two constants crates, not ideal.

My 2c: `kusama-runtime-constants` is the one crate all five runtimes already depend on, and it already carries non-constant content with the proxy module and an xcm-builder dependency, so it can host both the forwarder and the single pallet id. The simplest route  imo is to fold #1289 into #1287. Mid/Longer term (or already now?) the forwarder belongs in xcm-builder next to `TeleportForwarderForAccountId32`.
Wdyt?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- Kusama relay: `DustRemoval` resolves into a `pallet-accumulate-and-forward` account which, at most hourly and once at least 0.01 KSM has gathered, teleports to Asset Hub and burns there via `BurnAsset`. KSM issuance is tracked on Asset Hub, so burning on the relay lowered its total while leaving Asset Hub's total and its XCM checking account untouched, overstating the real supply. Asset Hub Kusama needs no change: it already trusts the relay as a KSM teleporter and grants it free execution. **The accumulation account must be funded with the existential deposit before this runtime goes live**, otherwise `can_deposit` rejects sub-existential-deposit dust and it is burned as before; the runtime logs an error and `try-runtime` fails if it is missing ([#1283](https://github.com/polkadot-fellows/runtimes/issues/1283)).
```
sigurpol: same as #1287: I would suggest one line +  link to 1289 and not 1283, it's too verbose as it is now

sigurpol:
Added some comments. I would consider fold #1289 into #1287. Or if you want to keep it separate, land #1287 first and put #1289 on top to reduce the duplication issue mentioned  [here](https://github.com/polkadot-fellows/runtimes/pull/1289#discussion_r4003497788)

CHANGELOG.md
Stephenlawrence00: Done, one line linking 1289

relay/kusama/constants/src/accumulate_and_forward.rs
Stephenlawrence00: Agreed on `kusama-runtime-constants`, moved the forwarder there in this PR so `acf/ksmt` and the adapter live in one place. We'd rather keep the two PRs separate and land #1287 first, then rebase this on top, dropping its copies in `system-parachains`/ and repointing the four system chains. Keeps the consolidation in one diff.

Also, the constants crate had no `try-runtime` feature, so `post_upgrade` on the funding guard wasn't compiling. Added and propagated from the relay.

Agreed the forwarder belongs in `xcm-builder` beside `TeleportForwarderForAccountId32` longer term.

sigurpol: sounds great, ty!

sigurpol:
@Stephenlawrence00 , I've funded F3opxRYLmNnYbWQsbYMryaRxM7PeLKRhKEorQVeDeF15KKy (acf/ksmt) on relay above ED - will retrigger the check migration job :popcorn:

relay/kusama/src/lib.rs
```diff
@@ -431,6 +432,28 @@ impl pallet_balances::Config for Runtime {
 	type DoneSlashHandler = ();
 }
 
+parameter_types! {
+	pub const AccumulateForwardPalletId: PalletId =
+		kusama_runtime_constants::account::ACCUMULATE_FORWARD_PALLET_ID;
+	/// Hourly at most, once 0.01 KSM has gathered; funds waiting still count towards issuance.
+	pub const ForwardPeriod: BlockNumber = HOURS;
+	pub const MinForwardAmount: Balance = UNITS / 100;
```
acatangiu: nit:


```suggestion
	pub const MinForwardAmount: Balance = UNITS;
```

acatangiu (approved):

relay/kusama/src/lib.rs
Stephenlawrence00: Applied, thanks. Also updated the comment above it, which still said 0.01 KSM

sigurpol:
/cmd bench --runtime kusama --pallet pallet_accumulate_and_forward

integration-tests/emulated/tests/assets/asset-hub-kusama/src/tests/burn.rs
```diff
@@ -0,0 +1,112 @@
...
+// limitations under the License.
+
+use crate::*;
+use frame_support::traits::{
+	fungible::{Inspect as FungibleInspect, Mutate as FungibleMutate},
+	Hooks,
+};
+use parachains_common::AccountId;
+
+/// KSM accumulated on the relay is teleported to Asset Hub and burned where issuance is
```
sigurpol: in  #1287 as per @kianenigma 's suggestion, we tried to make e2e tests more "realistic". Here you are still funding  the accumulation account synthetically and topping up the checking account by hand - which is ok by itself for this test but maybe we could  make a real teleport from Asset Hub to the relay first, instead. Wdyt?

sigurpol (commented):
Small comment - let me know if you think it makes sense to address.
Happy to approve once new weights land (I've just run the /cmd bench 🍿 )

integration-tests/emulated/tests/assets/asset-hub-kusama/src/tests/burn.rs
Stephenlawrence00: Done, it now teleports from Asset Hub into the accumulation account, then checks the forward, the burn with its exact fee, and that the checking account keeps `teleported - forwarded`.

Unlike 1287 it doesn't compare against relay issuance: the relay pays its arrival fee via `ToAuthor`, which the emulator burns for lack of an author.

sigurpol (approved):
Happy to approve, let's merge then when new weights land and CI is happy!

PR #1292, opened 2026-09-15
People Polkadot: Remove single-use Asset Hub subscription whitelist

andrzejSulkowski:
Runtime Upgrade `v2.5.0` was enacted on the people chain, and with it the single-use migration `SeedAssetHubSubscriptionWhitelist` ran.

This PR turns the migration dormant for future short-term review. I've setup a notification in 2 months to fully remove it.


kianenigma (approved):
Thanks, and good that you are removing this!

We generally write this kind of runtime migrations idempotent to guard against this. 

For pallet migrations, the `storage_version` and using a `VersionedMigration` is exactly for this. In this case, it had to be done manually I presume :)

georgepisaltu (approved):

sigurpol (approved):

andrzejSulkowski:
> Thanks, and good that you are removing this!
> 
> We generally write this kind of runtime migrations idempotent to guard against this.
> 
> For pallet migrations, the `storage_version` and using a `VersionedMigration` is exactly for this. In this case, it had to be done manually I presume :)

Thanks for the feedback! Noted and much appreciated 🙌🏻

acatangiu (approved):

Issue #696, opened 2025-04-29
Tracking issue for TODOs and leftover


bkontur:
Check: https://github.com/search?q=repo%3Apolkadot-fellows%2Fruntimes%20TODO&type=code
Looks like some of them can be remove now.

## TODO

- [ ] when all fixed, we can close
- [ ] thinking about bringing `FAIL-CI` grep regex feature as CI job, so we won't forget to fix anything which should be fixed as a part of PR

clangenb:
I have been thinking about this - the way it stands now, this issue will stay open forever, as new todos will keep emerging. So I thought how about a CI check that simply enforces that each todo/fixme also has a tracking issue on the same line? This is trivially to implement, and we at least know the actionability (or the absence thereof) of each todo. Also, having actual issues will make the todo more visible, and the people who scan actionable issues with AI will find them, and fix them if the issues are simple enough. What do you think?

I asked AI to give me a summary of the current state, we have 56 todos, where 21 are lacking an issue. (For some issues the actionability needs be researched deeper, this was just to get an overview).

Disclaimer, as part of this effort I have also started fixing some actionable todos.

  ## 1. Current state of TODOs

  | Bucket | Count | Action |
  |---|---|---|
  | Tracked against an **open** issue (sharing allowed) | 29 | None — compliant |
  | **Unlinked** (no reference at all) | 21 | File/link an issue, or delete |
  | Linked to a **closed runtimes** issue (stale) | 2 | Verify & remove, or reopen |
  | Linked to a **merged/closed SDK** item (condition met) | 4 | Remove the guarded workaround |
  | **Total markers** | **56** | **27 actionable** (9 already fixed in open PRs #1213/#1214/#1216/#1217 → 18 left) |

  Breakdown of the 29 compliant ones: #1137 "Configure DAP account as fee collector" (open) ×23, #840 "pallet-revive benchmark with AHK runtime" (open) ×4, #404 "Coretime Runtime API refactor" (open) ×2.

  ## 2. Full triage of the 27 actionable markers

  Grouped by theme; each group should become **one** issue (or link to an existing one) unless noted.

  ### A. Post-AHM (set/relocate after Asset Hub Migration) — 4 markers → 1 umbrella issue
  | File:Line | Text |
  |---|---|
  | `relay/kusama/src/lib.rs:560` | `we will set this post-AHM` |
  | `relay/kusama/src/xcm_config.rs:232` | `post-ahm move the Treasury funds … to sovereign account` |
[...]

PR #1293, opened 2026-09-16
Bridge Hub Polkadot: use upstream `ethereum_extrinsic` test

clangenb:
`snowbridge-runtime-test-common` has exposed `ethereum_extrinsic` since `stable2412`, which is what the local copy was waiting for. The only behavioural difference was an extra `run_to_block(3)` before the invalid sync committee update, which the test no longer needs.

Drops the now-unused `snowbridge-pallet-ethereum-client-fixtures` dev-dependency along with it.

Part of #696.


claravanstaden (approved):
Nice! The `snowbridge-pallet-ethereum-client-fixtures` crate can now be removed from the workspace dependencies too: https://github.com/polkadot-fellows/runtimes/blob/main/Cargo.toml#L271

acatangiu (approved):

PR #1299, opened 2026-09-16
Migrate from `integration-tests-helpers` to upstream's `emulated-integration-tests-common` crate where possible

### LINKED ISSUE #363: Remove emulated test macros
TODO from here:  

https://github.com/polkadot-fellows/runtimes/blob/8ec28f96eeb30fbba30d29006d75e1a3fa1cea1c/integration-tests/emulated/helpers/src/lib.rs#L31-L33

cc @bkontur

clangenb:
`integration-tests-helpers` had become a re-export facade, the macros it existed for were backported. For compatibilty reasons `pub use emulated_integration_tests_common::*` was kept. Although, the test crates have mostly imported from the `emulated_integration_tests_common` directly.

This PR does the following:
- Remove the re-export to prevent the schizophrenic import pattern. Almost all crates depend now only on the upstream version.
- The snowbridge stuff is re-exported where possible instead of redefined
- Replace the test macro with a generic test function for better typechecks, mirroring the more recent pattern in the codebase.
- One unrelated line: `.gitignore`'s `**/chains/` also matches `integration-tests/emulated/chains/`, so
files added to the emulated chain crates are silently untracked. Happy to split it out.

Closes #363


.gitignore
```diff
@@ -17,6 +17,7 @@
 **/.criterion/
 **/*.rs.bk
 **/chains/
+!integration-tests/emulated/chains/
```
clangenb, line 20: This kept biting me. The idea was to ignore the specs I assume. I guess it is fine that we whitelist this exact entry?

integration-tests/emulated/helpers/src/burn.rs
```diff
@@ -0,0 +1,165 @@
...
+/// Asserts that funds teleported into a chain's accumulation account are forwarded back to Asset
+/// Hub and burned there, leaving Asset Hub's checking account and the chain's issuance in step.
+///
+/// For Kusama-like chains, which burn. Polkadot chains forward to the DAP staging account instead
+/// and are covered by the SDK's `dap_helpers::test_accumulate_forward_transfers_to_asset_hub`.
+///
+/// `set_block_number` advances whatever clock the chain configures as
+/// `pallet_accumulate_and_forward::Config::BlockNumberProvider`, since that is what decides
+/// whether `on_idle` forwards. Today every such chain uses `System`.
+pub fn test_accumulated_funds_are_burnt_on_asset_hub<Sender, AH>(
```
clangenb, line 34: The content of the test is identical to the macro version that was there before except for:

- The unused `ASSET_HUB_KUSAMA_ED` is removed
- The other chains ED retrieval is moved to the pallet_balances.

The more recent tests in the repo have already started to work with generic functions rather than macros. So I thought that I will migrate it too while I am at it.

integration-tests/emulated/tests/bridges/bridge-hub-kusama/src/tests/burn.rs
```diff
@@ -15,9 +15,11 @@
...
-	ASSET_HUB_KUSAMA_ED,
-);
+#[test]
+fn accumulated_funds_are_burnt_on_asset_hub() {
+	use bridge_hub_kusama_runtime::System;
+	integration_tests_helpers::burn::test_accumulated_funds_are_burnt_on_asset_hub::<
+		BridgeHubKusama,
+		AssetHubKusama,
+	>(AssetHubKusamaSender::get(), System::set_block_number);
+}
```
clangenb, line 25: `ASSET_HUB_KUSAMA_ED` has been unused and the `BRIDGE_HUB_KUSAMA_ED` is taken from balances.

The `set_block_number` is for chains that might eventually swith to `RelaychainBlockNumberProvider`.

integration-tests/emulated/tests/coretime/coretime-kusama/Cargo.toml
```diff
@@ -13,7 +13,9 @@ codec = { workspace = true, default-features = true }
 # Substrate
 sp-runtime = { workspace = true, default-features = true }
 frame-support = { workspace = true, default-features = true }
+pallet-accumulate-and-forward = { workspace = true, default-features = true }
 pallet-balances = { workspace = true, default-features = true }
+pallet-collator-selection = { workspace = true, default-features = true }
```
clangenb, line 18: The only crate that relied on some re-exports that were imported via integration-test helpers before.

integration-tests/emulated/helpers/src/lib.rs
```diff
@@ -13,26 +13,11 @@
...
-pub use pallet_xcm;
-pub use xcm::prelude::{
-	AccountId32, Assets, Junction, Location, VersionedAssetId, VersionedAssets, Weight, WeightLimit,
-};
-pub use xcm_runtime_apis::fees::runtime_decl_for_xcm_payment_api::XcmPaymentApiV2;
-
-// Cumulus
-pub use cumulus_pallet_xcmp_queue;
-pub use emulated_integration_tests_common::*;
-pub use xcm_emulator::Chain;
```
clangenb, line 36: Non of those where imported anywhere except for one narrow case. They existed purely for macro import, which made me remove them.

integration-tests/emulated/helpers/src/burn.rs
clangenb: On another note, I guess the idea is that we move this test upstream too?

PR #1301, opened 2026-09-17
Coretime Kusama: burn revenue through accumulate-and-forward and retire py/ctbrn

### LINKED ISSUE #1290: Coretime Kusama: retire py/ctbrn and route OnRevenue through accumulate-and-forward
Part of #1283, follow-up to #1287.

After #1287, Coretime Kusama has two teleport-and-burn paths: bulk revenue via `py/ctbrn` and the daily `on_new_timeslice sweep`, and dust via `pallet-accumulate-and-forward`. Collapse them by routing `OnRevenue` to `AccumulateAndForward` and retiring `py/ctbrn` with an idempotent migration, mirroring Coretime Polkadot in #1282.

This also closes a gap: `py/ctbrn` is swept to zero daily with `Preservation::Expendable`, so sub-ED dust arriving while it sits empty is burned outright — `can_deposit` refuses a deposit leaving an account below ED. The accumulation account keeps its ED.

Must land after #1287, which introduces the pallet this depends on.

Stephenlawrence00:
Coretime Kusama now routes bulk revenue through `pallet-accumulate-and-forward` like its dust, dropping the `py/ctbrn` pot and its daily sweep with an idempotent migration mirroring Coretime Polkadot from #1282, so there is a single path to Asset Hub (closes #1290).


sigurpol:
(cc @Ank4n )

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 
+- Coretime Kusama: bulk revenue now accumulates with dust and is burnt on Asset Hub through `pallet-accumulate-and-forward`, and a migration retires the `py/ctbrn` holding account ([#1290](https://github.com/polkadot-fellows/runtimes/issues/1290)).
```
sigurpol: Better to point to 1301 and not to 1290

system-parachains/coretime/coretime-kusama/src/tests.rs
```diff
@@ -91,36 +91,81 @@ fn bulk_revenue_is_burnt() {
...
+		// AND a rerun changes nothing.
+		RetireCoretimeBurnAccount::on_runtime_upgrade();
+		assert_eq!(Balances::balance(&accumulation_account), ExistentialDeposit::get() + 5 * UNITS);
+		assert!(!System::account_exists(&burn_account));
+		assert_eq!(Balances::total_issuance(), issuance_before);
+	});
+}
+
 #[test]
 fn timeslice_period_is_sane() {
```
sigurpol: remove it - not needed anymore ,right? The comment below "assumption in burning logic" is also misleading now. I think we can drop it

integration-tests/emulated/tests/coretime/coretime-kusama/src/tests/burn.rs
```diff
@@ -15,82 +15,102 @@
...
+	},
+	frame_system, pallet_accumulate_and_forward, pallet_balances, pallet_collator_selection,
+	pallet_message_queue, pallet_xcm,
 };
-use kusama_runtime_constants::{system_parachain::coretime::TIMESLICE_PERIOD, time::DAYS};
-use pallet_broker::CoretimeInterface;
 use sp_runtime::traits::AccountIdConversion;
 
-/// Coretime sweeps its revenue holding account and burns it on Asset Hub.
+/// Revenue paid through the broker's `OnRevenue` hook reaches the accumulation account and is
```
sigurpol: Same point @kianenigma  made on #1287: this test mints the revenue and tops up the checking account by hand, while the macro `test_accumulated_funds_are_burnt_on_asset_hub!`  does a real teleport in (and you are using at line 118 of this file already). 
I see 2 options:
1. make the test more realistic (Could it teleport KSM to a buyer, run a real purchase using the sale setup from bulk_revenue_is_accumulated, and then assert the macro's invariant for example)? 
2. If that is too awkward / complex in the emulator, dropping this test _might_ be fine too, since the macro covers the forward and the unit test covers OnRevenue.

sigurpol:
`on_new_timeslice` is now a no-op, so CT  `pallet_broker` weights overestimate, will re-run benchmark there

integration-tests/emulated/tests/coretime/coretime-kusama/src/tests/burn.rs
sigurpol: there is also option 3: leave it as it is now of course 😄

Stephenlawrence00:
> `on_new_timeslice` is now a no-op, so CT `pallet_broker` weights overestimate, will re-run benchmark there

Thanks. It's a no-op now, so the current pallet_broker weights overestimate it until you re-run them.

system-parachains/coretime/coretime-kusama/src/tests.rs
Stephenlawrence00: Removed, along with the imports it left unused

integration-tests/emulated/tests/coretime/coretime-kusama/src/tests/burn.rs
Stephenlawrence00: Went with option 1: it now teleports KSM from Asset Hub to a buyer, runs a real sale and purchase, and checks the macro's invariant (checking-account delta equals Coretime's issuance delta) plus the exact burn and fee. The revenue it forwards comes entirely from that purchase.

sigurpol (approved):

acatangiu (approved):

sigurpol (approved):
Great stuff as usual 🙇

PR #1302, opened 2026-09-21
Polkadot & Kusama: remove stale downward message queues

### LINKED ISSUE #513: Cleanup stale HRMP messages
We need to clean up some stale HRMP messages after https://github.com/paritytech/polkadot-sdk/pull/6604 was applied on chain.

clangenb:
Before paritytech/polkadot-sdk#6604, DMP queued messages for paras without a head. Remove the queues and queue heads left behind for the 8 affected paras per relay, skipping any that has a head by the time of the upgrade.

Closes #513


relay/kusama/src/migrations/stale_dmp_queues.rs
```diff
@@ -0,0 +1,176 @@
...
+	alloc::vec::Vec,
+	codec::{Decode, Encode},
+	frame_support::ensure,
+	sp_runtime::TryRuntimeError,
+};
+
+const LOG_TARGET: &str = "runtime::kusama::migrations";
+
+/// Paras with a downward message queue but no head in live state at block 35279280.
+const STALE_PARAS: [u32; 8] = [0, 1, 49, 100, 2030, 2034, 2042, 2125];
```
clangenb, line 40: Notes regarding the approach: I have diagnosed the para ids with a [script](https://github.com/encointer/runtimes/blob/20a803f067e9d5f3fa80b42b2a8bcad64aa476f0/stale-dmp-queues.py), and chose to only tackle those paras explicitly instead of iterating through all storage keys, as the bug has been fixed, and no new stale DMQs should occur.

relay/kusama/src/migrations/stale_dmp_queues.rs
```diff
@@ -0,0 +1,176 @@
...
+pub struct RemoveStaleDmpQueues;
+
+impl OnRuntimeUpgrade for RemoveStaleDmpQueues {
+	fn on_runtime_upgrade() -> Weight {
+		let mut removed = 0;
+		for para in STALE_PARAS.map(ParaId::from) {
+			if has_head(&para) {
+				log::warn!(target: LOG_TARGET, "Keeping the downward message queue of {para:?}, it has a head");
+				continue;
+			}
```
clangenb, line 68: Some paras could be reregistered in the meantime. So we have to ensure that we do not clear those messages.

PR #1305, opened 2026-09-21
[AHMv2] Shared migrator types and balance primitives

Ank4n:
Extends the scaffold PR [#1225](https://github.com/polkadot-fellows/runtimes/pull/1225).
No-op to the runtimes, these pallets aren't used anywhere yet.

## Context

AHMv2 moves the Relay Chain's remaining state to the Coretime chain and Asset Hub. This PR contains the common things needed by all data stages (accounts, proxies, sweep, Registrar and HRMP, TI correction). It carries the shared items and nothing stage-specific. Split out of the reference PoC (#1284).

## Changes

- **[New]** `migrator-types`: the common types shared between `rc2-migrator` and `ct-migrator`.
- **[Existing]** `pallet-rc2-migrator`: adds a ledger type to track what has migrated and what's left.
- **[Existing]** `pallet-ct-migrator`: receive-batch logic and types to track successes and failures.

## Also see 
- #1285
- #1305 (this)

All PRs can be reviewed independently. I've left `TODO(ahm-v2)` markers that we should go through before wiring everything up.

The reason for splitting is simply to make reviewing and iterating on each change easier.

---

- [x] Does not require a CHANGELOG entry — no runtime is affected.

---


pallets/ct-migrator/src/lib.rs
```diff
@@ -64,6 +79,34 @@ pub mod pallet {
...
+
+		/// The overarching hold reason type.
+		///
+		/// The `From<PortableHoldReason>` bound is where the runtime declares what each migrated
+		/// Relay Chain hold becomes locally.
+		type RuntimeHoldReason: From<HoldReason> + From<PortableHoldReason>;
+	}
+
+	#[pallet::composite_enum]
+	pub enum HoldReason {
```
sigurpol, line 89: [nitpick] can we collapse PortableHoldReason and HoldReason and remove some boilerplate from runtime integration? `impl From<PortableHoldReason>` for `HoldReason` in the pallet and a single `From<HoldReason>` would probably do . Unless there is a good reason to get each runtime freedom to do something different.

pallets/ct-migrator/src/lib.rs
```diff
@@ -77,3 +120,68 @@ pub mod pallet {
...
+		(count_good, count_bad)
+	}
+
+	/// Move `amount` of `who`'s free balance under `reason`.
+	///
+	/// Not `MutateHold::hold`: that reduces the free balance first and books the hold second,
+	/// and pallet-balances reaps an account that is below the ED with nothing on hold, burning
+	/// the remainder. Deposit holders arrive here with sub-ED free balance next to their
+	/// deposit, so the hold is booked first. Total issuance is unchanged, as with `hold`.
+	pub fn place_hold(
```
sigurpol: it would fail with `FundsUnavailable` when the remaining free is below the ED unless the caller bumped a provider ref, right? Should it be stated clearly in the doc or should we bump it here?

sigurpol (commented):
Overall, LGTM - one nit, and one philosophical doubt at this round 😄

pallets/ct-migrator/src/lib.rs
Ank4n: Yeah, this is what the PoC is doing (bumping the provider). But I think thats also incorrect. We should just hold as much as we can, leaving ED free, and then the pallets do their holds on a best-effort basis up to the hold (CT deposits are much lower, so in normal cases this should be fine). I will rework that part of the PoC to make sure everything's good there, and update here too. 

Thanks for flagging it (as always) 🙏

pallets/ct-migrator/src/lib.rs
Ank4n: Yeah, I think we can. This is how it was in v1 so followed that.

Ank4n: Done

pallets/ct-migrator/src/lib.rs
Ank4n: removed the helpers. We will just do normal hold/release now (in the accounts stage)

sigurpol: makes sense!

sigurpol (approved):

pallets/migrator-types/src/lib.rs
```diff
@@ -0,0 +1,147 @@
...
+	MaxEncodedLen,
+)]
+pub enum PortableHoldReason {
+	/// Registrar and HRMP deposits. Re-attributed to the owning pallet when its state arrives.
+	#[codec(index = 0)]
+	UnnamedReserve,
+	/// Proxy deposit of a delegator whose definitions travel too. Resized to the destination's
+	/// rates when they arrive.
+	#[codec(index = 1)]
+	ProxyDeposit,
```
muharem, line 87: why registrar and HRMP grouped? while proxy is separate?

pallets/rc2-migrator/src/lib.rs
```diff
@@ -26,9 +26,37 @@ pub use pallet::*;
...
+	Default,
+	PartialEq,
+	Eq,
+	Debug,
+	TypeInfo,
+	MaxEncodedLen,
+)]
+pub struct MigratedBalances {
+	/// Balance that remains on the Relay Chain.
+	pub kept: u128,
```
muharem, line 49: also later burned? or?

pallets/migrator-types/src/lib.rs
muharem: or bc its unnamed and proxy is named?

Ank4n: All three are unnamed. But registrar and HRMP records note how much deposit is for them, which we will send across to CT as well (see eg: https://github.com/Ank4n/runtimes/blob/ankn-min-relay-int/pallets/migrator-types/src/lib.rs#L136) so those two can share a pool but not take away deposit from other's share. Proxy stage would run first and it does not have an explicit deposit field.

I see this is a little odd setup and cleaner would be to have three separate hold reasons. I will make the change.

pallets/rc2-migrator/src/lib.rs
Ank4n: Yes, at the end, all remaining balances (Manager account, pot balances) will be swept and teleported to AH.

cirko33 (approved):
LGTM! After improving docs it will be complete, as @Ank4n said that he will improve!

pallets/migrator-types/src/lib.rs
Ank4n: https://github.com/polkadot-fellows/runtimes/pull/1305/changes/5d3684587f517ee4e51d35711ea062e57766338a

pallets/rc2-migrator/src/lib.rs
Ank4n: https://github.com/polkadot-fellows/runtimes/pull/1305/changes/5d3684587f517ee4e51d35711ea062e57766338a

muharem (approved):

cirko33 (approved):

sigurpol (approved):

pallets/migrator-types/src/lib.rs
seadanda: This is not a pallet, maybe this is not the right path for this. turbo nit, but if we let some of these things pile up and the repo becomes harder to navigate

pallets/rc2-migrator/primitives/src/lib.rs
Ank4n: Good point. I can move this inside `rc2-migrator/primitives`, which seems a better place for it (rc2-migrator owning the wire format)

Ank4n: https://github.com/polkadot-fellows/runtimes/pull/1305/changes/bcbea27476f980d9781c2d5ed87a4d67c75b81e5

cirko33 (approved):

seadanda (approved):

sigurpol (approved):

sigurpol (approved):

